PR #10, opened 2018-03-15
Add Python 3-only classifier

hugovk:
https://pypi.python.org/pypi?%3Aaction=list_classifiers


ambv:
I thought the 3.6 classifier was explicit enough but sure, no harm in adding "Only".

ambv:
Thanks! ✨ 🍰 ✨

PR #12, opened 2018-03-15
Update code formatting

hugovk:
!XXX doesn't render as code formatting, XXX does.

Before:

image

After:

image


ambv:
Right, thank you!

PR #24, opened 2018-03-15
add python_requires to setup.py

kennethreitz:
This makes Black play nicer with tools like Pipenv :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,8 @@ setup(
     package_data={'blib2to3': ['*.txt']},
     python_requires=">=3.6",
     zip_safe=False,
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -52,6 +52,7 @@ def get_version():
     python_requires=">=3.6",
     zip_safe=False,
     install_requires=['click', 'attrs'],
+    python_requires='>=3.6.0',
```
renstrom: There's already `python_requires` one line 52.

kennethreitz: oops!

ambv: :-)

PR #25, opened 2018-03-15
Add piping from stdin to stdout with a -

treuherz:
Being able to format code by piping it through the formatter makes it much easier to integrate with tools like google/vim-codefmt or Chiel92/vim-autoformat.

My Python is okay-but-not-fantastic and I'm not toally happy with the way I've modified main(), and am not sure how best to test this. Happy to put the work into improving the PR, though.

### DIFF
--- a/black.py
+++ b/black.py
@@ -175,7 +175,8 @@ def format_stdin_to_stdout(line_length: int, fast: bool) -> bool:
     except NothingChanged:
         return False
 
-    sys.stdout.write(contents)
+    finally:
+        sys.stdout.write(contents)
     return True
 
 
--- a/black.py
+++ b/black.py
@@ -3,13 +3,22 @@ import asyncio
 from asyncio.base_events import BaseEventLoop
 from concurrent.futures import Executor, ProcessPoolExecutor
 from functools import partial
-import io
 import keyword
 import os
 from pathlib import Path
 import tokenize
 from typing import (
-    Dict, Generic, Iterable, Iterator, List, Optional, Set, Tuple, Type, TypeVar, Union
+    Dict,
+    Generic,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    TextIO,
+    Tuple,
+    TypeVar,
+    Union,
 )
 import sys
 
@@ -174,21 +183,21 @@ def format_stdin_to_stdout(line_length: int, fast: bool) -> bool:
 
 
 def format_file(
-    src_buffer: Type[io.TextIOBase], line_length: int, fast: bool
+    src: TextIO, line_length: int, fast: bool
 ) -> Tuple[FileContent, Encoding]:
     """Reformats a file and returns its contents and encoding."""
-    src_contents = src_buffer.read()
+    src_contents = src.read()
     if src_contents.strip() == '':
-        raise NothingChanged(src_buffer.name)
+        raise NothingChanged(src.name)
 
     dst_contents = format_str(src_contents, line_length=line_length)
     if src_contents == dst_contents:
-        raise NothingChanged(src_buffer.name)
+        raise NothingChanged(src.name)
 
     if not fast:
         assert_equivalent(src_contents, dst_contents)
         assert_stable(src_contents, dst_contents, line_length=line_length)
-    return dst_contents, src_buffer.encoding
+    return dst_contents, src.encoding
 
 
 def format_str(src_contents: str, line_length: int) -> FileContent:
--- a/black.py
+++ b/black.py
@@ -9,19 +9,8 @@ from pathlib import Path
 import tokenize
 import sys
 from typing import (
-    Dict,
-    Generic,
-    Iterable,
-    Iterator,
-    List,
-    Optional,
-    Set,
-    TextIO,
-    Tuple,
-    TypeVar,
-    Union,
+    Dict, Generic, Iterable, Iterator, List, Optional, Set, Tuple, TypeVar, Union
 )
-import sys
 
 from attr import dataclass, Factory
 import click
--- a/black.py
+++ b/black.py
@@ -83,27 +83,24 @@ class CannotSplit(Exception):
 def main(ctx: click.Context, line_length: int, fast: bool, src: List[str]) -> None:
     """The uncompromising code formatter."""
     sources: List[Path] = []
-    if len(src) == 1 and src[0] == '-':
-        sources.append(Path('-'))
-    else:
-        for s in src:
-            p = Path(s)
-            if p.is_dir():
-                sources.extend(gen_python_files_in_dir(p))
-            elif p.is_file():
-                # if a file was explicitly given, we don't care about its extension
-                sources.append(p)
-            elif s == '-':
-                err('Cannot format both stdin and other paths')
-            else:
-                err(f'invalid path: {s}')
+    for s in src:
+        p = Path(s)
+        if p.is_dir():
+            sources.extend(gen_python_files_in_dir(p))
+        elif p.is_file():
+            # if a file was explicitly given, we don't care about its extension
+            sources.append(p)
+        elif s == '-':
+            sources.append(Path('-'))
+        else:
+            err(f'invalid path: {s}')
     if len(sources) == 0:
         ctx.exit(0)
     elif len(sources) == 1:
         p = sources[0]
         report = Report()
         try:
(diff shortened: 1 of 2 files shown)


black.py
```diff
@@ -3,13 +3,15 @@
 from asyncio.base_events import BaseEventLoop
 from concurrent.futures import Executor, ProcessPoolExecutor
 from functools import partial
+import io
```
ambv: Unnecessary.

black.py
```diff
@@ -3,13 +3,15 @@
...
 import keyword
 import os
 from pathlib import Path
 import tokenize
 from typing import (
-    Dict, Generic, Iterable, Iterator, List, Optional, Set, Tuple, TypeVar, Union
+    Dict, Generic, Iterable, Iterator, List, Optional, Set, Tuple, Type, TypeVar, Union
```
ambv: Unnecessary.

black.py
```diff
@@ -147,18 +162,28 @@ def format_file_in_place(src: Path, line_length: int, fast: bool) -> bool:
...
+    except NothingChanged:
+        return False
+
+    sys.stdout.write(contents)
+    return True
+
+
 def format_file(
-    src: Path, line_length: int, fast: bool
+    src_buffer: Type[io.TextIOBase], line_length: int, fast: bool
```
ambv: I work on typing, too, so I'm wondering what made you choose `Type[]` here and why you chose `io.TextIOBase` specifically. Both choices are wrong here and I will explain why.

1. `Type[]` is to signify the caller is going to pass a *class*, not an *object*. Here we want a regular object.
2. `io.TextIOBase` is overly abstract, it doesn't have `name` for example.

Instead, you would use `typing.TextIO` to describe what you wanted to achieve. Don't worry if you didn't get this until now, this is very new tech. So I really wonder about your first experiences here. I should also add `mypy` to CI to make stuff like this more apparent :-)

# Practical suggestions

1. Rename `src_buffer` back to `src`. It's an abstract enough name and the function's name suggests we want a file-like object anyway.
2. Change the docstring to say "an open file" instead of "a file" to make this more obvious.
3. Change the type to `TextIO` (import it from `typing` instead of `Type` above).

tests/test_black.py
```diff
@@ -10,12 +11,16 @@
...
 ll = 88
-ff = partial(black.format_file, line_length=ll, fast=True)
 fs = partial(black.format_str, line_length=ll)
 THIS_FILE = Path(__file__)
 THIS_DIR = THIS_FILE.parent
 
 
+def ff(file_):
+    with tokenize.open(file_) as buf:
+        black.format_file(buf, line_length=ll, fast=True)
```
ambv: This is fine. Also, you can just use `file`. The builtin name `file` is deprecated, we only keep it to not break old code.

black.py
```diff
@@ -64,28 +66,38 @@ class CannotSplit(Exception):
 @click.argument(
     'src',
     nargs=-1,
-    type=click.Path(exists=True, file_okay=True, dir_okay=True, readable=True),
+    type=click.Path(
+        exists=True, file_okay=True, dir_okay=True, readable=True, allow_dash=True
+    ),
 )
 @click.pass_context
 def main(ctx: click.Context, line_length: int, fast: bool, src: List[str]) -> None:
```
ambv: I agree this got a bit messy now, let's see suggestions below.

black.py
```diff
@@ -64,28 +66,38 @@ class CannotSplit(Exception):
...
-        p = Path(s)
-        if p.is_dir():
-            sources.extend(gen_python_files_in_dir(p))
-        elif p.is_file():
-            # if a file was explicitly given, we don't care about its extension
-            sources.append(p)
-        else:
-            err(f'invalid path: {s}')
+    if len(src) == 1 and src[0] == '-':
+        sources.append(Path('-'))
```
ambv: Remove this special case. Dedent the for loop as it was before.

black.py
```diff
@@ -64,28 +66,38 @@ class CannotSplit(Exception):
...
+    else:
+        for s in src:
+            p = Path(s)
+            if p.is_dir():
+                sources.extend(gen_python_files_in_dir(p))
+            elif p.is_file():
+                # if a file was explicitly given, we don't care about its extension
+                sources.append(p)
+            elif s == '-':
+                err('Cannot format both stdin and other paths')
```
ambv: *Here* just do `sources.append(Path('-'))`.

black.py
```diff
@@ -64,28 +66,38 @@ class CannotSplit(Exception):
...
+            else:
+                err(f'invalid path: {s}')
     if len(sources) == 0:
         ctx.exit(0)
     elif len(sources) == 1:
         p = sources[0]
         report = Report()
         try:
-            changed = format_file_in_place(p, line_length=line_length, fast=fast)
+            if str(p) == '-':
```
ambv: Since `-` is an unlikely real file name, it would be better to say `if not p.is_file() and str(p) == '-':`

black.py
```diff
@@ -138,7 +150,10 @@ def main(ctx: click.Context, line_length: int, fast: bool, src: List[str]) -> No
 def format_file_in_place(src: Path, line_length: int, fast: bool) -> bool:
     """Format the file and rewrite if changed. Return True if changed."""
     try:
-        contents, encoding = format_file(src, line_length=line_length, fast=fast)
+        with tokenize.open(src) as src_buffer:
```
ambv: Note: with my suggested changes to `main()`, if a user passes more than one file name, the tool will fail with "FileNotFoundError: [Errno 2] No such file or directory: '-'" and I think this is clear enough of an indication that it wasn't treated as stdin by the tool then. Since the report at the end will show other files being properly handled, this should be enough of a nudge to the user that `-` is special.

black.py
treuherz: Ah, I clearly misunderstood the purpose of `Type`. I thught it was to specify an object of a class *or subclass* with the given type. Not running mypy much led to me never finding out I was getting this wrong! :laughing: Will have to read throgh `typing`'s docs a bit more thoroughly.

treuherz:
Thanks for taking the time to go through this so thoroughly! Will take on board your changes and resubmit at some point in the next few days.

Easiest way to test might be to do reassign `sys.stdin` and `sys.stdout` to `StringIO`s, and pipe black.py through it with the same `assertRaises(NothingChanged)` as the other tests do.

treuherz:
While writing the test, I realised I wasn't piping output if no changes were made, which isn't how pipes are supposed to work. Tried doing it in a try/finally, but of course that doesn't work as the variable is only assigned when format_file returns. Current ideas are just replace format_file in the stdin/stdout case with a function that does the same thing but writes regardless (too duplicative) or copying the contents of stdin to another buffer, so that we know what the input was even if NothingChanged is raised. Neither of these seems great. Any better ideas?

ambv:
Yeah, we need to always pipe stdin to stdout. Well, for now we're going to open the file in format_file_in_place() and rename format_file() into format_file_contents() which it now does. That way format_stdin_to_stdout() becomes rather easy to understand and the separation of concerns is symmetrical.

This is now merged. Thanks! ✨ 🍰 ✨

joslarson:
@treuherz @ambv piping support is looking great, but when I run `cat file.py | black -` I'm still getting `reformatted -` as the last line of stdout. I assume we don't want the report generated when piping things around.

black.py
```diff
@@ -97,9 +101,12 @@ def main(
...
-            changed = format_file_in_place(
-                p, line_length=line_length, fast=fast, write_back=not check
-            )
+            if not p.is_file() and str(p) == '-':
+                changed = format_stdin_to_stdout(line_length=line_length, fast=fast)
+            else:
+                changed = format_file_in_place(
+                    p, line_length=line_length, fast=fast, write_back=not check
+                )
             report.done(p, changed)
```
joslarson: Removing the report text from stdout when piping looks as simple as indenting this one line in under the else block.

ambv:
@joslarson, you are mistaking stdout with stderr. Black never outputs anything to stdout (because I foresaw the "-" use case).

ambv:
In other words, if you say `cat file.py | black - 2>/dev/null` it will do what you expected. If you pipe things further, it would work as expected, too.

This is why we don't really need "--quiet" or similar.

joslarson:
@ambv oh, haha... fail. Sorry about that. I somehow thought I had accounted for that, but nope.

PR #36, opened 2018-03-17
Line breaks before logical operators

Autophagy:
This addresses https://github.com/ambv/black/issues/21 where the PEP 8 guidelines state that it's preferred for lines to be broken before a logical operator, rather than afterwards. 

However, flake8 yells at me with ``W503 line break before binary operator``, so :sweat_smile:

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,7 +2,7 @@
 # Keep in sync with setup.cfg which is used for source packages.
 
 [flake8]
-ignore = E266, E501
+ignore = E266, E501, W503
 max-line-length = 80
 max-complexity = 15
 select = B,C,E,F,W,T4,B9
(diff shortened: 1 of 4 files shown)


ambv:
This is amazing. Thanks a ton! 👍 

Since you have a fresh view at this project, can you share what you found surprising, badly commented (incl. docstrings), etc.? We can improve things for future newcomers!

ambv:
Seal of Approval

Autophagy:
@ambv I actually found the docstrings really well written and helpful, and everything else was pretty intuitive (also everything is named really well, which helps!)

I think the only thing, and this is more of a personal preference, but huge single-file python scripts can be a little intimidating! I dont know if splitting it up into logical components would make much sense, but yeah, with one large python file I find it a little difficult to grasp the structure at a glance :slightly_smiling_face:

PR #38, opened 2018-03-17
Add files generated during test to ignores

giuscri:


ambv:
Thanks for your pull request! However, @giuscri, you should rather have those in a global gitignore. Those rules are universal to any project you'll work on. See this for example: https://github.com/ambv/.dot_files/blob/master/gitignore

giuscri:
+1

SethMichaelLarson:
I thought these would actually be standard in gitignore? GitHub's Python gitignore template includes [`__pycache__/`](https://github.com/github/gitignore/blob/master/Python.gitignore#L2) and [`*.egg-info`](https://github.com/github/gitignore/blob/master/Python.gitignore#L23).

PR #50, opened 2018-03-21
Add --no-error flag to allow for 0 exit code when reformatting

MichalPodeszwa:
When using the package with IntelliJ's file watcher the console pops up on command errors.
Currently the package returns a 1 exit code when it reformatted at least one file.

This flag adds the possibility to still get 0 after that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 .coverage
+*.egg-info
+__pycache__
--- a/black.py
+++ b/black.py
@@ -71,6 +71,14 @@ class CannotSplit(Exception):
     is_flag=True,
     help='If --fast given, skip temporary sanity checks. [default: --safe]',
 )
+@click.option(
+    '--no-error',
+    is_flag=True,
+    help=(
+        "Don't return 1 if there were any files changed but no error. "
+        "This is useful when running black in a file watcher in your IDE"
+    ),
+)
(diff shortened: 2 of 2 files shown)


ambv:
I worry about adding many options because it makes the tool appear complicated to use. I'll leave this PR open and think a bit more about how we should cover this case. Funnily enough, originally Black wasn't using exit code 1 for changes but somebody asked for it because some other workflow needs it. So your intuition to add an option is probably right. I just need to make sure we end up with an obvious UI for this.

In the mean time, you can work around your issue with "Show console" -> "Never" in "Advanced Options" of your File Watcher.

eplaut:
@ambv I think most of unix utilities are not returning an error exit code if there job been done properly, only for invalid input. IMHO, this should be the default behavior (not just a flag)

SethMichaelLarson:
Maybe there should only be an error code when running with `--check`? That's the only scenario I can think of you'd want an error code (such as running on CI).

ambv:
I like Seth's idea to only return 1 if --check is used.

ambv:
Thank you, @MichalPodeszwa, for you pull request. Sorry we didn't end up merging it this time. I hope this doesn't discourage you from contributing in the future, and more importantly, that our current approach solves your issue.

Issue #70, opened 2018-03-23
create Sphinx docs and ReadTheDocs hosting


carljm:
We should have Sphinx documentation in the repo, and probably host them on RTD.

carljm:
@willingc - @ambv mentioned you are working on this? Thank you!!

carljm:
Some writing on why Black opposes backslashes that would be good to include in some form: https://github.com/ambv/black/issues/64#issuecomment-375429106

willingc:
Thanks @carljm. I'll likely push a skeleton PR in the hour. I'll include support for both rst and markdown sources so folks can pick their favorites.

ambv:
Alright: http://black.readthedocs.io/en/latest/

Carol, thanks a lot for your help here, this is a great start! ✨ 🍰 ✨

willingc:
Sweet. Thanks @ambv. Now for the fun part - iterating. :sunny:

PR #71, opened 2018-03-23
add sphinx docs skeleton

willingc:
Partially addresses #70.

A first pass at docs.


ambv:
Looks great, let's iterate on it!

PR #75, opened 2018-03-25
Normalize string quotes

### LINKED ISSUE #51: Consider being even more opinionated about quotes
One thing I've come to appreciate from JS land is that even though ' and " are interchangeable, you can set the linters to require one or the other. My Python code is littered with alternating ' and " strings, sometimes even in the same dictionary, like so:

```python
{
    'My key': "has a value",
    "my quote": 'is doubled',
}
```
I don't have an opinion on which is best (and I don't want to), but I wish to hell my code was consistent. There are good reasons to use one or the other (if the string contains other quotes, for example), but in general, I wish they were all the same. Maybe it could adjust any string that is enclosed in `'` characters and does not contain a `"` character to be wrapped with `"` characters (or vice versa):

e.g
`'hello' becomes "hello"`
`'She said "yes"'` doesn't change
`"Don't do that"` never changes.

Should also be able to change all triple quotes to the same character as well unless the string contains the inverse quote.

zsol:
Convert simple double-quoted strings to single-quoted. Convert triple (single) quoted strings to triple (double) quoted. Do not touch any strings that have backslashes or quotes inside the string.

Fixes #51.

### DIFF
--- a/black.py
+++ b/black.py
@@ -47,9 +47,9 @@ LeafID = int
 Priority = int
 Index = int
 LN = Union[Leaf, Node]
-SplitFunc = Callable[['Line', bool], Iterator['Line']]
+SplitFunc = Callable[["Line", bool], Iterator["Line"]]
 out = partial(click.secho, bold=True, err=True)
-err = partial(click.secho, fg='red', err=True)
+err = partial(click.secho, fg="red", err=True)
 
 
 class NothingChanged(UserWarning):
@@ -94,15 +94,15 @@ class FormatOff(FormatError):
 
 @click.command()
 @click.option(
-    '-l',
-    '--line-length',
+    "-l",
+    "--line-length",
     type=int,
     default=DEFAULT_LINE_LENGTH,
-    help='How many character per line to allow.',
+    help="How many character per line to allow.",
     show_default=True,
 )
 @click.option(
-    '--check',
+    "--check",
     is_flag=True,
     help=(
         "Don't write back the files, just return the status.  Return code 0 "
@@ -111,13 +111,13 @@ class FormatOff(FormatError):
     ),
 )
 @click.option(
-    '--fast/--safe',
+    "--fast/--safe",
     is_flag=True,
-    help='If --fast given, skip temporary sanity checks. [default: --safe]',
+    help="If --fast given, skip temporary sanity checks. [default: --safe]",
 )
 @click.version_option(version=__version__)
 @click.argument(
-    'src',
+    "src",
     nargs=-1,
     type=click.Path(
         exists=True, file_okay=True, dir_okay=True, readable=True, allow_dash=True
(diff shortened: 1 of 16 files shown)


carljm:
FWIW, I am strongly opposed to this. It will take code that currently makes consistent use of quotes (eg a list of human-facing UI text strings that are all double quoted because they are likely to contain single-quote “apostrophes”, even though only some of them actually do), and make it needlessly inconsistent (the strings that happen to contain single quotes will remain double-quoted, the other ones will be converted to single quotes.) In addition to making the code uglier, this hurts maintainability; now I have to manually switch back to double quotes if I rewrite one of the strings in a way that adds a single quote. 

I am not at all convinced that changing quoting of strings should even be considered in scope for Black at all. If we really think it is, let’s please think through a more thorough solution before pushing something actively harmful.

zsol:
> now I have to manually switch back to double quotes if I rewrite one of the strings in a way that adds a single quote.

IMO you should just add a `\'` to your string, then black should figure this out and convert the string to double quotes.

This discussion is relevant here: https://github.com/prettier/prettier/issues/973

zsol:
Well I'm glad I'm not going to make things worse :) I expect to look at this again next weekend, stay tuned

carljm:
In terms of the test case I mentioned above, I don't think the full proposal outlined here improves anything. I don't want Black to ever turn this well-formatted and consistent code:

```
messages = {
    MESSAGE.one: "Please click the button.",
    MESSAGE.two: "Please don't look over there.",
    MESSAGE.three: "Stop and smell the roses.",
}
```

into this inconsistent mess:

```
messages = {
    MESSAGE.one: 'Please click the button.',
    MESSAGE.two: "Please don't look over there.",
    MESSAGE.three: 'Stop and smell the roses.',
}
```

because presence or absence of an apostrophe in a given message is an accidental and arbitrary, not inherent, characteristic of the string, and it's highly likely that some string in this list may be changed to include an apostrophe in the future. This is a concrete harm to both consistency and maintainability; the opposite of Black's goal. If a string changes from not having an apostrophe to having one, I either have to change its quoting manually or remember to type an escape, or else I've created a parsing error. In either case, Black has directly made maintenance of this code noticeably worse than it was before; it's not just a matter of aesthetic preference.

This issue would also be resolved if Black standardizes on double quotes instead of single. IMO the prevalence of apostrophes alone far outweighs every other argument for preferring single quotes. The extra keystroke on US keyboards is a non-issue; just type single quotes and let Black convert them.

ambv:
Ha, Carl used my own argument against me here... and so I agree with him. Let's default to double quotes always.

renstrom:
I thought I would mention that we need to consider [PEP 257](https://www.python.org/dev/peps/pep-0257/) (Docstring Convention) as well, specifically this:

>For consistency, always use `"""triple double quotes"""` around docstrings.

zsol:
I pushed a new implementation that tries to minimize escaping while preferring `"` and `"""`. A couple of caveats:
- I punted on reformatting strings nested inside f-strings for now. I think we can add this on later, let's agree on the simple(r?) cases first
- The code doesn't force `"""` for docstrings, but will lean towards them if it causes less escaping. This docstring is considered well-formatted, for example: `'''Docstring "'''`, instead of reformatting it to `"""Docstring \""""`

I tried adding important edge cases to the test case, let me know if I missed anything!

ambv:
This is great. Thanks! ✨ 🍰 ✨ 

We'll iterate on it if anything comes up but looks good. Now for some documentation and README updates.

alanhamlett:
Don't use single quote inside human readable strings, use `’` instead. Problem solved.

PR #77, opened 2018-03-26
Include .out file(s) in the distribution

polyzen:
> FileNotFoundError: [Errno 2] No such file or directory: '/home/user/pkg/build/black/src/black-18.3a4/tests/debug_visitor.out'


ambv:
Good catch. Thanks! ✨ 🍰 ✨

PR #82, opened 2018-03-28
document Black classes, functions, exceptions

willingc:


ambv:
This is great, thank you!

PR #83, opened 2018-03-28
[WIP] use sphinx-click to autodoc cli

willingc:
Opening as a WIP since the PR doesn't really improve on what we have currently. It's an option to autodoc the CLI now or in the future.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1,0 +1,7 @@
+Command Line Interface
+======================
+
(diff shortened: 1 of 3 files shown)


ambv:
Thanks for this suggestion!

I am not sure about this since we need the CLI --help in the README anyway. Since we are copying this automatically from the README now, I think we can pass on this idea.

PR #88, opened 2018-03-29
Fix --check with multiple files

anowlcalledjosh:
Passing multiple files with --check would previously result in the report being printed as if the files had been written to:

```console
$ black --check foo.py 
would reformat foo.py
$ black --check foo.py bar.py 
reformatted foo.py
bar.py already well formatted, good job.
All done! ✨ 🍰 ✨
1 file reformatted, 1 file left unchanged.
```

Now the output is correct when multiple files are checked.

### DIFF
--- a/black.py
+++ b/black.py
@@ -181,7 +181,7 @@ async def schedule_formatting(
     }
     await asyncio.wait(tasks.values())
     cancelled = []
-    report = Report()
+    report = Report(check=not write_back)
     for src, task in tasks.items():
         if not task.done():
             report.failed(src, 'timed out, cancelling')


ambv:
Thanks! ✨ 🍰 ✨

ikatanic:
Hi, are you sure this fixed the issue?

For me it is still broken, and it seems using enums like that doesn't work:
```
ivan@Ivans-MacBook-Pro code $ python3
Python 3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 03:03:55)
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from enum import Enum
>>> class WriteBack(Enum):
...     NO = 0
...     YES = 1
...     DIFF = 2
...
>>> x = WriteBack.NO
>>> not x
False
```

PR #89, opened 2018-03-30
use conda for readthedocs

willingc:
This PR adds 2 yaml files so that RTD will use conda to build the docs environment. 

[Rendered test docs on RTD](http://test-black.readthedocs.io/en/doc-conda/reference/reference_summary.html)

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -2,7 +2,7 @@ name: black_docs
 channels:
   - conda-forge
 dependencies:
-- python==3.6
+- python>=3.6
 - Sphinx==1.7.2
 - pip:
   - recommonmark==0.4.0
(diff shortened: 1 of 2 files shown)


docs/environment.yml
```diff
@@ -0,0 +1,9 @@
+name: black_docs
+channels:
+  - conda-forge
+dependencies:
+- python==3.6
```
ambv: Would `>=3.6` not work? Then we'll get 3.6.4 or 3.6.5 which is preferable :)

ambv:
This is great! Just this one question I have on the environment (see inline comment).

docs/environment.yml
willingc: Sure that would work

PR #94, opened 2018-03-31
Lines now break before all delimiters

### LINKED ISSUE #73: Lines should break before all binary operators
Style of line breaks is not consistent.

```python
(
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa |
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
)

(
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa &
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
)

(
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa -
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
)

(
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    and aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
)

(
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    or aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
)
```

According to [PEP8](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator), lines should break before all binary operators.

Autophagy:
The default behaviour is that now all lines break *before* delimiters,
instead of afterwards. The special cases for this are commas and
behaviour around args.

Resolves #73 

I've also added my full name to the contributors list :wink:

### DIFF
--- a/black.py
+++ b/black.py
@@ -451,6 +451,7 @@ MATH_OPERATORS = {
     token.DOUBLESTAR,
     token.DOUBLESLASH,
 }
+VARARGS = {token.STAR, token.DOUBLESTAR}
 COMPREHENSION_PRIORITY = 20
 COMMA_PRIORITY = 10
 LOGIC_PRIORITY = 5
@@ -492,32 +493,12 @@ class BracketTracker:
             leaf.opening_bracket = opening_bracket
         leaf.bracket_depth = self.depth
         if self.depth == 0:
-            delim = is_delimiter(leaf)
-            if delim:
-                self.delimiters[id(leaf)] = delim
-            elif self.previous is not None:
-                if leaf.type == token.STRING and self.previous.type == token.STRING:
-                    self.delimiters[id(self.previous)] = STRING_PRIORITY
-                elif (
-                    leaf.type == token.NAME
-                    and leaf.value == "for"
-                    and leaf.parent
-                    and leaf.parent.type in {syms.comp_for, syms.old_comp_for}
-                ):
-                    self.delimiters[id(self.previous)] = COMPREHENSION_PRIORITY
-                elif (
-                    leaf.type == token.NAME
-                    and leaf.value == "if"
-                    and leaf.parent
-                    and leaf.parent.type in {syms.comp_if, syms.old_comp_if}
-                ):
-                    self.delimiters[id(self.previous)] = COMPREHENSION_PRIORITY
-                elif (
-                    leaf.type == token.NAME
-                    and leaf.value in LOGIC_OPERATORS
-                    and leaf.parent
-                ):
-                    self.delimiters[id(self.previous)] = LOGIC_PRIORITY
+            after_delim = is_split_after_delimiter(leaf, self.previous)
+            before_delim = is_split_before_delimiter(leaf, self.previous)
+            if after_delim > before_delim:
+                self.delimiters[id(leaf)] = after_delim
+            elif before_delim > after_delim and self.previous is not None:
+                self.delimiters[id(self.previous)] = before_delim
         if leaf.type in OPENING_BRACKETS:
             self.bracket_match[self.depth, BRACKET[leaf.type]] = leaf
             self.depth += 1
@@ -1374,17 +1355,35 @@ def preceding_leaf(node: Optional[LN]) -> Optional[Leaf]:
     return None
 
 
-def is_delimiter(leaf: Leaf) -> int:
-    """Return the priority of the `leaf` delimiter. Return 0 if not delimiter.
+def is_split_after_delimiter(leaf: Leaf, previous: Leaf = None) -> int:
+    """Return the priority of the `leaf` delimiter, given a line break after it.
+
+    The delimiter priorities returned here are from those delimiters that would
+    cause a line break after themselves.
 
     Higher numbers are higher priority.
     """
     if leaf.type == token.COMMA:
         return COMMA_PRIORITY
 
-    if leaf.type in COMPARATORS:
-        return COMPARATOR_PRIORITY
+    if (
+        leaf.type in VARARGS
+        and leaf.parent
+        and leaf.parent.type in {syms.argument, syms.typedargslist}
+    ):
+        return MATH_PRIORITY
 
+    return 0
+
+
+def is_split_before_delimiter(leaf: Leaf, previous: Leaf = None) -> int:
+    """Return the priority of the `leaf` delimiter, given a line before after it.
+
+    The delimiter priorities returned here are from those delimiters that would
+    cause a line break before themselves.
+
+    Higher numbers are higher priority.
+    """
     if (
         leaf.type in MATH_OPERATORS
         and leaf.parent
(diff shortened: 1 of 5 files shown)


Autophagy:
Just quickly fixed the flake8 here https://github.com/ambv/black/pull/94/commits/f5b9e0f946344ef4d70681560c785727577f5660#diff-a75051ebf8e3452bf3d590c186944252L1386 - but I'm not sure why black missed it. Will look into it...

black.py
```diff
@@ -443,11 +443,13 @@ def show(cls, code: str) -> None:
     token.DOUBLESTAR,
     token.DOUBLESLASH,
 }
+KWARGS = {token.STAR, token.DOUBLESTAR}
```
ambv: I'd call it VARARGS since you mean both `*args` and `**kwargs`.

black.py
```diff
@@ -1364,27 +1346,83 @@ def preceding_leaf(node: Optional[LN]) -> Optional[Leaf]:
...
         return COMMA_PRIORITY
 
-    if leaf.type in COMPARATORS:
-        return COMPARATOR_PRIORITY
+    if (
+        leaf.type in KWARGS
+        and leaf.parent
+        and leaf.parent.type in {syms.argument, syms.typedargslist}
+    ):
+        return KWARGS_PRIORITY
```
ambv: I actually don't quite get why you added this. Can you add a unit test that showcases what splitting on stars does?

black.py
```diff
@@ -1364,27 +1346,83 @@ def preceding_leaf(node: Optional[LN]) -> Optional[Leaf]:
     return None
 
 
-def is_delimiter(leaf: Leaf) -> int:
-    """Return the priority of the `leaf` delimiter. Return 0 if not delimiter.
+def is_split_after_delimiter(leaf: Leaf, previous: Leaf = None) -> int:
+    """Returns the priority of the `leaf` delimiter, given that it is a delimiter
```
ambv: nit: Have you noticed how I modified all docstrings to be imperative? "Return" rather than "Returns".

black.py
```diff
@@ -1364,27 +1346,83 @@ def preceding_leaf(node: Optional[LN]) -> Optional[Leaf]:
...
+        and leaf.parent
+        and leaf.parent.type in {syms.argument, syms.typedargslist}
+    ):
+        return KWARGS_PRIORITY
+
+    return 0
+
 
+def is_split_before_delimiter(leaf: Leaf, previous: Leaf = None) -> int:
+    """Returns the priority of the `leaf` delimiter, given that it is a delimiter
```
ambv: nit: Ditto about "Return" vs. "Returns".

Also, prefer PEP 257-style "single line description followed by a bigger body" docstrings. I know I don't *always* follow this myself but I definitely try.

black.py
```diff
@@ -1364,27 +1346,83 @@ def preceding_leaf(node: Optional[LN]) -> Optional[Leaf]:
...
+        return COMPREHENSION_PRIORITY
+
+    if (leaf.type == token.NAME and leaf.value in LOGIC_OPERATORS and leaf.parent):
+        return LOGIC_PRIORITY
+
     return 0
 
 
+def is_delimiter(leaf: Leaf, previous: Leaf = None) -> int:
+    """Returns the priority of the `leaf` delimiter. Returns 0 if not delimiter.
```
ambv: nit: Ditto about "Return" vs. "Returns".

black.py
```diff
@@ -1364,27 +1346,83 @@ def preceding_leaf(node: Optional[LN]) -> Optional[Leaf]:
...
+
+    if (
+        leaf.type == token.NAME
+        and leaf.value == 'if'
+        and leaf.parent
+        and leaf.parent.type in {syms.comp_if, syms.old_comp_if}
+    ):
+        return COMPREHENSION_PRIORITY
+
+    if (leaf.type == token.NAME and leaf.value in LOGIC_OPERATORS and leaf.parent):
```
ambv: nit: parens are spurious here. We will teach Black to do this transformation on its own 😏

Autophagy:
@ambv Thank you for the review ! I've pushed a fixup to address your comments.

I was wondering: how do you feel with the use of ``previous`` in the function definition here: https://github.com/ambv/black/pull/94/files#diff-a75051ebf8e3452bf3d590c186944252R1348

I added it to keep the function signature consistent with the other delimiter check, but it's never actually used (and I dont think it ever will be, considering these delimiters dont need to know about the previous leaf).

What would your preference be? Consistent signature but an unused parameter, or inconsistent signatures?

Autophagy:
Just squashed my changes + rebased from master.

ambv:
Oh, since you're here, I still don't quite get why you need the VARARGS check in `is_split_after_delimiter()`. Can you give me an example of what would go wrong if we removed this check?

Autophagy:
Not having it was causing some weird line break issues with * and ** in arguments that I couldn't totally figure out why they were happening. For example, without the VARARGS check:

```
def function_signature_stress_test(
    number: int,
    no_annotation=None,
    text: str = "default",
    *,
    debug: bool = False,
    **kwargs,
) -> str:
    return text[number:-1]

call(this_is_a_very_long_variable_which_will_force_a_delimiter_split, arg, another, kwarg='hey', **kwargs)
```

Would produce:
```
def function_signature_stress_test(
    number: int,
    no_annotation=None,
    text: str = "default", *,
    debug: bool = False, **kwargs,
) -> str:
    return text[number:-1]


call(
    this_is_a_very_long_variable_which_will_force_a_delimiter_split,
    arg,
    another,
    kwarg="hey", **kwargs,
) 
```

Instead of the desired result of:
```
def function_signature_stress_test(
    number: int,
    no_annotation=None,
    text: str = "default",
    *,
    debug: bool = False,
    **kwargs,
) -> str:
    return text[number:-1]


call(
    this_is_a_very_long_variable_which_will_force_a_delimiter_split,
    arg,
    another,
    kwarg="hey",
    **kwargs,
)
```

ambv:
Alright, makes sense. I'll merge this and will look into why what you're describing is happening.

ambv:
Thanks! ✨ 🍰 ✨

ambv:
Tomorrow's release is going to be the best ever.

Autophagy:
:tada: Looking forward to it!

PR #97, opened 2018-04-02
Use IntEnum instead of Enum

dbarrosop:
The code is failing to detect properly that we are running in check mode because the `Enum` class is not being used properly. It's defined as:

```
class WriteBack(Enum):
    NO = 0
    YES = 1
    DIFF = 2
```

And then in the code you can see this:

```
    report = Report(check=not write_back)
```

However, that code always yields `False`:

```
>>> class WriteBack(Enum):
...     NO = 0
...     YES = 1
...     DIFF = 2
...
>>> not WriteBack.NO
False
>>> not WriteBack.YES
False
>>> not WriteBack.DIFF
False
```

This issue basically causes `black --check .` to always return 0.

Using `IntEnum` instead causes the logic to work as expected:

```
>>> class WriteBack(IntEnum):
...     NO = 0
...     YES = 1
...     DIFF = 2
...
>>> not WriteBack.NO
True
>>> not WriteBack.YES
False
>>> not WriteBack.DIFF
False
```

### DIFF
--- a/black.py
+++ b/black.py
@@ -3,7 +3,7 @@
 import asyncio
 from asyncio.base_events import BaseEventLoop
 from concurrent.futures import Executor, ProcessPoolExecutor
-from enum import Enum
+from enum import IntEnum
 from functools import partial, wraps
 import keyword
 import logging
@@ -95,7 +95,7 @@ class FormatOff(FormatError):
     """Found a comment like `# fmt: off` in the file."""
 
 
-class WriteBack(Enum):
+class WriteBack(IntEnum):
     NO = 0
     YES = 1
     DIFF = 2


ambv:
Good catch! Thank you! The bug is because `write_back` was originally a `bool`. However, instead of using `IntEnum` (which is just a broken `Enum`), I'd rather fix the logic problem.

Why are IntEnums bad? Look no further than [enum documentation](https://docs.python.org/3/library/enum.html#flag) which says:
> **Note:** For the majority of new code, Enum and Flag are strongly recommended, since IntEnum and IntFlag break some semantic promises of an enumeration (by being comparable to integers, and thus by transitivity to other unrelated enumerations). IntEnum and IntFlag should be used only in cases where Enum and Flag will not do; for example, when integer constants are replaced with enumerations, or for interoperability with other systems.

So, I'm going to close this but feel free to open a new one that fixes the `Report()` creation in `schedule_formatting()`. Bonus points for unit tests of `schedule_formatting()` which is one of the few places not covered by tests yet ;-)

PR #101, opened 2018-04-04
Fix --check for multiple files

ikatanic:
Report would always be created with `check=False` in case of multiple files. Now we pass the correct flag value instead of guessing it.

### DIFF
--- a/black.py
+++ b/black.py
@@ -205,7 +205,7 @@ def main(
         try:
             return_code = loop.run_until_complete(
                 schedule_formatting(
-                    sources, line_length, write_back, check, fast, quiet, loop, executor
+                    sources, line_length, write_back, fast, quiet, loop, executor
                 )
             )
         finally:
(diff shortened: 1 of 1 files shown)


black.py
```diff
@@ -217,6 +217,7 @@ def main(
     sources: List[Path],
     line_length: int,
     write_back: WriteBack,
+    check: bool,
```
ambv: This is unnecessary.

black.py
```diff
@@ -246,7 +247,7 @@ def main(
     loop.add_signal_handler(signal.SIGTERM, cancel, _task_values)
     await asyncio.wait(tasks.values())
     cancelled = []
-    report = Report(check=not write_back, quiet=quiet)
+    report = Report(check=check, quiet=quiet)
```
ambv: Just write `check=write_back is WriteBack.NO`

ikatanic: Alright, I'll do it, I assume you mean `check=write_back is not WriteBack.NO`.

In my opinion, it is a little hacky and error-prone. I wouldn't be surprised if it breaks again in the future.

ambv:
Thanks! ✨ 🍰 ✨

PR #104, opened 2018-04-04
Add integration for pre-commit.com

### LINKED ISSUE #103: Adapter for pre-commit
I'm not sure if its doable today (given explicit Python 3.6 requirement), but would be great to provide out of the box support for the pre-commit framework: https://pre-commit.com/

ambv:
Fixes #103


ambv:
@asottile, does this look sensible?

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,6 @@
+-   id: black
+    name: black
+    entry: black
+    language: python
+    language_version: python3.6
+    files: \.(py|gyp|gypi|tac|wsgi)$
```
asottile: Note that [`types`](https://pre-commit.com/#filtering-files-with-types) was added recently and offers a more flexible way to do this

I think what would work best here is:

```yaml
    types: [python]
```

Note that this'll also allow it to apply to extensionless files which have the proper `python` shebang

ambv: At the moment "python" only matches .py though:
https://github.com/chriskuehl/identify/blob/master/identify/extensions.py#L86

asottile: Indeed, I'll work with @chriskuehl to get a few more extensions added there 👍

asottile: but yeah, don't block on that -- go ahead and ship this as is and we can iterate in `identify` separately

asottile: Sweet, thanks for the list :D

ambv: Last question: why is `[python]` not enough? I thought it would implicitly assume it's a text file (per identify)?

asottile: Oh yeah [python] would be enough. I was misremembering from  an earlier implementation where symlinks to files would be classified as the thing they point to. This isn't the case so [python] should be good!

PR #107, opened 2018-04-04
Add a description for `black`

asottile:
Bikeshedding welcome, this is the string appears on the hooks page

See also: https://github.com/pre-commit/pre-commit.github.io/pull/160


ambv:
Thanks for pre-commit, as you see we are using it now ourselves!

asottile:
Awesome!  Let us know if you run into anything strange -- always happy to help :)

PR #115, opened 2018-04-06
Add support for all valid string literals

### LINKED ISSUE #3: Cannot parse `rf""`

zsol:
Fixes #3. This was the cause for black not being able to support `rf` strings. Now this forked version of `lib2to3` only parses valid string literals as defined by https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

### DIFF
--- a/blib2to3/pgen2/tokenize.py
+++ b/blib2to3/pgen2/tokenize.py
@@ -48,6 +48,8 @@ except NameError:
 def group(*choices): return '(' + '|'.join(choices) + ')'
 def any(*choices): return group(*choices) + '*'
 def maybe(*choices): return group(*choices) + '?'
+def combinations(*l):
+    return [(x + y).strip() for x in l for y in l + (" ",) if x.casefold() != y.casefold()]
 
 Whitespace = r'[ \f\t]*'
 Comment = r'#[^\r\n]*'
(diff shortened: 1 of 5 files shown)


blib2to3/pgen2/tokenize.py
```diff
@@ -107,59 +109,27 @@ def maybe(*choices): return group(*choices) + '?'
 pseudoprog = re.compile(PseudoToken, re.UNICODE)
 single3prog = re.compile(Single3)
 double3prog = re.compile(Double3)
+
+_strprefixes = set(combinations('r', 'R', 'f', 'F') +
+                   combinations('r', 'R', 'b', 'B') +
+                   ['u', 'U'])
```
ambv: You're missing `'uR', 'UR', 'Ur', 'ur'` which is valid on Python 2.

zsol: uR right :rofl: I was wondering why tokenize used to accept this in the regex above.

ambv: Funnily enough, since we are still supporting `u""` in Python 3, not supporting `ur""` is likely just an omission of whoever brought `u""` back :-)

blib2to3/pgen2/tokenize.py
```diff
@@ -48,8 +48,10 @@
 def group(*choices): return '(' + '|'.join(choices) + ')'
 def any(*choices): return group(*choices) + '*'
 def maybe(*choices): return group(*choices) + '?'
-def combinations(*l):
-    return [(x + y).strip() for x in l for y in l + (" ",) if x.casefold() != y.casefold()]
+def _combinations(*l):
+    return set(
+        x + y for x in l for y in l + ("",) if x.casefold() != y.casefold()
```
zsol: That needs to be a `(" ",)` in the middle

zsol: oh nevermind

ambv:
Thanks! ✨ 🍰 ✨

PR #128, opened 2018-04-12
Handle unnecessarily escaped strings

### LINKED ISSUE #120: Different output between first and second pass if original string escapes quotes unnecessarily
Example string: `'he said \"hello\"'`
First pass formats it as `"he said \"hello\""`
Second pass formats it as `'he said "hello"'`

zsol:
Remove extraneous backslashes from the original string. This stabilizes formatting output, so fixes #120.

### DIFF
--- a/black.py
+++ b/black.py
@@ -2380,10 +2380,11 @@ def shutdown(loop: BaseEventLoop) -> None:
 
 
 def sub_twice(regex: Pattern[str], replacement: str, original: str) -> str:
-    """Replaces `regex` with `replacement` twice on `original`.
+    """Replace `regex` with `replacement` twice on `original`.
 
     This is used by string normalization to perform replaces on
-    overlapping matches."""
+    overlapping matches.
+    """
     return regex.sub(replacement, regex.sub(replacement, original))
 
 
--- a/black.py
+++ b/black.py
@@ -24,6 +24,7 @@ from typing import (
     Iterator,
     List,
     Optional,
+    Pattern,
     Set,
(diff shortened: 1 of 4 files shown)


black.py
```diff
@@ -1958,6 +1959,12 @@ def normalize_prefix(leaf: Leaf, *, inside_brackets: bool) -> None:
     leaf.prefix = ""
 
 
+def _sub_twice(regex: Pattern[str], replacement: str, string: str) -> str:
```
ambv: 1. Utility functions belong at the bottom of the file.
2. Add a reference to this function in docs/reference/functions.rst.
3. Note that I don't use any underscored names for functions in black.py. It's an application so the concept of private/public is not necessary. `sub_twice` reads a tiny bit nicer.
4. `string` would be more obvious if named `original`. We see it's a string from the type hint right next to it.
5. In the docstring, add a single leading line explaining what it does. I get that this is spurious for a single line function but it's going to be visible in the docs. Something simple like "Replace instances of `regex` in `original` with `replacement`, twice."

tests/string_quotes.py
```diff
@@ -38,6 +38,8 @@
 "x = '''; y = \"\"\"\""
 "x = ''''; y = \"\"\"\"\""
 "x = '' ''; y = \"\"\"\"\""
+'unnecessary \"\"escaping'
```
ambv: Could you add some in this vein for the opposite case?
`"unnecessary \'\'escaping"`

PR #131, opened 2018-04-14
Add install instructions for Vim plugin

### LINKED ISSUE #129: Plugin installation via Vim plugin managers (Vundle, etc.)
It would be nice to have `black` vim plugin installed via popular vim package managers e.g. Vundle.

anowlcalledjosh:
As far as I can see, there are basically two options here:

- put the Vim plugin in the root of the repo (in `plugin/black.vim` rather than `vim/plugin/black.vim`)

  That makes installing it really easy:

  ```
  Plug 'ambv/black'
  ```

  but means you have a Vim-specific `plugin` directory in the root of the repo.

- leave the Vim plugin where it is, in the `vim` directory

  This makes installation marginally less easy (and I also don't know how this method works with e.g. Pathogen or `packadd`):

  ```
  Plug 'ambv/black', {'rtp': 'vim'}
  ```

  but keeps the repo "cleaner"(?)

I've done the latter, but I'm happy to defer to anybody's preference.

(I guess you could also have a separate `black.vim` repo that has the normal Vim plugin structure; in some ways that's the best of both worlds, but I can also see why you might want to keep all Black-related code in one repo.)

Fixes #129.


LuRsT:
:+1: to the latter option, creating a separate black.vim repo just for the plugin.

ashwinvis:
How about:
```vim
Plug 'ambv/black', {'for': 'python', 'rtp': 'vim'}
```

ambv:
In #129 @JackLeo also suggests making a separate repo for the Vim plugin. This is what I'll do then.

laixintao:
@ambv I think "put the Vim plugin in the root of the repo" is a good idea, put vim plugin to a separate repo is confusing.

Use `Plug 'ambv/black', {'rtp': 'vim'}` is even better!

JackLeo:
If it would be `'ambv/black.vim'` and linked in docs, there would be no confusion. Equaly like many other options, flake8 does not have vim plugin in it self, but there are plenty available ones.

laixintao:
@JackLeo Make sense, then people can develop varies plugins not only the official one.

ambv:
Can you move vim/plugin to plugin/ and update the README accordingly?

anowlcalledjosh:
Done.

ambv:
Thanks! ✨ 🍰 ✨

PR #138, opened 2018-04-17
Parse complex expressions in parameters after * and **

### LINKED ISSUE #2: Cannot parse `f(*[] or [])`

zsol:
Allow `test`s after * and **, instead of only `expr`s in lib2to3's grammar. Fixes #2


black.py
```diff
@@ -1255,7 +1255,7 @@ def whitespace(leaf: Leaf) -> str:  # noqa C901
                     # that, too.
                     return prevp.prefix
 
-        elif prevp.type == token.DOUBLESTAR:
+        elif prevp.type in {token.DOUBLESTAR, token.STAR}:
```
ambv: `prevp.type in STARS`

blib2to3/Grammar.txt
```diff
@@ -138,8 +138,8 @@ arglist: argument (',' argument)* [',']
 # that precede iterable unpackings are blocked; etc.
 argument: ( test [comp_for] |
             test '=' test |
-	    '**' expr |
-	    star_expr )
+	    '**' test |
+            '*' test )
```
ambv: Yup, that's what CPython does.

zsol: where does this happen in CPython?

ambv: https://github.com/python/cpython/blob/master/Grammar/Grammar#L136-L139

black.py
```diff
@@ -1364,7 +1364,7 @@ def whitespace(leaf: Leaf) -> str:  # noqa C901
             if not prevp or prevp.type == token.LPAR:
                 return NO
 
-        elif prev.type == token.EQUAL or prev.type == token.DOUBLESTAR:
+        elif prev.type in {token.EQUAL} | STARS:
```
ambv: You could have used your newly found power:
```
elif prev.type in {*STARS, token.EQUAL}:
```
😉

ambv:
In the future merge by squashing. Merges look bad in history.

PR #143, opened 2018-04-18
Fix parsing of async for in comprehensions

### LINKED ISSUE #125: Can't parse `return (i * 2 async for i in arange(n))`
Title says all.

zsol:
This basically merges in changes to lib2to3 from
https://github.com/python/cpython/commit/ac317700ce7439e38a8b420218d9a5035bba92ed

Fixes #125


ambv:
This change by Jelle was invalid and was reverted by him later.

zsol:
Interesting. Thanks for the context, I'll dig some more later

Issue #149, opened 2018-04-20
Make check and diff not mutually exclusive


tomelliff:
Operating system: Ubuntu 16.04
Python version:
```
$ python --version
Python 3.6.3
```
Black version:
```
$ black --version
black, version 18.4a2
```
---

Running `black --check --diff` exits with:
> Error: Options --check and --diff are mutually exclusive
In CI it's nice to have auto-formatters explain exactly what it wants to change so running `--check` and `--diff` would be useful. Sadly `--diff` alone exits 0 if it needs to change things so I can't just use that.

ambv:
What's the use case for `--check --diff`?

ambv:
Oh, I just read the description in the blockquote. Sure, we can do this.

tomelliff:
Ha @csurfer, looks like you beat me to it and also got around to adding tests too. I'll leave it to you if you're happy to close this? Thanks for working on it by the way :)

csurfer:
@tomelliff : Ya its closed now. :)

PR #155, opened 2018-04-21
[#149] Make check and diff not mutually exclusive

csurfer:
This change adds the ability to use both `--check` and `--diff` together to see the proposed changes and adds a test to verify that these options can indeed be called together.

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -596,13 +596,22 @@ class BlackTestCase(unittest.TestCase):
             black.write_cache({}, [])
 
     def test_check_diff_use_together(self) -> None:
-        with cache_dir() as workspace:
-            src = (workspace / "test.py").resolve()
-            with src.open("w") as fobj:
-                fobj.write("print('hello_bad_indented')")
-            result = CliRunner().invoke(black.main, [str(src), "--diff", "--check"])
+        with cache_dir():
+            # Files which will be reformatted.
+            src1 = (THIS_DIR / "string_quotes.py").resolve()
+            result = CliRunner().invoke(black.main, [str(src1), "--diff", "--check"])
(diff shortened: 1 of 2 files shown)


tests/test_black.py
```diff
@@ -595,6 +595,15 @@ def test_write_cache_write_fail(self) -> None:
             mock.side_effect = OSError
             black.write_cache({}, [])
 
+    def test_check_diff_use_together(self) -> None:
```
ambv: Instead of creating a temp file (which is slow), just run -the command on two existing files. I suggest`tests/string_quotes.py` and `tests/composition.py`. The former should return exit code 1, the latter should return exit code 0. The former should produce a diff, the latter should not produce a diff.

ambv:
Thanks for your pull request!

csurfer:
No luck. Let me know if I need to do something to make it easier.

ambv:
Oh man, this PR now includes way more changes than you intended. You will probably want to create a new pull request, which will probably be faster for you than recovering from this mess. This time create a branch for your changes which will make rebasing easier and less error-prone.

To not lose your changes, pipe `git show YOUR_REVISION >patch.diff` of your revision to a file. Then on a clean branch you can reapply the diff with `patch -p1 <patch.diff`.

CMCDragonkai:
Is there a MR for this? I'd like to use `--check` and `--diff` together in CI to show the team what they need to change to pass MR review?

hugovk:
They're already available in the released version. 

https://github.com/psf/black/blob/master/README.md#installation-and-usage

hugovk:
They're already available in the released version. 

https://github.com/psf/black/blob/master/README.md#installation-and-usage

PR #161, opened 2018-04-23
[#149] Make check and diff not mutually exclusive

csurfer:


csurfer:
@ambv : Here you go. Raised request from a separate branch. For some reason my local repo is screwed up.

PR #163, opened 2018-04-24
Make cache work with non-default line lenghts

### LINKED ISSUE #158: Line length changes are not taken into account for cached files
After running `black file.py`, `black file.py -l 80` doesn't even try reformatting `file.py` if it hasn't changed since the previous invocation.

ojii:
Fixes #158

Stores the cache per-line-length-setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,9 +448,9 @@ location of the file depends on the black version and the system on which black
 is run. The file is non-portable. The standard location on common operating systems
 is:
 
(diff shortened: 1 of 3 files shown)


ambv:
From AppVeyor:

C:\Python36\python.exe -m mypy black.py tests/test_black.py
tests\test_black.py:631:5: error: Function is missing a type annotation

Install pre-commit on your checkout, Jonas :-)

ojii:
> Install pre-commit on your checkout, Jonas :-)

Done and sorry about that!

PR #165, opened 2018-04-24
Support parsing of async generators in non-async functions

### LINKED ISSUE #125: Can't parse `return (i * 2 async for i in arange(n))`

zsol:
This is a new syntax added in python3.7, so black can't verify that reformatting will not change the ast unless black itself is run with 3.7. We'll need to change the error message black gives in this case. @ambv any ideas?

Fixes #125.

### DIFF
--- a/blib2to3/pgen2/tokenize.py
+++ b/blib2to3/pgen2/tokenize.py
@@ -516,13 +516,14 @@ def generate_tokens(readline):
                         stashed = tok
                         continue
 
-                    if token == 'def':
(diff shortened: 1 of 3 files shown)


zsol:
Yeah I had some 8 hours to dive deeper into how this works in cpython itself, and it's the same thing. Await expressions still raise syntax error with this PR. So I think for the purposes of black this is sufficient

ambv:
Let's do it.

PR #178, opened 2018-04-29
Format subscriptions in a PEP-8 compliant way

### LINKED ISSUE #157: Slicing is not formatted according to PEP 8
I always found what other formatters are doing here wrong and currently Black is also doing the wrong thing.

### Needs changing

The explicit PEP 8 recommendation is:

**YES**:
```py3
ham[lower + offset : upper + offset]
ham[: upper_fn(x) : step_fn(x)], ham[:: step_fn(x)]
```

~NO~:
```
ham[lower + offset:upper + offset]
ham[:upper_fn(x):step_fn(x)], ham[::step_fn(x)]
```

Essentially, we should treat `:` as an "operator" inside slices when any operand of the slice includes punctuation (dots, brackets, and other math, binary or logic operators).

More examples:
**YES**
```py3
ham[lower * unit + offset * unit : upper * unit + offset * unit]
```

~NO~:
```
ham[lower*unit+offset*unit : upper*unit+offset*unit]
```

Note that PEP 8 allows for the following that [we cannot properly enforce](https://github.com/ambv/black/issues/148):
```
ham[lower+offset : upper+offset]
```

Black won't do that because "lower", "upper", or "offset" can be arbitrary complexity and then operator hugging isn't improving readability anymore.

### Already well formatted
Black is already formatting the other PEP 8 recommendation well:

**YES**:
```py3
ham[1:9], ham[1:9:3], ham[:9:3], ham[1::3], ham[1:9:]
ham[lower:upper], ham[lower:upper:], ham[lower::step]
```

~NO~:
```
ham[1: 9], ham[1 :9], ham[1:9 :3]
ham[lower : : upper]
ham[ : upper]
```

(edited: a previous version of this issue suggested hugging operators but I since realized this cannot be performed consistently)

zsol:
Fixes #157


zsol:
Looks like we'll need to disable `E203`

ambv:
E203 is spaces before commas. Where do we need to do that?

zsol:
There's an E203 on this line: https://github.com/ambv/black/blob/1c8a6200c03e8452dde91aeda12f0c6f0738f8dd/black.py#L2104

zsol:
this seems relevant https://github.com/PyCQA/pycodestyle/issues/373

ambv:
Welp, OK, then the rule is stupid. We need to add it to docs as one of the stupid rules (next to W503).

README.md
```diff
@@ -302,6 +302,19 @@ This behaviour may raise ``W503 line break before binary operator`` warnings in
 style guide enforcement tools like Flake8. Since ``W503`` is not PEP 8 compliant,
 you should tell Flake8 to ignore these warnings.
 
+### Whiespace In Subscripts and ``:``
+
+PEP 8 [recommends](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements)
+to treat ``:`` in slices as a binary operator with the lowest priority, and to
+leave an equal amount of space on either side, except if a parameter is omitted
+(e.g. ``ham[1 + 1:]``). It also states that for extended slices, both ``:``
```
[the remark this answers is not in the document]
zsol: Yes, this is a typo. Well spotted :)

black.py
```diff
@@ -1349,7 +1388,7 @@ def whitespace(leaf: Leaf) -> str:  # noqa C901
 
         elif prevp.type == token.COLON:
             if prevp.parent and prevp.parent.type in {syms.subscript, syms.sliceop}:
-                return NO
+                return SPACE if complex_subscript else NO
```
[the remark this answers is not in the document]
zsol: This needs to return `SPACE` because otherwise control would flow to this block of `if`s https://github.com/ambv/black/blob/8f6b5048fcb71bc59399fd7b5511d6e229033b4c/black.py#L1414

Unless you prefer to handle this case down there too

black.py
```diff
@@ -698,6 +715,13 @@ def maybe_decrement_after_lambda_arguments(self, leaf: Leaf) -> bool:
 
         return False
 
+    def get_open_lsqb(self, leaf: Leaf) -> Optional[Leaf]:
+        """Returns the most recent opening square bracket at `leaf` (if any)."""
```
ambv: 1. Adding a `leaf` argument to `get_open_lsqb()` creates a rather bizarre signature. It suggests that `leaf`'s position on the line has something to do with the result but it doesn't. It's simply a special case if `leaf` itself is a LSQB and is not otherwise used.  I think that path should be handled directly in `is_complex_subscript()`.

2. nit: docstrings should be in imperative: "Return" rather than "Returns".

black.py
```diff
@@ -726,7 +750,9 @@ def append(self, leaf: Leaf, preformatted: bool = False) -> None:
         if self.leaves and not preformatted:
             # Note: at this point leaf.prefix should be empty except for
             # imports, for which we only preserve newlines.
-            leaf.prefix += whitespace(leaf)
+            leaf.prefix += whitespace(
```
ambv: Yeah, much better.

black.py
```diff
@@ -920,6 +946,22 @@ def remove_trailing_comma(self) -> None:
                 self.comments[i] = (comma_index - 1, comment)
         self.leaves.pop()
 
+    def is_complex_subscript(self, leaf: Leaf) -> bool:
+        """Returns True iff `leaf` is part of a slice with non-trivial exprs."""
```
ambv: Ditto about imperative.

black.py
```diff
@@ -1328,7 +1377,14 @@ def whitespace(leaf: Leaf) -> str:  # noqa C901
...
 
         if t == token.COLON:
-            return SPACE if prevp.type == token.COMMA else NO
+            if prevp.type == token.COLON:
+                return NO
+
+            elif prevp.type != token.COMMA and not complex_subscript:
+                return NO
+
+            else:
```
ambv: Much better!

nit: `else` is unnecessary

PR #180, opened 2018-04-29
Removing empty parentheses after class name

skapil:
Hi @ambv,
This is WIP.
would like to get early feedback on this as this is my first PR request.
This is working but 2 tests are failing, I am working on them to fix(if you have any suggestion to pass, would be appreciated). I am also in working on adding new tests after fixing the failed tests.

### DIFF
--- a/black.py
+++ b/black.py
@@ -744,6 +744,9 @@ class Line:
         if not has_value:
             return
 
+        if token.COLON == leaf.type and self.is_class_parenth_empty:
+            del self.leaves[-2:]
+
         if self.leaves and not preformatted:
             # Note: at this point leaf.prefix should be empty except for
             # imports, for which we only preserve newlines.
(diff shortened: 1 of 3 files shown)


ambv:
Nice! Thanks for working on this. ✨ 🍰 ✨  

High-level comments:
- LineGenerator doesn't need `leaves` and also shouldn't modify the line directly, that breaks the line's BracketTracker;
- your transformation fits `Line.append()` best:
   - before you append the leaf, check if it's a token.COLON
   - if it is, check whether:
      - you are a class definition; and
      - the current number of leaves is 4; and
      - the previous two leaves are token.LPAR, token.RPAR
   - if so, pop() the two last leaves

That will ensure you are only deleting the right colons, and there is always the right amount of leaves so we will have no index errors, and that BracketTracker is happy.

skapil:
@ambv please have a look 👀

ambv:
Alright, this looks good to go!

Thanks! ✨ 🍰 ✨

PR #189, opened 2018-05-04
Remove unnecessary shebang lines

tiran:
Since black.py is not marked as executable, the shebang in black.py serves
no purpose. black should be invoked through its entry point any way.

token.py is an internal module without a __name__ == '__main__' block or
other executable code. It contains just list of constants and small
helper functions.

### DIFF
--- a/black.py
+++ b/black.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 import asyncio
 import pickle
 from asyncio.base_events import BaseEventLoop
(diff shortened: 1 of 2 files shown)


tiran:
Full disclosure:

I like to use black in a couple of my Fedora / Red Hat projects. To make it easier consumable for my co-workers, I'm currently packaging black for Fedora. The shebangs and executable bit are in violation of Fedora's packaging rules for Python packages. This PR would allow me to drop a couple of lines from https://cheimes.fedorapeople.org/python-black/python-black.spec.

ambv:
FINE.

(Thanks! ✨ 🍰 ✨ )

PR #190, opened 2018-05-05
Output something when no files are reformatted

tiran:
Just executing ``black`` without any argument does not print any message
to stdout or stderr. It's rather confusing, because the user doesn't
know what happened.

In ``len(sources) == 0`` case, black now prints ``No files
reformatted``.


tiran:
Is the message ok with you or would you prefer something like ``Nothing to do!``? Let the bike shedding begin!

tiran:
YOLO :)

PR #191, opened 2018-05-05
Add more files/directories to .gitignore

tiran:
Ignore .tox, black.egg-info and __pycache__ directories.


tiran:
I don't have common files in my global ignore, because I don't expect users to have a sane global configuration. Your argument is valid, too.

Issue #198, opened 2018-05-09
Standardize case for string prefixes


JelleZijlstra:
In implementing a fix for #188 it occurred to me that black could also standardize the case of string literal prefixes (presumably to lowercase). For example, change `R"string"` to `r"string"`.

@ambv is this something you'd be open to doing?

ambv:
Yes, that's an obvious omission from our previous standardization. Fortunately way fewer people actually use upper-case string prefixes :)

cc @zsol

JelleZijlstra:
Sounds good, I can implement this in #199.

PR #199, opened 2018-05-09
Remove u prefix if unicode_literals is present

### LINKED ISSUE #188: Don't use u"..." in Python 2.7 when there's `from __future__ import unicode_literals`
Since all literals will be unicode in this case, it should be fine to just use normal double quoted strings rather than prefix them with `u`.

Operating system: CentOS7
Python version: 3.6
Black version: 18.4a4

JelleZijlstra:
Fixes #188, #198.

### DIFF
--- a/black.py
+++ b/black.py
@@ -2357,6 +2357,44 @@ def is_python36(node: Node) -> bool:
     return False
 
 
+def get_future_imports(node: Node) -> Set[str]:
+    """Returns a set of __future__ imports in the file."""
+    imports = set()
+    for child in node.children:
+        if child.type != syms.simple_stmt:
+            break
+        first_child = child.children[0]
+        if isinstance(first_child, Leaf):
+            # Continue looking if we see a docstring; otherwise stop.
+            if (
+                len(child.children) == 2
+                and first_child.type == token.STRING
+                and isinstance(child.children[1], Leaf)
+                and child.children[1].type == token.NEWLINE
+            ):
+                continue
+            else:
+                break
+        elif isinstance(first_child, Node) and first_child.type == syms.import_from:
+            module_name = first_child.children[1]
+            assert isinstance(module_name, Leaf)
+            if module_name.value == "__future__":
+                for import_from_child in first_child.children[3:]:
+                    if isinstance(import_from_child, Leaf):
+                        if import_from_child.type == token.NAME:
+                            imports.add(import_from_child.value)
+                    elif isinstance(import_from_child, Node):
+                        assert import_from_child.type == syms.import_as_names
+                        for leaf in import_from_child.children:
+                            if isinstance(leaf, Leaf) and leaf.type == token.NAME:
+                                imports.add(leaf.value)
+            else:
+                break
+        else:
+            break
+    return imports
+
+
 PYTHON_EXTENSIONS = {".py"}
 BLACKLISTED_DIRECTORIES = {
     "build", "buck-out", "dist", "_build", ".git", ".hg", ".mypy_cache", ".tox", ".venv"
--- a/black.py
+++ b/black.py
@@ -1207,8 +1207,7 @@ class LineGenerator(Visitor[Line]):
             else:
                 normalize_prefix(node, inside_brackets=any_open_brackets)
                 if node.type == token.STRING:
-                    if self.remove_u:
-                        normalize_string_prefix(node)
+                    normalize_string_prefix(node, remove_u=self.remove_u)
                     normalize_string_quotes(node)
                 if node.type not in WHITESPACE:
                     self.current_line.append(node)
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
The CI failure looks like a flake (it got stuck for whatever reason).

black.py
```diff
@@ -409,7 +409,8 @@ def format_str(src_contents: str, line_length: int) -> FileContent:
     """
     src_node = lib2to3_parse(src_contents)
     dst_contents = ""
-    lines = LineGenerator()
+    future_imports = get_future_imports(src_node)
```
ambv: We should rather move `py36 = ...` here. If the file is 3.6+ we don't even have to look at the future imports then.

black.py
```diff
@@ -2125,6 +2128,22 @@ def normalize_prefix(leaf: Leaf, *, inside_brackets: bool) -> None:
     leaf.prefix = ""
 
 
+def normalize_string_prefix(leaf: Leaf, remove_u: bool = False) -> None:
+    """Lowercases all string prefixes.
```
ambv: Use imperative verb forms. "Make all string prefixes lowercase."

black.py
```diff
@@ -1138,6 +1139,7 @@ class LineGenerator(Visitor[Line]):
     in ways that will no longer stringify to valid Python code on the tree.
     """
     current_line: Line = Factory(Line)
+    remove_u: bool = False
```
ambv: I would call this field `remove_u_prefix`. `LineGenerator` deals with all nodes, so what "u" means here is not obvious.

black.py
```diff
@@ -2357,6 +2376,44 @@ def is_python36(node: Node) -> bool:
     return False
 
 
+def get_future_imports(node: Node) -> Set[str]:
+    """Returns a set of __future__ imports in the file."""
```
ambv: Again, imperative form.

black.py
```diff
@@ -2357,6 +2376,44 @@ def is_python36(node: Node) -> bool:
...
+    for child in node.children:
+        if child.type != syms.simple_stmt:
+            break
+        first_child = child.children[0]
+        if isinstance(first_child, Leaf):
+            # Continue looking if we see a docstring; otherwise stop.
+            if (
+                len(child.children) == 2
+                and first_child.type == token.STRING
+                and isinstance(child.children[1], Leaf)
```
ambv: This isinstance check is unnecessary. If the `type` is equal to `token.NEWLINE`, it's got to be a leaf.

black.py
```diff
@@ -2357,6 +2376,44 @@ def is_python36(node: Node) -> bool:
...
+                and child.children[1].type == token.NEWLINE
+            ):
+                continue
+            else:
+                break
+        elif isinstance(first_child, Node) and first_child.type == syms.import_from:
```
ambv: Does mypy need this `isinstance` check? If it's not a Leaf (we're in the elif portion after all), it's got to be a Node.

black.py
```diff
@@ -2357,6 +2376,44 @@ def is_python36(node: Node) -> bool:
...
+            module_name = first_child.children[1]
+            assert isinstance(module_name, Leaf)
```
ambv: That might not be true. For example:
```
from a.b.c import X
```
a.b.c here is a Node of type `syms.dotted_name`.

Fold this into the `if` check below.

black.py
```diff
@@ -2357,6 +2376,44 @@ def is_python36(node: Node) -> bool:
     return False
 
 
+def get_future_imports(node: Node) -> Set[str]:
```
ambv: I find that the number of isinstance checks here makes the function harder to read than necessary. I wonder if we can do anything to minimize those checks.

black.py
```diff
@@ -2357,6 +2376,44 @@ def is_python36(node: Node) -> bool:
...
+                for import_from_child in first_child.children[3:]:
+                    if isinstance(import_from_child, Leaf):
+                        if import_from_child.type == token.NAME:
+                            imports.add(import_from_child.value)
+                    elif isinstance(import_from_child, Node):
+                        assert import_from_child.type == syms.import_as_names
+                        for leaf in import_from_child.children:
+                            if isinstance(leaf, Leaf) and leaf.type == token.NAME:
+                                imports.add(leaf.value)
+            else:
```
ambv: If you reversed the if/else logic here, you would save one indentation level.

black.py
```diff
@@ -2357,6 +2376,44 @@ def is_python36(node: Node) -> bool:
...
+                break
+        else:
```
ambv: Same here about saving an indenatation level.

black.py
zsol: Because we want to strip that `u` in 3.6+ code, right?

ambv: Yes, why keep it for code that has f-strings et al.

black.py
JelleZijlstra: Mypy made me add some but I think I overdid it. Your other comments helped me find a few unnecessary ones, so I think it's better now.

black.py
JelleZijlstra: I think in this case it would make the structure of the function less obvious. It makes more sense to me to keep the block looking for docstrings and the one looking for import from nodes at the same indentation.

black.py
JelleZijlstra: Good catch, I added a test case for this.

black.py
JelleZijlstra: I just made it do the removal in 3.6-only code, but really of course it should do it in all Python 3-only code. Black currently doesn't have functionality for detecting whether any Python 3 feature is used, and adding that seems a bit too big for this PR. Of the top of my head, we could look for: annotations, class keyword arguments, keyword-only arguments, async/await, print as a function without a future import, extended unpacking.

PR #200, opened 2018-05-09
Adding Jupyter Notebook magic command

csurfer:


ambv:
Thanks, this is helpful! :-)

PR #202, opened 2018-05-10
Check for broken symlinks before checking file data

miggaiowski:
New unit test fails without the check in the code, passed afterwards.

### DIFF
--- a/black.py
+++ b/black.py
@@ -2494,7 +2494,7 @@ def gen_python_files_in_dir(path: Path) -> Iterator[Path]:
 
             yield from gen_python_files_in_dir(child)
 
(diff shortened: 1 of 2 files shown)


miggaiowski:
@zsol: That would require passing the report object down to that function, or passing the quiet param. Do you think it's that necessary?

ambv:
Thanks for the new test! For the actual fix, instead of adding an `exists()` in a hidden place in cache handling, we should fix even attempting to add this broken symlink to the list of files.

You can change line 2497 to just say `elif child.is_file() and child.suffix in PYTHON_EXTENSIONS:`. `is_file()` returns False for broken symlinks.

PR #219, opened 2018-05-17
Class new line between docstrings / vars / methods

Sterbic:
Input:
```
class ClassWithInit:
    def __init__(self):
        pass
```

Current output:
```
class ClassWithInit:
+
    def __init__(self):
        pass
```

Desired output:
```
class ClassWithInit:
    def __init__(self):
        pass
```

Input:
```
class ClassWithDocstring:
    """docstring"""
    var = 10
    def __init__(self):
        pass
```

Current output:
```
class ClassWithDocstring:
    """docstring"""
    var = 10

    def __init__(self):
        pass
```

Desired output:
```
class ClassWithDocstring:
    """docstring"""

    var = 10

    def __init__(self):
        pass
```

### DIFF
--- a/black.py
+++ b/black.py
@@ -912,8 +912,13 @@ class Line:
     @property
     def is_docstring(self) -> bool:
         """Is the line a triple quoted docstring?"""
-        return self.leaves[0].type == token.STRING and self.leaves[0].value.startswith(
-            '"""'
+        return (
+            bool(self)
+            and self.leaves[0].type == token.STRING
+            and (
+                self.leaves[0].value.startswith('"""')
+                or self.leaves[0].value.startswith("'''")
+            )
         )
 
     def contains_standalone_comments(self, depth_limit: int = sys.maxsize) -> bool:
--- a/black.py
+++ b/black.py
@@ -1181,6 +1181,12 @@ class EmptyLineTracker:
                 return 0, 0
 
             if (
+                self.previous_line.is_class
+                and self.previous_line.depth != current_line.depth
+            ):
+                return 0, 0
+
+            if (
                 self.previous_line.is_comment
                 and self.previous_line.depth == current_line.depth
                 and before == 0
--- a/black.py
+++ b/black.py
@@ -910,7 +910,7 @@ class Line:
         )
 
     @property
-    def is_docstring(self) -> bool:
+    def is_triple_quoted_string(self) -> bool:
         """Is the line a triple quoted docstring?"""
         return (
             bool(self)
@@ -1214,7 +1214,7 @@ class EmptyLineTracker:
         if (
             self.previous_line
             and self.previous_line.is_class
-            and current_line.is_docstring
+            and current_line.is_triple_quoted_string
         ):
             return before, 1
 
--- a/black.py
+++ b/black.py
@@ -909,6 +909,13 @@ class Line:
             and self.leaves[3].value == ")"
         )
 
+    @property
+    def is_docstring(self) -> bool:
+        """Is the line a triple quoted docstring?"""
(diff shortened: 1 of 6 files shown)


tests/class_methods_new_line.py
```diff
@@ -0,0 +1,42 @@
...
+
+# output
+
+
+class ClassSimplest:
+    pass
+
+
+class ClassWithInit:
+
```
Sterbic: This new line needs to be removed

tests/class_methods_new_line.py
```diff
@@ -0,0 +1,42 @@
...
+class ClassWithInitAndVars:
+    cls_var = 100
+
+    def __init__(self):
+        pass
+
+
+class ClassWithInitAndVarsAndDocstring:
+    """Test class"""
+    cls_var = 100
```
Sterbic: @ambv, I'd usually put an empty line between docstrings and the rest of the code below. What's your view on this?

ambv: Black currently does not enforce that. What does PEP 257 say?

Sterbic: > There's no blank line either before or after the docstring.

> Insert a blank line after all docstrings (one-line or multi-line) that document a class -- generally speaking, the class's methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line.

(src: https://www.python.org/dev/peps/pep-0257/)

black.py
```diff
@@ -909,6 +909,13 @@ def is_class_paren_empty(self) -> bool:
             and self.leaves[3].value == ")"
         )
 
+    @property
+    def is_docstring(self) -> bool:
+        """Is the line a triple quoted docstring?"""
```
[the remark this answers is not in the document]
Sterbic: 1. With this you mean that I should check the ending quotes as well?
2. I don't think `is_multiline_string` is applicable here. The docstring can be in a single line but we would still want an empty line after it and `is_multiline_string` returns `False` for `"""test"""`.
3. Will fix.

Sterbic:
@ambv, anything else I should take care of here? PEP 257 is now enforced and class def and first method are not separated by a new line any more.

renstrom:
Sorry for the too-late comment. But I think this is wrong, and the old behaviour was correct.

>Desired output:
>
>```
>class ClassWithInit:
>    def __init__(self):
>        pass
>```

For one, having a blank line is more consistent with the behaviour you get if you have a class docstring. And second, it's consistent with the PEP8 rule the methods should be surrounded by a blank line.

renstrom:
PEP 8 also says this:

>Method definitions inside a class are surrounded by a single blank line.

It doesn't go into much detail, but it doesn't mention any exception like the `class` definition. Therefore I think it should be treated as there should _always_ be blank line before and after the method.

Sterbic:
I believe this should be consistent regardless of what the first line after class declaration is, be it a method, variable or docstring. And you wouldn't put an empty line between class and docstring, as per PEP 257.

If you check the wording in PEP 257 it actually clarifies that the one empty line rule is meant to be applied between 2 methods only:
> ... generally speaking, the class's methods are separated from **each other** by a single blank line
(src: https://www.python.org/dev/peps/pep-0257/)

Various PEPs use this style in examples, including PEP 8, PEP 484 and PEP 525.

ambv:
@renstrom, if you decide to deconstruct particular wording in PEP 8 and PEP 257, you might be technically right but I don't think that would be right in the spirit of those documents, they aren't written in a particularly strict style (pun not intended). The authors of those PEPs write code that looks like what Luka implemented in this PR. Previous behavior was very unpopular and a big criticism of Black for "putting unnecessary empty lines" between the first method and the class signature.

PR #231, opened 2018-05-19
Move Grammar cache to user's cache directory

tiran:
blib2to3.pygram assumes that blib2to3's directory is writeable. This is
not the case for global installation of black on Linux distributions.

Fixes: https://github.com/ambv/black/issues/192

### DIFF
--- a/black.py
+++ b/black.py
@@ -38,17 +38,170 @@ from attr import dataclass, Factory
 import click
 
 # lib2to3 fork
+import blib2to3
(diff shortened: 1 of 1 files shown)


tiran:
Alternative approach for #203

If you like the approach, I'll add some unit tests for the cache, too.

ambv:
I like this approach! I used a different technique to minimize the number of changes to black.py and blib2to3 but your idea is what inspired it! Thanks.

PR #249, opened 2018-05-22
Add --pyi flag to force pyi mode when reformatting stdin.

### LINKED ISSUE #244: Make it possible to pass --pyi and --py36
1. When a file is passed on standard input, there is no way currently to discover it's a typing stub file and should be formatted with the .pyi style. Provide a command-line option that forces this.

2. When a file isn't using Python 3.6 features, Black will avoid putting trailing commas after stars in function signatures and function calls. This raises question from team members who think this is a Black bug. Allow for a command-line option to force Python 3.6 behavior.

carljm:
Partially resolves #244.

Current behavior here is that the `--pyi` flag is silently ignored unless srcs is `-`. Other options would be to make it an error to provide `--pyi` when reformatting files in place, or to have `--pyi` actually force pyi mode (regardless of file extension) when formatting files in place. Let me know what you prefer.


carljm:
@ambv This is complete (I think) and ready for review.

I think it might be time to consolidate `line_length`, `pyi`, and `py36` into a `BlackConfig` class rather than passing the three of them separately to so many methods, but I feel it will be easier to review if we first merge this PR and then do that as a separate follow-up.

ambv:
This is all good except that I don't like test data kept inside the test code. We have a lot of test files around already, could you move the data from your new tests to new data files instead?

Also, please update the Change Log.

PR #268, opened 2018-05-28
Add isort args to README

jasonfriedland:


ambv:
What are you trying to accomplish here? You changed existing docs that mention names of options as spelled out in .isort.cfg to cmdline options. How is that better?

zsol:
I suppose this is somewhat easier to copy paste.

jasonfriedland:
Because they don't match. I had to needlessly grep through the CLI output
to find the corresponding equivalents.

I'll include both if you think that would be better.

ambv:
It's surprising that the spelling is different between command-line options and the configuration file. I guess now this can never be fixed due to backwards compatibility.

The reason I prefer the spelling of a config file is that you only need to put that in once and future invocations will keep your config. Command-line spelling means you have to provide this every time. Another reason why Black is trying so hard not to have options and configuration.

I'd be fine with documenting both variants. Just use simplified English, please. I don't use British spelling in the docs nor in the code.

jasonfriedland:
> I'd be fine with documenting both variants. Just use simplified English, please. I don't use British spelling in the docs nor in the code.

OK sure I'll add both. 

> The reason I prefer the spelling of a config file is that you only need to put that in once and future invocations will keep your config. Command-line spelling means you have to provide this every time. Another reason why Black is trying so hard not to have options and configuration.

I fully appreciate the aim to have less (no?) config options, however this is more about bending other tools' to match Black's way of formatting :-) My feeling is that (at least in my CI environment...) isort's invocation is typically handled by build/test scripts. That way the args are persisted and version-controlled per-project without having to maintain a separate config for it.

Thanks for the feedback.

Issue #118, opened 2018-04-09
Single quotes option


bofm:
Hi! Although Black _now_ prefers doubles, can we have an option to keep single quotes? Forcing double quotes would make this great project unusable for many users who picked the rule of using single quotes.

Operating system: MacOS
Python version: 3.6
Black version: 18.4a0
Does also happen on master: yes

gaggle:
I'm not a maintainer just an interested user, I respectfully voted down because I think black's power is to annihilate all these discussions forever. Out of curiosity are there objective reasons that you need to stick with single-quotes?

bofm:
> are there objective reasons that you need to stick with single-quotes?

It is an approved code style I have to use at work. There are millions of lines of code with single quotes in the existing code base which are not going to be reformatted. I believe this is also the case for other people.

ambv:
Please discuss the approved code style at work, Black's documentation is using several arguments as to why double quotes are preferred.

If you have strong reasons to prefer single quotes, list them here. "Approved code style" is not a reason alone.

bofm:
The reasons to add single quotes options:

* Black is unusable when contributing to an existing project written with single quotes
* Python official documentation uses single quotes in code examples
* default __repr__() produces single-quoted string literals for most standard types (dict, list, tuple, str, bytes, namedtuple, etc)
* many popular projects and their docs use single quotes
* PEP proposals mostly use single quotes in code examples
* Guido used single quotes in his latest commits on Github

Let me emphasize - the point is that restricting the quotes variants to only double quotes makes this cool project unusable for a big part of Python programmers.

ambv:
All your reasons can be summarized like this: "Black shouldn't have an opinion because other projects don't have an opinion."

This is [an appeal to authority](https://en.wikipedia.org/wiki/Argument_from_authority) and as such doesn't give us anything actionable to discuss.

I also don't buy the argument that this makes it "unusable for a big part of Python programmers". Why? The code is going to still work, programmers may even continue writing their code with single quotes (and let Black convert it as needed).


Note that Black is using a variant of wrapping lines with brackets that isn't even covered as an option in PEP 8. Does this make Black unusable?

bofm:
What if a project _has_ an opinion but a different one? Is Black’s intention to change opinions or to be a helpful tool?

kalekseev:
[...]

PR #272, opened 2018-05-29
The very exact Henry Ford’s quote.

batisteo:


ambv:
Thanks for your contribution but we'll keep the existing wording :-)

batisteo:
I guess arguing here would be bikeshedding anyway :slightly_smiling_face:

batisteo:
I just though some people in the #118 thread didn’t get the reference of this project name.

jgirardet:
I must say I didn't know this sentence and I now better understand @ambv's state of mind.
your PR wasn't useless for me ;-)

ambv:
Let's keep this as a not very well hidden Easter egg. It's more satisfying when people find out on their own ;-)

Issue #270, opened 2018-05-29
--include= and --exclude=


ambv:
I always steered people towards `find | grep -E | grep -Ev | xargs black` to do custom file matching. But that doesn't work on Windows. We need support for `--include=` and `--exclude=` within Black.

They hold regular expressions. The former is "what gets included on recursive searches", the latter "what gets dropped on recursive searches". You can pass multiple `--include` and multiple `--exclude`. Checks for `--exclude` are done first, checks for `--include` later. 

`--include=` would have a default that is now stores in PYTHON_EXTENSIONS ("\.pyi?$" in regex form)
`--exclude=` would have a default that is now stored in BLACKLISTED_DIRECTORIES

This functionality is required by https://github.com/pypa/pip/pull/5425.

ambv:
Re 1: Agreed.

Re 2: Yes, your suggestion is good to append / when we discover `entry.is_dir()` :-)

PR #281, opened 2018-05-30
Added --include and --exclude cli options

autophagy:
These 2 options allow you to pass in regular expressions that determine whether files/directories are included or excluded in the recursive search.

This should address #270. I put the change log entry for this under `18.6b0` since I don't think it has been released yet - but let me know if this was incorrect! :sparkles:

### DIFF
--- a/black.py
+++ b/black.py
@@ -238,21 +238,19 @@ def main(
 ) -> None:
     """The uncompromising code formatter."""
     sources: List[Path] = []
+    try:
+        include_regex = re.compile(include)
+    except re.error:
+        err(f"Invalid regular expression for include given: {include!r}")
+        ctx.exit(2)
+    try:
+        exclude_regex = re.compile(exclude)
+    except re.error:
+        err(f"Invalid regular expression for exclude given: {exclude!r}")
+        ctx.exit(2)
     for s in src:
         p = Path(s)
         if p.is_dir():
-            try:
-                include_regex = re.compile(include)
-            except Exception as exc:
-                raise SyntaxError(
-                    "Invalid regular expression for include given: {}".format(include)
-                )
-            try:
-                exclude_regex = re.compile(exclude)
-            except Exception as esc:
-                raise SyntaxError(
-                    "Invalid regular expression for exclude given: {}".format(exclude)
-                )
             sources.extend(gen_python_files_in_dir(p, include_regex, exclude_regex))
         elif p.is_file():
             # if a file was explicitly given, we don't care about its extension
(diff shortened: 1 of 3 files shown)


Autophagy:
Also, I genuinely do not know how I would check the types of `exclude` and `include` here https://github.com/ambv/black/pull/281/files#diff-a75051ebf8e3452bf3d590c186944252R2791 since I am passing in compiled regular expressions which have a type of `_sre.SRE_Pattern`, but apparently `AttributeError: module '_sre' has no attribute 'SRE_Pattern'` :sweat_smile:

JelleZijlstra:
Try `typing.Pattern[str]`.

Autophagy:
@JelleZijlstra Ah! Thank you a bunch.

black.py
```diff
@@ -2750,32 +2788,24 @@ def get_future_imports(node: Node) -> Set[str]:
...
+def gen_python_files_in_dir(
+    path: Path, include: Pattern[str], exclude: Pattern[str]
+) -> Iterator[Path]:
+    """Generate all files under `path` whose paths are not excluded by the
+    `exclude` regex, but are included by the `include` regex.
     """
     for child in path.iterdir():
         if child.is_dir():
-            if child.name in BLACKLISTED_DIRECTORIES:
+            if not exclude.search(str(child) + "/"):
```
ambv:
```
$ mkdir -p a/b/c/d
$ touch a/b/c/d/file.py
$ cd a
$ black .
No paths given. Nothing to do 😴
```

(This works on master.)

black.py
```diff
@@ -46,6 +46,10 @@
 
 __version__ = "18.5b1"
 DEFAULT_LINE_LENGTH = 88
+DEFAULT_EXCLUDES = (
+    r"build/|buck-out/|dist/|_build/|\.git/|\.hg/|\.mypy_cache/|\.tox/|\.venv/"
```
ambv: There's a problem to solve here. In the current model, "not_a_build/" is also matching. It would be more robust to prepend all those directory names with slashes. However, if the user specified a relative directory, we might not have a leading slash always. If they said `black .`, it's fine. If they said `black *`, it's not.

So up to you how we solve this but it will have to be addressed before we land this.

autophagy: Good point, I think prepending the directories with slashes would be the best route. I'll play around with it and see if I can come up with an elegant way of doing this.

autophagy:
@ambv Added some logic to allow for prepended directories in the filter, and confirmed it works well with both `*` and `.`. I'm building up the path to search from using the children, rather than the path, as it's a little easier when the path itself is something like `.`.

ambv:
Makes sense to just mention: "use forward-slashes on Windows, too."

ambv:
Awesome! ✨ 🍰 ✨

PR #288, opened 2018-06-01
Preserve line endings when formatting a file in place

### LINKED ISSUE #258: Line endings on Windows.
It's possible to work around using for example Git configuration, but it would be nice if the behavior was configurable. Some of us like to keep Unix line endings even on Windows.

zsol:
Fixes #258 

Note that this only changes behavior when black writes back to the file. In other cases like when using `--diff` or writing to stdout, I think it's reasonable to expect the caller to handle line endings.

### DIFF
--- a/black.py
+++ b/black.py
@@ -454,7 +454,9 @@ def format_file_in_place(
         return False
 
     if write_back == write_back.YES:
-        with open(src, "w", encoding=src_buffer.encoding) as f:
+        with open(src, encoding=src_buffer.encoding, newline="") as src_buffer:
+            newline = "\r\n" if "\r\n" in src_buffer.readline() else "\n"
+        with open(src, "w", encoding=src_buffer.encoding, newline=newline) as f:
             f.write(dst_contents)
     elif write_back == write_back.DIFF:
         src_name = f"{src}  (original)"
@@ -569,8 +571,7 @@ def lib2to3_parse(src_txt: str) -> Node:
     """Given a string with source, return the lib2to3 Node."""
     grammar = pygram.python_grammar_no_print_statement
     if src_txt[-1] != "\n":
-        nl = "\r\n" if "\r\n" in src_txt[:1024] else "\n"
-        src_txt += nl
+        src_txt += "\n"
     for grammar in GRAMMARS:
         drv = driver.Driver(grammar, pytree.convert)
         try:
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -910,6 +910,18 @@ class BlackTestCase(unittest.TestCase):
             result = CliRunner().invoke(black.main, ["-", option, "**()(!!*)"])
             self.assertEqual(result.exit_code, 2)
 
+    def test_preserves_line_endings(self) -> None:
+        with TemporaryDirectory() as workspace:
+            test_file = Path(workspace) / "test.py"
+            for nl in ["\n", "\r\n"]:
+                contents = nl.join(["def f(  ):", "    pass"])
+                test_file.write_bytes(contents.encode())
+                ff(test_file, write_back=black.WriteBack.YES)
+                updated_contents = test_file.read_bytes()
+                self.assertIn(nl.encode(), updated_contents)
+                if nl == "\n":
+                    self.assertNotIn(b"\r\n", updated_contents)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/README.md
+++ b/README.md
@@ -715,6 +715,8 @@ More details can be found in [CONTRIBUTING](CONTRIBUTING.md).
 * fixed stdin handling not working correctly if an old version of Click was
   used (#276)
 
+* *Black* now preserves line endings when formatting a file in place (#258)
+
 
 ### 18.5b1
 


ambv:
I'll wait with landing this until I fix the tests on Travis.

zsol:
In the meanwhile I'll figure out how to install precommit on windows and fix everything I've broken

black.py
```diff
@@ -454,7 +454,9 @@ def format_file_in_place(
         return False
 
     if write_back == write_back.YES:
-        with open(src, "w", encoding=src_buffer.encoding) as f:
+        with open(src, encoding=src_buffer.encoding, newline="") as src_buffer:
```
ambv: You don't have to open it again, that's slow. You already have `src_contents` that you can check. Suggestion:
```py3
windows_newlines = src_contents.split("\n", 1)[0][-1:] == "\r"
newline = "\r\n" if windows_newlines else "\n"
```

zsol: I'm not sure if opening a file for reading one more time right after reading through it is a huge performance hit, but if it turns out to be the case, we can reproduce the 10 or so lines of code in `tokenize.open()` to make sure we only do it once.

ambv: Right. OK, leave as is but open it in `"rb"` so it doesn't need to decode anything. The first line must be readable anyway since if it's not UTF8 (or ASCII on Python 2) it needs the `# -*- coding: X -*-` comment.

And as a total nit: instead of `b"\r\n" in readline` you can do `readline[-2:] == b"\r\n"` which is marginally faster still.

tests/test_black.py
```diff
@@ -910,6 +910,18 @@ def test_invalid_include_exclude(self) -> None:
...
 
+    def test_preserves_line_endings(self) -> None:
+        with TemporaryDirectory() as workspace:
+            test_file = Path(workspace) / "test.py"
+            for nl in ["\n", "\r\n"]:
+                contents = nl.join(["def f(  ):", "    pass"])
+                test_file.write_bytes(contents.encode())
+                ff(test_file, write_back=black.WriteBack.YES)
+                updated_contents: bytes = test_file.read_bytes()
+                self.assertIn(nl.encode(), updated_contents)  # type: ignore
```
zsol: mypy complains about this line like so:
```
Cannot infer type argument 1 of "assertIn" of "TestCase"
```
I think it's confused because both arguments are `bytes`. Seems like a bug in mypy to me

JelleZijlstra: I think the typeshed stub is wrong. It has:
```
    def assertIn(self, first: _T, second: Iterable[_T],
                 msg: Any = ...) -> None: ...
    def assertNotIn(self, first: _T, second: Iterable[_T],
                    msg: Any = ...) -> None: ...
```
This should actually be `Container`, not `Iterable`.

JelleZijlstra: https://github.com/python/typeshed/pull/2186 should fix it.

jgirardet:
@zsol 

> Note that this only changes behavior when black writes back to the file. In other cases like when using --diff or writing to stdout, I think it's reasonable to expect the caller to handle line endings.

I think It makes things very "compromising" and I thought it wasn't the purpose of black.
As the readme proposes to use `black -` to integrate black on editors, it gives some different behaviour depending the platform.

I think it would be more consistant that black  preserves line ending with `black -` and `black --diff`. 

Anyway If you disagree It should be mentioned in the readme

ambv:
@zsol, I think @jgirardet is right. Especially on piping stdin to stdout we should be preserving existing newline style. But also on --diff, not taking existing format into account will lead to bugs in intergrations (like sublack).

I don't understand @jgirardet's comment about compromise. I just think the current behavior is still effectively buggy.

jgirardet:
Just a joke,  but that was probably bad English speaking :blush:

zsol:
fair enough, let me iterate on this

jgirardet:
Hi guys,
I'm not  sure anymore my sublack bug was this black bug. I tried it on windows python idle : 

```python
>> p = subprocess.Popen(
                ['black', '-'],
                stdin=subprocess.PIPE,
                stdout=subprocess.PIPE,
                stderr=subprocess.PIPE,)
>>> p.communicate(b"hgello\n")
(b'hgello\r\n', b'- already well formatted, good job.\r\n')

>>> p = subprocess.Popen(
                ['black', '-'],
                stdin=subprocess.PIPE,
                stdout=subprocess.PIPE,
                stderr=subprocess.PIPE,
		universal_newlines=True)
>>> p.communicate('hello\n')
('hello\n', '- already well formatted, good job.\n
```
Maybe this is just the popen standard behaviour ?

ambv:
@jgirardet, wow, this is strange. I don't work with Windows so I don't really understand what the expected behavior should be with regards to stdin/stdout. Maybe what @zsol is already doing is valid after all on Windows?

zsol:
@jgirardet what you're seeing here is just the effect of `universal_newlines` being on or off IMO.

```
>>> subprocess.Popen(['python', '-c', r'import sys; sys.stdout.write("\r\n")'], stdout=subprocess.PIPE).communicate()
(b'\r\n', None)
>>> subprocess.Popen(['python', '-c', r'import sys; sys.stdout.write("\r\n")'], stdout=subprocess.PIPE, universal_newlines=True).communicate()                                                                                              
('\n', None)
```

There _is_ a bug in black in that it doesn't preserve newline styles as you can see in your first `communicate()` call. Without `universal_newlines` you get back the bytestream black outputs without any decoding or newline conversion. I bet if you do `p.communicate(b'hello\r\n')` you'll get the exact same output as `p.communicate(b'hello\n')`. After I fix this issue, they will be different.

zsol:
Alright this is ready for review @ambv. I had to rewrite most of the code because we can't afford the luxury of consuming stdin twice

ambv:
Thanks! ✨ 🍰 ✨

jgirardet:
thanks @zsol.
This solved my problem in sublack for stdout.
I still have the issue with stderr:
All done! \u2728 \U0001f370 \u2728
1 file left unchanged.

ambv:
@jgirardet, stderr is user-facing messages. It should have \r\n on Windows, this is how people are going to see them in the console.

jgirardet:
ok
thats clear thanks

blkpanther:
If I understand this correctly, black should preserve line endings when used with - (stdin). But I think 18.6b4 (the version that is currently supplied by pip) won't. I input code over stdin with \r\n line endings, I get output with \n. [am I wrong or missing something?]

ambv:
@zsol, can you try it directly from PowerShell on Windows (without WSL)?

zsol:
I have no idea how to verify this in PowerShell (or cmd.exe for that matter). I do know our [test](https://github.com/ambv/black/blob/master/tests/test_black.py#L1210) passes, but that doesn't use `-`

zsol:
OK I wrote a new test for standardin and it fails. Investigating

PR #301, opened 2018-06-05
✨ 🍰 ✨ isn't appropriate when it fails, fixes #300

### LINKED ISSUE #300: ✨ 🍰 ✨ isn't appropriate when it fails
It's not a bug, more of an enhancement/feature request. 
I run `black --check dir/with/files` and get this output:
```
...
would reformat /app/tests/test_ingest_backfill.py
would reformat /app/tests/test_main_models.py
would reformat /app/tests/test_ingest_sqs.py
All done! ✨ 🍰 ✨
23 files would be reformatted, 11 files would be left unchanged.
```

It's "all done" as in it's finished checking all the files. But the ✨ 🍰 ✨ emojis are way to happy when in fact the whole check failed.

peterbe:
It's whimsical but this output feels more appropriate:
```bash
▶ python black.py --check /tmp/foo.py
would reformat /tmp/foo.py
All done!💥 💔 💥
1 file would be reformatted.
▶ python black.py /tmp/foo.py
reformatted /tmp/foo.py
All done! ✨ 🍰 ✨
1 file reformatted.
▶ python black.py --check /tmp/foo.py
All done! ✨ 🍰 ✨
```

PR has no tests. Tips on where to put the tests appreciated.

### DIFF
--- a/black.py
+++ b/black.py
@@ -337,7 +337,7 @@ def main(
         finally:
             shutdown(loop)
     if verbose or not quiet:
-        out("All done! ✨ 🍰 ✨")
+        out("All done!" + ("💥 💔 💥" if report.return_code else " ✨ 🍰 ✨"))
         click.secho(str(report), err=True)
     ctx.exit(report.return_code)
 


ambv:
Thanks! 🎖

PR #302, opened 2018-06-05
correct email for Peter Bengtsson

peterbe:
Thanks for the [ack](https://github.com/ambv/black/commit/26b92e66aca6124ba8761f690ec7ee18757411b7) but I'd like to use my non-work email if you don't mind.


ambv:
Sure thing!

PR #310, opened 2018-06-06
vim: add "--skip-string-normalization" support

dotwaffle:
Since 18.6b0 was released, there has been a new option to skip string
normalization when Black is called, but it wasn't able to be specified
from within the vim plugin. This commit adds that functionality.

Tested with g:black_skip_string_normalization set to 0 (off) and 1 (on).


ambv:
Thanks! ✨ 🍰 ✨

PR #316, opened 2018-06-07
Add `-h` as a shortcut for `--help`

sixninetynine:
I always type `black -h` and then kick myself for not typing `black --help`, this PR being accepted would mean I don't have to kick myself :)


ambv:
Thanks! ✨ 🍰 ✨

PR #324, opened 2018-06-08
fix when black_skip_string_normalization is set to 0

xoolive:
mode is undefined if black_skip_string_normalization is set to 0 which raises an error in vim


xoolive:
already solved by previous commit...

PR #339, opened 2018-06-12
Ignore symbolic links pointing outside of the root directory

### LINKED ISSUE #338: Crash if there is a symlink to a resource outside of the root directory
Hello there, thank you for this great tool!

Black crashes when it finds a symbolic link pointing to a file or a directory outside of the root directory.

**Steps to reproduce**

- Create the following structure:
```
base
  |-- repo
  |     |-- dummy.py
  |     `-- resource -> ../resource
  `-- resource
```
- Initiate `base/repo` as a Git repository (so as to set the root directory);
- Call `black .` within `base/repo`;
- Expected behavior: *Black* ends as expected;
- Current behavior: throws an exception:
```
  File "/home/neraste/.virtualenvs/test_black/bin/black", line 11, in <module>
    sys.exit(main())
  File "/home/neraste/.virtualenvs/test_black/lib/python3.6/site-packages/click/core.py", line 722, in __call__
  File "/home/neraste/.virtualenvs/test_black/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/neraste/.virtualenvs/test_black/lib/python3.6/site-packages/click/core.py", line 895, in invoke
  File "/home/neraste/.virtualenvs/test_black/lib/python3.6/site-packages/click/core.py", line 535, in invoke
  File "/home/neraste/.virtualenvs/test_black/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
  File "/home/neraste/.virtualenvs/test_black/lib/python3.6/site-packages/black.py", line 347, in main
  File "/home/neraste/.virtualenvs/test_black/lib/python3.6/site-packages/black.py", line 2942, in gen_python_files_in_dir
    normalized_path = "/" + child.resolve().relative_to(root).as_posix()
  File "/usr/lib64/python3.6/pathlib.py", line 872, in relative_to
    .format(str(self), str(formatted)))
ValueError: '/data/test_black/base/resource' does not start with '/data/test_black/base/repo'
```

As pointed out in the backtrace, this is due to the line 2942, where the resolved path of the files is made relative to the root directory. This is indeed impossible if the file is outside of the root directory. One can wonder if it makes sense to have an external link within a repository, but, well I am in this case.

Adding the conflicting file/folder to the exclude list does not work.

**Possible workaround**

A possible workaround could be to simply ignore this case and do not handle this simlinked resource:
```python
        try:
            normalized_path = "/" + child.resolve().relative_to(root).as_posix()

        except ValueError:
            if child.is_symlink():
                continue                                                        

            raise
```

**Additional information**

Operating system: GNU/Linux OpenSUSE Tumbleweed
Python version: 3.6
*Black* version: 18.6b2

Neraste:
Within the structure of files to analyze, if there is a symbolic link pointing to a file or a directory outside of the root directory, this symlink is now ignored, as it appears to be out of the scope of the analyzed project.

This PR fixes #338.

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -1162,7 +1162,7 @@ class BlackTestCase(unittest.TestCase):
         with self.assertRaises(AssertionError):
             black.assert_equivalent("{}", "None")
 
-    def test_symlink_within_structure(self) -> None:
+    def test_symlink_out_of_root_directory(self) -> None:
         # prepare argumens
         path = MagicMock()
         root = THIS_DIR
@@ -1186,6 +1186,24 @@ class BlackTestCase(unittest.TestCase):
         path.iterdir.assert_called_once()
         child.resolve.assert_called_once()
         child.is_symlink.assert_called_once()
+
+        # set the behavior of mock arguments
+        # child should behave like a strange file which resolved path is clearly
+        # outside of the root directory
+        child.is_symlink.return_value = False
+
+        # call the method
+        # it should raise a ValueError
+        with self.assertRaises(ValueError):
+            list(black.gen_python_files_in_dir(path, root, include, exclude, report))
+
+        # check the call of the methods of the mock objects
+        path.iterdir.assert_called()
+        self.assertEqual(path.iterdir.call_count, 2)
+        child.resolve.assert_called()
+        self.assertEqual(child.resolve.call_count, 2)
+        child.is_symlink.assert_called()
+        self.assertEqual(child.is_symlink.call_count, 2)
 
 
 if __name__ == "__main__":
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -11,7 +11,7 @@ import sys
 from tempfile import TemporaryDirectory
 from typing import Any, BinaryIO, Generator, List, Tuple, Iterator
 import unittest
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
 
 from click import unstyle
 from click.testing import CliRunner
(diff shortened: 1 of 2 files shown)


tests/test_black.py
```diff
@@ -1162,6 +1162,49 @@ def test_assert_equivalent_different_asts(self) -> None:
...
+
+        # check the call of the methods of the mock objects
+        path.iterdir.assert_called_once()
+        child.resolve.assert_called_once()
+        child.is_symlink.assert_called_once()
+
+        # set the behavior of mock arguments
+        # child should behave like a strange file which resolved path is clearly
+        # outside of the root directory
+        child.is_symlink.return_value = False
```
alvarocavalcanti: From this point on I would suggest a new test case, which could share the setup with the current one.

tests/test_black.py
```diff
@@ -1162,6 +1162,49 @@ def test_assert_equivalent_different_asts(self) -> None:
...
+
+        # call the method
+        # it should raise a ValueError
+        with self.assertRaises(ValueError):
+            list(black.gen_python_files_in_dir(path, root, include, exclude, report))
+
+        # check the call of the methods of the mock objects
+        path.iterdir.assert_called()
```
alvarocavalcanti: Personally, I don't like test assertions like these, where the test deeply knows the tested method implementation.

I would suggest having simple assertions as if an error is raised or not based solely on the symlink validity. In other words, assert the behavior, not the implementation. That way, if the code changes in the future, but yields the same result, the tests would still be valid.

Neraste: I am using mock objects for the directory and the symlink inside it. I have set up a scenario to use them, according to their actual use within the function (which allows to perform the behavior test), so at the end of the test, I just check if the scenario was accurate.

I am not very experienced with mocking, but it seems to me that if I use mock objects, I somewhat make the test aware about the implementation of the tested function.

alvarocavalcanti: You are correct on that point, just by using mocks you are already surfacing the inwards of the implementation, and thus asserting on the mock calls isn't wrong as well.

At work I have the same scenario on several (hundreds) tests, mocking 3rd-party calls (which relates the `os` file module here) at `setUp` for all test cases and only adding `side_effect` or new `return_value` when needed. And sometimes we assert on the mocks as you did. But we strive to assert on the behavior.

I guess my point here is mere semantics, and a little bit of food for thought. 😁 

Anyway, thank you for this PR, I was trying to use Black for the first time when I got this error, but it felt great to see someone already fixing it. 👍

ambv:
Thanks! ✨ 🍰 ✨

PR #347, opened 2018-06-13
Harmonise "PEP 8" spelling/formatting in README

katrinleinweber:
There was one instance without a whitespace in between, see diff ;-)


ambv:
Thanks! ✨ 🍰 ✨

PR #352, opened 2018-06-14
correct author name

ofek:
cc @ambv wait until new pip release ofc :slightly_smiling_face:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,9 +36,7 @@ name = "black"
 version = "18.6b2"
 description = "The uncompromising code formatter."
 readme = "README.md"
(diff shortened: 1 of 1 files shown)


ambv:
As you can see on the AppVeyor failure in this pull request, we can't merge this. It's not enough for us to wait for a fix to be released with https://github.com/pypa/pip/pull/5482 merged because users will likely have older pip 10 releases. So for at least a long while, we will have to stick to ASCII there.

However, you can replace "L" with a Python-like `"\uXXXX"` Unicode escape since TOML uses those. Thanks for your case about Unicode cleanliness!

ambv:
This is now fixed.

PR #357, opened 2018-06-15
Add PyPI Python version badge to readme

Mariatta:


ambv:
We have an issue in Black where many users assume that you can only format Python 3 code with it. This badge may mislead them even more into thinking that.

Mariatta:
Personally I've used it to quickly identify the minimum Python version needed before using a library. However, I can see where people might misunderstood the meaning of the badge..

ambv:
Yeah, thanks for your PR but I think I'm going to pass on this badge :)

PR #360, opened 2018-06-16
add blank line after constants in stub file

### LINKED ISSUE #340: pyi mode: blank line after constants
In pyi mode, I think there should be a blank line here
```python
x: int
def y() -> None: ...
```
I can make this change.

JelleZijlstra:
Fixes #340


JelleZijlstra:
I made coverage 0.003% better! 🎂

Now why is Travis failing? Looks like it's:
```
black.py:1374:5: C901 'EmptyLineTracker._maybe_empty_lines' is too complex (19)
```
I don't think splitting up this function would be particularly helpful; @ambv should we just disable this flake error?

ambv:
I made the code as hackable as it is by splitting functions when they hit the complexity limit ;-)

JelleZijlstra:
Fair enough, I split it up.

PR #361, opened 2018-06-16
Add pyls-black to README

rupert:
I recently wrote a Python Language Server plugin for [Black](https://github.com/rupert/pyls-black). In the `README` you mentioned you were working on a Atom/Nuclide plugin so this may be of interest. To use it from Atom you'll need to install the `atom-ide-ui` and `ide-python` packages.

It would be great if `pyls-black` could be added to the `README` so other people can try it out!


ambv:
Thanks! ✨ 🍰 ✨

PR #382, opened 2018-06-27
pre-commit: use exclusion instead of ever-growing regex

asottile:


zsol:
Thanks! Sorry for taking so long to get to this

asottile:
No problem!

Issue #415, opened 2018-07-22
vim plugin hijacks the comma, and for all filetypes too


dimbleby:
version 18.6b4, also true on master.

`black.vim` [contains](https://github.com/ambv/black/blob/e94a41f92a568706700522aaad48ebd137fe1d8b/plugin/black.vim#L126-L127):

```
nmap ,= :Black<CR>
vmap ,= :Black<CR>
```

But there are problems with this:

* First, this hijacks the comma - which is a standard vim command, that some of us find useful.  It's not OK for for plugins to do this!
* Second, this mapping applies to all files - so even when I'm not doing anything python-related at all, `black.vim` has hijacked my comma

I'd suggest a couple of options:

* My preference: simply don't provide a mapping at all.  Users can perfectly well set up their own mapping for `:Black` according to what works for them.
  * Suggest a mapping in the README, if you like

* If you think it's important to provide a mapping out of the box, then my understanding of the conventions (though I'm not completely sure of this) is that:
  * `black.vim` should be filetype-specific - that is, you should move it to `ftplugin/python.vim`
  * the mapping that you provide should be scoped to the current `<buffer>`
  * the mapping should begin with `<localleader>`

That way your mapping will only affect python files, and not clash with any standard mappings.

dimbleby:
Actually it's probably desirable to move the plugin to `ftplugin/python.vim` *anyway* - so that I don't have to pay to load it, in the case that I'm not editing a python file.

ambv:
Everything you say sounds reasonable. Happy to merge PRs with this.

PR #416, opened 2018-07-22
The vim plugin is for the python filetype

dimbleby:
Per comment in #415 - only load the vim plugin for python files.

I hope this bit is uncontroversial.  I intend to follow up with a pull request that removes the mappings, which may want more discussion.


dimbleby:
On further reflection, an even better approach would be to put most of the plugin in `autoload/`, so as to get the benefit of lazy loading - while keeping the commands such as `:BlackVersion` always available, even if one hasn't yet loaded a python file.

I doubt that it makes sense to do that in this repository - it's already a *bit* weird to overlay the vim plugin and the formatter in a single repository, and once we start to make more use of vim's directory conventions that's only going to get more awkward.

So I've put together [`dimbleby/black.vim`](https://github.com/dimbleby/black.vim) to show what I think this should look like.

My hope is that you'll want to take this back and make it an "official" `ambv/black.vim`...

ambv:
I like preloading this so that by the time you load a Python file, it's ready for you. It's in the vein of Carmack's execute-and-inhibit style. Does this affect startup time bad enough for you that you need this change?

dimbleby:
Motivation was more along the lines of believing that this was the Right Way to Do Things, rather than specific pain with load times.  Computers are fast!

FWIW I see this version of the plugin taking ~175ms to load, and my variation taking about 90ms.  

Most of the 90ms is I think doing the python 3.6 check at start of day - I reckon if I deferred that until such time as we actually tried to use the formatter, the startup time would become completely negligible.  Maybe I'll take a look at doing this...

ambv:
That is an amazing find! Can you abandon this pull request and do another one *just with this py_version check change*? That makes everybody happy and doesn't disrupt anything :-)

dimbleby:
Not sure what you're asking for.  You want the dimbleby/black.vim version of the vim plugin, but superimposed on top of the ambv/black project?  That seems quite bizarre to me - probably I've misunderstood.

Feel free to take whatever you like from dimbleby/black.vim; but according to my current understanding I'd still most prefer for you to adopt that wholesale as ambv/black.vim.

goerz:
I would strongly recommend taking `dimbleby/black.vim` wholesale. It also solves another significant problem: With the original plugin, if `vim` has Python 3.5 instead of 3.6, it shows a very nasty error message every time vim starts:
~~~
Error detected while processing function provider#python3#Call:
line   18:
  File "<string>", line 17
    return venv_path / 'lib' / f'python{pyver[0]}.{pyver[1]}' / 'site-packages'
                                                            ^
SyntaxError: invalid syntax
~~~
This is because the plugin is written with Python 3.6 syntax, and the code has to be parsed before it can run to check the Python version, which already causes a syntax error. @dimbleby checks the Python version in vimscript, so it doesn't have this problem. It also doesn't needlessly complain at every startup, only when I'm actually trying to use the plugin.

I find this rather important, as I like to have a single vim configuration that I clone everywhere, but I may run a relatively ancient vim on some HPC cluster I'm logged in to. In that case, I don't care if some of the plugins don't work, but they should fail gracefully, not with a list of error messages that need to be confirmed with Enter every time the editor starts.

blueyed:
FWIW: this provides a good enough command for me, supporting (visual) ranges also:
```vim
function! s:black(line1, line2, ...) range
  normal! m`
  exe a:line1.','.a:line2.'!black --quiet '.join(a:000).' -'
  normal! ``
  if v:shell_error
    let error = getline('.')
    echohl ErrorMsg | echom error | echohl None
    undo
  endif
endfunction
command! -range=% -nargs=* Black :call s:black(<line1>, <line2>, <f-args>)
```

dimbleby:
Yeah - I'm the originator of this pull request and even I don't use 'my' version of the Vim plugin any more.  (I just set `formatprg` and be done).  

While I think that dimbleby/black.vim is an improvement on the original per the reasons in this thread, I wonder whether the sensible thing from this project's point of view wouldn't be to drop out of the editor plugin game altogether.

- the Vim plugin is a slight oddity in an otherwise conventional python project
- why stop at Vim?  Would ambv/black accept a pull request to add (say) a Sublime plugin to this codebase?  (I guess not)
- these days there are anyway plenty of other plugins providing this function in the Vim world 
  - one can use black as an ale fixer
  - vim-autoformat has code for black
  - @blueyed's solution looks sensible
  - per my current practice there's always `formatprg`.  
  - I've probably missed several more

So I'd suggest removing the plugin code altogether, maybe pointing at one or two of the above instead, and stop at that.

illfated:
Just a friendly reminder: this PR branch needs a rebase to resolve the file conflict.

dimbleby:
@illfated No-one in this thread wants the pull request in its current form any more, the conversation has moved on.  I don't intend to rebase.

cooperlees:
This PR has gone stale. https://github.com/psf/black/tree/master/ftplugin/python/black.vim 404's. Closing.

PR #417, opened 2018-07-22
Remove mappings from Vim plugin.

dimbleby:
They clashed with standard mappings.  Simplest just to let users define
their own.


ambv:
Thanks! ✨ 🍰 ✨

PR #418, opened 2018-07-23
added instructions for PyCharm File Watcher setup

beniwohli:
With these steps, PyCharm will run black on every file save.

Not suuuper happy with the amount of indentation, but it's the best I could come up with without adding a new section.


ambv:
Thanks! ✨ 🍰 ✨

PR #430, opened 2018-07-31
not enforcing python3.6 for precommit hook

miki725:
this should allow precommit hooks to be used with py37

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,5 @@
     description: 'Black: The uncompromising Python code formatter'
     entry: black
     language: python
-    language_version: python3.6
+    language_version: python3
     types: [python]


ambv:
I understand that the problem is that we should be using Python 3.7 instead. However, the change as is will inevitably end with somebody getting Python 3.5 and then complaining here. Anthony, any way we can specify *minimum* version?

asottile:
Not currently, `pre-commit` itself doesn't actually know or care about the values of `language_version` and just passes them through to the underlying tool (`virtualenv` / `nodeenv` / `rbenv` / etc.).

tbh, `language_verison: python3` and then docs about requiring python3.6 and how to set that probably makes the most sense.  At least for macos we've had to override this in our repositories to `language_verison: python3`

ambv:
@miki725 is right, Black will just fail to install on an old Python 3 so we're good anyway. Let's have this.

ambv:
Thanks! ✨ 🍰 ✨

miki725:
no problem. glad to help, at least in some small way. and thanks for merging!

btw the project is awesome!!! thanks for sharing it with everyone

PR #433, opened 2018-08-02
vim: Restore cursor/window position after format

Jonty:
Without this the cursor jumps to the top of the window after formatting occurs.


ambv:
Cool, this problem bugged me, too, but never gotten to find out how to implement it. Thanks! ✨ 🍰 ✨

PR #443, opened 2018-08-11
Fix minor typos

ceh:
This is a minor proposal to align the style of comments somewhat.

(Most of the changes being found by a furious comment/documentation formatter that I'm trying to bootstrap.)

This is of course very subjective, and may spur bikeshedding. Although I found that the majority of the code in black were already on par with the changes being made.

Please just close this PR if these kind of styling attempts are frowned upon.

### DIFF
--- a/black.py
+++ b/black.py
@@ -169,7 +169,7 @@ def read_pyproject_toml(
     "--line-length",
     type=int,
     default=DEFAULT_LINE_LENGTH,
-    help="How many character per line to allow.",
+    help="How many characters per line to allow.",
     show_default=True,
 )
 @click.option(
@@ -1150,7 +1150,7 @@ class Line:
             self.remove_trailing_comma()
             return True
 
-        # Otheriwsse, if the trailing one is the only one, we might mistakenly
+        # Otherwise, if the trailing one is the only one, we might mistakenly
         # change a tuple into a different type by removing the comma.
         depth = closing.bracket_depth + 1
         commas = 0
@@ -1363,7 +1363,7 @@ class EmptyLineTracker:
                 newlines = 1
             elif current_line.is_class or self.previous_line.is_class:
                 if current_line.is_stub_class and self.previous_line.is_stub_class:
-                    # No blank line between classes with an emty body
+                    # No blank line between classes with an empty body
                     newlines = 0
                 else:
                     newlines = 1
@@ -3059,7 +3059,7 @@ class Report:
         - otherwise return 0.
         """
         # According to http://tldp.org/LDP/abs/html/exitcodes.html starting with
-        # 126 we have special returncodes reserved by the shell.
+        # 126 we have special return codes reserved by the shell.
         if self.failure_count:
             return 123
 


black.py
```diff
@@ -2199,20 +2197,20 @@ def right_hand_split(
     body.should_explode = should_explode(body, opening_bracket)
     bracket_split_succeeded_or_raise(head, body, tail)
     if (
-        # the body shouldn't be exploded
+        # The body shouldn't be exploded.
```
zsol: I know this is subjective but for me these read better in the original style. I read all these comments in this `if` expression as if they were a single huge sentence, but introducing those full stops break this for me at least. I would capitalize the first comment and put a `.` at the end of the last one, but otherwise leave everything as is.

zsol:
Looks good to me in general, let me know what you think of my slight bikeshedding attempt in the middle.

black.py
ceh: Updated. Thanks for the comment!

black.py
```diff
@@ -40,7 +40,6 @@
 import click
 import toml
 
-# lib2to3 fork
```
ambv: Please don't remove those.

black.py
```diff
@@ -56,7 +55,6 @@
 CACHE_DIR = Path(user_cache_dir("black", version=__version__))
 
 
-# types
```
ambv: ...and those.

black.py
```diff
@@ -319,7 +317,7 @@ def main(
                 gen_python_files_in_dir(p, root, include_regex, exclude_regex, report)
             )
         elif p.is_file() or s == "-":
-            # if a file was explicitly given, we don't care about its extension
+            # If a file was explicitly given, we don't care about its extension.
```
ambv: This is a good change...

black.py
```diff
@@ -455,7 +453,7 @@ def reformat_one(
             loop.add_signal_handler(signal.SIGINT, cancel, pending)
             loop.add_signal_handler(signal.SIGTERM, cancel, pending)
         except NotImplementedError:
-            # There are no good alternatives for these on Windows
+            # There are no good alternatives for these on Windows.
```
ambv: ...and so is this, esp. that the starting letter was already a capital...

black.py
```diff
@@ -785,13 +783,13 @@ def show(cls, code: Union[str, Leaf, Node]) -> None:
 STARS = {token.STAR, token.DOUBLESTAR}
 VARARGS_PARENTS = {
     syms.arglist,
-    syms.argument,  # double star in arglist
-    syms.trailer,  # single argument to call
+    syms.argument,  # Double star in arglist.
+    syms.trailer,  # Single argument to call.
     syms.typedargslist,
-    syms.varargslist,  # lambdas
+    syms.varargslist,  # Lambdas.
```
ambv: But those above? Those aren't sentences, they aren't meant to be sentences. They are semantic explanations of the node codes in this context.

ceh:
Cool. Thanks for the thoughtful feedback @ambv. I updated the PR to only fix the minor typos.

Sorry for the noise.

ambv:
Thanks, this is helpful!

PR #447, opened 2018-08-14
write to cache even if the file is unchanged

JelleZijlstra:
Maybe I'm missing something, but it seems useful to write to the cache even if the file is unchanged, so that the next time we run Black, we don't need to look at the file.

However, there is no need to write to the file if the result is already cached.


JelleZijlstra:
#453 supersedes this.

PR #453, opened 2018-08-16
write cache in --check mode

### LINKED ISSUE #448: `--check` mode does not write cache
I noticed that the cache is not written when `--check` mode is on. This makes sense if the check fails, but if the file being checked is well-formatted, it would seem to be safe to write the cache. I propose that in `--check` mode we write to the cache only if the checked file is well-formatted. 

My proposed behavior is useful when running Black in CI with `--check`, where we will frequently check the same file, and almost all will usually be well-formatted.

JelleZijlstra:
Fixes #448.

This diff makes us always write to the cache in normal mode, except
if the file is already in the cache, and it makes us write to the
cache in --check mode if the file is already well formatted.

I also fixed some related docstrings.

WriteBack.NO is now used only in tests.

### DIFF
--- a/black.py
+++ b/black.py
@@ -94,11 +94,12 @@ class WriteBack(Enum):
     NO = 0
     YES = 1
     DIFF = 2
+    CHECK = 3
 
     @classmethod
     def from_configuration(cls, *, check: bool, diff: bool) -> "WriteBack":
         if check and not diff:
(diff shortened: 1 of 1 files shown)


ambv:
WHAAA COVERAGE DROPPED BY 0.06%.

PR #454, opened 2018-08-16
normalize numeric literals

### LINKED ISSUE #452: Proposal: Add underscores in long numeric literals.
I am bad at counting zeros, so it's hard for me to tell at a glance what value "100000000" is. It would be nice if Black automatically added underscores (in versions of Python that support it), so that this would become "100_000_000".

JelleZijlstra:
Fixes #452

I ended up making a couple of other normalizations to numeric literals
too (lowercase everything, don't allow leading or trailing . in floats,
remove redundant + sign in exponent). I don't care too much about those,
so I'm happy to change the behavior there.

For reference, here is Python's grammar for numeric literals:
https://docs.python.org/3/reference/lexical_analysis.html#numeric-literals

I ran black with this patch on CPython's test suite (which I'm guessing
provides a good variety of numeric identifiers) and found no new crashes.


ambv:
Amazing. Thanks! ✨ 🍰 ✨

PR #460, opened 2018-08-18
blackd: a HTTP server for blackening

### LINKED ISSUE #349: blackd: a server mode to circumvent startup time costs
I love black, and have championed its adoption where I work (ravelin.com).

However, it is currently not a viable save hook in the same way `gofmt` and `rustfmt` are: taking into account the startup time of the Python interpreter, it is simply too slow. After every save, I need to be conscious that I need to wait a second or two for black to do its thing and let my editor autoreload.

Would you be open to adding a server/daemon mode, where a persistent instance of Black is running and listening to formatting requests?

(Of course, ideally we would compile Black into a very quickly starting binary, but that is probably quite a difficult task.)

zsol:
This implements https://github.com/ambv/black/issues/349#issuecomment-411396604

As expected, it shaves off a couple hundred milliseconds (the python startup time) when individually formatting files. In my extremely dubious testing, this takes about 16-17 ms (after starting `python blackd.py`):

```
curl http://localhost:45484 -d hello
```

While doing it the old-fashioned way takes about 280-290 ms:
```
echo hello | python black.py -
```

I'm putting this early version up to get feedback, but will add 

- [x]  a `--bind-host` and a `--bind-port` option
- [x] 100% test coverage
- [x] CI config updates
- [x] timestamps in the server output

before merging.

Fixes #349


jgirardet:
Adding the check option would also be great. Is it planned ? Need a PR ?

blackd.py
```diff
@@ -0,0 +1,116 @@
...
+        req_bytes = await reader.read()
+        black.out(f"[{peer}] got {len(req_bytes)} bytes")
+        req_str = req_bytes.decode("utf8")
+        formatted = black.format_file_contents(
+            req_str, line_length=line_length, fast=fast, mode=mode
+        ).encode("utf8")
+        writer.write(b"OK\n" + formatted)
+        black.out(f"[{peer}] format done, sending {len(formatted)} bytes")
+    except black.NothingChanged:
+        writer.write(b"OK\n" + req_bytes)
```
JelleZijlstra: Would it be cleaner to use a different status string for this case? There could be a file that just only contains an int literal.

zsol:
@jgirardet I didn't plan to include `--check` as it wouldn't have any benefit. `blackd` is not modifying files, which is the primary reason for that flag's existence. I'll change the status string so it's easier to tell when there was any formatting applied.

jgirardet:
Of course, my bad. I was thinking about plugin usage with blackd and I merged twice in my head...

zsol:
@ambv convinced me to do this on top of HTTP which is a good middleground between our own TCP protocol and something like the language server protocol

jgirardet:
Is there a reason to put parameters like --py36 or others in server_side rather than request_side ?

This force to always reload blackd when changing current project work if parameters are differents between project.

If this a performance issue It seems a good option otherwise maybe it could be discussed more ?

ambv:
This looks very promising.

I think `blackd` should accept options for --py36, -S, --line-length as request parameters instead of hard-coding this to an entire daemon. Ideally we'd like to have a single daemon that can provide us with any request.

ambv:
Haha, what @jgirardet said essentially.

jgirardet:
@zsol I can take care of it if you want

zsol:
Agreed, these should be part of the request. I'll take care of it, thanks @jgirardet  :)

zsol:
Ready to review

README.md
```diff
@@ -744,6 +745,71 @@ affect your use case.
...
+`charset` is specified, `blackd` assumes `UTF-8`.
+
+There are a few HTTP headers that control how the source is formatted. These
+correspond to command line flags for *Black*. There is one exception to this:
+`X-Protocol-Version` which if present, should have the value `1`, otherwise the
+request is rejected with `HTTP 501` (Not Implemented). 
+
+The headers controlling how code is formatted are:
+
+ - `X-Line-Length`: corresponds to the `--line-length` command line flag.
```
JelleZijlstra: Why aren't these just URL parameters, like `POST /?line_length=42&skip_string_normalization=1`? It's going to be a little easier for most clients to set a URL parameter rather than an HTTP header.

zsol: The reason I chose headers is I think there's less confusion around them compared to POST requests with query parameters. I remember being bitten by frameworks trying to be smart and merging query parameters with the request body.

For most at least somewhat sophisticated HTTP client APIs, AFAIK the only difference between the two approaches is passing a string->string mapping as the `headers` parameter vs `params` in the API call. I don't think this is a big deal.

ambv: I agree with Zsolt, GET params on POST requests are non-standard. The typical way is to form-encode them and this is admittedly more work for us since we would really need to support multipart/form-data requests at this point.

I agree with Jelle that using headers for this control is kinda weird, too. But this is definitely the easiest thing for us to support on both ends.

blackd.py
```diff
@@ -0,0 +1,93 @@
...
+        fast = False
+        if request.headers.get(FAST_OR_SAFE_HEADER, "safe") == "fast":
+            fast = True
+        mode = black.FileMode.from_configuration(
+            py36=py36, pyi=pyi, skip_string_normalization=skip_string_normalization
+        )
+        req_bytes = await request.content.read()
+        charset = request.charset if request.charset is not None else "utf8"
+        req_str = req_bytes.decode(charset)
+        formatted = black.format_file_contents(
```
ambv: This needs to happen in a ProcessPoolExecutor, otherwise blackd is not concurrent at all.

blackd.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def make_app() -> web.Application:
+    app = web.Application()
+    app.add_routes([web.post("/", handle)])
+    return app
+
+
+async def handle(request: web.Request) -> web.Response:
+    try:
```
ambv: This "try/except" block is too far reaching.

blackd.py
zsol: good catch

blackd.py
zsol: Apparently any uncaught exceptions here will cause no response to be sent back to the client. Arguably this should be a middleware, and I'm happy to refactor it that way. Did you have any other concerns?

README.md
```diff
@@ -872,7 +942,6 @@ More details can be found in [CONTRIBUTING](CONTRIBUTING.md).
 
 * fixed formatting of lambda expressions with default arguments (#468)
 
-
```
ambv: Why though.

ambv:
```
0.31s$ coverage run tests/test_black.py
  File "tests/test_black.py", line 26, in <module>
    from aiohttp.test_utils import TestClient, TestServer
ModuleNotFoundError: No module named 'aiohttp'
```

Since blackd is an extra, we need to make the tests for it skippable if aiohttp is not installed. They shouldn't crash on import.

README.md
```diff
@@ -745,6 +746,73 @@ affect your use case.
...
 This can be used for example with PyCharm's [File Watchers](https://www.jetbrains.com/help/pycharm/file-watchers.html).
 
+## blackd
+
+`blackd` is a small HTTP server that exposes *Black*'s functionality over
+a simple protocol. The main benefit of using it is to avoid paying the
+cost of starting up a new *Black* process every time you want to blacken
+a file.
+
+### Usage
```
ambv: You need to mention `Black` needs to be installed with `pip install black[d]` rather than just `pip install black`.

black.py
```diff
@@ -90,6 +90,10 @@ class CannotSplit(Exception):
     """
 
 
+class InvalidInput(Exception):
```
ambv: Should subclass ValueError though.

black.py
```diff
@@ -90,6 +90,10 @@ class CannotSplit(Exception):
...
+    """Raised by :func:`lib2to3_parse` when its input fails all parse attempts."""
```
ambv: You shouldn't say *who* raises the exception but what it signifies. Who raises the exception can change in the future.

zsol: Fair enough, will change the one above this as well, then.

tests/test_black.py
```diff
@@ -1289,6 +1318,195 @@ def test_shhh_click(self) -> None:
             except RuntimeError as re:
                 self.fail(f"`patch_click()` failed, exception still raised: {re}")
 
+    @async_test
+    async def test_blackd_request_needs_formatting(self) -> None:
+        if not has_blackd_deps:
+            self.skipTest("blackd's dependencies are not installed")
+            return
```
ambv: unittest has this decorator called @unittest.skipIf that you could use here instead of this.

tests/test_black.py
```diff
@@ -1289,6 +1318,195 @@ def test_shhh_click(self) -> None:
...
+            self.assertEqual(response.status, 204)
+            self.assertEqual(await response.read(), b"")
+
+    @async_test
+    async def test_blackd_request_syntax_error(self) -> None:
+        if not has_blackd_deps:
+            self.skipTest("blackd's dependencies are not installed")
+            return
+        app = blackd.make_app()
+        async with TestClient(TestServer(app)) as client:  # type: ignore
```
ambv: Looks like mypy is unhappy with the type ignores. Did you put them because flake8-mypy wanted them?

ambv:
Coverage fell a lot because we aren't installing blackd on CI. We should be installing blackd on CI and test with it.

tests/test_black.py
zsol: When `aiohttp` is installed, mypy says
```
tests/test_black.py:1325:20: error: Call to untyped function "TestClient" in typed context
```
but flake8-mypy still keeps complaining about the unused type: ignore.

When `aiohttp` is not installed, the ignores are unnecessary. Any tips for working this around?

ambv: Make sure we install the blackd extra on CI, we need it anyway for coverage... and, you know, continuous integration of blackd tests ;-)

zsol: Yeah I forgot that when splitting it out to an extra.
For the type errors I'll just add aiohttp ignores in mypy.ini 😢

ambv: How? mypy.ini module-specific sections are about errors raised *within* said modules, not about their usage outside.

zsol: Well, this did the trick: https://github.com/ambv/black/pull/460/commits/03b77bd0ea3f8803aaf80f8f439f7468cec0c02e

PR #463, opened 2018-08-20
wrap atoms in invisible parens to split adjacent strings

### LINKED ISSUE #441: Multi-line string separated by \ results in line exceeding line length
Operating system: macOS Sierra (10.12.6)
Python version: 3.7.0
*Black* version: 18.6b4
Does also happen on master: yes (but I also see one test failure on `master` so I might be doing this wrong)

Input file:

```python
def myfn():
    val = "string"
    my_str_w_long_name = f"This is a string string string string string {val}"\
                         f"string string string string string {val}"
```

Output file:

```python
def myfn():
    val = "string"
    my_str_w_long_name = f"This is a string string string string string {val}" f"string string string string string {val}"

```

Unexpected outcomes:
* Two separate f-strings in one line
* Resulting line exceeds max line length
* Additional empty line added making flake8 say: `W391: blank line at end of file`

Other observations:
* Does not happen when the second line of the string assignment also terminates in `\`
* Does not happen when the `\` isn't there
* Does happen when there is a space character before the `\`

JelleZijlstra:
Fixes #441.

Adjacent string literals, which are concatenated by Python, create atom nodes rather than expr nodes like actual operators, so we weren't wrapping them in invisible parens and we weren't considering them for splitting into multiple lines.


JelleZijlstra:
I ran this on our codebase and found one example that seems unfortunate:
```
-    uid_to_followed_tid = {}  # long comment that makes the line long enough to split
+    uid_to_followed_tid = (
+        {}
+    )  # long comment that makes the line long enough to split
```
But Black already behaves similarly in cases like this when there's for example an int being assigned, so I don't think this is a problem.

PR #474, opened 2018-08-22
Add reference to atom-python-black

quic0:
Pull request #456 linked a Black for Atom named [atom-black](https://github.com/hauntsaninja/atom-black) in the Readme.  A plugin named [python-black](https://github.com/mikehoyio/atom-python-black) already existed. I suggest to link both in the Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -727,7 +727,7 @@ the [Python Language Server](https://github.com/palantir/python-language-server)
 
 ### Atom/Nuclide
 
-Use [atom-black](https://github.com/hauntsaninja/atom-black).
+Use [atom-black](https://github.com/hauntsaninja/atom-black) or [python-black](https://github.com/mikehoyio/atom-python-black).
 
 
 ### Other editors


quic0:
That makes sense. 

Both packages provide similar functionality:
* Formatting with a shortcut
* Toggle for formatting on save
* Format shortcut in `menu -> packages`.

Minor advantages for `python-black`:
- Supports non-default line length
- Python-black is more commonly used. 422 downloads vs 47 downloads.

Python-black would be my preference, since:
- I was using it already (and so were others),
- It supports non-default line lengths.

In the end, it's your choice.

hauntsaninja:
Hello, author of atom-black here. There's no sense in having two plugins; I clearly could have googled better. I would just archive mine, but it has some minor advantages (better handling of unsaved files, support for all of black's options by changing the black command used). That being said, this repo isn't the place to discuss this. How about I contact the author of atom-python-black and attempt to merge the relevant changes in, and update this PR once things have been resolved?

ambv:
If you can resolve this between the authors of the two plugins and then let me know, that's definitely the way to go. Keeping my fingers crossed!

PR #481, opened 2018-08-27
ISSUE_TEMPLATE.md: Add mention of online formatter

madig:
People can try out https://black.now.sh/?version=master to test against master. That should make issue reporting easier.

See https://github.com/jpadilla/black-playground/issues/6#issuecomment-416088863. Thanks @jpadilla!


zsol:
Nice, thanks!

PR #483, opened 2018-08-28
Change my email in the README

treuherz:
Would prefer my personal email here. I realise it's still in the git log but c'est la vie.


treuherz:
Cheers!

PR #490, opened 2018-08-28
Update the Poetry section in pyproject.toml

sdispater:
This PR updates the `tool.poetry` section of the `pyproject.toml` file to properly package the project.

This PR should make Poetry the only tool needed to develop and package the project and as such the following files should no longer be needed:

- `MANIFEST.in`
- `Pipfile` and `Pipfile.lock`
- `setup.py`
- `setup.cfg`

I did not remove them in the PR since they might still want to be used by the maintainers, so I let this decision to them.

Also, this will now package the `*.pyi` files, which is not currently the case. If this is not desired, it's possible to exclude them by adding the following line to the `tool.poetry` section:

```toml
exclude = ["**/*.pyi"]
```

Also, if the maintainers want to only rely on Poetry I am willing to update the necessary files (like the CI configuration files) to make it work.

Finally, I added the `pyproject.lock` file to the `.gitignore` file because `black` is a library and as such does not need it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ __pycache__
 black.egg-info
 build/
 dist/
+pyproject.lock
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,11 @@ classifiers=[
     "Topic :: Software Development :: Libraries :: Python Modules",
     "Topic :: Software Development :: Quality Assurance",
 ]
-# include = ["blib2to3/**/*.py", "blib2to3/*.txt", "blib2to3/LICENSE"]
+
+packages = [
+    {include = "black.py"},
+    {include = "blib2to3"}
+]
 
 [tool.poetry.scripts]
 black = "black:main"


jgirardet:
I think black is mostly used as standalone runner than a library so pyproject.lock should be keep.

About this PR could the maintainers tell us why there are Pipfile AND pyproject ? maybe a choice should be done?

sdispater:
@jgirardet This is the same thing. The `pyproject.lock` file will be ignored when packaging `black` and thus whenever it is installed the resulting dependencies might differ from the lock file. Basically, the lock file will only be used when developing or testing `black` but not when installing it from PyPI.

ambv:
@sdispater, thanks for your PR, it took me a while to get to this. I agree with @jgirardet that it's worth keeping pyproject.lock so that CI runs can use the exact same dependencies, esp. the dev ones. I cleaned this up in a follow-up commit.

Oh man, it is indeed **tempting** to remove MANIFEST.in, setup.py, setup.cfg, Pipfile, and Pipfile.lock. Having one tool for all purposes (development, packaging, publishing, virtualenv management) is the dream.

First we'll need to move all of their functionality to pyproject.toml first and see how Poetry as a command-line tool fits our workflows. What I see for now is:
- we need the ability to set up `scripts` in pyproject.toml in a way that adds the `blackd` tool **only if** the `d` extra is installed (see how setup.py does this now)
- `poetry shell` is not documented, it starts a new shell but doesn't activate the virtualenv
- `packages` in pyproject.toml don't actually get installed with `poetry install` so it's not possible to, say, do `$ poetry run python tests/test_black.py` since it raises ModuleNotFoundError on `import black`
- `poetry develop` is not documented and does not support `-E`

So it looks like we're getting there but are not quite there yet.

It would also be nice for Poetry to support PEP 518-style `[build-system]` section of pyproject.toml so that we never have to generate `setup.py`. But that's not a blocker at this point.

sdispater:
@ambv Yes there is still some work to be done but we are getting there. The next `0.12` release will have many improvements.

> we need the ability to set up scripts in pyproject.toml in a way that adds the blackd tool only if the d extra is installed (see how setup.py does this now)

Honestly, I didn't even know that was possible so I learned something. Adding support for it should not be too difficult, something like this:

```toml
[tool.poetry.scripts]
blackd = {callable = "blackd:main", extras = ["d"]}
```

> poetry shell is not documented, it starts a new shell but doesn't activate the virtualenv

It should, if not it's a bug. And yes `shell` is missing in the documentation. It should be fixed shortly :-)

> packages in pyproject.toml don't actually get installed with poetry install so it's not possible to, say, do $ poetry run python tests/test_black.py since it raises ModuleNotFoundError on import black

`install` still needs improvement and will eventually be merged with `develop`.

> It would also be nice for Poetry to support PEP 518-style [build-system] section of pyproject.toml so that we never have to generate setup.py. But that's not a blocker at this point.

It already exists but has not been advertised yet because it still needs some testing. It will be released with the next `0.12` version. But just so you know it will look like this.

```toml
[build-system]
requires = ["poetry"]
build-backend = "poetry.masonry.api"
```

Issue #512, opened 2018-09-14
Black fails to detect tabbed standalone comment indent correctly


jleclanche:
Operating system: Linux
Python version: 3.7
*Black* version: master

Running black on the following:

```py
def test():
	if 1:
		return True

	# Note: we are not returning False here because a line might have *both*
	# a leading opening bracket and a trailing closing bracket.  If the
	# opening bracket doesn't match our rule, maybe the closing will.
```

Generates:

```py
def test():
    if 1:
        return True

        # Note: we are not returning False here because a line might have *both*
        # a leading opening bracket and a trailing closing bracket.  If the
        # opening bracket doesn't match our rule, maybe the closing will.
```

ambv:
This is a duplicate of #262.

PR #513, opened 2018-09-14
[WIP] Add a --use-tabs argument to use tabs instead of spaces

### LINKED ISSUE #47: Support for tab indentation
Black always output 4 spaces for indentation.

It would be nice to have a --use-tabs or similar to tell it to output tabs instead. By manually switching the '    '*self.depth into '\t'*self.depth in the Line.__str__() method it seems to support this rather nicely.

I'm not sure if there are other edge cases to consider (i see another '    ' literal in the code but i don't understand when it applies)

Would something like this be considered ?

jleclanche:
Fixes #47 

Yes, I know #47 was rejected. This is following up on [this Twitter thread](https://twitter.com/Adys/status/1040389461006266368). I figured it'd be better to just write the code if I were to restart this discussion.

This is a WIP. I'll finish it if I can get a thumbs up on the option itself. Blocking bits:
- [ ] Converting spaces to tabs causes instability errors with standalone comments following indented lines (only on the first pass). I tried my hand at debugging it and found the culprit: #512. Tried to fix it but been having a hard time, and it's an issue right now either way.
- [ ] Again in spaces<->tabs conversion, indented docstrings get mangled. This is an issue in Black today when going from tabs to spaces. [PEP0257](https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation) defines the algorithm to handle docstring indentation; Black should probably implement it (but it kinda sucks to have to touch the AST). This is probably out of scope of this PR as well.
- [ ] Tests are missing
- [ ] Documentation/changelog untouched

## Rationale

I'm going to avoid trying to spark a debate on whether tabs are better than spaces, as I don't believe that's a requirement in order to include this option. Regardless whether they are or not, projects do use them and currently Black is simply not available for those projects.  Development on prettier-python (which supports tabs) has slowed down, primarily because of Black. So this leaves the Python ecosystem in a bit of a shitty place for tab users.

Addressing in advance the main arguments I've heard against this landing…

> Argument 1: "Black's philosophy is not to be configurable".

I'm a big fan of very carefully evaluating any new option, in any medium. Too many options have very negative consequences: Analysis paralysis, bikeshedding, confusion, added maintenance cruft, etc...
I wrote a bit in the linked thread about what I called "bikeshed bullshit", which has been omnipresent in the linting&formatting space -- see tools like pylint/jslint/tslint, yapf, etc which have hundreds of options for whether there's whitespace and/or commas in a variety of combinatorial situations.
These aren't productive options; they exist because some people subjectively prefer code looking one way versus another.

Pragmatically speaking, there are reasons to use tabs. Pragmatically speaking, Black isn't available to the chunk of the Python ecosystem. This is what I aim to solve.
Black itself is pragmatic and admits that there can be *functional* reasons for certain code style options by having a --line-length option ("paid by the LOC"), and by having ways to disable the autoformatter for chunks of the code.

So Black's philosophy isn't "one codestyle to rule them all", and it shouldn't be. I think it's akin to prettier where there are no useless options. No "bikeshed bullshit", so to say. I believe there are excellent reasons to keep configuration to the absolute minimum, and that's been prettier's mantra as well.
Prettier does have a few more configuration options, for very pragmatic reasons (compatibility with other tools, primarily). It's now used by a massive chunk of the JS ecosystem and there's never any bikeshed in any of those projects. It includes --use-tabs. I think Black should too.

> Argument 2: Tabs prevent a consistent and predictable line-length.

*Note: This argument does not apply to "alignment" as neither black nor this PR implements tab alignment*

My suggestion (and implementation) here is to always treat a tab indent as 4 spaces. This essentially neuters that discussion. In practice, that follows the PEP8 4-space rule, and it avoids having a discussion on "oh now do we introduce an indent_size option", which kinda brings us back to "bikeshed bullshit". (If you want to have a different indent size than 4... you should be using tabs, and not impose your indent preference on everyone who reads your code, that's what it's all there for)

On the technical side, I moved the Line rendering functionality to a new `Line.render()` method which can take a `force_spaces=True` argument. This is used when counting characters in a line, so that a tab indent is always counted as 4 spaces.

### DIFF
--- a/black.py
+++ b/black.py
@@ -174,6 +174,14 @@ def read_pyproject_toml(
     show_default=True,
 )
 @click.option(
+    "--use-tabs",
+    is_flag=True,
+    help=(
+        "Use tabs instead of spaces for indentation. "
+        "Tabs are always equal to 4 spaces."
+    ),
+)
+@click.option(
     "--py36",
     is_flag=True,
     help=(
@@ -280,6 +288,7 @@ def read_pyproject_toml(
 def main(
     ctx: click.Context,
     line_length: int,
+    use_tabs: bool,
     check: bool,
     diff: bool,
     fast: bool,
@@ -333,6 +342,7 @@ def main(
         reformat_one(
             src=sources.pop(),
             line_length=line_length,
+            tabs=use_tabs,
             fast=fast,
             write_back=write_back,
             mode=mode,
@@ -346,6 +356,7 @@ def main(
                 schedule_formatting(
                     sources=sources,
                     line_length=line_length,
+                    tabs=use_tabs,
                     fast=fast,
                     write_back=write_back,
                     mode=mode,
@@ -366,6 +377,7 @@ def main(
 def reformat_one(
     src: Path,
     line_length: int,
+    tabs: bool,
     fast: bool,
     write_back: WriteBack,
     mode: FileMode,
@@ -381,19 +393,24 @@ def reformat_one(
         changed = Changed.NO
         if not src.is_file() and str(src) == "-":
             if format_stdin_to_stdout(
-                line_length=line_length, fast=fast, write_back=write_back, mode=mode
+                line_length=line_length,
+                tabs=tabs,
+                fast=fast,
+                write_back=write_back,
+                mode=mode,
             ):
                 changed = Changed.YES
         else:
             cache: Cache = {}
             if write_back != WriteBack.DIFF:
-                cache = read_cache(line_length, mode)
+                cache = read_cache(line_length, tabs, mode)
                 res_src = src.resolve()
                 if res_src in cache and cache[res_src] == get_cache_info(res_src):
                     changed = Changed.CACHED
             if changed is not Changed.CACHED and format_file_in_place(
                 src,
                 line_length=line_length,
+                tabs=tabs,
                 fast=fast,
                 write_back=write_back,
                 mode=mode,
@@ -402,7 +419,7 @@ def reformat_one(
             if (write_back is WriteBack.YES and changed is not Changed.CACHED) or (
                 write_back is WriteBack.CHECK and changed is Changed.NO
             ):
-                write_cache(cache, [src], line_length, mode)
+                write_cache(cache, [src], line_length, tabs, mode)
         report.done(src, changed)
     except Exception as exc:
         report.failed(src, str(exc))
@@ -411,6 +428,7 @@ def reformat_one(
 async def schedule_formatting(
     sources: Set[Path],
     line_length: int,
+    tabs: bool,
     fast: bool,
     write_back: WriteBack,
     mode: FileMode,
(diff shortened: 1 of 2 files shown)


ambv:
This is not desired functionality in Black. Thanks for understanding.

jleclanche:
@ambv It's definitely desired functionality seeing as at least a few people desire it. Whether you want it to be merged is something else... but I'd appreciate the chance to at least talk about the reasoning, rather than closing outright after putting in the effort on essentially a no-answer ...

jleclanche:
Okay well clearly this is not welcome as it's being ignored here and on Twitter. And at the end of the day, I still need this functionality, so I'm going to be using this branch. I forked it [into tan](https://github.com/jleclanche/tan) if anyone else wants to use it, much as I hate forking. `pip install tan` and `tan --use-tabs` should work just like black and hopefully play nice with it.

@ambv I will say this, the [code of conduct you yourself added](https://github.com/ambv/black/blob/master/.github/CODE_OF_CONDUCT.md) references the PSF one, which has this Paragraph I believe everybody should do their best to follow (regardless of written rules):

> ### Open
> Members of the community are open to collaboration, whether it's on PEPs, patches, problems, or otherwise. We're **receptive to constructive comment and criticism**, as the experiences and skill sets of other members contribute to the whole of our efforts. We're **accepting of all who wish to take part in our activities**, fostering an environment where anyone can participate and everyone can make a difference.

Leaving it up to you to think about whether being unwilling to even discuss a PR abides by that. I appreciate and admire your work on Black and this left a really bad taste in my mouth.

blooalien:
The lack of this exact option is literally the **only** thing stopping me from using black.

jleclanche:
@blooalien FYI I just published a new version of tan rebased on top of the latest black master.

Thanks to some upstream changes to FileMode, the diff required to implement this is actually very small now: https://github.com/jleclanche/tan/commit/4405c86284f7a8a5beccd9ac1cf028c7f0f79814 (`1 changed file with 40 additions and 11 deletions.`)

blooalien:
@jleclanche Sweet.  Will surely give your fork a try, seeing as how you're still updating it.  Thanks much for that.

DaRoee:
This discussion is kind of weird...
I get it, PEP 8 says we should use spaces, but some of us are working in big projects, that were created years ago (not by us) with a giant codebase and tabs configured. 
I want to implement dark throughout our projects, but I can't do that because the diff will just override every **commit and change** that was ever done. New projects are created with spaces as default, but it's too late for the old main projects. Either way, every normal IDE gives you the option to configure spaces/tabs, with a default for spaces.
I really don't see a reason why this shouldn't be merged, if black gives the options to configure the number of characters in line (although pep 8 recommends a specific number), why shouldn't this be configured as well?

ambv:
The existence of any additional configurable option opens unnecessary space for teams to bikeshed which side of the fence they want to be on. This project aims to solve bikeshedding. It's opinionated. Very much so.

Additionally, the community is (almost) entirely decided on this matter, we don't want to give the alternative the same prominence as the recommended default by placing it next to --line-length=.

The above is strong enough reason. However, there's one more nail to the coffin here, namely the argument you're making yourself: "since there's X as an option, why can't we have Y?" It's a flawed argument, it's a Pandora's box for people wanting to have their pet behavior reproduced by Black under a feature flag.

PR #518, opened 2018-09-18
Automatic test-case detection

zsol:
This PR changes the way test cases are defined. Test cases that assert formatting can now be dropped into `tests/data/python`. For pyi files it's `tests/data/pyi`, and for inputs that need to be forced to python3.6 mode, it's `tests/data/python36`.

For tests that should exercise the infrastructure around the formatting logic (like caching, reporting, diff generation, etc), nothing changes.

TODO before merge:

- [ ] add pytest to Pipfile
- [ ] figure out how to do coverage with pytest
- [ ] change CI to run tests through pytest
- [ ] mention how to add new test cases in `CONTRIBUTING.md`
- [ ] describe test case format in the docs


tests/test_black.py
```diff
@@ -74,6 +74,50 @@ def read_data(name: str, data: bool = True) -> Tuple[str, str]:
...
+        if "# output\n" in src:
+            src, _, expected = (t.strip() + "\n" for t in src.partition("# output\n"))
+        else:
+            expected = src
+    return (src, expected)
+
+
+@pytest.fixture(
+    params=list_fixtures("python"),
+    ids=[str(item.relative_to(THIS_DIR)) for item in list_fixtures("python")],
```
kaste: FWIW `ids` can take a function. Something like this `ids=lambda item: item.relative_to(THIS_DIR)` should work afair 🤞

PR #525, opened 2018-09-21
Add .nox directories to default exclude

tswast:
[Nox](https://nox.readthedocs.io/) is similar to Tox. It creates a .nox directory that contains virtualenv for testing with different Python versions.

### DIFF
--- a/black.py
+++ b/black.py
@@ -50,7 +50,7 @@ from blib2to3.pgen2.parse import ParseError
 __version__ = "18.6b4"
 DEFAULT_LINE_LENGTH = 88
 DEFAULT_EXCLUDES = (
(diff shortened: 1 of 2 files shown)


black.py
```diff
@@ -50,7 +50,7 @@
 __version__ = "18.6b4"
 DEFAULT_LINE_LENGTH = 88
 DEFAULT_EXCLUDES = (
-    r"/(\.git|\.hg|\.mypy_cache|\.tox|\.venv|_build|buck-out|build|dist)/"
+    r"/(\.git|\.hg|\.mypy_cache|\.tox|\.nox|\.venv|_build|buck-out|build|dist)/"
```
ambv: This list was previously sorted.

Issue #522, opened 2018-09-20
Bogus Error Message


AXGKl:
Operating system: (Debian) Linux
Python version: 3.7
*Black* version: 18.6b4
Does also happen on master: yes

First of all: Thanks for black!

Here a funny bogus error message:

```python
black root@ip-10-34-2-19:~r/mdv# cat r.py
errout = lambda *msg: print(" ".join(msg), file=sys.stderr) # <-no error here
correct = 'foo'
def foo():
    if 1:  # <- error here
    i = 2
```
```bash
black root@ip-10-34-2-19:~/mdv# black r.py
error: cannot format r.py: Cannot parse: 2:22: errout = lambda *msg: print(" ".join(msg), file=sys.stderr)
All done! 💥 💔 💥
1 file failed to reformat.
black root@ip-10-34-2-19:~/terminal_markdown_viewer/mdv# black --version
black, version 18.6b4
```

I'm not saying it should format it, I'm only saying that it reports the wrong place for the error, should be the `if1:\ni=2` place.

It seems to be a problem with print() in the ast - if I do `lambda: foo()` it works, if I do `lambda: print()` it does not:

```bash

$ echo -e 'lambda: foo()\nif i:\ni=1' | black -
lambda: foo()
if i:
i=1
error: cannot format -: Cannot parse: 3:0: i=1
All done! 💥 💔 💥
1 file failed to reformat.

$ # same with print function:

$ echo -e 'lambda: print()\nif i:\ni=1' | black -
lambda: print()
if i:
i=1
error: cannot format -: Cannot parse: 1:8: lambda: print()
All done! 💥 💔 💥
1 file failed to reformat.
```

-> a bit annoying when you have big file and and try to find the real place where you messed up...

zsol:
Yes, this is an unfortunate effect of how black tries several different python grammars before giving up and is unable to determine which is the "real" error: #423

AXGKl:
sorry for not having seen #423 - commenting there...

Issue #423, opened 2018-07-27
Syntax errors reported on the wrong line due to multiple grammars tested


noahstier:
Operating system: Ubuntu 16.04
Python version: 3.7.0
*Black* version: 18.6b4
Does also happen on master: yes

For this minimal example, the syntax error is reported on the second line instead of the first:
```python
print(f"spam", ham="eggs")
pancakes = [,,]
```
It happens with at least several other syntax errors too (eg `pancakes = [..]`, `pancakes =`).
It happens no matter how many valid statements are added before, after, or between the lines.
It only happens with the print function, and only when passing it a keyword argument.

JelleZijlstra:
What are you doing exactly? Black is not meant for reporting syntax errors, it's a code formatter.

zsol:
This is because of how black tries multiple possible python grammars when encountering a syntax error. It can't tell the "real" cause of the syntax error, because the last grammar it tries doesn't accept print as a function.

noahstier:
> Black is not meant for reporting syntax errors

Black reports syntax errors, and not by accident. I would argue that it is meant to report syntax errors, and if it reports them incorrectly that's a bug, even if reporting syntax errors is not listed in the mission statement.

> black tries multiple possible python grammars ... the last grammar it tries doesn't accept print as a function

Interesting. Although it apparently does accept print as a function, just not when it is given keyword arguments. Is that a bug in the last grammar?

ambv:
> Is that a bug in the last grammar?

No. Python 2 and Python 3 are different in this regard as you might now. Black can be used to format both Python 2 and Python 3 code. We default to the Python 3 grammar because we're not Luddites. However, for users who are still on Python 2 we try to be helpful to them, too, and when the Python 3 grammar fails to parse, we attempt Python 2 and see what happens.

The problem here is a bit tricky to solve because out of many syntax errors that we see, we don't know which syntax error is "the worst", in other words which is the one we should report to the users.

Let me give you an example:
```py3
async def g():
    print 'Hey'
```

This is clearly an invalid file but Black can't know what the user meant, e.g. which part is wrong and should be addressed by the user. It can be either.

You might think, maybe a sensible heuristic would be to choose the syntax error out of all available errors that comes from the farthest line in the file? That means the parser was able to get through all the previous "errors" that other grammars found problematic! But in this case, if somebody puts a Python 2-only piece of code somewhere high in their file, we will report valid Python 3 code below as invalid syntax. So, there's no silver bullet.

[...]

PR #528, opened 2018-09-23
Offer to choose a certain grammar

AXGKl:
Hi, this addresses #423 and #522, based on the suggestion I made in #423. 

If you accept, the user would have the choice to force Black to use only one grammar when trying to build the AST - so that syntax errors do match the *intended* source code grammar and are not misleading.

I had to do this via a cli switch only and set process wide - passing it through your config machinery proved hard, could not manage to make it w/o tons of tests failing. But I think process wide is ok: When formatting many files through the subproc scheduler, then syntax errors should not be the problem -this is more for live coding and formatting in the editor. In the vim plugin I just set

```
black.grammar_chosen.append('py3') after its import, don't need a conditional.
```

Thanks for Black & hope you == ok with it

image

### DIFF
--- a/black.py
+++ b/black.py
@@ -164,15 +164,14 @@ def read_pyproject_toml(
     return value
 
 
-GRAMMARS = [
-    ["py3", pygram.python_grammar_no_print_statement_no_exec_statement],
-    ["py2", pygram.python_grammar_no_print_statement],
-    ["py2_print_stmt", pygram.python_grammar],
+GRAMMARS: List[Tuple[str, Any]] = [
+    ("py3", pygram.python_grammar_no_print_statement_no_exec_statement),
+    ("py2", pygram.python_grammar_no_print_statement),
+    ("py2_print_stmt", pygram.python_grammar),
 ]
-
-grammar_choices = ["all"]  # the default: we try any, in the order given
+grammar_choices: List[str] = ["all"]  # the default: we try any, in the order given
 grammar_choices.extend([g[0] for g in GRAMMARS])  # click option
-grammar_chosen = []
+grammar_chosen: List[str] = []
 
 
 @click.command(context_settings=dict(help_option_names=["-h", "--help"]))
--- a/black.py
+++ b/black.py
@@ -164,6 +164,17 @@ def read_pyproject_toml(
     return value
 
 
+GRAMMARS = [
+    ["py3", pygram.python_grammar_no_print_statement_no_exec_statement],
+    ["py2", pygram.python_grammar_no_print_statement],
+    ["py2_print_stmt", pygram.python_grammar],
+]
+
+grammar_choices = ["all"]  # the default: we try any, in the order given
+grammar_choices.extend([g[0] for g in GRAMMARS])  # click option
+grammar_chosen = []
+
+
 @click.command(context_settings=dict(help_option_names=["-h", "--help"]))
 @click.option(
     "-l",
@@ -172,6 +183,23 @@ def read_pyproject_toml(
     default=DEFAULT_LINE_LENGTH,
     help="How many characters per line to allow.",
     show_default=True,
+)
+@click.option(
+    "-g",
+    "--grammar",
+    type=click.Choice(grammar_choices),
+    default="all",
+    help=(
+        "Restrict Black to using only *one* grammar when parsing source code."
+        "By default Black tries to apply various valid Python grammars "
+        "(currently: %s) and "
+        "will reformat code based on the first one not producing syntax errors. "
+        "Downside: Black might return error messages "
+        "not fitting your intended grammar when *any* of these detect errors. "
+        "Using the switch you can bind Black to using only one grammar. "
+        "Note: This switch must be passed via the command line."
+        % ", ".join([g[0] for g in GRAMMARS])
+    ),
 )
 @click.option(
     "--py36",
@@ -280,6 +308,7 @@ def read_pyproject_toml(
 def main(
     ctx: click.Context,
     line_length: int,
+    grammar: str,
     check: bool,
     diff: bool,
     fast: bool,
@@ -313,6 +342,8 @@ def main(
     report = Report(check=check, quiet=quiet, verbose=verbose)
     root = find_project_root(src)
     sources: Set[Path] = set()
+    if grammar != "all":
+        grammar_chosen.append(grammar)
     for s in src:
         p = Path(s)
         if p.is_dir():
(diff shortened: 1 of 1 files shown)


ambv:
I'm not sure about this. This exposes internals of the tool to the user and the only reason to do so is to present a better syntax error. This seems like a bad compromise to me. More people will now think about which option they should choose here than are actually affected by the syntax errors.

What I would rather do, and is *somewhat* in line with your thinking, is to only try the Python 3 grammar if `--py36` is passed.

AXGKl:
Hey.

My personal observations following, with a disclaimer: Please do take it as something from a guy who might be doing it wrong, since I'm not using Black for very long - I really don't want to sound "smart" just try to help remove the last WTFs from an otherwise close to perfect tool, which I want to get established in my company, i.e. part of default environments plus that global run of Black over our code base. And I need really good arguments for that.

> 'only....present a better syntax error'

I think you underrate this. I guess that is caused by different workflows of yours and mine.
You seem to have another tool at hand, giving you the right place for the error anyway, so you are not 
dependent on Blacks output? Like an IDE or another plugin in vim (?)

Because if you do *not* have such a tool, it is about right or wrong not about something 'better':

- Most people in my company work on remote servers, i.e. work without IDEs over remote mounts but in vim. 
- Large parts of our code base out there is Py2 - conservative customers ;-)

So: You are in the editor and in Py2, use print as a function - and Black practically 100% of time fails in giving you the right place of your syntax error, misleading you in your edit flow.

Acceptable - if the fix would not be so simple. 

Getting it fixed via the `--py36` switch? I'm ok with it.
I feels a bit funny to have to say --py36 for py2 code but main thing is to get the errors away, so yes, definitely, if you dont like `-g` then sure, drop this PR.

Regarding confusing people: Matter of presentation I think, I made it too prominent in `-h`, true.

I sort of thought you like the `-g`, seeing your cython effort and there are quite a lot of other things out there, e.g. Transcrypt, so maybe in the future such a grammar switch and a plugin system for contributed grammars would make sense to have for you, but that is clearly on you not on me :-)

Anyway, as said: Thanks for Black and hopefully we get rid of the syntax error and your way is fine as well :-)

ambv:
I appreciate your thoughts on this. My reluctance for a command-line option stems from the fact that it makes it harder for people to understand how to use the tool. They will now wonder if `-g` is required. They will sometimes choose the wrong one. They will wonder what `-g` gives them (typically nothing actually).

When I meant making this work with the `--py36` switch, I meant it would in this case assume it's a Python 3 only file. So for users of Python 3.6+, that fixes the problem. This is the main case that I'm interested in solving. I mean, the grammar ordering is "Python 3", "Python 2 without print stmt", "Python 2 with print stmt". So Python 2 users most often are getting the right syntax error message anyway.

ambv:
However, @zsol is actually in your camp. He would like a more granular --py= switch instead of the current --py36 and a future --py38 and so on. I'm not dead set on any option here yet.

AXGKl:
High, thanks for weighing this so thoroughly - at least you give outsiders the feeling they are heard.
=> Good leadership in deed, in this critical phase towards setting a de-facto standard. I've seen many come and go and the decisions in this phase are supercritical.

I thought a bit more about why *I* personally got intrigued by Black - and honestly, it was in deed the 'uncompromising' in the tagline - *that* caused the click even in my head seeing the advantage, being otherwise actually more a freestyler ;-)

Trying to say: I hate meanwhile how I proposed `-g`.
You can't possibly first claim "opinionated" and "uncompromising" - have tons of success by that - and then start offering degrees of freedom for things the normal user even comprehends what they mean (i.e. grammar? wtf?)

Thats this side of the coin.

On the other, I'm very frank now: Here is Black, proposing to the user to give up choices and freedom - for the promise of making better decisions and some guys I look up to confirm the claim => Ok, deal, I'll invest time trying it out - but, hey: That tool **must work** - solid and reliable as a rock - and I got that feeling after a the usual first tests, on some production files, which I guess everybody does while still in the shop.
But then, that error. Before I understood why it happens it was really one big downer, I got huge second thoughts about the idea of letting this thing reformat a significant code base with industrial/telco customers on the receiving end of subtle bugs it would produce. Can I trust it when it obviously cannot even see that print used as a function is correct (as said, I did not understand at that time why, just trying to sync my state last week with you)?

>  So Python 2 users most often are getting the right syntax error message anyway.
I'm not buying that, not in 2018 :-)
Rationale: I think all the kids and younger companies with not much backlog are long at 3.
Users of Py2 are using it still, not because they are hillibillies, who did not hear the shot but professional users with good reasons. Turning print into a function is the least problem and we did it long ago, like most others I think. Sidenote: The real conversion problem is the totally different I/O philosophy in Py3 (which works only because UTF8 is everywhere, thanks god). That is unrelated to Black but preventing professional code bases to upgrade - it is simply not financially feasible, you can trust me on that one...

----

Summing up the arguments up to here, my opinion at this point:

1. `-g` that prominent of an option?
No, collides with Black's main mission statement.
2. `--py36` set but still producing print as a statement related syntax errors, when print is used as a function? 
Bug, imho. Sorry but... no, its a bug :-)
3. py2 print statement as last syntax check, dictating the final error the user is presented with, assuming it matches most use cases?
I object. I would propose to change the default order.
4. Only `--py36`?
Looks strange, I think most users have been at least at 3 minors of PY3, and 3.7 is already out.

-----
Lastly, if you still have the patience, a train of thought regarding the selectable grammar option - not expecting any answer here, just a feed for you to think about:

Like @zsol I think they'll pump out new python minors pretty frequently also the next years... When I'm not bound to 2.7, for me the `--py36` is outdated already. Had to change the vim plugin accordingly...

Having looked superficially into how Black is architectured I must say, I may be not superattracted by the way it passes config around but I'm very impressed by the pluggable ways its components interplay.

=> You could easily cover a lot of more ground, question is if you want it.
The often quoted 'do one thing...' approach or become a platform with pluggable components even from third parties? 
If the latter would be an attractive idea then you need *something like* such a switch, determining internal or even external parsing and also synthesizing modes. E.g. allowing also something like [this](https://github.com/AXGKl/black/commit/d32d70e7b15d21e7b61552c60a0787216804c68f) - parametrizable quotes handling (which I would not even dare to propose to you without you == ok with such an amendment to the general vision).

Not on the command line but maybe only in project setup files.  The vision of one and the same editor setting for any PR in the python world with the Black label was invalidated anyway already with configurable line lengths - variations here cause huge diffs, far more severe than e.g. the comma after **kwargs.

virtuald:
I don't have this problem, but started reading the issue... why not have black install an additional script called 'black2.7' that is intended for parsing python 2.7 scripts (obviously underneath the hood it would be using the same main() function as 'black')? That would make it so that a command line option isn't needed (a goal of the tool) and it follows the way that python is deployed these days. Then, once 2.7 is really truly dead (sadly not yet), it can easily be removed?

zsol:
Since #618 has been merged, I think now's the right time to look at this PR again, if you have time @AXGKl

JelleZijlstra:
Is there still a need for this feature now that we have `--target-version`? As far as I know the main motivation was the fact that we try Python 2 grammars for Python 3 code, and you can prevent that by just limiting your `--target-version` to Python 3.

zsol:
@JelleZijlstra is right, `--target-version` limits the grammars we try, so improves the error message we get from black on syntax errors.

PR #533, opened 2018-09-26
Added a fallback to using a ThreadPoolExecutor

### LINKED ISSUE #524: Semaphore is not implemented.
Relates to #388

Operating system: Android 8.0.0 (OnePlus 3T, Termux)
Python version: ~~3.6.6~~ 3.6.5
*Black* version: 18.6b4

It has come to my attention that there is no semaphore synchronization primitive on Android. This means that black will not work on this platform whatsoever, which is disappointing, especially as I find this to be a fairly useful tool that I use regularly. I cannot get the traceback off of my phone while at work unfortunately, but I believe this issue has been put in before. It occurs on line 341 of black.py when initialising a ProcessPool. In master, this is at https://github.com/ambv/black/blob/master/black.py#L343

I have tested this in the Python console with both the multiprocessing pool and the concurrent.futures process pool. Both complain about said missing synchronization primitives.

[Python issue #3770](https://bugs.python.org/issue3770) discusses this issue, which appears to not just occur on Android, but also on the BSDs. This would mean most likely (haven't tested as I don't use BSD) that both Android and BSD cannot support Black.

Whilst this is not ideal, the following could be a work around for ensuring that this package would actually still work on these platforms. Again, I haven't tested, but both TPE and PPE in concurrent.futures share the same interface.

```py
import concurrent.futures
...
def generate_pool(*args, **kwargs) -> concurrent.futures.Executor:
    try:
        return concurrent.futures.ProcessPoolExecutor(*args, **kwargs)
    except ImportError:
        return concurrent.futures.ThreadPoolExecutor(*args, **kwargs)
...
# 344
executor = generate_pool(max_workers=os.cpu_count())
```

Alternatively, could we just run the tasks in series if this issue is detected, rather than using a pool to perform multiple things at once? As much as it is an inconvenience for things to run slowly or inefficiently, I would much prefer that and be able to use this tool at least somewhat on this platform.

Would this be worth considering?

Cheers 😄

dobilquoo:
This enables devices lacking a hardware semaphore implementation such as some Android Devices (looking at the OnePlus3T here!) to still use black.

Fixes #524

### DIFF
--- a/black.py
+++ b/black.py
@@ -1,6 +1,6 @@
 import asyncio
 from asyncio.base_events import BaseEventLoop
-from concurrent.futures import Executor, ProcessPoolExecutor
+from concurrent.futures import Executor, ProcessPoolExecutor, ThreadPoolExecutor
 from datetime import datetime
 from enum import Enum, Flag
 from functools import lru_cache, partial, wraps
@@ -340,7 +340,13 @@ def main(
         )
     else:
         loop = asyncio.get_event_loop()
-        executor = ProcessPoolExecutor(max_workers=os.cpu_count())
+
+        cpu_count = os.cpu_count()
+        try:
+            executor = ProcessPoolExecutor(max_workers=cpu_count)
+        except ImportError:
+            executor = ThreadPoolExecutor(max_workers=cpu_count)
+
         try:
             loop.run_until_complete(
                 schedule_formatting(


ambv:
Thanks for testing that this fixes the issue for you. This is important to fix but I think we'll go with a "SerialExecutor" instead as described in #524.

PR #534, opened 2018-09-26
blackd ignore pyproject

jgirardet:
I added it because, it was a question I'was  asking myself.


zsol:
@jgirardet can you make a PR against master?

Issue #250, opened 2018-05-22
isort==4.3.4 incompatible with black==18.5b0


mxr:
Operating system: Mac OS 10.13.4
Python version: 3.6.5 (through pyenv)
Black version: 18.4a4
Does also happen on master: yes

---

Given this setup:

```
$ tail -n+1 foo.py .pre-commit-config.yaml setup.cfg
==> foo.py <==
from a_really.really.really.really.really.long.package.namespace import a_really_long_package
from really.really.really.really.really.long.package.namespace import really_long_package as pkg

==> .pre-commit-config.yaml <==
repos:
- repo: https://github.com/pre-commit/mirrors-isort
  rev: v4.3.4
  hooks:
  - id: isort
- repo: https://github.com/ambv/black
  rev: 18.5b0
  hooks:
  - id: black
    python_version: python3.6

==> setup.cfg <==
[isort]
# Using config from
# https://github.com/ambv/black/blame/c7bc22388d30f1ba503eefd574e4bb794749b782/README.md
force_grid_wrap = 0
include_trailing_comma = true
line_length = 88
multi_line_output = 3
```

Running `isort` changes the file to

```python
from a_really.really.really.really.really.long.package.namespace import (
    a_really_long_package,
)
from really.really.really.really.really.long.package.namespace import \
    really_long_package as pkg
```

Running `black` changes the file to

```python
from a_really.really.really.really.really.long.package.namespace import (
    a_really_long_package
)
from really.really.really.really.really.long.package.namespace import (
    really_long_package as pkg
)
```

The difference being the trailing comma and the `\` line break

```diff
--- foo_isort.py	2018-05-22 15:46:53.000000000 -0400
+++ foo_black.py	2018-05-22 15:46:46.000000000 -0400
@@ -1,5 +1,6 @@
 from a_really.really.really.really.really.long.package.namespace import (
-    a_really_long_package,
+    a_really_long_package
 )
-from really.really.really.really.really.long.package.namespace import \
+from really.really.really.really.really.long.package.namespace import (
     really_long_package as pkg
+)
```

If I add this option to `setup.cfg`

```diff
--- setup.cfg	2018-05-22 15:39:19.000000000 -0400
+++ setup.cfg.new	2018-05-22 15:47:41.000000000 -0400
@@ -1,6 +1,5 @@
 [isort]
-# Using config from
-# https://github.com/ambv/black/blame/c7bc22388d30f1ba503eefd574e4bb794749b782/README.md
[...]

PR #547, opened 2018-10-02
Update isort config to use_parentheses instead of combine_as_imports

cas--:
The `combine_as_imports=True` modifies isort style as a side-effect and was not the intended purpose of the suggested change in #250. The problem was that isort was actually replacing the parens with backslash and using `combine_as_imports=True` happened to also produce the same result.

The actual setting should be `use_parentheses` as this tells isort to use parenthesis for line continuation instead of \ for lines over the allotted line length limit and matches precisely what black is outputting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Options:
   -S, --skip-string-normalization
                               Don't normalize string quotes or prefixes.
   -N, --skip-numeric-underscore-normalization
-                              Don't normalize underscores in numeric literals.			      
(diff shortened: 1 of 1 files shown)


cas--:
Can this be reviewed for merging?

cas--:
@ambv Please can you review this PR

cas--:
Anyone?

ambv:
Thanks! ✨ 🍰 ✨

PR #550, opened 2018-10-03
Deploy on ambv/black repo

sethmlarson:


SethMichaelLarson:
I'm unsure how this change would be impacting the 3.8-dev build. Guessing it's unrelated?

PR #553, opened 2018-10-09
set entry to black instead of python3 -m black

jgirardet:
On systems where python3.6 is not default python3  (still many linux distro) running `python3 -m black` will faill, trying using python 3.5 or whathever.


ambv:
Thanks! ✨ 🍰 ✨

PR #562, opened 2018-10-14
Fix "No module named 'tests.test_black'"

polyzen:
Occurs somewhat sporadically. See the comments[] on the AUR package for
more details.

Thanks to Kaizhao Zhang (zhangkaizhao).

[]: https://aur.archlinux.org/packages/python-black/#news


ambv:
Hm, there is no longer a requirement for `__init__.py` files to denote Python packages. So it's weird that this would work *sometimes*. I would like to know more about the issue before we put the file there.

polyzen:
@zhangkaizhao, any chance you could shed some light?

zhangkaizhao:
Hello! @ambv is right. I have just tested successfully without a `__init__.py` file on my up-to-date environment. Just don't know why it failed at that time. Anyway, it works without a `__init__.py` file now. Hi @polyzen I think you can remove that line and then wait a few days to see if anyone reports the `ModuleNotFoundError` error again.

polyzen:
Alright ┐('～`；)┌.

gyscos:
Still getting the module import error with python 3.7.1. Python 3.6 seems to work.

```
% git clone https://github.com/ambv/black/ && cd black
% python --version
Python 3.7.1
% python -m unittest tests/test_black.py

E
ERROR: test_black (unittest.loader._FailedTest)
ImportError: Failed to import test module: test_black
  File "/usr/lib/python3.7/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
ModuleNotFoundError: No module named 'tests.test_black'


Ran 1 test in 0.000s
```

gyscos:
It seems to be caused by a `tests` python package (in my case installed by `python-gdbgui`), which conflicts with `tests.test_black`. Uninstalling this package solves the module import error.

joelnb:
I had the same issue with the Arch package (in my case it was the `markdownify` module which had installed a tests module).

polyzen:
@joelnb, which package provides that module? Can't find a `markdownify` package in the repos nor AUR.

joelnb:
I'm not sure if there is one. I think I installed it using pip, which I normally would have done in a virtualenv but I have a script which uses it and I didn't want to have to enter a virtualenv every time I call the script.

In this case I don't think there is any change needed, I just wanted to notify people about another module which may cause this error.

PR #572, opened 2018-10-17
patch main to ensure click_patch() gets called

wooters:
The `patch_click()` function wasn't called when black was run as a command-line tool. This PR creates a new main function (`patched_main()`) that contains both calls to `patch_click()` and `main()`.


ambv:
Thanks! ✨ 🍰 ✨

PR #588, opened 2018-10-27
remove dead code

fkohlgrueber:
While reading through the source of `black.py`, I found some dead code. As you can see from the diff, I only performed three small changes where I'm pretty sure the unused code is unintentional.

There's also a bunch of functions that don't use some of their parameters, but I didn't know if this was on purpose so that related functions share the same signature. If you want me to remove these unused parameters as well, I'd be happy to do it.

Thanks in advance for your feedback.


ambv:
Thanks! ✨ 🍰 ✨

PR #589, opened 2018-10-29
Feature/issue 414: vim plugin respects pyproject.toml config

### LINKED ISSUE #414: Vim plugin should respect pyproject.toml config
vim-black does not respect configuration in pyproject.toml
Python version: 3.6
*Black* version: 18.3-alpha.0

pyproject.toml:
```
[tool.black]
line-length = 120
```

From vim `:Black` will format the file using the default configuration or the global config in vimrc such as `let g:black_skip_string_normalization = 1`
It would be great if the vim plugin could use the project specific configuration

pappasam:
This pull request makes the Vim plugin respect pyproject.toml without changing any public interfaces (resolves  #414)

I pulled some functionality out of read_pyproject_toml and put it into 2 separate functions: abspath_pyproject_toml and parse_pyproject_toml. These are used by the vim plugin to find and read a relevant pyproject.toml according the the already-established rules of the project.

## Regarding tests

I did not yet add tests for this modification; the original function (read_pyproject_toml) was not explicitly tested and I'm unsure what the preferred testing location would be for a bunch of pyproject.toml files.

I did run the tests using the technique recommended in the contributing guidelines and no existing tests break.

### DIFF
--- a/black.py
+++ b/black.py
@@ -138,6 +138,23 @@ class FileMode(Flag):
         return mode
 
 
+def abspath_pyproject_toml(path_search_start: str) -> Optional[str]:
+    """Find the absolute filepath to a pyproject.toml if it exists"""
+    path_project_root = find_project_root(path_search_start)
+    path_pyproject_toml = path_project_root / "pyproject.toml"
+    return str(path_pyproject_toml) if path_pyproject_toml.is_file() else None
+
+
+def parse_pyproject_toml(path_config: str) -> Dict[str, Any]:
+    """Parse a pyproject toml file, pulling out relevant parts for Black
+
+    If parsing fails, will raise a toml.TomlDecodeError
+    """
+    pyproject_toml = toml.load(path_config)
+    config = pyproject_toml.get("tool", {}).get("black", {})
+    return {k.replace("--", "").replace("-", "_"): v for k, v in config.items()}
+
+
 def read_pyproject_toml(
     ctx: click.Context, param: click.Parameter, value: Union[str, int, bool, None]
 ) -> Optional[str]:
@@ -148,16 +165,12 @@ def read_pyproject_toml(
     """
     assert not isinstance(value, (int, bool)), "Invalid parameter type passed"
     if not value:
-        root = find_project_root(ctx.params.get("src", ()))
-        path = root / "pyproject.toml"
-        if path.is_file():
-            value = str(path)
-        else:
+        value = abspath_pyproject_toml(ctx.params.get("src", ()))
+        if value is None:
             return None
 
     try:
-        pyproject_toml = toml.load(value)
-        config = pyproject_toml.get("tool", {}).get("black", {})
+        config = parse_pyproject_toml(value)
     except (toml.TomlDecodeError, OSError) as e:
         raise click.BadOptionUsage(f"Error reading configuration file: {e}", ctx)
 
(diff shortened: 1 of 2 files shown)


dimbleby:
would be good to reach a conclusion, one way or the other, on #416 before making very much change to the Vim plugin.

RohitK89:
Super excited to get this merged in! I have to stop using the plugin for now because I have project-specific configs which differ from the defaults. I feel like a pleb having to run the command from the CLI. . .

black.py
```diff
@@ -138,6 +138,23 @@ def from_configuration(
         return mode
 
 
+def abspath_pyproject_toml(path_search_start: str) -> Optional[str]:
```
JelleZijlstra: I feel like `find_pyproject_toml` would be a better name for this function. Also, why does it return a str instead of a Path?

plugin/black.vim
```diff
@@ -94,13 +94,44 @@ def _initialize_black_env(upgrade=False):
...
-  fast = bool(int(vim.eval("g:black_fast")))
-  line_length = int(vim.eval("g:black_linelength"))
+
+  path_pyproject_toml = black.abspath_pyproject_toml(vim.eval("fnamemodify(getcwd(), ':t')"))
+  try:
+    config_pyproject_toml = black.parse_pyproject_toml(path_pyproject_toml)
+  except Exception as exc:
+    config_pyproject_toml = {}
+
+  toml_line_length = config_pyproject_toml.get("line_length")
```
JelleZijlstra: It's not great that we have to hardcode all of these options here. Is there any way to avoid having separate code for each option? (Relatedly, #618 will change a few of these.)

plugin/black.vim
```diff
@@ -94,13 +94,44 @@ def _initialize_black_env(upgrade=False):
...
 
 def Black():
   start = time.time()
-  fast = bool(int(vim.eval("g:black_fast")))
-  line_length = int(vim.eval("g:black_linelength"))
+
+  path_pyproject_toml = black.abspath_pyproject_toml(vim.eval("fnamemodify(getcwd(), ':t')"))
+  try:
+    config_pyproject_toml = black.parse_pyproject_toml(path_pyproject_toml)
+  except Exception as exc:
```
JelleZijlstra: Catching any Exception is bad practice. What errors is this looking for specifically?

pappasam:
@JelleZijlstra no worries! Unfortunately, given the lack of configuration options for black, I've stopped using it. Additionally, given the amount of time I spent looking elsewhere for something that respects the pyproject.toml, I ended up writing my own [vim formatting plugin](https://github.com/pappasam/vim-filetype-formatter) that does everything I want for every language.

Feel free to close this pull request, or use it as a jumping off point in case you're interested in its content. Totally agree with your comments regarding exception handling, and with hardcoding options... see my plugin for a well-functioning alternative on POSIX-like systems and apologies for the rush job.

anthrotype:
it looks like this PR is now stale.. any plans to add support for respecting pyproject.toml in the official black.vim plugin?
Or can anybody recommend a good alternative solution?

pappasam:
@anthrotype if you are on a Unix-like system, my auto-formatting plugin has worked well for me. To use black for Python, just use the string "black -q -" when configuring.

https://github.com/pappasam/vim-filetype-formatter

dimbleby:
Several suggestions in [this comment](https://github.com/python/black/pull/416#issuecomment-478103036), in which I propose that it doesn't make a lot of sense to maintain a vim plugin in this repository.

Personally I just use `formatprg`.

anthrotype:
Thanks both, I’ll have a look!

TalAmuyal:
@ambv, @JelleZijlstra, @pappasam, https://github.com/psf/black/pull/1273 has been merged in stead of this one.
Please closes this.

PR #614, opened 2018-11-20
Refactor Travis

JelleZijlstra:
- Run separate jobs for mypy, self-formatting, flake8, and test runs.
- Don't run flake8 in 3.8 because it is broken (and we can't really expect flake8 to always keep up with 3.8 development).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,14 @@ sudo: required
 dist: xenial
 language: python
 cache: pip
+env:
+- TEST_CMD="coverage run tests/test_black.py"
 install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,24 +2,38 @@ sudo: required
 dist: xenial
 language: python
 cache: pip
+env:
+- TEST_CMD="coverage run tests/test_black.py"
 install:
 - pip install coverage coveralls flake8 flake8-bugbear mypy
```
JelleZijlstra: We could speed up builds slightly by only installing what we need, but Travis is pretty fast already and most of this is probably cached anyway, so I don't think it's really worth doing.

Issue #613, opened 2018-11-19
Black autodetection of 3.6 makes broken PyPy3 code


dstufft:
Operating system: macOs Mojave
Python version: 3.7.0
*Black* version: 18.9.0
Does also happen on master: Didn't check, but a scan of the code appears so.

It looks like Black is attempting to auto detect the version of Python that a specific file supports by looking for features that only exist in a certain version of Python. The doc string suggests that it is specifically looking for f strings, numeric underscores, and trailing commas as ``*args`` and ``**kwargs``.

This detection currently generates code that is broken for PyPy3 (which implements 3.5) because PyPy3 has backported f-string support (but not numeric underscores or trailing commas).

I'm not sure what the right answer here is. Certainly I understand the rationale behind auto detecting Python 3.6+ code, however in doing that there are now edge cases where broken code is being generated. Perhaps changing ``--py36`` to be a boolean instead of a flag, and add a ``--no-py36`` flag that projects that support PyPy3 as well would work?

rooterkyberian:
This seems somewhat relevant: https://github.com/ambv/black/issues/419#issuecomment-422090963

AFAIK introduction of option flags into `black`, especially because of some edge cases, seem to be in direct conflict to what this project is about - which is "uncompromising formatting". You would use this to make your code run on PyPy3, other user would use it to disable some things (s)he doesn't like.

Alternative solutions:
* do a feature detection using target interpreter instead of custom options
  * I believe this would reduce complexity of feature detection
  * increase complexity of debugging user reported bugs
* very issue specific solution: wait for PyPy3 to implement trailing comma
  * it seems as one of the easiest things to support and should require much less effort than f-string implementation
  * in the mean time users can just forgo using f-strings and other Python 3.6 (or higher) features if they interpreter does not implement full set

@dstufft  Regardless what happens in black, as a PyPy user it would be good to report it to them as well and that you have a need for this and/or that mixed Python version compatiblity can lead to auxiliary support tools incompatibilities. https://bitbucket.org/pypy/pypy/issues/2827/trailing-commas-are-occassionaly seems like a good place to start.

dstufft:
> AFAIK introduction of option flags into black, especially because of some edge cases, seem to be in direct conflict to what this project is about - which is "uncompromising formatting". You would use this to make your code run on PyPy3, other user would use it to disable some things (s)he doesn't like.

[...]

PR #618, opened 2018-11-23
Add `--target-version` option to allow users to choose targeted Python versions

JelleZijlstra:
This generalizes the `--py36` flag to allow the user to specify exactly which Python versions should be supported by the code Black produces. For example, you can now say `black -t cpy27 -t pypy35 setup.py` to produce code that is compatible with Python 2.7 and PyPy 3.5.

Advantages of this approach include:
- Better error messages: we'll no longer attempt to use a Python 2 grammar if the user has told us to target only Python 3, which would fix the example in #423.
- More granular version detection, fixing the PyPy3.5 issue reported in #613.
- The approach is more generalizable to other future Python features.

I haven't yet fully tested all the ways this could be used, but wanted to put it up to see if others (@ambv, @zsol) are OK with this approach.

### DIFF
--- a/black.py
+++ b/black.py
@@ -126,6 +126,9 @@ class TargetVersion(Enum):
         return self is not TargetVersion.CPY27
 
 
+PY36_VERSIONS = {TargetVersion.CPY36, TargetVersion.CPY37, TargetVersion.CPY38}
+
+
 class Feature(Enum):
     # All string literals are unicode
     UNICODE_LITERALS = 1
@@ -376,7 +379,7 @@ def main(
         else:
             versions = set(target_versions)
     elif py36:
-        versions = {TargetVersion.CPY36, TargetVersion.CPY37, TargetVersion.CPY38}
+        versions = PY36_VERSIONS
     else:
         # We'll autodetect later.
         versions = set()
--- a/black.py
+++ b/black.py
@@ -135,6 +135,7 @@ class Feature(Enum):
 
 
 VERSION_TO_FEATURES = {
+    TargetVersion.CPY27: set(),
     TargetVersion.PYPY35: {Feature.UNICODE_LITERALS, Feature.F_STRINGS},
     TargetVersion.CPY33: {Feature.UNICODE_LITERALS},
     TargetVersion.CPY34: {Feature.UNICODE_LITERALS},
@@ -162,7 +163,7 @@ VERSION_TO_FEATURES = {
 
 @dataclass
 class FileMode:
-    target_versions: Set[TargetVersion]
+    target_versions: Set[TargetVersion] = Factory(set)
     line_length: int = DEFAULT_LINE_LENGTH
     numeric_underscore_normalization: bool = True
     string_normalization: bool = True
@@ -518,7 +519,7 @@ async def schedule_formatting(
         lock = manager.Lock()
     tasks = {
         loop.run_in_executor(
-            executor, format_file_in_place, src, fast, write_back, mode, lock
+            executor, format_file_in_place, src, fast, mode, write_back, lock
         ): src
         for src in sorted(sources)
     }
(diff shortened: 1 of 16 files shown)


black.py
```diff
@@ -111,32 +112,86 @@ class Changed(Enum):
...
-            mode |= cls.NO_NUMERIC_UNDERSCORE_NORMALIZATION
-        return mode
+
+@dataclass
+class FileMode:
+    target_versions: Set[TargetVersion] = Factory(set)
+    line_length: int = DEFAULT_LINE_LENGTH
+    numeric_underscore_normalization: bool = True
+    string_normalization: bool = True
+    is_pyi: bool = False
```
zsol: Have you thought about folding `is_pyi` into `target_versions`?

black.py
```diff
@@ -2991,14 +3084,22 @@ def is_python36(node: Node) -> bool:
...
 
-    return False
+    return features
+
+
+def detect_target_versions(node: Node) -> Set[TargetVersion]:
+    """Detect the version to target based on the nodes used."""
+    features = get_features_used(node)
+    return {
+        version for version in TargetVersion if features <= VERSION_TO_FEATURES[version]
```
zsol: Beautiful ❤️

blackd.py
```diff
@@ -51,22 +52,42 @@ def make_app() -> web.Application:
...
+                        # the corresponding CPython version.
+                        if tag == "pypy" and not hasattr(
+                            black.TargetVersion, version_str
+                        ):
+                            version_str = f"CPY{major}{minor}"
+                        versions.add(black.TargetVersion[version_str])
+                    except (KeyError, ValueError):
+                        return web.Response(
+                            status=400,
+                            text=f"Invalid value for {PYTHON_VARIANT_HEADER}",
```
zsol: Let's have a more specific error message here. End users will likely encounter this if they misconfigure their clients

blackd.py
```diff
@@ -51,22 +52,42 @@ def make_app() -> web.Application:
...
-                    if int(major) == 3 and len(rest) > 0:
-                        if int(rest[0]) >= 6:
-                            py36 = True
-                except ValueError:
-                    return web.Response(
-                        status=400, text=f"Invalid value for {PYTHON_VARIANT_HEADER}"
-                    )
+                for version in value.split(","):
+                    tag = "cpy"
+                    if version.startswith("cpy"):
```
zsol: I don't really like the explicit parsing here. Is it possible to try to match on all possible enum values here?

black.py
JelleZijlstra: Yes, I'm not sure it's worth it though, since pyi mode is not quite like a Python version: you can't target say "pyi and CPython 3.6", it's always just pyi and nothing else.

blackd.py
JelleZijlstra: Yes, it's not great. I tried to work with the following constraints:
- Continue accepting everything the previous code accepted (for backward compatibility). That's why there's the default if the user just put "3" or "2" as the version.
- Don't be too liberal in what I accept, because then future backward compatibility will bind us to continue accepting it.
- Avoid exposing our internal enum values too much. Users shouldn't have to care that we have a separate PyPy 3.5 version, but not one for other PyPy versions.
I'll see what I can change but I'm not sure there's much more I can do keeping these constraints.

blackd.py
JelleZijlstra: Will do

JelleZijlstra:
There's some issue with asyncio in 3.8 that I don't fully understand now—I wonder whether it's even related to this diff, or whether it's just 3.8 instability.

asvetlov:
The observed issue is an aiohttp bug.
I have a plan to fix it in the next aiohttp 3.5.x release

JelleZijlstra:
Thanks for confirming! I was thinking in the same direction.

The reason it came up in this PR is that currently we actually skip all blackd tests because `aiohttp-cors`, which was recently introduced as a dependency, is missing from `setup.py`. I added the missing dep in this PR (not sure why). I backed it out and will make the change in a separate PR.

Issue #622, opened 2018-11-27
Add CORS support to blackd


tuchandra:
**Issue description**
Suppose you have blackd running locally. When sending requests to blackd that originate from another site (i.e., anywhere that's not localhost), you cannot use custom headers, despite the fact that blackd is configured to accept them (e.g., `X-Line-Length` or `X-Python-Variant` -- see the [docs](https://black.readthedocs.io/en/stable/blackd.html#protocol)).

**Explanation**
This is because blackd, by default, does not allow cross-origin requests. (This itself is a property of aiohttp -- see the [docs](https://docs.aiohttp.org/en/stable/web_advanced.html#cors-support).) When sending a cross-origin request to blackd, you'll get a CORS error. If you try to get around this by sending a `no-cors` request, you cannot send custom headers. See this [StackOverflow answer](https://stackoverflow.com/a/45640164) for why this is the case; but this essentially means that if a request comes from an origin different from where blackd is running, the custom headers will not be accepted.

**Example -- how it should always work**
Run blackd locally in one terminal. From another terminal, these will produce the following output:
```
$ curl http://127.0.0.1:45484 -X POST -d 'print("something very long", x, y = 122, z= 8)'
print("something very long", x, y=122, z=8)
```
```
$ curl http://127.0.0.1:45484 -X POST -d 'print("something very long", x, y = 122, z= 8)' -H "X-Line-Length: 40"
print(
    "something very long", x, y=122, z=8
)
```
Both of these are expected behavior. You can verify that the same happens when sending requests in Postman, if you'd like.

**Example -- how it doesn't work from other origins**
Run blackd locally in a terminal. Go into the console on any website on the newest version of Chrome (or probably any other modern browser, but I didn't try). We can use the fetch API (see this [StackOverflow answer](https://stackoverflow.com/a/16786805) for more info) to send requests.

Type this in the console:
```
fetch("http://127.0.0.1:45484", { method : "POST", body: 'print("something very long", x, y = 122, z= 8)' })
```
You'll get a CORS error:
```
Access to fetch at 'http://127.0.0.1:45484/' from origin <website you're on> has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```
We can set the mode to `no-cors` to suppress the error:
```
[...]

PR #627, opened 2018-11-29
Add CORS support to blackd

tuchandra:
See #622. This is my first time contributing to a project that uses pipenv, so please let me know if there's something I did wrong.

Additionally, I'm not sure how to add a test for this. I'd expect it to test that a request sent from a client on a different host works correctly, but I read through the docs for the aiohttp [TestClient](https://docs.aiohttp.org/en/stable/testing.html?highlight=testing#test-client) and couldn't come up with anything. Setting the host param in the constructor for TestClient `async with TestClient(TestServer(app), host=some_string)` gives a TypeError about the host parameter being invalid.

### DIFF
--- a/blackd.py
+++ b/blackd.py
@@ -52,7 +52,7 @@ def make_app() -> web.Application:
         resource.add_route("POST", partial(handle, executor=executor)),
         {
             "*": aiohttp_cors.ResourceOptions(
-                allow_headers=(*BLACK_HEADERS, "Content-Type")
+                allow_headers=(*BLACK_HEADERS, "Content-Type"), expose_headers="*"
             )
         },
     )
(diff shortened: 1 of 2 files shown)


tuchandra:
@zsol, when you get a chance could you give this a look?

blackd.py
```diff
@@ -35,7 +45,18 @@ def main(bind_host: str, bind_port: int) -> None:
...
     executor = ProcessPoolExecutor()
-    app.add_routes([web.post("/", partial(handle, executor=executor))])
+
+    cors = aiohttp_cors.setup(app)
+    resource = cors.add(app.router.add_resource("/"))
+    cors.add(
+        resource.add_route("POST", partial(handle, executor=executor)),
+        {
+            "*": aiohttp_cors.ResourceOptions(
+                allow_headers=(*BLACK_HEADERS, "Content-Type")
```
zsol: We'll probably also want to expose some headers. At least `Content-Type`, but maybe just `*`?

tuchandra: Fine with me; fixed in the most recent commit.

zsol:
Thanks!

PR #632, opened 2018-12-06
Fix multiprocessing support for Windows binary

allhailwesttexas:
The black binary generated for Windows builds would fail on reformatting multiple files due to a Windows-specific multiprocessing issue. Fix by calling freeze_support() as described in [Python docs](https://docs.python.org/dev/library/multiprocessing.html#multiprocessing.freeze_support).

### DIFF
--- a/black.py
+++ b/black.py
@@ -8,7 +8,7 @@ import io
 import itertools
 import keyword
 import logging
-from multiprocessing import Manager
+from multiprocessing import Manager, freeze_support
 import os
 from pathlib import Path
 import pickle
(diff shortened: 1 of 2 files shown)


allhailwesttexas:
I didn't create an issue for this, not sure if that's a required part of the workflow?  If that's preferred, let me know and I'll do it and link this PR to it.

zsol:
You probably want to add this in `blackd.py` as well

allhailwesttexas:
@zsol Added now. At the moment, this binary is not built in CI, but I built it locally with `C:\\Python36\\python.exe -m PyInstaller --clean -F --add-data blib2to3/;blib2to3 blackd.py` and verified that it fails before the fix and works after.

zsol:
Thanks!

PR #633, opened 2018-12-08
Add a --single-quotes option to prefer single quotes

### LINKED ISSUE #594: Add --single-quote option
[Prettier](https://prettier.io), the most successful "opinionated code formatter", probably an inspiration for this project, advertises exactly the same values as this project. In many ways they make even stricter formatting decisions than black.

Yet, even prettier supports a `--single-quote` [option](https://prettier.io/docs/en/options.html#quotes).

Please add a `--single-quote` option to black.

And before immediately closing this ticket, please understand there is a huge need from users to have this option. At least go as far as to search across Github for `pyproject.toml` file's `[tool.black]` and check how many [major projects](https://github.com/Pylons/pyramid/blob/master/pyproject.toml#L6) have to use black with `skip-string-normalization = true`. Or simply listen to your users here.

davvid:
Now there are no excuses not to use black.
Fixes #594

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ Options:
   --pyi                       Format all input files like typing stubs
                               regardless of file extension (useful when piping
                               source on standard input).
+  -s, --single-quotes         Use single quotes during string
+                              normalization.
   -S, --skip-string-normalization
                               Don't normalize string quotes or prefixes.
   -N, --skip-numeric-underscore-normalization
@@ -670,6 +672,7 @@ Commands and shortcuts:
 Configuration:
 * `g:black_fast` (defaults to `0`)
 * `g:black_linelength` (defaults to `88`)
+* `g:black_single_quotes` (defaults to `0`)
 * `g:black_skip_string_normalization` (defaults to `0`)
 * `g:black_virtualenv` (defaults to `~/.vim/black`)
 
@@ -805,6 +808,7 @@ request is rejected with `HTTP 501` (Not Implemented).
 The headers controlling how code is formatted are:
 
  - `X-Line-Length`: corresponds to the `--line-length` command line flag.
+ - `X-Single-Quotes`: corresponds to the `--single-quotes` command line flag.
  - `X-Skip-String-Normalization`: corresponds to the `--skip-string-normalization`
     command line flag. If present and its value is not the empty string, no string
     normalization will be performed.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -79,6 +79,8 @@ Options:
   --pyi                       Format all input files like typing stubs
                               regardless of file extension (useful when piping
                               source on standard input).
+  -s, --single-quotes
+                              Use single-quotes during string normalization.
```
hugovk:
```suggestion
                              Use single quotes during string normalization.
```

black.py
```diff
@@ -199,6 +203,12 @@ def read_pyproject_toml(
         "(useful when piping source on standard input)."
     ),
 )
+@click.option(
+    "-s",
+    "--single-quotes",
+    is_flag=True,
+    help="Use single-quotes during string normalization.",
```
hugovk:
```suggestion
    help="Use single quotes during string normalization.",
```

ambv:
Thank you for your contribution but we won't be adding this to Black.

richardARPANET:
Support for this has been added to https://github.com/odwyersoftware/brunette :+1:

rspadim:
Could the brunette --single-quotes be imported to black?

y4rr:
Is there any explanation as to why this option not being added?

Your main page says,
> Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. **You will save time and mental energy for more important matters.**

Actually, after using your tool, I don't have any time and mental energy left **at all**, because I use single quotes everywhere, as they are easier to type on my keyboard.

harahu:
> Actually, after using your tool, I don't have any time and mental energy left **at all**, because I use single quotes everywhere, as they are easier to type on my keyboard.

Not sure I understand why you feel this is an issue, @y4rr. After all, since `black` will gladly swap out your quotes, there's nothing stopping you from continuing to type single quotes. Nobody is forcing you to _type_ double quotes.

manulera:
This feels really strange. Why not allow maintainers to harmonise string quotes to single if they want to do so? Really sad to see that a working implementation was discarded on the basis of what seems to be personal preference of the lead maintainer.

Zananok:
Black is like the original vim. It's the maintainer's project, not a community project like nvim.
I respect the maintainer's decision and am simply moving to a community project.

My humble suggestion on further friction is to always keep your preference as the default and explain why your opinion is best in the readme.md, yet allow PRs that don't affect your default as a form of extra config when it is highly desirable.

arjun-menon:
I know it's been forever, but I'd appreciate if PSF changes their mind, and merges this PR.

PR #641, opened 2018-12-13
black.vim: put cursor in last line if old position is invalid

suola:
Formatting the following file in vim with `:Black` when cursor is on the last line

```python
print(
    "foooo")
```

produces error

```
Traceback...
...
vim.error: cursor position outside buffer
```

Fix: put cursor on start of last line when this happens.


daturkel:
Confirming that this fixes #734.

Issue #282, opened 2018-05-30
Breaks multiline function type annotation comments


ned:
Operating system: Fedora 27  4.13.9-300.fc27.x86_64)
Python version: 3.6.2
Black version: 18.5b1
Does also happen on master: yes

PEP484 was amended (https://github.com/python/typing/issues/186) to extend Python 2 compatible type annotation comments to allow placing each function parameter on a separate line with its own type annotation and the return type at the end. The motivation for this was to be able to write type annotations which would otherwise extend beyond the maximum line limit. But when it will fit in the maximum line length, Black combines all the arguments onto one line, including the type annotations, which become malformed.

Given this input

```python
def f(a,  # type: int
      b,  # type: str
      c,  # type: bool
      ):  # type: (...) -> None
    ...
```

Black produces this

```python
def f(a, b, c):  # type: int  # type: str  # type: bool  # type: (...) -> None
    ...
```

which has a malformed type annotation.

I think the correct behaviour would be either leave the function parameters on separate lines (do nothing) or restructure the type annotation like this:

```python
def f(a, b, c):  # (int, str, bool) -> None
    ...
```

or if the resultant line is too long, this

```python
def f(a, b, c): 
    # (int, str, bool) -> None
    ...
```

---

This might deserve a separate issue, but it'd be nice if Black did the reverse as well: given a very long function type annotation on one line, annotate each parameter separately.

ambv:
Yes, this is unfortunate and I should probably have special handling for it. Not quite clear how I would special-case it yet.

perrygeo:
Noting that this is a more general issue with inline comments. For example, `black` changes the meaning of this comment intended to be associated with a particular kwarg:

```
$ cat example.py
myfunc(
    a,
    b=1,  # this is critical due to #123
    c=None)

$ black example.py
reformatted example.py

$ cat example.py
myfunc(a, b=1, c=None)  # this is critical due to #123
```

ambv:
@perrygeo, that's intentional.

tony-cox:
There are many cases where neither of the suggestions (apart from 'do nothing') from @ned will work without making a line that is too long.  Usually when I use a multiline type annotation it's because it won't fit on a single line after the function.  That is,

```
def f(a, b, c): 
    # (int, str, bool) -> None
```

in some cases would cause the type annotation `# (int, str, bool) -> None` to extend beyond the character limit.  In these cases, Black should probably do nothing at all.

whatisaphone:
[...]

PR #642, opened 2018-12-15
Add support for special comments in multiline functions

Utsav2:
This is a starter diff to deal with #282, and I'm happy to iterate on it.

We want to ensure that type comments (https://www.python.org/dev/peps/pep-0484/) for Python 2 are formatted in a valid way, so that we can use Black in our organization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.idea/
 .coverage
 _build
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 .coverage
 _build
 .DS_Store
--- a/black.py
+++ b/black.py
@@ -2113,7 +2113,7 @@ def split_line(
 
     # we don't want to split special comments like type annotations
     # https://github.com/python/typing/issues/186
-    has_special_comment: bool = False
+    has_special_comment = False
     for leaf in line.leaves:
         for comment in line.comments_after(leaf):
             if leaf.type == token.COMMA and is_special_comment(comment):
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
gvanrossum: Not sure adding this is a good idea -- I'd use my global .gitignore for this.

JelleZijlstra: Yes, if nothing else this change doesn't belong in this PR.

tests/data/comments6.py
```diff
@@ -0,0 +1,56 @@
+# test type comments
+def f(a, b, c, d, e, f, g, h, i):
+    # type: (int, int, int, int, int, int, int, int, int) -> None
+    pass
+
+
+def f(
+    a,  # type: int
```
JelleZijlstra: Can you also test this with other kinds of arguments (those with defaults, *args, **kwargs, etc.)?

black.py
```diff
@@ -2110,8 +2110,19 @@ def split_line(
         return
 
     line_str = str(line).strip("\n")
-    if not line.should_explode and is_line_short_enough(
-        line, line_length=line_length, line_str=line_str
+
+    # we don't want to split special comments like type annotations
+    # https://github.com/python/typing/issues/186
+    has_special_comment: bool = False
```
JelleZijlstra: No need for the explicit type.

tests/data/comments6.py
```diff
@@ -0,0 +1,56 @@
...
+    b,  # type: int
+    c,  # type: int
+    d,  # type: int
+):
+    # type: (...) -> None
+    e = (
+        a
+        + b
```
JelleZijlstra: Would be better to use longer variable names so you don't need quite as many to get the line to be long enough.

PR #643, opened 2018-12-17
add .gitignore support

### LINKED ISSUE #475: Automatically ignore files in .gitignore
I think a feature to automatically ignore files and directories listed in `.gitignore` would be useful. This file normally contains a long list of auto-generated files or files that are otherwise unnecessary to format. By pruning these big directories, black will run much faster. By adding such a feature, black will require even less configuration as most projects will no longer require passing `--exclude` on the command line. 

Other opinionated code format tools already do this or are discussing this. For example:

JavaScript Standard: https://standardjs.com/#how-do-i-ignore-files

> Paths in a project's root .gitignore file are also automatically ignored.

PHP-CS-Fixer: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3947#issuecomment-408342692

> I think it can be fixed in a broader, generalised way, by parsing (and excluding the results of) ignore files like .gitignore.

jgirardet:
fix #475


jgirardet:
@zsol

black.py
```diff
@@ -190,6 +191,25 @@ def supports_feature(target_versions: Set[TargetVersion], feature: Feature) -> b
...
+    gitignore = root / ".gitignore"
+    if not gitignore.is_file():
+        return value
+    # parse gitignore
+    excluded_in_gitignore = []
+    for line in gitignore.read_text().splitlines():
+        line = line.strip().strip("/")
+        if line and not line.startswith("#"):
+            excluded_in_gitignore.append(  # noqa  W605 incorrect
+                fnmatch.translate(line).lstrip("(?s:").rstrip(r")\Z")
```
ambv: I see what you're doing here but the required surgery on the resulting regular expression sucks a bit. Removing \Z is necessary but won't (?s:...) be harmless?

black.py
```diff
@@ -190,6 +191,25 @@ def supports_feature(target_versions: Set[TargetVersion], feature: Feature) -> b
...
+            )
+    # merge gitignore and others exclude
+    return r"/(" + r"|".join(excluded_in_gitignore) + "|" + value.strip()[2:]
```
ambv: `value` is not necessarily going to be in the format `/(something)/` so assuming that will blow up when somebody just passed `--exclude=.git`.

jgirardet: Indeed. 
the gitignore values could be merged with :  a regex value (default_include and pyproject) or a more simple regex '--exclude=.git'.
That's not obvious to that we should guess which pattern it is to merge our gitigore's items in it. There could be many pattern so it's not very predictable.

Solutions ? :
- never merge gitignore with the others and try to match twice in `gen_python_files_in_dir`: one as it is now and one and one for gitignore
- Enforce an exclude "style".  say to people to always have at least "/()/" in their regex.
- guess the regex pattern but I don't like it.

ambv: I think the first solution will be the most robust one.

black.py
```diff
@@ -190,6 +191,25 @@ def supports_feature(target_versions: Set[TargetVersion], feature: Feature) -> b
     return all(feature in VERSION_TO_FEATURES[version] for version in target_versions)
 
 
+def merge_exclude(ctx: click.Context, param: click.Parameter, value: str) -> str:
+    """Injects .gitignore in excluded files"""
```
gaborbernat: As far as I know pythons fnmatch does not implement gits wildmatch. So using the fnmatc is not a good idea, leaves you exposed to a lot of errors. I would suggest something like https://github.com/ambv/retype/pull/15 via parsepath library.

gaborbernat: @JelleZijlstra @ambv shall we take the implementation from https://github.com/ambv/retype/pull/15 instead then?

jgirardet: If it's ok I'll have some time to finish it this week

gaborbernat: As I said above I don't think using fnmatch is good, we need a proper wildmatch, currently provided via pathspec with recursive ``.gitignore`` discovery. https://github.com/ambv/retype/pull/15/files already does this so we can just take the solution from there and use it here.

jgirardet: I totally agree and I was waiting for a reply from @ambv

jgirardet: I started a new PR #878 with pathspec support.
Sorry It was much easier than continuing this old one.

benjaoming:
Instead of scanning for `.gitignore` and parsing them, would it be possible to invoke `git status --ignored` for instance if black is invoked with `--gitignore`? Seems to be able to save a lot of time and fetch the results from the correct single source of truth?

ambv:
Closing in favor of #878.

ambv:
> would it be possible to invoke git status --ignored

Subprocessing is always slow.

PR #645, opened 2018-12-20
Fix instructions for PyCharm integration

johnnymetz:
The instructions for navigating through PyCharm to enable black integration are slightly off. The first navigation is for a Windows user. However, the latter two are for Mac users; though the latter two are missing the first step in the navigation.

Only difference between PyCharm for Mac and Windows is navigating to the Preferences (or Settings) menu:

Mac: `PyCharm -> Preferences`
Windows: `File -> Settings`

Please note I adjusted the instructions for a Mac user. Figure this could help someone down the road as it took me some time to figure out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -635,7 +635,7 @@ $ where black
 %LocalAppData%\Programs\Python\Python36-32\Scripts\black.exe  # possible location
 ```
 
-3. Open External tools in PyCharm with `File -> Settings -> Tools -> External Tools`.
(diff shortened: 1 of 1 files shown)


Heaford:
These instructions are designed for both Windows and Mac users. I think by making these suggested changes it is now incorrect for windows users as it no longer tells them to go `File -> Settings`

I have made a pull request fixing an addtional issue and combined a fix for this issue too.

ambv:
We adopted the other PR that fixed this. Thanks!

PR #651, opened 2018-12-29
Remove 'bytes' workaround for Python <= 2.5

jdufresne:
Black only supports Python 3.6+ where the bytes type is always available.

### DIFF
--- a/blib2to3/pgen2/tokenize.py
+++ b/blib2to3/pgen2/tokenize.py
@@ -38,13 +38,6 @@ __all__ = [x for x in dir(token) if x[0] != '_'] + ["tokenize",
            "generate_tokens", "untokenize"]
 del token
 
-try:
-    bytes
-except NameError:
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
This is harmless, but maybe we'd prefer to touch `blib2to3` as little as possible (the file is not even black-formatted).

jdufresne:
Just curious, does `blib2to3` track changes to the upstream `lib2to2` and rebase on them? Presumably, overtime, this module will require some amount of maintenance. If this is handled through upstream, then I agree, Black should only apply minimal patches. But if not, might as well maintain it in this repo. Removing dead code will help make make maintenance easier in the future.

ambv:
We are tracking upstream and indeed many requirements of Black were *sources* of upstream changes :-)

ambv:
Thanks for your contribution though!

Issue #637, opened 2018-12-10
Black fails to run if pyproject.toml has " }" after a boolean


jasongi-actu:
Operating system: OSX High Sierra
Python version: 3.7.1
*Black* version:  18.9b0
Does also happen on master: Yes

Black gives the following error when you try to run in a folder with the following pyproject.toml:

```
[tool.poetry]
name = "test"
version = "0.1"
description = "test repo"
authors = ["Jason Giancono <jgiancono@actu.org.au>"]

[tool.poetry.dependencies]
python = "^3.7"

[tool.poetry.dev-dependencies]
black = { python=">3.6", version=">=18.9b0", allow_prereleases=true }
```
```
Error: <click.core.Context object at 0x100f02630>
```
However if you change the black line to not have a space after allow_prereleases=true then it works fine e.g

```
[tool.poetry]
name = "test"
version = "0.1"
description = "test repo"
authors = ["Jason Giancono <jgiancono@actu.org.au>"]

[tool.poetry.dependencies]
python = "^3.7"

[tool.poetry.dev-dependencies]
black = {python=">3.6", version=">=18.9b0", allow_prereleases=true}
```

This happens if any config in pyproject.toml has "something=true }"

GandaG:
Same error happens on mine as long as there is a boolean:
```
[tool.flit.metadata]
module = "pyfomod"
author = "Daniel Nunes"
author-email = "daniel.henri.nunes@gmail.com"
home-page = "https://github.com/GandaG/pyfomod"
description-file = "README.rst"
requires-python = ">=3"
classifiers = [
    "Intended Audience :: Developers",
    "License :: OSI Approved :: Apache Software License",
    "Programming Language :: Python :: 3",
    "Topic :: Software Development :: Libraries",
]
requires = [
    "lxml",
]

[tool.flit.metadata.requires-extra]
dev = [
    "black",
    "isort",
]
test = [
    "pytest",
    "pytest-cov",
    "pytest-sugar",
]

[tool.isort]
multi_line_output = 3
force_grid_wrap = 0
line_length = 88
include_trailing_comma = True
use_parentheses = True
```
If the last two lines are removed, everything works.

zsol:
Waiting for a new release of [toml](https://pypi.org/project/toml/) to fix this.

AlexisGoodfellow:
Just bumped into this myself after upgrading my system to try to understand the walrus operator - has `toml` really not released a new version in 15+ months?

hugovk:
[...]

PR #653, opened 2018-12-31
Improve an error message when failed to load pyproject.toml

thombashi:
This PR aims to improve the information content of an error message when failed to load `pyproject.toml` file.

Like #637, current `black 18.9b0 (and master branch)` output the following error message when `pyproject.toml` malformed or `toml` package considers the format is invalid:

```
$ black .
Usage: black [OPTIONS] [SRC]...

Error: <click.core.Context object at 0x7f51c84264e0>
```

I think that it is difficult to understand the cause of the error from the message.
After applying the patch, the message looks like below:

```
$ black .
Error: Could not open file /home/.../pyproject.toml: Error reading configuration file: Invalid date or number (line 36 column 1 char 478)
```

### DIFF
--- a/black.py
+++ b/black.py
@@ -160,7 +160,9 @@ def read_pyproject_toml(
         pyproject_toml = toml.load(value)
         config = pyproject_toml.get("tool", {}).get("black", {})
     except (toml.TomlDecodeError, OSError) as e:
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Thank you!

PR #654, opened 2019-01-01
Don't use Python 3.6 grammar on files using future-fstrings

radek-sprta:
Projects that support older versions of Python and use future-fstrings package to backport fstrings are currently broken when reformatted with Black, because it uses Python 3.6-only grammar. The patch checks for '# -*- coding: future_fstrings -*-' in the files, so they don't get falsely flagged as Python 3.6 compatible.

I am considering how to best implement the tests for this. Any advice in that regard would be welcome!

### DIFF
--- a/black.py
+++ b/black.py
@@ -632,7 +632,11 @@ def format_str(
     dst_contents = ""
     future_imports = get_future_imports(src_node)
     is_pyi = bool(mode & FileMode.PYI)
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
#618 may provide a nicer way to support this.

radek-sprta:
> #618 may provide a nicer way to support this.

That looks great! I'm closing this PR then and I'll use my patched version until #618 is merged.

PR #666, opened 2019-01-10
Use g:pymode_python-defined interpreter if defined and exists, otherwise use existing defaults

pmacosta:
This is helpful when using custom-compiled interpreters, or alternative Python interpreters in non-standard locations

### DIFF
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -42,8 +42,11 @@ import sys
 import vim
 
 def _get_python_binary(exec_prefix):
-  default = vim.eval("g:pymode_python").strip()
-  if os.path.exists(default):
+  try:
+      default = vim.eval("g:pymode_python").strip()
+  except:
+      default = ""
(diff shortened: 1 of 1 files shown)


plugin/black.vim
```diff
@@ -37,10 +37,17 @@ if !exists("g:black_skip_string_normalization")
...
 
 python3 << endpython3
+import os
 import sys
 import vim
 
 def _get_python_binary(exec_prefix):
+  try:
+      default = vim.eval("g:pymode_python").strip()
+  except:
```
ambv: Pokemon-style "except:" is an anti-pattern. Can you catch the actual exception you are expecting?

pmacosta: Thank you for the suggestion, I have fixed this in my latest commit to this PR.

ambv:
Thanks! ✨ 🍰 ✨

PR #675, opened 2019-01-18
Turn off pre-commit's automatic parallelization for black

asottile:
black internally uses multiprocessing for speed.  In pre-commit 1.13.0 this is automated by the framework itself however if both pre-commit and black are forking processes this is slower and hits race-conditions in `black`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,4 +4,5 @@
     entry: black
     language: python
     language_version: python3
+    require_serial: true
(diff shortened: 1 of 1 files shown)


zsol:
Thanks!

asottile:
@zsol mind taking a look at #675  as well? it's related :D

zsol:
I did actually, just didn't have a chance to play around with it. Anyway I saw Jelle already merged it. Thanks for that, too :)

PR #679, opened 2019-01-23
Wrap `loop.run_in_executor` up in `asyncio.ensure_future` for reliable cross-platform berhavior.

### LINKED ISSUE #494: Black is not formatting code on Python 3.7 on some platform configurations (Windows, Ubuntu)
Howdy! Sorry you're having trouble. To expedite your experience,
provide some basics for me:

Operating system: Windows 10 64bit
Python version: Python 3.7.0 (32Bit)
*Black* version: 18.6b4

When running black for directory it fails without any error when formatting.

As far as I can tell it happens here:
`black.py@445`

```py
tasks = {
        loop.run_in_executor(
            executor,
            format_file_in_place,
            src,
            line_length,
            fast,
            write_back,
            mode,
            lock,
        ): src
        for src in sorted(sources)
    }
```

So I'm not sure if it's an bug in black or somewhere deeper. Or it might just be a feature of running 32bit python3 on windows 64bit.
It doesn't happen when using py3.7 64bit. Nor on py3.6 64bit. When I have a moment I'll test it on py3.6 32bit.

tomchristie:
Closes #494

Task completion should also remove the task from `pending`.

Only replicates on some platforms. (eg. Can replicate on Python 3.7+, with either Windows or whatever default Linux distro Travis uses.)


tomchristie:
Not obvious to me how to write a test for this, but it should be self-evidently more correct/robust.

I assume that there is differing behavior between platforms for what happens when *already complete* tasks are included in the `asyncio.wait`, which would account for why #494 fails in some platform configurations, but not others.

JelleZijlstra:
I don't think this does anything, because `pending` remains a reference to the same `tasks.keys()` object. Are you sure this makes a practical difference?

JelleZijlstra:
Just to demonstrate how this works:
```
In [749]: x = {1: 2}

In [751]: y = x.keys()

In [752]: len(y)
Out[752]: 1

In [753]: x.pop(1)
Out[753]: 2

In [754]: len(y)
Out[754]: 0
```

tomchristie:
Surprising that this change breaks some caching behavior. I don't understand the tests fully enough to see what might be at issue there.

What would be super helpful would be to have a failing test case for #494 - that's slightly hampered by the fact that I can reproduce #494 in Travis's build environment, but not on my local machine.

black.py
```diff
@@ -464,15 +464,17 @@ def reformat_one(
...
-        loop.run_in_executor(
-            executor,
-            format_file_in_place,
-            src,
-            line_length,
-            fast,
-            write_back,
-            mode,
-            lock,
+        asyncio.create_task(
```
JelleZijlstra: create_task is new in 3.7, so we can't use it yet in Black, since we support 3.6 (https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)

tomchristie:
Right then, here we go.

I'm not sure how I would write a test for #494, in any case it *wouldn't* replicate on xenial, which black is using in it's travis builds.

It *appears* that on some platforms `run_in_executor` is returning a raw coroutine, rather than a Future/Task. In those cases the items in the `done` result returned by `asyncio.wait` do *not* correspond to entries in the `tasks` dictionary.

As a result, `tasks.pop(task)` raises a KeyError, like so:

```shell
$ python black.py foo --check
  File "black.py", line 3689, in <module>
    patched_main()
  File "black.py", line 3685, in patched_main
    main()
  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/click/core.py", line 764, in __call__
  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/click/core.py", line 956, in invoke
  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/click/core.py", line 555, in invoke
  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/click/decorators.py", line 17, in new_func
  File "black.py", line 375, in main
    executor=executor,
  File "/opt/python/3.7-dev/lib/python3.7/asyncio/base_events.py", line 495, in run_until_complete
    return future.result()
  File "black.py", line 489, in schedule_formatting
    src = tasks.pop(task)
KeyError: <Task finished coro=<BaseEventLoop.run_in_executor() done, defined at /opt/python/3.7-dev/lib/python3.7/asyncio/base_events.py:653> result=False>
```

Obviously it'd be good to have a test case here to prevent a regression, but I can't see how to proceed with that.
Wrapping `loop.run_in_executor` up in `asnycio.ensure_future` resolves this.

I've been able to confirm this using a pull request against this trivial repository... https://github.com/tomchristie/testblack/pull/1

Note the last two commits on that pull request.

1. Dropping `asyncio.ensure_future`... https://github.com/tomchristie/testblack/pull/1/commits/159bf4a10219a006297c5162978f9368ce572bd1 the [travis tests fail](https://travis-ci.org/tomchristie/testblack/builds/489959759?utm_source=github_status&utm_medium=notification) on python 3.7.
2. Re-introducing `asyncio.ensure_future`... https://github.com/tomchristie/testblack/pull/1/commits/d1019e7ef68b1925fbe1a02e19326d6c264745ec the [travis tests succeed](https://travis-ci.org/tomchristie/testblack/builds/489960709?utm_source=github_status&utm_medium=notification).

tomchristie:
It's possible that this resolution is only required due to somewhat buggy implementations - afterall `run_in_executor` [is documented as returning a `Future`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor), so it's not clear that it should require the `ensure_future` wrapping.

Perhaps whatever build Travis use for `3.7-dev` has an issue that is also shared by the Windows 32 bit 3.7.0 build. Either way this change would result in more reliable behavior, so you *may* want it in even if the root cause is a bug somewhere deeper inside particular python builds.

JelleZijlstra:
Sorry for the delay, this looks good now.

PR #680, opened 2019-01-23
Add PyCharm setup step

gsganden:
The file watcher is unusable on PyCharm without this additional step -- it constantly triggers in the middle of editing, which then creates conflicts between what is in memory and what is on disk.


JelleZijlstra:
I believe you but as with #681 I'd prefer if somebody else who uses PyCharm could confirm this is accurate.

D0cR3d:
I use PyCharm, Just set this up and can confirm first hand that keeping the option checked does as OP describes but unchecking the option solves the issue.

JelleZijlstra:
Thanks for confirming!

PR #681, opened 2019-01-23
Indicate that PyCharm instructions also work with IntelliJ

gsganden:


JelleZijlstra:
This looks reasonable to me, but I don't use either PyCharm or IntelliJ so I can't confirm this is correct. I'm willing to merge it if somebody else can confirm this is accurate.

ambv:
Thanks! ✨ 🍰 ✨

PR #684, opened 2019-01-29
chore: Fix noqa comment

taion:
Omitting the colon makes Flake8 ignore all errors, rather than the specific code.


JelleZijlstra:
Thanks!

PR #692, opened 2019-02-06
Fix end of codeblock to re-show rest of README

hugovk:
https://github.com/ambv/black/pull/663 removed the triple-backticks, so the codeblock never ends and the rest of the README is hidden in the `<summary>`'s `<details>`.

# Before

image

https://github.com/ambv/black/blob/9d749280bb1051991d391e0ee70174a613da16fc/README.md

# After

image

...

image

https://github.com/hugovk/black/blob/5eafb5ef49831f3d9e886d3b3e0676c6c8f2ac17/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -590,6 +590,7 @@ exclude = '''
                      # the root of the project
 )
 '''
+```
 
 </details>
 


JelleZijlstra:
I just merged #695 which applied the same fix. Thanks for your PR!

PR #696, opened 2019-02-06
Remove numeric underscore normalization

### LINKED ISSUE #549: Remove numeric underscore normalization
Below is the original issue, which was arguing for "doing something about `--skip-numeric-underscore-normalization`." Since then, a consensus seems to have formed amongst users that the who feature should just be removed, because grouping by threes is not appropriate for numeric literals such as object IDs, phone numbers, dates, and much more. Please read the comments for a more detailed explanation.

----

Hey! As I understand, this option was added to keep the possibility to write 'localized' code, respecting the developers' regional numeric format.

### Localized code formatting

I disagree with the premise that this is something to be considered. Even though my local date format is not ISO, I'm using `YYYY-MM-DD` in code. Even though my local format marks decimal places with a comma, I'm fine with writing `3.14` instead of `3,14`.

### The black philosophy

The README makes some promises: 

>Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, […]. You will save time and mental energy for more important matters.

>Blackened code looks the same regardless of the project you're reading. Formatting becomes transparent after a while and you can focus on the content instead.

Adding options like this, while having only a minor effect, goes directly against these promises. Each option is a new discussion to have. Each option is something new to discover: consider if I want to make a quick edit to a constant in someone's blackened open source project on the github.com in-browser editor. I know how to write black-compliant code, so I can risk not cloning the project and committing something without running black. But now that this option is there, I either need to look for the project's config first to see their preference for literals, or guess and risk a CI failure wasting my time.

And yes, I honestly believe this *is* a slippery slope.

You might say, well, that's okay, everyone using black would just use the defaults unless they are in an aforementioned non-three digit grouping region, right? Well…

### Developers are a lost cause

We had a team at my company bump the black version a couple days ago. I glanced at their project today and was horrified to find `skip-numeric-underscore-normalization = true` in their config. Apparently they saw the option in the changelog and 

- held a poll on whether to keep this on or off
- got a 50-50% split for/against in the results
- said 'we may remove the option if there is consensus between our developers'

Really, developers are a lost cause when it comes to code style arguments. Just don't let them have them, please.

### Possible fixes

1. Remove the `--skip-numeric-underscore-normalization` option.
1. Remove the whole feature of numeric-underscore-normalization; if it's not a universally enforceable standard, maybe it has no right to be a part of black.
1. Rewrite the style of documentation of these options. Mention restrictive rules on when you as a developer are allowed to use it. Something like this:

   > ## Configuring black
   >
   > Some special circumstances might require project specific configuration. But if you're using black, to not undermine all the benefits black gives you and the community, you must agree to only use these options if we permit it here.
   >
   > ### `--skip-numeric-underscore-normalization`
   >
   > **You may use this if:** Your codebase is worked on only by developers from regions where the local number format does not group by three digits.
   > **Effect:** Numbers can be written both like `100_000_000` and `1_0000_0000`
   >
   > ### `--skip-string-normalization`
   > **You may use this if:** Your codebase predates black with pre-existing string conventions, and adoption would be unreasonably difficult if all strings had to instantly be reformatted.
   > **Effect:** Strings can be written both like `'hello'` and `"hello"`

zsol:
Fixes #549.


README.md
```diff
@@ -381,14 +379,8 @@ an adoption helper, avoid using this for new projects.
...
-styled as `2L` instead of `2l` to avoid confusion between `l` and `1`. In
-Python 3.6+, *Black* adds underscores to long numeric literals to aid
-readability: `100000000` becomes `100_000_000`.
-
-For regions where numerals are grouped differently (like [India](https://en.wikipedia.org/wiki/Indian_numbering_system)
-and [China](https://en.wikipedia.org/wiki/Chinese_numerals#Whole_numbers)),
-the `-N` or `--skip-numeric-underscore-normalization` command line option
-makes *Black* preserve underscores in numeric literals.
+styled as `2L` instead of `2l` to avoid confusion between `l` and `1`.
+In Python 3.6 and later, *Black* preserves underscores in numeric literals.
```
JelleZijlstra: Maybe add "unchanged"? Otherwise, I'm not sure it's worth mentioning something Black *doesn't* change.

tests/data/fmtonoff.py
```diff
@@ -4,13 +4,13 @@
 
 from third_party import X, Y, Z
 
-from library import some_connection, \
-                    some_decorator
+from library import some_connection, some_decorator
```
JelleZijlstra: This looks like you accidentally blackened the whole file.

tests/data/function.py
```diff
@@ -4,52 +4,98 @@
 
 from third_party import X, Y, Z
 
-from library import some_connection, \
-                    some_decorator
-f'trigger 3.6 mode'
+from library import some_connection, some_decorator
```
JelleZijlstra: Same here.

README.md
zsol: Hm, fair enough. I'll just remove this sentence.

imomaliev:
Hi, when this will be released?

PR #699, opened 2019-02-07
add missing aiohttp dep

### LINKED ISSUE #690: Installing black[d] from github misses aiohttp-cors dependency
My Pipfile:

```ini
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[dev-packages]

[packages]
black = {git = "https://github.com/ambv/black.git",editable = true, extras = ["d"]}
aiohttp-cors = "*"

[requires]
python_version = "3.7"
```
If I leave out "aiohttp-cors", the install process fails, complaining about the needed aiohttp-cors dependency.

Python 3.7.1, using `pipenv install`.

JelleZijlstra:
Also mark 3.8 as allowed to fail for now; it will fail due to an aiohttp bug.

Fixes #690

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,8 @@ matrix:
       python: 3.6
     - name: "3.7"
       python: 3.7
+  # aiohttp currently has a bug affecting 3.8
+  allow_failures:
     - name: "3.8-dev"
       python: 3.8-dev
 before_deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,8 @@ matrix:
       python: 3.6
     - name: "3.7"
       python: 3.7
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
This Travis configuration just doesn't run 3.8 at all, which doesn't seem ideal. Let me try something else.

.travis.yml
```diff
@@ -31,6 +31,8 @@ matrix:
       python: 3.6
     - name: "3.7"
       python: 3.7
+  # aiohttp currently has a bug affecting 3.8
+  allow_failures:
```
asottile: For this to run it needs to also be in the matrix

JelleZijlstra:
Thanks for fixing!

PR #701, opened 2019-02-07
Fix PyCharm instructions in README

russelldavis:
Without this change, PyCharm won't refresh the file in the editor after Black runs.


ambv:
Thanks! ✨ 🍰 ✨

PR #722, opened 2019-02-22
Format all files in a repo

fangyi-zhou:
Use `git ls-files *.py` for git repos and `hg locate *.py` for mercurial repos to gather all python files when looking for files. This addresses the issue that when you do `black .` under a virtualenv, black starts to format files in the site-package in the virtualenv because it doesn't honour `.gitignore`. This proposed approach can honour `.gitignore` correctly.

### DIFF
--- a/black.py
+++ b/black.py
@@ -13,6 +13,7 @@ from pathlib import Path
 import pickle
 import re
 import signal
+import subprocess
 import sys
 import tempfile
 import tokenize
@@ -316,6 +317,9 @@ def read_pyproject_toml(
     show_default=True,
 )
 @click.option(
+    "--entire-repo", is_flag=True, help=("Format all python files in the repo.")
+)
+@click.option(
     "-q",
     "--quiet",
     is_flag=True,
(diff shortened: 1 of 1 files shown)


black.py
```diff
@@ -3172,6 +3183,30 @@ def get_imports_from_children(children: List[LN]) -> Generator[str, None, None]:
...
 
 
+def gen_python_files_via_vcs(root: Path) -> Iterator[Path]:
+    """Generate all python files in a repo under VCS
+    using `git ls-files` for Git Repos and `hg locate` for Mercurial Repos
+
+    `include` and `exclude` are not considered.
+    """
+    cwd = os.getcwd()
+    os.chdir(root)
```
JelleZijlstra: Better to use the `cwd=` argument to subprocess.run so you don't have to chdir.

fangyi-zhou:
Thanks for the comments. The code is not yet complete and this is a simple prototype

Do you think this feature is going to useful?

Fangyi

— sent from a mobile device, apologies for typos

zsol:
I would build this logic on top of black, not inside it. Separation of concerns, and all that

fangyi-zhou:
Hi Zsolt

Thanks for the comments. So this PR should be closed?

Fangyi

— sent from a mobile device, apologies for typos

zsol:
Yes

PR #732, opened 2019-03-04
Changes default logger used by blib2to3 Driver

apljungquist:


zsol:
lgtm, but please add this change to the readme of blib2to3

zsol:
Also maybe consider upstreaming it ;)

apljungquist:
"upstreaming it"?

zsol:
Yes, blib2to3 is a fork of https://github.com/python/cpython/tree/master/Lib/lib2to3

It's a good idea to make changes there as well if they make sense

apljungquist:
Good point, I'll look into it.

ambv:
Thanks! ✨ 🍰 ✨

PR #740, opened 2019-03-09
Default to current directory if no paths are specified

hugovk:
I usually run Black on the current directory:

```
black .
```

Similarly for Flake8, the other Python CLI tool I often run:

```
flake8 .
```

And Flake8 [defaults to search in the current directory](https://gitlab.com/pycqa/flake8/blob/88caf5ac484f5c09aedc02167c59c66ff0af0068/src/flake8/checker.py#L203). This is the same:

```
flake8
```

I'd like to ask for Black to do the same:

```
black
```

Benefits:
* Less typing
* Consistency

### DIFF
--- a/black.py
+++ b/black.py
@@ -55,6 +55,7 @@ DEFAULT_EXCLUDES = (
     r"/(\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|_build|buck-out|build|dist)/"
 )
 DEFAULT_INCLUDES = r"\.pyi?$"
+DEFAULT_SRC = (".",)
 CACHE_DIR = Path(user_cache_dir("black", version=__version__))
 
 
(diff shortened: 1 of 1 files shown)


uranusjr:
Passerby here, I want to voice that I wouldn’t like this change as a user. Unlikely flake8, Black actively modifies the checked code base (unless `--check` is passed). This makes the operation much more dangerous, and allowing omittion of the target directory makes it much easier to make mistakes. The path argument should be mandatory IMO.

hugovk:
Thanks, that's a fair point. 

I'm trying to think of any Unix-like commands that modify without a path argument but coming up blank.

At least `black .` is much easier than the situation with pyupgrade :)

```sh
pyupgrade `find . -name "*.py" -type f`
```

ambv:
I think at this point the tool is popular enough that @uranusjr has a point: existing users might be surprised. Another small point: invalid editor integration that uses a variable as the filename might cause unexpected files to be formatted that way. Consider:
```py3
os.system(f"black {file}")
```
If `{file}` is an empty string for any reason, currently we will just error out. If we change the behavior, it would reformat stuff in the current directory which will be unexpected.

Thanks for your contribution @hugovk but we will have to pass on this change.

PR #758, opened 2019-03-15
Describe how to add black to Wing IDE

ArjaanBuijk:


README.md
```diff
@@ -676,6 +677,36 @@ $ where black
...
+$ pip install black
+```
+
+2. Make sure it runs from the command line, e.g.
+
+```console
+$ black --help
+```
+
+3. In Wing IDE, activate **OS Commands** panel  and define the command  **black** with e.g. the following values, to execute black on the currently selected file, using key binding **F1**:
```
akuchling: Thanks for doing this! I wanted to configure Black for use with Wing and had to track down how, and was going to write basically your patch.  Note that the Wing 7.0.0 manual refers to Black in its section on PEP8 formatting: https://wingware.com/doc/edit/pep8 )

Wording change: 'activate OS' -> 'activate the OS'.  

I suggest rewording the second half of the second as just: `define the command **black** to execute black on the currently selected file:`

Then change the key-binding line below to something like: - Key Binding: F1 <or enter a different key or sequence of keys>

README.md
```diff
@@ -676,6 +677,36 @@ $ where black
...
+
+- right-click on a panel -> Insert Tool -> OS Commands
```
akuchling: 'on a panel' ->the Wing docs (e.g. https://wingware.com/doc/custom/gui-layout ) seem to call these "toolboxes".   So this could be: "right-click on the toolbar area of a toolbox".

It could also just be "use the Tools > OS Commands menu selection", which I think will add 'OS Commands' to its default location if it's not already present, and then display it.

README.md
```diff
@@ -676,6 +677,36 @@ $ where black
...
+- click on **+** in **OS Commands** -> New: Command line..
+  - Title: black
+  - Command Line: black %s
+  - I/O Encoding: Use Default 
+  - Key Binding: F1
+  - [x] Raise OS Commands when executed
+  - [x] Auto-save files before execution
+  - [x] Line mode
+
+4. Select a file in the editor and press **F1** , or whatever Key Binding you selected in step 3, to reformat the file.
```
akuchling: Should probably lowercase 'Key Binding', i.e. `whatever key binding you selected ...`.

README.md
ArjaanBuijk: Thanks for the review.

I updated the descriptions with these suggestions, and also included a link to the wing documentation on pep8 formatting.

README.md
ArjaanBuijk: I also corrected it with these changes. Switched to the more simple description "use the Tools->OS Commands" method.  It will indeed display the 'OS Commands' tool in the default location.

README.md
ArjaanBuijk: Also made this change.

ambv:
Thanks! ✨ 🍰 ✨

PR #761, opened 2019-03-15
add `black -c "code"`

JelleZijlstra:
Useful as a way to quickly try out what Black does on a chunk of code. A `-c` option with similar behavior is also in mypy and python itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ black {source_file_or_directory}
 black [OPTIONS] [SRC]...
 
 Options:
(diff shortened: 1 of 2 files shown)


zsol:
Personally, I use `echo code | black -` for this, but I'm not against this option. Can we rename the long option to `--code` though?

JelleZijlstra:
I used `--command` because that's what mypy uses as the long form, but `--code` does make more sense. I'll make the change.

Issue #759, opened 2019-03-15
black adds trailing comma despite using --target-version py35


bjoernricks:
Hi,

black 19.3b0 adds trailing commas for function signatures containing `*` despite setting target version 3.5. This creates invalid code for python 3.5 which will raise a SyntaxError. Setting
target version to py34 or using the auto detection doesn't add a trailing comma in this case.

```python3
def foo_func(bar, *, lorem=None, ipsum=None, dolor=None, sit=None, amet=None, consectetur=None):
    print('lorem ipsum dolor sit amet')
```
running
```sh
> black --target-version py35 --line-length 80 foo.py
```
creates

```python3
def foo_func(
    bar,
    *,
    lorem=None,
    ipsum=None,
    dolor=None,
    sit=None,
    amet=None,
    consectetur=None,
):
    print("lorem ipsum dolor sit amet")
``` 

Operating system: Ubuntu 18.04
Python version: Python 3.5/3.6
*Black* version: 
```sh
> black --version
black, version 19.3b0
```
Does also happen on master: yes

JelleZijlstra:
Looks like we improperly merged two different syntax changes: `f(**x,)` (trailing comma in a call) first became legal in 3.5, but `def f(**x,): pass` (trailing comma in a definition) became legal only in 3.6. I'll try to disentangle the two in the code.

bjoernricks:
Thanks a lot!

mikenerone:
Seems like there's a regression -I'm seeing this behavior with 20.8b1. It could be something more specific: my affected function has a return type annotation.

ichard26:
@mikenerone could you please post the bug-inducing code? Might be a case of #1356.

mikenerone:
> @mikenerone could you please post the bug-inducing code? Might be a case of #1356.

I believe you right. My triggering sample below closely resembles a case shown there.
```
class LongishClassNameToHelpTriggerTheBug:
    ...


def only_seems_to_trigger_bug_if_definition_is_long_enough_to_require_wrapping(
    *some_args: LongishClassNameToHelpTriggerTheBug
) -> LongishClassNameToHelpTriggerTheBug:
    pass
```

PR #762, opened 2019-03-15
fix incorrect call to `right_hand_split()`

JelleZijlstra:
The second argument to `right_hand_split()` is an int (line length), but we were passing `supports_trailing_commas` as a bool. Mypy didn't catch this because bool is a subclass of int. Found this while working on a fix for #759 because I changed the type of `supports_trailing_commas`.

This affects one test that is now formatted differently, and it also causes a fair number of changes in our codebase. Now that I look at it, in most cases I actually feel like the old formatting is better, so maybe we need to do something different than just fix the type mismatch.

### DIFF
--- a/black.py
+++ b/black.py
@@ -2206,7 +2206,7 @@ def split_line(
             # All splits failed, best effort split with no omits.
             # This mostly happens to multiline strings that are by definition
             # reported as not fitting a single line.
-            yield from right_hand_split(line, supports_trailing_commas)
+            yield from right_hand_split(line, line_length, supports_trailing_commas)
 
         if line.inside_brackets:
             split_funcs = [delimiter_split, standalone_comment_split, rhs]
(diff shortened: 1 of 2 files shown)


ambv:
Oh my, this one's embarrassing! It dates all the way back to 3eea3aad when I was fixing trailers (content with brackets) being unnecessarily exploded into their own lines after a dedented closing bracket.

ambv:
@JelleZijlstra, please add a tiny description of the change to the change log for 19.4c0, please.

rouge8:
> This affects one test that is now formatted differently, and it also causes a fair number of changes in our codebase. Now that I look at it, in most cases I actually feel like the old formatting is better, so maybe we need to do something different than just fix the type mismatch.

I had the same experience in one of our codebases where the old formatting was generally better. Is there an issue or plans to address this?

JelleZijlstra:
Filed #781 to follow up on this.

Issue #755, opened 2019-03-14
Vim plugin needs updates for black 19.3b0


llchan:
With black 19.3b0, I now see a `AttributeError: type object 'FileMode' has no attribute 'AUTO_DETECT'` when I call `:Black` in vim. I think black.vim needs some updates to work with the new API.

It may also be a good idea while we're looking at this to spec out a minimal stable API between vim and the black module, since there's really no guarantee that black.py and black.vim are on the same version.

Operating system: Linux
Python version: 3.7.1
*Black* version: 19.3b0
Does also happen on master: Yes (well, haven't tried, but just based on black.vim code vs black.py)

gsong:
For those looking for a workaround for now:

> If you need to do anything special to make your virtualenv work and install Black (for example you want to run a version from master), create a virtualenv manually and point g:black_virtualenv to it. The plugin will use it.

Just install `black==18.9b0` for now.

pquentin:
Thanks to @tlv and @JelleZijlstra, this was fixed in https://github.com/ambv/black/pull/766. This issue can now be closed!

PR #766, opened 2019-03-16
fix vim plugin for 19.3b0 (#755)

tlv:


plugin/black.vim
```diff
@@ -98,15 +98,13 @@ if _initialize_black_env():
...
   start = time.time()
   fast = bool(int(vim.eval("g:black_fast")))
-  line_length = int(vim.eval("g:black_linelength"))
-  mode = black.FileMode.AUTO_DETECT
-  if bool(int(vim.eval("g:black_skip_string_normalization"))):
-    mode |= black.FileMode.NO_STRING_NORMALIZATION
-  if vim.current.buffer.name.endswith('.pyi'):
-    mode |= black.FileMode.PYI
+  mode = black.FileMode()
+  mode.line_length = int(vim.eval("g:black_linelength"))
```
JelleZijlstra: Could you pass these to the constructor instead of setting them later?

tlv: done

JelleZijlstra:
Thank you!

PR #807, opened 2019-04-16
[WIP] Allow a single blank line between class and function definitions

### LINKED ISSUE #619: One space between class definition and the first inner method
Can Black have a rule to **not remove** the whitespace at line `2` from here:
```python
class C:
    
    def __init__(self):
        pass
```
so that it will keep the blank line between class definition and any first inner entity (as it does already with the docstrings and assignment attributes)?

It's extremely ugly to have that whitespace consistent in the entire codebase and just in this particular situation, where you don't have anything to add between class definition and the first inner method (which usually is the `__init__` one), to end up with "glued" headers of class and method definition.

Now if I'm formatting this:
```python
class C:
    
    def __init__(self):
        pass
    def func(self):
        pass
```
I get this:
```python
class C:
    def __init__(self):
        pass

    def func(self):
        pass
```
So I expect the rule of _one empty line before and after class methods_ to be consistent here. Everyone likes to have symmetry in their code.

sersorrel:
Fixes #619.

However:

- it "fixes" it by allowing up to one blank line between `class` and `def`; it won't choose one over the other (nevertheless, I believe Black already does this for blank lines inside functions, so it's not totally without precedent :slightly_smiling_face: – in particular, both of the strict behaviours have been tried in the past (before #219, one blank line was enforced; now, Black removes the blank line), and both have been determined to be suboptimal)
- one test is currently failing due to the above
- there are no new tests specifically for this behaviour (is there documentation on how to write a test, or do I have to copy-paste an existing one?)
- I don't know if the way I've implemented this is actually sensible

There are several cases that weren't covered in the aforementioned issue; for example, should the following (currently rejected) be accepted? Clearly in this case it's worse than the version with no blank line, but if either class had more definitions I think it would make sense:

```python
class C:

    class Inner:
        pass
```

(which I guess would lead to something like "if the previous line is a `class`, and this line is a class/function/decorator, allow 0 or 1 leading blank lines"? should this also apply when the previous line is a `def`?)

tldr: I kind of feel like this is a bit of a hack when a proper solution would be more widely applicable, see https://github.com/ambv/black/issues/450#issuecomment-436244142. (at the same time, this is about the only irritation I run into on a regular basis, so I don't want to get in the way of merging this too much!)

### DIFF
--- a/black.py
+++ b/black.py
@@ -1451,6 +1451,11 @@ class EmptyLineTracker:
         if self.previous_line.is_decorator:
             return 0, 0
 
+        if self.previous_line.is_class and (
+            current_line.is_def or current_line.is_decorator
+        ):
+            return before, 0
+
         if self.previous_line.depth < current_line.depth and (
             self.previous_line.is_class or self.previous_line.is_def
         ):


cooperlees:
This PR has become very stale. Please reopen if you wish to finish it.

PR #818, opened 2019-04-30
consider all optional brackets when splitting

### LINKED ISSUE #781: Reconsider formatting that was changed by #762
#762 fixed broken code, but in many cases the old, broken code produced subjectively better formatting, especially (in my experience) around assert statements.

Let's consider whether there's a way to improve the formatting in cases that were affected by #762.

It would be helpful to add examples to this issue of formatting changes caused by #762 that may not be improvements.

zsol:
I'm not married to this solution, I put it together in half an hour. This will likely cause some code churn, but hopefully for the better.

TODO:
- [x] fix tests
- [ ] verify formatting changes on a (few?) large codebase(s)
- [ ] add changelog
- [ ] better commit message

Fixes #781.

### DIFF
--- a/black.py
+++ b/black.py
@@ -479,8 +479,11 @@ def reformat_one(
                 res_src = src.resolve()
                 if res_src in cache and cache[res_src] == get_cache_info(res_src):
                     changed = Changed.CACHED
-            if changed is not Changed.CACHED and format_file_in_place(
-                src, fast=fast, write_back=write_back, mode=mode
+            if (
+                changed is not Changed.CACHED
+                and format_file_in_place(
+                    src, fast=fast, write_back=write_back, mode=mode
+                )
             ):
                 changed = Changed.YES
             if (write_back is WriteBack.YES and changed is not Changed.CACHED) or (
@@ -550,8 +553,9 @@ async def schedule_formatting(
                 changed = Changed.YES if task.result() else Changed.NO
                 # If the file was written back or was successfully checked as
                 # well-formatted, store this information in the cache.
-                if write_back is WriteBack.YES or (
-                    write_back is WriteBack.CHECK and changed is Changed.NO
+                if (
+                    write_back is WriteBack.YES
+                    or (write_back is WriteBack.CHECK and changed is Changed.NO)
                 ):
                     sources_to_cache.append(src)
                 report.done(src, changed)
@@ -1349,8 +1353,9 @@ class Line:
 
             if subscript_start.type == syms.subscriptlist:
                 subscript_start = child_towards(subscript_start, leaf)
-        return subscript_start is not None and any(
-            n.type in TEST_DESCENDANTS for n in subscript_start.pre_order()
+        return (
+            subscript_start is not None
+            and any(n.type in TEST_DESCENDANTS for n in subscript_start.pre_order())
         )
 
     def __str__(self) -> str:
@@ -1451,8 +1456,9 @@ class EmptyLineTracker:
         if self.previous_line.is_decorator:
             return 0, 0
 
-        if self.previous_line.depth < current_line.depth and (
-            self.previous_line.is_class or self.previous_line.is_def
+        if (
+            self.previous_line.depth < current_line.depth
+            and (self.previous_line.is_class or self.previous_line.is_def)
         ):
             return 0, 0
 
@@ -1875,10 +1881,10 @@ def whitespace(leaf: Leaf, *, complex_subscript: bool) -> str:  # noqa: C901
 
             prevp_parent = prevp.parent
             assert prevp_parent is not None
-            if prevp.type == token.COLON and prevp_parent.type in {
-                syms.subscript,
-                syms.sliceop,
-            }:
+            if (
+                prevp.type == token.COLON
+                and prevp_parent.type in {syms.subscript, syms.sliceop}
+            ):
                 return NO
 
             elif prevp.type == token.EQUAL and prevp_parent.type == syms.argument:
@@ -3655,7 +3661,10 @@ def can_omit_invisible_parens(line: Line, line_length: int) -> bool:
         for _index, leaf, leaf_length in enumerate_with_length(line):
             length += leaf_length
             if leaf is last.opening_bracket:
-                if seen_other_brackets or length <= line_length:
+                if seen_other_brackets or length <= int(0.5 * line_length):
+                    # Omit optionals if there are multiple brackets to split on
+                    # or if the first bracket starts in the first half of the line.
+                    # This tends to result in subjectively good results :)
                     return True
 
             elif leaf.type in OPENING_BRACKETS:
--- a/black.py
+++ b/black.py
@@ -553,9 +553,8 @@ async def schedule_formatting(
                 changed = Changed.YES if task.result() else Changed.NO
                 # If the file was written back or was successfully checked as
                 # well-formatted, store this information in the cache.
-                if (
-                    write_back is WriteBack.YES
-                    or (write_back is WriteBack.CHECK and changed is Changed.NO)
+                if write_back is WriteBack.YES or (
+                    write_back is WriteBack.CHECK and changed is Changed.NO
                 ):
                     sources_to_cache.append(src)
                 report.done(src, changed)
@@ -1353,9 +1352,8 @@ class Line:
 
             if subscript_start.type == syms.subscriptlist:
                 subscript_start = child_towards(subscript_start, leaf)
-        return (
-            subscript_start is not None
-            and any(n.type in TEST_DESCENDANTS for n in subscript_start.pre_order())
+        return subscript_start is not None and any(
+            n.type in TEST_DESCENDANTS for n in subscript_start.pre_order()
         )
 
     def __str__(self) -> str:
@@ -1456,9 +1454,8 @@ class EmptyLineTracker:
         if self.previous_line.is_decorator:
             return 0, 0
 
-        if (
-            self.previous_line.depth < current_line.depth
-            and (self.previous_line.is_class or self.previous_line.is_def)
+        if self.previous_line.depth < current_line.depth and (
+            self.previous_line.is_class or self.previous_line.is_def
         ):
             return 0, 0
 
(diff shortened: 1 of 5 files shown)


JelleZijlstra:
I'm running it on our code now. A couple of things I noticed:

```
-from qcore.asserts import assert_eq, assert_dict_eq, assert_is, assert_unordered_list_eq
+from qcore.asserts import (
+    assert_eq,
+    assert_dict_eq,
+    assert_is,
+    assert_unordered_list_eq,
+)
```
Here the old line was exactly 88 characters long. I'm guessing we add the parentheses, that pushes it over 88 chars, and then we split it into multiple lines. This issue is causing a lot of churn in our codebase.

```
-    assert isinstance(value, return_type), (
-        "Value is not in one of the exepected types. Type: %s. Value: %r"
-        % (type(value), value)
+    assert isinstance(
+        value, return_type
+    ), "Value is not in one of the exepected types. Type: %s. Value: %r" % (
+        type(value),
+        value,
     )
```
Here the old formatting seems better to me.

JelleZijlstra:
This diff shows more clearly that the optional parens are pushing the line over the limit:
```
-                    for experiment, settings in exp_settings.future_experiments.items():
+                    for (
+                        experiment,
+                        settings,
+                    ) in exp_settings.future_experiments.items():
```
The old line length was again exactly 88.

Also overall impact:
```
1559 files reformatted, 8134 files left unchanged.
```

tests/data/composition.py
```diff
@@ -31,13 +31,15 @@ def test(self) -> None:
...
                     ):
                         print(i)
-        xxxxxxxxxxxxxxxx = Yyyy2YyyyyYyyyyy(
-            push_manager=context.request.resource_manager,
-            max_items_to_push=num_items,
-            batch_size=Yyyy2YyyyYyyyyYyyy.FULL_SIZE,
-        ).push(
-            # Only send the first n items.
-            items=items[:num_items]
+        xxxxxxxxxxxxxxxx = (
```
zsol: I'm not sure about this one. I slightly prefer the old one

tests/data/composition.py
```diff
@@ -154,26 +166,30 @@ def tricky_asserts(self) -> None:
...
                       2 LOAD_CONST               1 (1)
                       4 COMPARE_OP               2 (==)
                       6 LOAD_FAST                0 (self)
                       8 STORE_ATTR               0 (x)
                      10 LOAD_CONST               0 (None)
                      12 RETURN_VALUE
-        """ % (
-            _C.__init__.__code__.co_firstlineno + 1,
+        """
+            % (_C.__init__.__code__.co_firstlineno + 1,)
```
zsol: This seems wrong. Haven't figured out what causes this yet

tests/data/expression.py
```diff
@@ -491,7 +492,7 @@ def gen():
     addr_proto,
     addr_canonname,
     addr_sockaddr,
-) in socket.getaddrinfo("google.com", "http"):
+) in (socket.getaddrinfo("google.com", "http")):
```
zsol: superfluous parens should be stripped here

zsol:
I updated this PR with another iteration on this idea, which hopefully causes significantly less churn, and gets some pathological cases right. There are still two bugs to figure out though.

black.py
```diff
@@ -1363,6 +1363,8 @@ def __str__(self) -> str:
         first = next(leaves)
         res = f"{first.prefix}{indent}{first.value}"
         for leaf in leaves:
+            if leaf.type == token.LPAR and not leaf.value and leaf.parent:
+                maybe_make_parens_invisible_in_atom(leaf.parent)
```
zsol: that's a noop. `leaf` is already invisible here

tests/data/composition.py
JelleZijlstra: Yes, the extra set of parentheses doesn't really help anything here.

tests/data/expression.py
```diff
@@ -285,10 +285,11 @@ def gen():
 lambda a, b, c=True, *, d=(1 << v2), e="str": a
 lambda a, b, c=True, *vararg, d=(v1 << 2), e="str", **kwargs: a + b
 manylambdas = lambda x=lambda y=lambda z=1: z: y(): x()
-foo = lambda port_id, ignore_missing: {
-    "port1": port1_resource,
-    "port2": port2_resource,
-}[port_id]
+foo = (
```
JelleZijlstra: I think the old code is a lot nicer here.

JelleZijlstra:
I tried it out again on our codebase. Here's a few notes:

```
                             assert (
                                 component[0] in VALID_DIRECTIVES
-                            ), "invalid directive %r" % (component,)
+                            ), ("invalid directive %r" % (component,))
```
These are just useless parens.

```
-        ] = self.detector.get_busy_time_proportion(None, time_us)
+        ] = (self.detector.get_busy_time_proportion(None, time_us))
```
Same here.


```
-        if not set(column_renames.keys()).issubset(columns_to_drop) or not set(
-            column_renames.values()
-        ).issubset(columns_to_add):
+        if (
+            not set(column_renames.keys()).issubset(columns_to_drop)
+            or not set(column_renames.values()).issubset(columns_to_add)
+        ):
```
This is great.

```
-            z += h.Button(
-                class_=["ui", "button"],
-                id="clear-search-button",
-                onclick="resetFilter()",
-            )(h.text("Clear"))
+            z += (
+                h.Button(
+                    class_=["ui", "button"],
+                    id="clear-search-button",
+                    onclick="resetFilter()",
+                )(h.text("Clear"))
+            )
```
Another unnecessary set of parentheses added.

JelleZijlstra:
Also in terms of churn the effect is similar:
`2289 files reformatted, 10702 files left unchanged, 6 files failed to reformat.`

The crashes are due to instability. IThe first pass is:

```             data[
                 "share_data"
-            ] = yield a.app.view.share.mobile.get_native_share_action.asynq(share)
+            ] = (yield a.app.view.share.mobile.get_native_share_action.asynq(share))
```
And the second is:
```
-            data[
-                "share_data"
-            ] = (yield a.app.view.share.mobile.get_native_share_action.asynq(share))
+            data["share_data"] = (
+                yield a.app.view.share.mobile.get_native_share_action.asynq(share)
+            )
```
The eventual result is an improvement though. The other crashes are also cases where the first pass adds parentheses, and the second one rearranges the structure of an assignment + yield.

zsol:
This latest one seems promising. I've run it on a large codebase and it did produce a bunch of changes, but all of the changes look better than without this PR.

Most of the changes follow this pattern:
```
-        if not self.ignore_interrupt and isinstance(
-            sys.exc_info()[1], KeyboardInterrupt
+        if (
+            not self.ignore_interrupt
+            and isinstance(sys.exc_info()[1], KeyboardInterrupt)
         ):
```

Some asserts have changed as well:

```
-        assert isinstance(
-            types, collections.Iterable
-        ), "types must be iterable, is %r" % (types,)
+        assert isinstance(types, collections.Iterable), (
+            "types must be iterable, is %r" % (types,)
+        )
```

The most questionable changes are around manually split strings:
```
-        sql = "INSERT INTO status_type (type_name, seq)  " "VALUES ('{}', {})".format(
-            key, val
+        sql = (
+            "INSERT INTO status_type (type_name, seq)  "
+            "VALUES ('{}', {})".format(key, val)
         )
```

All in all, I think this PR makes the formatting better, so ready for review.

zsol:
As for the amount of churn: on a codebase with half a million LOC across 3k+ files this PR generates
```
41 files changed, 188 insertions(+), 151 deletions(-)
```

JelleZijlstra:
I ran it on our codebase again. This one is a little questionable:
```
     def table_string(self):
-        return "%5d  %5d  %5d  %5d  %5d  %5d  %5d  %5d  %5d  %5d" % (
-            self.task_count,
-            self.delete_count,
-            self.suspend_count,
-            self.call_count,
-            self.send_count,
-            self.throw_count,
-            self.return_count,
-            self.cancel_count,
-            self.pause_count,
-            self.resume_count,
+        return (
+            "%5d  %5d  %5d  %5d  %5d  %5d  %5d  %5d  %5d  %5d"
+            % (
+                self.task_count,
+                self.delete_count,
+                self.suspend_count,
+                self.call_count,
+                self.send_count,
+                self.throw_count,
+                self.return_count,
+                self.cancel_count,
+                self.pause_count,
+                self.resume_count,
+            )
         )
```
On the other hand there are similar cases to this where the tuple after % fits on one line and it looks nicer, so maybe this is the price we have to pay.

This still inserts unnecessary parens:
```
         assert email_type not in {
             aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
             aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
             aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
-        }, "should not have xxxxxxxxxxxxxxxxxxxxxxx %s" % (email_type)
+        }, ("should not have xxxxxxxxxxxxxxxxxxxxxxx %s" % (email_type))
```

Overall impact: `604 files reformatted, 12022 files left unchanged.`

I agree that most of the changes now are improvements, so I'd be OK with applying it (except for the unnecessary parens in the second diff snippet above).

zsol:
I tried hard to remove the unnecessary parens you highlight, but I realized this is a bug that's better addressed in a separate issue. I'm reasonably sure the root cause is the same as with this one:

Black on master does this:
```
-        assert looooooooooooooooooooong in { looooooooooooooooooooooooooooooooooooooooooooooooooooooong}, "wot. a bug?!" % loooooooong
+        assert looooooooooooooooooooong in {
+            looooooooooooooooooooooooooooooooooooooooooooooooooooooong
+        }, ("wot. a bug?!" % loooooooong)
```

nchammas:
@zsol - I read [your comment about fluent interfaces](https://github.com/psf/black/issues/781#issuecomment-490466808) and came here to check this PR out. Could you share an example or two of how this PR might improve the formatting of fluent-style programs?

My team heavily uses Pandas, Apache Spark, and SQLAlchemy, so we are writing code in this style all the time. I would be very happy to see any improvements to how Black handles this style of code, since it would make it easier for me to sell Black to my team.

ambv:
Rejected in favor of #1092. Maybe we'll revisit the approach in the future.

PR #820, opened 2019-05-02
Fix Flake8: B011 Do not call assert False

hugovk:
Fixes the Flake8 failure on the CI:

`black.py:3149:17: B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().`

https://travis-ci.com/python/black/builds/110413516


zsol:
Thanks!

PR #825, opened 2019-05-05
Remove deprecated license_file from setup.cfg

jdufresne:
Starting with wheel 0.32.0 (2018-09-29), the "license_file" option is
deprecated.

https://wheel.readthedocs.io/en/stable/news.html

The wheel will continue to include LICENSE, it is now included
automatically:

https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,2 @@
 [bdist_wheel]
 python-tag = py36
-
-[metadata]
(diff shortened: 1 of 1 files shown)


ambv:
Thanks! ✨ 🍰 ✨

PR #826, opened 2019-05-05
tweak collection literals to explode with trailing comma

durin42:
This is a change I discussed with @ambv at PyCon - the commit sequence should make things clear. I didn't go to the effort of fixing up existing tests (they'll fail a lot right now), but I can if this looks like a valid approach. This is somewhere between "proof of concept" and something I'd feel happy about, but I wanted feedback on how I'm tackling this early in development since I'm very much noideadog on this code.

### DIFF
--- a/black.py
+++ b/black.py
@@ -1241,6 +1241,61 @@ class Line:
         ]
 
     @property
+    def is_collection_with_optional_trailing_comma(self) -> bool:
+        """Is this line a collection literal with a trailing comma that's optional?
+
+        Note that the trailing comma in a 1-tuple is not optional.
+        """
+        if not self.leaves or len(self.leaves) < 4:
+            return False
+        # Look for and address a trailing colon.
+        if self.leaves[-1].type == token.COLON:
+            closer = self.leaves[-2]
+            close_index = -2
+        else:
+            closer = self.leaves[-1]
+            close_index = -1
+        if closer.type not in CLOSING_BRACKETS or self.inside_brackets:
+            return False
+        if closer.type == token.RPAR:
+            # Tuples require an extra check, because if there's only
+            # one element in the tuple removing the comma unmakes the
+            # tuple.
+            #
+            # We also check for parens before looking for the trailing
+            # comma because in some cases (eg assigning a dict
+            # literal) the literal gets wrapped in temporary parens
+            # during parsing. This case is covered by the
+            # collections.py test data.
+            opener = closer.opening_bracket
+            for _open_index, leaf in enumerate(self.leaves):
+                if leaf is opener:
+                    break
+            else:
+                # Couldn't find the matching opening paren, play it safe.
+                return False
+            commas = 0
+            comma_depth = self.leaves[close_index - 1].bracket_depth
+            for leaf in self.leaves[_open_index + 1 : close_index]:
+                if leaf.bracket_depth == comma_depth and leaf.type == token.COMMA:
+                    commas += 1
+            if commas > 1:
+                # We haven't looked yet for the trailing comma because
+                # we might also have caught noop parens.
+                return self.leaves[close_index - 1].type == token.COMMA
+            elif commas == 1:
+                return False  # it's either a one-tuple or didn't have a trailing comma
+            if self.leaves[close_index - 1].type in CLOSING_BRACKETS:
+                close_index -= 1
+                closer = self.leaves[close_index]
+                if closer.type == token.RPAR:
+                    # TODO: this is a gut feeling. Will we ever see this?
+                    return False
+        if self.leaves[close_index - 1].type != token.COMMA:
+            return False
+        return True
+
+    @property
     def is_def(self) -> bool:
         """Is this a function definition? (Also returns True for async defs.)"""
         try:
@@ -2339,6 +2394,7 @@ def split_line(
     if (
         not line.contains_uncollapsable_type_comments()
         and not line.should_explode
+        and not line.is_collection_with_optional_trailing_comma
         and (
             is_line_short_enough(line, line_length=line_length, line_str=line_str)
             or line.contains_unsplittable_type_ignore()
--- a/black.py
+++ b/black.py
@@ -1356,60 +1356,45 @@ class Line:
 
     def maybe_remove_trailing_comma(self, closing: Leaf) -> bool:
         """Remove trailing comma if there is one and it's safe."""
+        if not (self.leaves and self.leaves[-1].type == token.COMMA):
+            return False
+        # Temporary: for now we unconditionally leave trailing commas
+        # in imports. This goes away on the next change when trailing
+        # commas imply exploding collections to multiple lines.
+        if closing.type == token.RPAR and self.is_import:
+            self.remove_trailing_comma()
+            return True
+        # We remove trailing commas only in the case of importing a
+        # single name from a module.
         if not (
             self.leaves
+            and self.is_import
+            and len(self.leaves) > 4
             and self.leaves[-1].type == token.COMMA
             and closing.type in CLOSING_BRACKETS
+            and self.leaves[-4].type == token.NAME
+            and (
+                # regular `from foo import bar,`
+                self.leaves[-4].value == "import"
+                # `from foo import (bar as baz,)
+                or (
+                    len(self.leaves) > 6
+                    and self.leaves[-6].value == "import"
+                    and self.leaves[-3].value == "as"
+                )
+                # `from foo import bar as baz,`
+                or (
+                    len(self.leaves) > 5
+                    and self.leaves[-5].value == "import"
+                    and self.leaves[-3].value == "as"
+                )
+            )
+            and closing.type == token.RPAR
         ):
             return False
 
-        if closing.type == token.RBRACE:
-            self.remove_trailing_comma()
-            return True
-
-        if closing.type == token.RSQB:
-            comma = self.leaves[-1]
-            if comma.parent and comma.parent.type == syms.listmaker:
-                self.remove_trailing_comma()
-                return True
-
-        # For parens let's check if it's safe to remove the comma.
-        # Imports are always safe.
-        if self.is_import:
-            self.remove_trailing_comma()
-            return True
-
-        # Otherwise, if the trailing one is the only one, we might mistakenly
-        # change a tuple into a different type by removing the comma.
-        depth = closing.bracket_depth + 1
-        commas = 0
-        opening = closing.opening_bracket
-        for _opening_index, leaf in enumerate(self.leaves):
-            if leaf is opening:
-                break
-
-        else:
-            return False
-
-        for leaf in self.leaves[_opening_index + 1 :]:
-            if leaf is closing:
-                break
-
-            bracket_depth = leaf.bracket_depth
-            if bracket_depth == depth and leaf.type == token.COMMA:
-                commas += 1
-                if leaf.parent and leaf.parent.type in {
-                    syms.arglist,
-                    syms.typedargslist,
-                }:
-                    commas += 1
-                    break
-
-        if commas > 1:
-            self.remove_trailing_comma()
-            return True
-
-        return False
+        self.remove_trailing_comma()
+        return True
 
     def append_comment(self, comment: Leaf) -> bool:
         """Add an inline or standalone comment to the line."""
--- a/black.py
+++ b/black.py
@@ -1413,12 +1413,6 @@ class Line:
         """Remove trailing comma if there is one and it's safe."""
         if not (self.leaves and self.leaves[-1].type == token.COMMA):
             return False
-        # Temporary: for now we unconditionally leave trailing commas
-        # in imports. This goes away on the next change when trailing
-        # commas imply exploding collections to multiple lines.
-        if closing.type == token.RPAR and self.is_import:
-            self.remove_trailing_comma()
-            return True
         # We remove trailing commas only in the case of importing a
         # single name from a module.
         if not (
--- a/tests/data/comments2.py
+++ b/tests/data/comments2.py
@@ -63,7 +63,7 @@ def inline_comments_in_brackets_ruin_everything():
         parameters.children = [
             children[0],  # (1
             body,
-            children[-1],  # )1
+            children[-1]  # )1
         ]
         parameters.children = [
             children[0],
@@ -142,7 +142,7 @@ short
         syms.simple_stmt,
         [
             Node(statement, result),
-            Leaf(token.NEWLINE, '\n'),  # FIXME: \r\n?
+            Leaf(token.NEWLINE, '\n')  # FIXME: \r\n?
         ],
     )
 
--- a/tests/data/expression.diff
+++ b/tests/data/expression.diff
@@ -11,7 +11,7 @@
  True
  False
  1
-@@ -29,63 +29,84 @@
+@@ -29,30 +29,37 @@
  ~great
  +value
  -1
@@ -59,16 +59,13 @@
  (1, 2)
  (1, 2, 3)
  []
- [1, 2, 3, 4, 5, 6, 7, 8, 9, (10 or A), (11 or B), (12 or C)]
--[1, 2, 3,]
-+[1, 2, 3]
+@@ -60,32 +67,46 @@
+ [1, 2, 3,]
  [*a]
  [*range(10)]
--[*a, 4, 5,]
--[4, *a, 5,]
+ [*a, 4, 5,]
+ [4, *a, 5,]
 -[this_is_a_very_long_variable_which_will_force_a_delimiter_split, element, another, *more]
-+[*a, 4, 5]
-+[4, *a, 5]
 +[
 +    this_is_a_very_long_variable_which_will_force_a_delimiter_split,
 +    element,
@@ -118,18 +115,7 @@
  call(**self.screen_kwargs)
  call(b, **self.screen_kwargs)
  lukasz.langa.pl
-@@ -94,23 +115,23 @@
- 1.0 .real
- ....__class__
- list[str]
- dict[str, int]
- tuple[str, ...]
--tuple[str, int, float, dict[str, int],]
-+tuple[str, int, float, dict[str, int]]
- very_long_variable_name_filters: t.List[
-     t.Tuple[str, t.Union[str, t.List[t.Optional[str]]]],
- ]
- xxxx_xxxxx_xxxx_xxx: Callable[..., List[SomeClass]] = classmethod(  # type: ignore
+@@ -104,13 +125,13 @@
      sync(async_xxxx_xxx_xxxx_xxxxx_xxxx_xxx.__func__)
  )
  xxxx_xxx_xxxx_xxxxx_xxxx_xxx: Callable[..., List[SomeClass]] = classmethod(  # type: ignore
@@ -199,7 +185,7 @@
 +    .filter(
 +        models.Customer.account_id == account_id, models.Customer.email == email_address
 +    )
-+    .order_by(models.Customer.id.asc())
++    .order_by(models.Customer.id.asc(),)
 +    .all()
 +)
  Ø = set()
@@ -391,3 +377,4 @@
      return True
  last_call()
  # standalone comment at ENDMARKER
+ 
--- a/tests/data/expression.diff
+++ b/tests/data/expression.diff
@@ -11,7 +11,7 @@
  True
  False
  1
-@@ -29,30 +29,37 @@
+@@ -29,63 +29,96 @@
  ~great
  +value
  -1
@@ -59,13 +59,28 @@
  (1, 2)
  (1, 2, 3)
  []
-@@ -60,32 +67,46 @@
- [1, 2, 3,]
+ [1, 2, 3, 4, 5, 6, 7, 8, 9, (10 or A), (11 or B), (12 or C)]
+-[1, 2, 3,]
++[
++    1,
++    2,
++    3,
++]
  [*a]
  [*range(10)]
- [*a, 4, 5,]
- [4, *a, 5,]
+-[*a, 4, 5,]
+-[4, *a, 5,]
 -[this_is_a_very_long_variable_which_will_force_a_delimiter_split, element, another, *more]
++[
++    *a,
++    4,
++    5,
++]
++[
++    4,
++    *a,
++    5,
++]
 +[
 +    this_is_a_very_long_variable_which_will_force_a_delimiter_split,
 +    element,
@@ -115,7 +130,20 @@
  call(**self.screen_kwargs)
  call(b, **self.screen_kwargs)
  lukasz.langa.pl
-@@ -104,13 +125,13 @@
+@@ -94,23 +127,25 @@
+ 1.0 .real
+ ....__class__
+ list[str]
+ dict[str, int]
+ tuple[str, ...]
+-tuple[str, int, float, dict[str, int],]
++tuple[
++    str, int, float, dict[str, int],
++]
+ very_long_variable_name_filters: t.List[
+     t.Tuple[str, t.Union[str, t.List[t.Optional[str]]]],
+ ]
+ xxxx_xxxxx_xxxx_xxx: Callable[..., List[SomeClass]] = classmethod(  # type: ignore
      sync(async_xxxx_xxx_xxxx_xxxxx_xxxx_xxx.__func__)
  )
  xxxx_xxx_xxxx_xxxxx_xxxx_xxx: Callable[..., List[SomeClass]] = classmethod(  # type: ignore
@@ -132,7 +160,7 @@
  slice[0:1:2]
  slice[:]
  slice[:-1]
-@@ -134,113 +155,171 @@
+@@ -134,113 +169,171 @@
  numpy[-(c + 1) :, d]
  numpy[:, l[-2]]
  numpy[:, ::-1]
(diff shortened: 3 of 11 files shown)


durin42:
If it's of interest, https://phab.mercurial-scm.org/D6342 shows some results on the hg codebase (which is motivating this change).

JelleZijlstra:
I take it the idea is to explode collection literals on multiple lines when they contain a trailing comma in the input, regardless of their size? I'm not sure I like that idea, because it kind of breaks Black's principle of not depending on the original formatting.

Your diff on hg also seems to show that it's exploding very short tuples unnecessarily: see the two big dicts in mercurial/policy.py.

JelleZijlstra:
I talked about this some more with @ambv and I'm OK with the change now.

Please fix the CI problems though. If you need help and you're still at PyCon, Lukasz and I are still around, so feel free to find us.

martinvonz:
Related to @JelleZijlstra's first comment, should we really split up collections within a line like this does in the policy.py example? It would probably be good to add a test case for that either way (i.e. how things like `{{1,2},{3,4},}` get formatted). Perhaps also test how it gets formatted if some inner collection should get exploded. For example, how should `{{1,2,},{3,4}}` get formatted? What if the outer collection has a trailing comma? clang-format adds a newline after the trailing comma in all these cases (and doesn't when there is no trailing comma).

durin42:
I plan on getting back to this tonight. I want to try and stop exploding
the sub-containers that lack trailing commas - if anyone has advice for how
I should approach that it would be appreciated.

I'm not at sprints, but plan to be hacking on this at least some of today
and much of Wednesday, and I'm in the same time zone as the sprints in case
some sort of real time communication would be helpful.

black.py
```diff
@@ -2161,6 +2106,22 @@ def make_comment(content: str) -> str:
     return "#" + content
 
 
+def _line_has_trailing_comma(l: Line) -> bool:
```
ambv: Nit: the current API would put "has_trailing_comma()" as a method on Line.

durin42: This turned into `is_collection_with_optional_trailing_comma` as a property on Line.

(Standalone tests for `is_collection_with_optional_trailing_comma` should be coming soon, but I wanted to get feedback from Travis on the rest of the CI status...)

durin42:
Please take a look - I think this is workable now.

I can't say I'm 100% thrilled with my approach in some corners, but I've also got no idea what I'm doing. :)

https://phab.mercurial-scm.org/D6342 is updated with the new results from this patch.

JelleZijlstra:
I ran your branch on my codebase and noticed some changes like this:
```
-    DEFAULT_DIRS = (str(AIRFLOW_DIR / "lib/floq"),)
+    DEFAULT_DIRS = (
+        str(AIRFLOW_DIR / "lib/floq"),
+    )
```
This seems undesirable; we need the trailing comma to make the tuple, but that doesn't mean we need to explode it.

JelleZijlstra:
Another example where it's especially ugly:
```
-                pieces.append("awaiting %r" % (coro.cr_await,))
+                pieces.append(
+                    "awaiting %r"
+                    % (
+                        coro.cr_await,
+                    )
+                )
```

durin42:
Huh. I guess my defense against one-element tuples isn't good enough. I'll try and figure it out, but if @ambv has any ideas I'd love to hear them. :/

ambv:
Unfortunately no genius suggestions here but I agree we need to not explode them.

durin42:
Oh, I see. I explicitly detected the case of [LPAR, one_token, COMMA, RPAR], so I'll miss any one-tuple that's got a more complicated construct inside. :(

Is there any prior art for matching a one-tuple in the codebase?

ambv:
Use `black.is_one_tuple()`.

durin42:
I'm a little lost - that takes a Node or Leaf, but I have a Line in the place where I'm doing the detection (look at `is_collection_with_optional_trailing_comma`) - it doesn't look like `is_one_tuple` quite fits? Or am I missing something obvious?

ambv:
You're right. Feeding leaves to a `Line` is flattening the structures so you cannot use `is_one_tuple()` anymore. Before I was handling this in lines 1303 - 1324 that you removed. Maybe you can reuse that code?

durin42:
That code was helpful, moved it to another property on Line because it let the code read more clearly.

See what you think? I added test cases based on the misformattings upthread.

JelleZijlstra:
I'm still seeing some issues in my codebase:
```
-        "sort_key": ("pid",),
-        "dist_key": ("pid",),
+        "sort_key": (
+            "pid",
+        ),
+        "dist_key": (
+            "pid",
+        ),
```

```
             yield create_alert(
                 "later_updates_not_blank",
-                "incorrect updates returned after message sent: %r" % (updates,),
+                "incorrect updates returned after message sent: %r"
+                % (
+                    updates,
+                ),
             )
```

```
-                ["git", "clone", "https://github.com/quora/%s.git" % (external_name,)]
+                [
+                    "git",
+                    "clone",
+                    "https://github.com/quora/%s.git"
+                    % (
+                        external_name,
+                    ),
+                ]
```
Seems like it still explodes one-tuples within other collections.

tests/data/comments2.py
```diff
@@ -142,7 +142,7 @@ def inline_comments_in_brackets_ruin_everything():
         syms.simple_stmt,
         [
             Node(statement, result),
-            Leaf(token.NEWLINE, '\n'),  # FIXME: \r\n?
+            Leaf(token.NEWLINE, '\n')  # FIXME: \r\n?
```
JelleZijlstra: What caused this change? Black should put a trailing comma here.

durin42: You're right, but you're commenting on the pre-formatted part of the test case, not the result. The spirit of the test appeared to be that the comment moved correctly with the collapsed line, so I removed the trailing comma in the input.

JelleZijlstra: Thanks! Sorry for that, it's not always easy to tell from changes in the test suite what is actually changing.

durin42:
Addressed those cases too. I'm off to bed now (got very nerdsniped by this one-tuple problem), but the tests I constructed based on your examples pass. I'm decreasingly happy with the code in `is_trailing_comma_because_of_one_tuple`, so maybe take a look at that if you have time and see if there's something cleaner I could do? Detecting the one-tuple case is harder than I expected.

JelleZijlstra:
The current version crashes on this import file:
```
import core, time, a

i = 20000
```
The stack trace ends:
```
  File "/home/jelle/repos/black/black.py", line 821, in visit
    yield from getattr(self, f"visit_{name}", self.visit_default)(node)
  File "/home/jelle/repos/black/black.py", line 1602, in visit_default
    self.current_line.append(node)
  File "/home/jelle/repos/black/black.py", line 1144, in append
    self.maybe_remove_trailing_comma(leaf)
  File "/home/jelle/repos/black/black.py", line 1355, in maybe_remove_trailing_comma
    or (self.leaves[-6].value == "import" and self.leaves[-3].value == "as")
IndexError: list index out of range
```
It also caused a bunch of changes to my (already formatted) codebase, which is unexpected. It's hard to tell for me right now exactly what the changes are because I messed up distinguishing this diff from the other changes in Black master (I'll redo the formatting later and try to distinguish the changes properly), but I *think* it caused this ugly change:
```
-                    assert False, (
-                        "could not insert even though method found: \n%s"
-                        % "".join(ret[-10:])
+                    assert (
+                        False
+                    ), "could not insert even though method found: \n%s" % "".join(
+                        ret[-10:]
                     )
```

durin42:
Alrighty, my discomfort with how I was handling oneples wore me down. I have reworked the optional-trailing-comma detection pretty extensively, and it feels better to me. At this point, the "meat" of the change is all in the `is_collection_with_optional_trailing_comma` property.

@JelleZijlstra I'm curious if this works on your codebase now. You're right that you shouldn't see changes on already-blackened code. Thanks for all the test cases - it's helped a lot.

JelleZijlstra:
I tried it again and got a similar set of changes, like this:
```
-    IGNORED_TYPES_FOR_ATTRIBUTE_CHECKING = (
-        Config.IGNORED_TYPES_FOR_ATTRIBUTE_CHECKING
-        | {pylons.controllers.WSGIController}
-    )
+    IGNORED_TYPES_FOR_ATTRIBUTE_CHECKING = Config.IGNORED_TYPES_FOR_ATTRIBUTE_CHECKING | {
+        pylons.controllers.WSGIController
+    }
```
It also still crashes on the file with `import core, time, a`; am I using the wrong commit? This was running commit `f1fb54003469ae846e4d9d2ff137c546d648e35f`.

durin42:
Sigh, the import thing. That was a latent bug, now tested and fixed.

As for the other issue: the line in your left hand diff side gets reformatted by black for me without my patches applied. Check again? I even added that exact case as a test to collections.py and you can see it gets reformatted even in the base commit of this chain, where I haven't changed any black behavior yet.

Beefy-Swain:
@JelleZijlstra Can you please explain what changed your opinion on this matter?

One of the most appealing things about Black is that (with few exceptions?) the only thing that affects code formatting is the AST and line length. This obviously goes against that.

ambv:
Black is also using pre-existing formatting for comment placement and blank lines. Yeah, we are trying to avoid this as much as possible and the original release of Black did not have either blank line retention nor comment placement retention (it used to be more aggressive moving comments to where it thought programmers should put them).

This has proven to be a tremendous area of contention. People are really dependent on their blank lines and comments for clarity. So I conceded at some point and it was a good, pragmatic decision. A slew of open source projects adopted Black right after.

The remaining issue that I deeply resonate with is: for collections that you *know* will grow, there was no way to anticipate it by exploding them into one-element-per-line. Why would you want that? Because it minimizes diffs and that is a thing Black advertises as a feature: minimizing diff noise.

Last but not least, developers of Mercurial, Django and Twisted listed this as a problem that is holding them back from adopting Black. I trust notable open source projects with decades of experience to have reasonable and well informed needs.

Summing up, yes, we are giving up a little purity for a lot of pragmatism.

Beefy-Swain:
@ambv After reading your response and giving it some consideration, I agree this change. Thanks!

cmmartti:
I'd like to suggest an alternative approach: rather than exploding when there's a trailing comma, explode when there is a newline between the opening bracket and the first element of the collection. Has this been considered?

This is the approach that Prettier takes with objects literals, and it seems to work well. There's an explanation [on the rationale page](https://prettier.io/docs/en/rationale.html#multi-line-objects).

durin42:
Based on my experiences with black's codebase, I don't think that's doable in black. I agree it seems like a reasonable approach, but done is better than perfect, and I'm not going to be able to justify sinking additional time into another approach that appears to be significantly harder (black maintainers et al please feel encouraged to correct me if I'm overlooking a way to detect this that'd be easy.)

zsol:
I agree the approach described in [the prettier docs](https://prettier.io/docs/en/rationale.html#multi-line-objects) significantly harder to do with `blib2to3` mainly because while the trailing `,` is kept as part of the parsed syntax tree, but newlines would have to be inferred from the line/column metadata attached to nodes.

cmmartti:
> done is better than perfect

I definitely agree.

I admit that I don't know anything about Black's codebase, but stumbled across this page by chance and thought I'd throw it out there in case it was easier. From an end-user's perspective, I think the newline approach is slightly more intuitive, but not enough that it really matters.

durin42:
Friendly ping: is there any further work expected of me?

KerryJones:
Any change we could get this merged into Black? This was the only thing that was holding back our ICs from agreeing to use Black

tests/data/expression.diff
```diff
@@ -116,14 +128,16 @@
...
  1.0 .real
  ....__class__
  list[str]
  dict[str, int]
  tuple[str, ...]
 -tuple[str, int, float, dict[str, int],]
-+tuple[str, int, float, dict[str, int]]
++tuple[
++    str, int, float, dict[str, int],
++]
```
merwok: Isn’t this a bug?

tests/data/expression.diff
```diff
@@ -199,7 +213,7 @@
 +    .filter(
 +        models.Customer.account_id == account_id, models.Customer.email == email_address
 +    )
-+    .order_by(models.Customer.id.asc())
++    .order_by(models.Customer.id.asc(),)
```
merwok: this too

tests/data/fmtonoff.py
```diff
@@ -150,7 +150,7 @@ def single_literal_yapf_disable():
     BAZ = {
         (1, 2, 3, 4),
         (5, 6, 7, 8),
-        (9, 10, 11, 12),
+        (9, 10, 11, 12)
     }  # yapf: disable
```
merwok: we want trailing comma here too?

tests/data/function.py
```diff
@@ -230,7 +230,7 @@ def trailing_comma():
     }
 
 
-def f(a, **kwargs) -> A:
+def f(a, **kwargs,) -> A:
```
merwok: Is this intended?  I don’t think human devs would write a one-line function signature like this.

tests/data/expression.diff
durin42: No, this is literally the point of this change. You're looking at a diff-of-diffs, maybe that's the confusing bit?

tests/data/expression.diff
durin42: Again, because it's a diff-of-diffs it's confusing, but this is expected behavior: the trailing comma is preserved, but because it's in parens it's not exploded (because it looks like a one-tuple).

tests/data/function.py
durin42: I don't see it as avoidable: we're now preserving trailing commas, and it's pretty hard to remove them only in function signatures as far as I can tell. I'm very open to suggestions, but as you say I don't think human devs would write a function like this so it's moot.

tests/data/fmtonoff.py
durin42: You're commenting on the input side. To preserve the spirit of the test, I had to make a change to the input so that things would still get formatted.

tests/data/expression.diff
merwok: Yes this is confusing.

Is black going to change one-line typing declarations to be on two lines?

tests/data/fmtonoff.py
merwok: Ah great then!

tests/data/expression.diff
durin42: If the input has a trailing comma (as in this example), yes.

tests/data/function.py
merwok: Preserving seems ok!  Even if people have to point that out in code review if it happens, there is still a lot of time saved and arguments avoided thanks to all the other things that black changes.

Thanks for replying!

zsol:
I think it should be possible to check the leaves' parents to see if they're part of a function definition or not. Collection literals' parents will be `syms.atom` and arguments in function definition will be descendents of `syms.parameters` (as well as `syms.funcdef`, but maybe parameters would be enough).

I have a general comment about the approach though: I'm wondering if assuming the entire `Line` to be a collection literal is enough or not. What if a `Line` has multiple collection literals? For example I think the PR as is would not explode this:
```
{[1,2,3,]}
```

Is that what we want? Or should we just explode all collection literals, even if there are multiple of them in the same `Line` (or maybe they span across them).

durin42:
> I think it should be possible to check the leaves' parents to see if they're part of a function definition or not. Collection literals' parents will be `syms.atom` and arguments in function definition will be descendents of `syms.parameters` (as well as `syms.funcdef`, but maybe parameters would be enough).
> 
> I have a general comment about the approach though: I'm wondering if assuming the entire `Line` to be a collection literal is enough or not. What if a `Line` has multiple collection literals? For example I think the PR as is would not explode this:
> 
> ```
> {[1,2,3,]}
> ```
> 
> Is that what we want? Or should we just explode all collection literals, even if there are multiple of them in the same `Line` (or maybe they span across them).

Yeah, the PR as written only handles the outermost collection literal in a line - interior trailing commas get preserved, but don't force wrapping. It's unclear to me how I could make that happen, and it feels like less of a pressing concern for non-top-level collections.

I see three options, none of which enthuse me:

1. Leave this PR as-is
2. Try and clean up trailing commas in non-top-level collection literals (This sounds tricky?)
3. Try and explode non-top-level collection literals with trailing commas without exploding the top level one (I'm not sure this is even possible or sensible)

The virtue of option 2 is that it means if we ever figure out how to do option 3 we don't cause format changes in already-formatted files. It sounds hard but doable. Option 1 has the virtue of being done, and treating all trailing commas consistently today.

What's the preference here?

KerryJones:
Vote: 1️⃣

martinvonz:
I'd vote for 3 (but 1 now and 3 later also makes sense). Clang does 3, AFAICT (see example in my earlier comment). I skimmed the tests to see if there was already a test case for it but I couldn't tell. It would probably be good to have a test case for it already even if it doesn't get formatted as it should (IMO) yet.

KerryJones:
Bump -- any chance we could get this one in? @ambv

KerryJones:
@merwok @JelleZijlstra what can we do to make this happen?

KerryJones:
@durin42 Nice! It got approved (thanks merwok!), looks like some conflicts appeared but otherwise good to go?

durin42:
I won’t have time until Tuesday, but can rebase the series then.

durin42:
Rebased. Tests should pass at all four commits, and hopefully the individual commits make this easier to review. Do note that this effectively backed out 9394de150ebf. I'm a little unclear if that's bad, given the special case it implies in the new trailing-comma-preserving regime, but I'm happy to try and revise b3b0d92873bd to preserve the behavior for one-argument functions. What's the preference?

durin42:
(Explicitly tagging @merwok since they already approved this once, and I have a blocking question in my previous.)

durin42:
I found a defect in this (on the hg code) where it's not stable, so I'll try and fix it up this week.

durin42:
The problem code, in case someone spots something obvious:
```    # all three of these should format the same, but they don't
    ec2client.get_waiter('instance_stopped').wait(
        InstanceIds=[instance.id],
        WaiterConfig={
            'Delay': 5,
        })
    ec2client.get_waiter("instance_stopped").wait(
        InstanceIds=[instance.id],
        WaiterConfig={"Delay": 5,},
    )
    ec2client.get_waiter("instance_stopped").wait(
        InstanceIds=[instance.id], WaiterConfig={"Delay": 5,},
    )
```

KerryJones:
> The problem code, in case someone spots something obvious:
> 
> ```
>     # all three of these should format the same, but they don't
>     ec2client.get_waiter('instance_stopped').wait(
>         InstanceIds=[instance.id],
>         WaiterConfig={
>             'Delay': 5,
>         })
>     ec2client.get_waiter("instance_stopped").wait(
>         InstanceIds=[instance.id],
>         WaiterConfig={"Delay": 5,},
>     )
>     ec2client.get_waiter("instance_stopped").wait(
>         InstanceIds=[instance.id], WaiterConfig={"Delay": 5,},
>     )
> ```

Are you presenting how they ended up being formatted? Or is there a before/after?

durin42:
That's the problematic input. It converges after two runs to a reasonable
result, but all three statements should end up the same after one run.

KerryJones:
@durin42 Sorry, it's unclear to me if you're saying more changes need to be done in your PR or something else is funky? Still want to merge at this point?

durin42:
I definitely need to make more changes, but I haven't had the time due to other commitments. I intend to come back to it, but if someone beats me to figuring out the problem that's fine. I expect to get to it in a couple of weeks, but it _might_ happen this week if I get lucky.

durin42:
I'm really confused now. I had internalized this mental model that lines get un-wrapped, formatted, and then wrapped if too long or marked explode. but the problematic code block I get back as multiple lines, eg:

```
    ec2client.get_waiter('instance_stopped').wait(
        InstanceIds=[instance.id], WaiterConfig={'Delay': 5,})

```

comes through the formatting logic (specifically `Line.is_collection_with_optional_trailing_comma`) as two separate lines. At some point, a new comma gets added by black (!), resulting in 

```
    ec2client.get_waiter("instance_stopped").wait(
        InstanceIds=[instance.id],
        WaiterConfig={"Delay": 5,},
    )
```


which then formats on a second pass to 
```
    ec2client.get_waiter("instance_stopped").wait(
        InstanceIds=[instance.id], WaiterConfig={"Delay": 5,},
    )
```

As far as I can tell it's the comma after the 5 that causes things to explode, but I'm confused why I ever get a line with the leaves

```
[Leaf(NAME, 'WaiterConfig'), Leaf(EQUAL, '='), Leaf(LBRACE, '{'), Leaf(STRING, '"Delay"'), Leaf(COLON, ':'), Leaf(NUMBER, '5'), Leaf(COMMA, ','), Leaf(RBRACE, '}'), Leaf(COMMA, ',')]
```

That's the second pass through `is_collection_with_optional_trailing_comma` - so I guess the line is getting split (why?) and then we're getting confused by an extra trailing comma that never existed in the source. I'm pretty stuck on this for now, so I think I'm going to take a break now. So far all my attempts to stabilize this have made things materially worse. I'll go ahead and push the new version of things with the failing test in case anyone gets inspired by my rambling here. :(

durin42:
Okay, took a bit, cleared my head, and I found it.

The bad news: something has regressed in master and there are now two files in hg that break black, but don't as of the release version. That's been filed as #993, and I don't think there's any other lurking failures hiding behind that. I'd appreciate it if other people could kick the tires on _this_ PR and make sure it doesn't have any fun regressions.

durin42:
@KerryJones @merwok this is ready for review and additional testing to make sure it doesn't regress anyone's codebases etc.

kerrick-lyft:
@durin42 does this change address the below test case?
```
print(
    {
        "123456789": 123456789,
        "987654321": 987654321,
        "qwertyuiop": "qwertyuiop",
        "asdfghjkl": "asdfghjkl",
    }
)
```

I think `black -t py36` should add a trailing comma on the second-to-last line here, and I think your PR is trying to fix things like this (although I may be mistaken)?

What I'm seeing is that on master, `black -t py36` will not change this code, and if I add a comma, black will remove it. On your branch, `black -t py36` will not change the code regardless of whether there is a trailing comma (which is bad, right? It shouldn't be inconsistent).

So I think you should handle the above case and add a trailing comma. Let me know if this is unrelated to your PR and I should file a separate bug report though.

durin42:
It doesn't address that, and after some reflection I'm pretty confident that it shouldn't: this patch isn't about adding new commas, it's about using existing trailing commas as evidence something should be exploded. In your particular example, the function call makes this a little ambiguous IMO.

I'd be happy for someone to do a follow-up that could consistently add that comma in supported versions of Python (IIRC trailing commas aren't always legal?), but it's not something I'm planning to address. Thanks!

KerryJones:
@merwok @JelleZijlstra can we get this merged? We've been excited for this for months 😎

KerryJones:
@ambv for good measure ^

zsol:
I'd love to have this feature in master - being able to hint to the formatter about exploding a collection literal would be helpful fairly often in practice.

Having said that, I have two concerns with this PR as it stands:
1. The dictionary literal in `[{"a": 1, "b": 2,}]` should be exploded. I think all collection literals with a trailing comma should be, regardless of the level of nesting. This is super helpful when you have a list of dictionaries where most items are long (so Black naturally expands them), but a couple are short enough to fit on one line, so they get lost in-between the expanded ones.

2. If an expression is not exploded, Black should remove the trailing comma. The same example applies: `[{"a": 1, "b": 2,}]` is considered well-formatted by the current PR. This is even more annoying in function signatures, as pointed out [above](https://github.com/psf/black/pull/826/files#r292184883).

durin42:
@zsol (1) is, as far as I can tell, not fixable without radically rewriting the line-splitting logic in black, which I'm not signing up to do. (2) is a necessary consequence of this PR. I agree that fixing (1) would be a better solution, but I don't see a way to do it.

KerryJones:
@zsol Thoughts?

durin42:
Is there anything I can do to get this un-stuck? At this point it's blocking some cleanup work on Mercurial's code, which is about to become pressing.

durin42:
Friendly ping on this! I'd love to get this un-stuck, but it's not clear from me what (if anything) I can do to help. Thanks!

/cc @merwok @ambv @JelleZijlstra

merwok:
I’ve only given an outsider approval after some of my questions were answered, but I am not a black maintainer.

ambv:
This will be merged next week.

-- 
Best regards,
Łukasz Langa

durin42:
Awesome, thanks!

KerryJones:
@durin42 any chance you could update for conflicting file in case ambv gets to it today/tomorrow?

durin42:
Done. That was less than a pain than I thought it would be.

KerryJones:
@ambv Any chance we could get this merged? Still waiting on this

KerryJones:
@zsol any chance you can bring this to @ambv? It's been a couple of weeks

thinker3:
> I ran your branch on my codebase and noticed some changes like this:
> 
> ```
> -    DEFAULT_DIRS = (str(AIRFLOW_DIR / "lib/floq"),)
> +    DEFAULT_DIRS = (
> +        str(AIRFLOW_DIR / "lib/floq"),
> +    )
> ```
> 
> This seems undesirable; we need the trailing comma to make the tuple, but that doesn't mean we need to explode it.

need an custom option whether to explode one-element tuple

PR #834, opened 2019-05-06
remove unnecessary parens around yield

JelleZijlstra:


JelleZijlstra:
This produces unstable formatting when there are *two* sets of parens around the yield. Will fix that.

PR #852, opened 2019-05-16
Remove happiness of error message

glasnt:
In the even that `black` encounters an error, the output gleefully tells the user it is "All done!", followed by the emoji ligature of sadness. 

This PR corrects that by ensuring that there are no mixed feelings on error, that indeed two collision emoji and a broken heart emoji is a very not good state to be in, and now returns "Oh no!"

Sample output: 

Before: 

```
error: cannot format app.py: Cannot parse: 13:33: @app.route("/", methods=["GET"]):
All done! 💥 💔 💥
1 file failed to reformat.
```

After: 

```
error: cannot format app.py: Cannot parse: 13:33: @app.route("/", methods=["GET"]):
Oh no! 💥 💔 💥
1 file failed to reformat.
```

### DIFF
--- a/black.py
+++ b/black.py
@@ -457,8 +457,7 @@ def main(
         )
 
     if verbose or not quiet:
-        bang = "💥 💔 💥" if report.return_code else "✨ 🍰 ✨"
-        out(f"All done! {bang}")
+        out("Oh no! 💥 💔 💥" if report.return_code else "All done! ✨ 🍰 ✨")
         click.secho(str(report), err=True)
     ctx.exit(report.return_code)
 


JelleZijlstra:
Love the branch name

zsol:
Haha thanks for making my day :) What if there are multiple input files, some of them just fine, and some with errors?

glasnt:
Hi @zol. I haven't changed any of the functionality; there's one error_code returned by one report object, so the functionality is the same: if there's any errors at all, sadness is eternal. 

```
Oh no! 💥 💔 💥
1 file left unchanged, 2 files failed to reformat.
```

zsol:
Fair enough. Thanks 💯

PR #858, opened 2019-05-20
minor performance improvement (~2% speedup in unit tests)

mflaxman:
I only clocked a 2% improvement in unit tests (meh) but if my understanding is correct this matters more for large files (where performance is most important).

https://stackoverflow.com/questions/39675898/is-python-string-concatenation-bad-practice
https://stackoverflow.com/a/3055541

I don't feel strongly (I was surprised the performance improvement was so small), so I won't mind if you decide to not merge this.


mflaxman:
Thanks @JelleZijlstra!

PR #859, opened 2019-05-20
Add doc clarifying that there is no blackd client

### LINKED ISSUE #854: Documentation request: best way to setup a blackd client tool?
I tried to finally get setup with blackd running as a systemd user unit, and I'm not clear from the docs how I should be setting up the client side of this.

<details><summary>I'm able to setup a working unit with this config</summary>

```
$ cat ~/.config/systemd/user/blackd.service
[Unit]
Description=blackd autoformatting server
[Service]  # installed in a virtualenv
ExecStart=/home/sirosen/bin/blackd-env/bin/blackd --bind-port 9090
[Install]
WantedBy=default.target
```
</details>

<details><summary>and then test this with simple curl</summary>

```
$ curl -sf -XPOST localhost:9090 -d "$(cat setup.py)" > /dev/null; echo $?
0
```
</details>

So it works, hooray!

I wanted to start using this in pre-commit hooks and other places.

I almost asked "how do I setup `black` to call out to `blackd`?" but that obviously defeats the whole point of blackd. How are people doing this? Is there some standard client tool in use?

I think even a doc note that "you need to write your own blackd client" is fine, just to alleviate any confusion about this.

sirosen:
Resolves #854

The first sentence of this is pretty uncontroversial. (Though I wasn't sure exactly where in the text to put it.)

I thought it would also be nice to show the `curl` test with a tiny statement that actually reformats.
If this worries anyone (e.g. "makes the readme too much longer"), I'll just cut this back down to the first sentence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -844,6 +844,14 @@ Options:
   -h, --help                      Show this message and exit.
 ```
 
+There is no official blackd client tool (yet!). You can test that blackd is
+working using `curl`:
+
+```
+blackd --bind-port 9090 &  # or let blackd choose a port
+curl -s -XPOST "localhost:9090" -d "print('valid')"
+```
+
 ### Protocol
 
 `blackd` only accepts `POST` requests at the `/` path. The body of the request


zsol:
Thank you!

PR #865, opened 2019-05-24
fix some out-of-date docstrings; other cleanup

JelleZijlstra:


black.py
```diff
@@ -3446,8 +3447,6 @@ def parse_ast(src: str) -> Union[ast3.AST, ast27.AST]:
 def assert_equivalent(src: str, dst: str) -> None:
     """Raise AssertionError if `src` and `dst` aren't equivalent."""
 
-    import traceback
```
JelleZijlstra: One of Black's dependencies already imports this, so there's no point in deferring the import here.

zsol:
lgtm

PR #866, opened 2019-05-24
Document cache location configuration

### LINKED ISSUE #574: Make cache location configurable
hello,

It would be nice to make cache loccation configurable by cli argument. 

Here's my use case: We use bitbucket pipelines, and in their workflow I can only cache/restore directories relative to the project's path.

So, if you could please make the cache location configurable, i could add `black --check --cache-dir=./black_cache` on bitbucket pipeline, cache that directory, and make black execution really really fast 😸 

Thanks, awesome project!!

yhoiseth:
Fixes #574.


zsol:
Thanks!

Issue #869, opened 2019-05-24
Inconsistent README formatting


yhoiseth:
Operating system: Not relevant
Python version: Not relevant
*Black* version: Not relevant
Does also happen on master: Not relevant

The README formatting appears inconsistent to me. Most lines seem to be limited to about 80 characters, but some are much longer. For example, the first line in the _Configuration format_ section is 124 characters.

This is by no means a big problem, but it did cause me to pause when I proposed a change to the documentation in #866.

I'll try to propose a fix.

hugovk:
https://github.com/psf/black/pull/874 has been merged, this issue can be closed.

---

Tip: https://github.com/psf/black/pull/874#issuecomment-544280566 says:

> Closes #869.

That needs to be in the PR description to autoclose, not in a comment.

PR #870, opened 2019-05-24
[WIP] Automatic markdown formatting

yhoiseth:
This is a proof of concept for how #869 could be fixed. Considering this project's philosophy, I think it's natural to format the documentation automatically.

# Implementation

I was unable to find a Python package to handle the formatting, so I went with the Node package [Prettier](https://prettier.io/), which seems to be a de facto standard for formatting markup. This adds a development dependency on Node which I think is worthwhile, but I may be wrong.

To demonstrate the effect, I ran `node_modules/prettier/bin-prettier.js README.md --prose-wrap=always --write --print-width=88`.

# Further work

If we're going to use this implementation, we would need to do a few more things:

- [ ] Formatting check in CI
- [ ] Pre-commit hook
- [ ] Update `CONTRIBUTING.md`
- [ ] Format all Markdown files

There may be others that I haven't thought of.


yhoiseth:
@ambv May I ask you to have a look at this? If you agree that this should be done and approve of my approach, I can try to finish the pull request. If not, that's fine – I won't take offense.

acdha:
@yhoiseth Since Black is already using [pre-commit](https://pre-commit.com), you could avoid needing to manage the dependency twice by simply adding that to the `.pre-commit-config.yaml` file:

```yaml
repos:
    - repo: https://github.com/prettier/prettier
      rev: 1.17.1
      hooks:
          - id: prettier
```

That will automatically manage the installation for you and works with the existing infrastructure for running it locally, during CI, etc. If you want to run it ad hoc, you can just do `pre-commit run --file …` or `pre-commit run --all-files`.

yhoiseth:
@acdha That's absolutely brilliant. I had no idea pre-commit could do that. Thanks a lot!

I've created #874 as a replacement for the current pull request.

acdha:
Awesome, glad I could help! I've been making up for lost time moving as many things into pre-commit lately myself.

PR #887, opened 2019-06-06
Fix behavior for empty directories.

### LINKED ISSUE #886: Misleading message when black checks only empty directories
If I provide to black a directory or several directories that exist but they are empty (all of them) black reports the following:
> No paths given. Nothing to do  

I used the following command
`black example1 example2`

The message should be different  because I have actually provided the paths. They are empty, though.

Operating system: Windows
Python version: 3.6.3
*Black* version: 19.3b0

lightlace:
Fix the behavior if any of the given directories is empty.
Fixes #886.

### DIFF
--- a/black.py
+++ b/black.py
@@ -434,12 +434,16 @@ def main(
             sources.update(
                 gen_python_files_in_dir(p, root, include_regex, exclude_regex, report)
             )
+            if len(os.listdir(p) ) == 0:
+                if verbose or not quiet:
+                    out(f"Directory {p} is empty.")
+                    continue
         elif p.is_file() or s == "-":
             # if a file was explicitly given, we don't care about its extension
             sources.add(p)
         else:
             err(f"invalid path: {s}")
-    if len(sources) == 0:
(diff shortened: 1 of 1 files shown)


black.py
```diff
@@ -434,12 +434,16 @@ def main(
             sources.update(
                 gen_python_files_in_dir(p, root, include_regex, exclude_regex, report)
             )
+            if len(os.listdir(p) ) == 0:
```
JelleZijlstra: This could be expensive if the directory is large. We should do this kind of check instead below where we print the message about "No paths given", so we don't incur more I/O cost in the normal case.

zsol: Or, we could just specialize the logic below to say
```python
if len(sources) == 0:
  if verbose or not quiet:
    if len(src) == 0:
      out("No paths given. Nothing to do 😴")
    else:
      out("Paths given contain no files to be formatted. 🤷")
```

And there's no need to `listdir()` even in the error case

JelleZijlstra:
We fixed this in #999 instead.

Issue #880, opened 2019-06-02
Trailing comma for function with one arg


cosinequanon:
A function with one argument keeps the trailing comma but if there are two it is stripped.

In:
```
def one(
    a,
):
    pass

def two(
    a,
    b,
):
    pass
```
Out
```
def one(a,):
    pass


def two(a, b):
    pass
```
From https://black.now.sh/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ACGAFNdAD2IimZxl1N_WlbvK5V-4TOOTwYyJEleH-JCdis69M0BXQy1O1PWCxUpRscYAufHo8BjrjIj1dM6YL_cyWPemFXxt1cpBhl0f8lWpI3sZ5u99cYAAACFuqAGuJSBnwABb4cBAAAAtWKNgbHEZ_sCAAAAAARZWg==

zsol:
That's a bug, `one` should be formatted like so:
```
def one(a):
    pass
```

drunkwcodes:
It is still runnable.

stagedatalosstest:
test123

dylanjblack:
I made pull request #891 with a potential fix. Please let me know if the approach makes sense and if you have any suggestions for improvement.

Jma353:
Looks like this was fixed in 9394de150ebf0adc426523f46dc08e8b2b2b0b63.  Can this be closed?

zsol:
yep.

PR #891, opened 2019-06-09
Fix trailing comma for function with one arg (#880)

dylanjblack:
Modified maybe_remove_trailing_comma to remove trailing commas for typedarglists (in addition to arglists), and updated line split logic to ensure that all lines in a function definition that contain only one arg have a trailing comma.

Some discussion points:

- There are likely some edge cases to consider with the use of `no_commas`. Would be great to hear any other suggestions for this test. 
- The new test data file should probably be folded into one of the existing function test files, are there any guidelines around this? 
- It looks like this will clash with Pull request #826.

### DIFF
--- a/black.py
+++ b/black.py
@@ -1352,7 +1352,10 @@ class Line:
             bracket_depth = leaf.bracket_depth
             if bracket_depth == depth and leaf.type == token.COMMA:
                 commas += 1
-                if leaf.parent and leaf.parent.type == syms.arglist:
+                if leaf.parent and leaf.parent.type in {
+                    syms.arglist,
+                    syms.typedargslist,
+                }:
                     commas += 1
                     break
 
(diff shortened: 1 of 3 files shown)


illfated:
(ref. issue #880 )

black.py
```diff
@@ -2488,9 +2491,12 @@ def bracket_split_build_line(
         if leaves:
             # Since body is a new indent level, remove spurious leading whitespace.
             normalize_prefix(leaves[0], inside_brackets=True)
-            # Ensure a trailing comma for imports, but be careful not to add one after
-            # any comments.
-            if original.is_import:
+            # Ensure a trailing comma for imports and standalone function arguments, but
+            # be careful not to add one after any comments.
+            no_commas = original.is_def and not [
```
JelleZijlstra: Use `any()`

tests/test_black.py
```diff
@@ -264,6 +264,14 @@ def test_function2(self) -> None:
         black.assert_equivalent(source, actual)
         black.assert_stable(source, actual, black.FileMode())
 
+    @patch("black.dump_to_file", dump_to_stderr)
+    def test_function3(self) -> None:
```
JelleZijlstra: It's fine to have another test file, but could you name it something less generic, like `test_function_trailing_comma`?

black.py
dylanjblack: Good call, cheers.

PR #895, opened 2019-06-20
Fix for #886

woopsix:


JelleZijlstra:
This got fixed by #999 instead.

Issue #907, opened 2019-06-24
type errors with newest mypy


JelleZijlstra:
CI is currently failing because of some mypy errors, likely because of the recent mypy release: https://travis-ci.org/python/black/jobs/549813115

PR #909, opened 2019-06-25
Force parentheses between unary op and binary power.

### LINKED ISSUE #646: combination of unary and power operator leads to misleading formatting.
Operating system: `Ubuntu 18.04.1 LTS`
Python version: `Python 3.6.7`
*Black* version: `18.9b0`
Does also happen on master: as of 2018-12-20T09:32, yes.

(as per my tweet: https://twitter.com/llanga/status/1075484480436080640 )

The code

    -2**8

blacks to

    -2 ** 8

I think this is misleading because the grammar reduction rules mean that the code actually parses as `-(2**8)`.
The result of this code, however you format it, is -256, not 256.

I'm not sure what to suggest. Perhaps _not_ altering the space around `**`?
Though I can see that almost any suggestion will be controversial.

brandtbucher:
Fixes #646. This patch forces expressions containing unary and power ops, like this...

```py
+a ** -b ** ~c ** -d  # This is legal blackened code.
```

...to be reformatted with parentheses to make operator precedence clear, like this...

```py
+(a ** -(b ** ~(c ** -d)))  # This is legal blackened code.
```

### DIFF
--- a/black.py
+++ b/black.py
@@ -1644,6 +1644,19 @@ class LineGenerator(Visitor[Line]):
             node.children[2].value = ""
         yield from super().visit_default(node)
 
+    def visit_factor(self, node: Node) -> Iterator[Line]:
+        """Force parentheses between a unary op and a binary power:
+
+        -2 ** 8 -> -(2 ** 8)
+        """
+        child = node.children[1]
+        if child.type == syms.power and len(child.children) == 3:
+            lpar = Leaf(token.LPAR, "(")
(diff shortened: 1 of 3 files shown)


brandtbucher:
The two remaining mypy errors are due to #907, unrelated.

brandtbucher:
Is there anything more I need to do for this? This is my first time contributing, so I'm not sure if I missed something!

brandtbucher:
Thanks!

PR #910, opened 2019-06-25
Ignore broken E203

Juanlu001:
See https://github.com/python/black/issues/565


Juanlu001:
Thanks @JelleZijlstra!

PR #911, opened 2019-06-25
Fix mypy errors.

### LINKED ISSUE #907: type errors with newest mypy

brandtbucher:
Fixes #907.

### DIFF
--- a/black.py
+++ b/black.py
@@ -1,5 +1,4 @@
 import asyncio
-from asyncio.base_events import BaseEventLoop
 from concurrent.futures import Executor, ProcessPoolExecutor
 from datetime import datetime
 from enum import Enum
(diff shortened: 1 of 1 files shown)


black.py
```diff
@@ -504,7 +504,7 @@ def reformat_many(
     report: "Report",
 ) -> None:
     """Reformat multiple files using a ProcessPoolExecutor."""
-    loop = asyncio.get_event_loop()
+    loop = cast(asyncio.BaseEventLoop, asyncio.get_event_loop())
```
JelleZijlstra: We should avoid `cast` where possible. Can you change the argument types of schedule_formatting() and shutdown() instead?

brandtbucher: Good suggestion. Done!

Issue #251, opened 2018-05-23
isort removes newlines before standalone comments if the following line is an import of the same group


gaborbernat:
Operating system: Arch Linux
Python version: Python 3.6.6
Black version: 18.5b0
Does also happen on master: yes

Given: 

``` python
import a
# noinspection PyProtectedMember
import _a
```

Black formats this to (note the introduced empty line):

``` python
import a
# noinspection PyProtectedMember

import _a
```

However isort correctly (in the recommended mode 3) formats it to:
``` python
import a
# noinspection PyProtectedMember
import _a
```

ambv:
Black puts an empty line after every group of imports. A standalone comment signifies an end of a group of imports for it.

So, your example is wrong, Black will do this instead with the original file:
```diff
--- /tmp/asd.py  (original)
+++ /tmp/asd.py  (formatted)
@@ -1,4 +1,5 @@
 import a
+
 # noinspection PyProtectedMember
 import _a
```

The behavior between isort and Black is indeed different here. I haven't decided yet how to resolve this.

gaborbernat:
Thanks for the explanation, would be nice to somehow support this use case though.

skorokithakis:
I am affected by this as well :/ Just to clarify, what is the problem here? Is Black wrong, or is isort? Is the newline supposed to be there?

tibz7:
I'm also affected. what is the current status about this bug?

thisfred:
I think black is being too prescriptive here: yes comments between imports can be used as headers, but they can also be things like: 

`# XXX: unused import, that however shouldn't be removed because our noddy framework depends on it`

Sharadh:
Hello @ambv ! Do you have an update on this?
> The behavior between isort and Black is indeed different here. I haven't decided yet how to resolve this.

Perhaps you can talk through your thoughts so we understand if a fix is likely?

Also, re:
> Black puts an empty line after every group of imports. A standalone comment signifies an end of a group of imports before it.

Is there some documentation that describes this behavior? I couldn't find it in [the code style](https://black.readthedocs.io/en/stable/the_black_code_style.html) - I ask because without knowing the reasoning behind this behavior, I find it hard to suggest a way forward. I do find the rest of the decisions well documented, and do a great job of answering my questions and feature requests before I make them ;)

Sharadh:
Does anyone know what the best way to get a contributor's attention here is? I'm happy to submit a proposal + fix, it's just that the radio silence makes me nervous about investing time :)

[...]

PR #912, opened 2019-06-25
fix issue #251: isort removes newlines before standalone comments if …

EgorChernik:
With this tested commit, I'd like to give the User the ability to choose if he wants to use default Black's behavior or wants to use it with Isort, accepting its behavior. Run test:
<pre>pytest ./tests/test_black.py -k 'test__maybe_empty_lines'</pre>

For using with Isort  it's just enough to add in command line option <code>--import-mode=1</code>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ _build
 .vscode
 docs/_static/pypi.svg
 .tox
+.idea
 __pycache__
 black.egg-info
 build/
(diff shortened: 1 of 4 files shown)


EgorChernik:
How do you see for any refactor? I suppose that black.py is too big to support. When some contributors are solving different issues in one module at the same time merge issues can occur. And not only that

JelleZijlstra:
We're generally not interested in adding command-line flags that affect formatting based on user preferences. The isort project is apparently interested in releasing a Black-compatible configuration, which would be a better solution.

PR #918, opened 2019-07-03
Remove rpar case from can_remove_invisible_parens

### LINKED ISSUE #824: Line break rules (on binary operators)
Python version: 3.7.1
*Black* version: 18.9b0

For a code snippet like

```python
conv1_out = model.conv1.forward(input_onehots).detach().numpy() - model.conv1.bias.detach().numpy().reshape(-1, 1)
```

currently, it will be formatted as

```python
conv1_out = model.conv1.forward(
    input_onehots
).detach().numpy() - model.conv1.bias.detach().numpy().reshape(-1, 1)
```

and

```python
conv1_out = model.conv1.forward(
    input_onehots
).detach().numpy() - model.conv1.bias.detach().numpy().reshape(
    -1, 1
)
```

These look wired. Shall we make it to be

```python
conv1_out = (
    model.conv1.forward(input_onehots).detach().numpy()
    - model.conv1.bias.detach().numpy().reshape(-1, 1)
)
```

alex-xnor:
#824 is blocking adoption of Black here at @xnorai, so I tried to poke around inside Black to see what I could do to fix it. After a number of attempts I found the `RPAR` line here to be the sole culprit:

https://github.com/python/black/blob/f3bb22a828495c66d8d2dae1edf45e15fc2daddb/black.py#L3793-L3796

After removing that line, I found the resulting output to be much more palatable. The test cases that broke when I changed that, also (in my subjective opinion) are more readable after the change as well (with the exception of the `dis_c_instance_method` case, which seems different but neither better nor worse).

I looked in the history to see if there was any reason for it to be the way it is now, and I found 87b8df28c48353d3d08d7e88d178c7a567de816a, but there's not a whole lot of context offered in the commit message or surrounding changes there, other than that it's a big overhaul of how invisible parentheses were handled.

Fixes #824.

### DIFF
--- a/black.py
+++ b/black.py
@@ -482,8 +482,11 @@ def reformat_one(
                 res_src = src.resolve()
                 if res_src in cache and cache[res_src] == get_cache_info(res_src):
                     changed = Changed.CACHED
-            if changed is not Changed.CACHED and format_file_in_place(
-                src, fast=fast, write_back=write_back, mode=mode
+            if (
+                changed is not Changed.CACHED
+                and format_file_in_place(
+                    src, fast=fast, write_back=write_back, mode=mode
+                )
             ):
                 changed = Changed.YES
             if (write_back is WriteBack.YES and changed is not Changed.CACHED) or (
@@ -585,8 +588,9 @@ async def schedule_formatting(
                 changed = Changed.YES if task.result() else Changed.NO
                 # If the file was written back or was successfully checked as
                 # well-formatted, store this information in the cache.
-                if write_back is WriteBack.YES or (
-                    write_back is WriteBack.CHECK and changed is Changed.NO
+                if (
+                    write_back is WriteBack.YES
+                    or (write_back is WriteBack.CHECK and changed is Changed.NO)
                 ):
                     sources_to_cache.append(src)
                 report.done(src, changed)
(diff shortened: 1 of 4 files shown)


ambv:
We cannot accept this change. Not introducing extra parentheses for asserts which are very common in pytest was something Black is designed to do.

akhmerov:
Sorry, can you please elaborate why? Black does sometimes introduce parentheses, and it's unclear to me which of the two cases is preferable. Is there perhaps a motivation for this in the docs—I've looked [here](https://black.readthedocs.io/en/stable/the_black_code_style.html#parentheses), but the motivation seems unclear.

ambv:
I commented on the issue. We're trying not to blow up a single line into very many lines if we can help it. Your example on the issue produces the same number of lines but there are countless others which would become longer with this PR. This PR includes some of them.

You have to understand that there's 100+ million lines of Python code being auto-formatted with Black right now. If we make big formatting changes like the one you're proposing in this PR, it will be hugely disruptive for very many people.

akhmerov:
> I commented on the issue.

Your reply in the issue was about refactoring the specific example of when black doesn't do a good job right now, not an explanation why this proposal against the black design.

> We're trying not to blow up a single line into very many lines if we can help it.

But this isn't always true: in method chaining black doesn't try to minimize the number of lines, and if it does split method chaining into multiple lines it also introduces parentheses. So minimizing the number of lines doesn't seem to have top priority in black.

> If we make big formatting changes like the one you're proposing in this PR, it will be hugely disruptive for very many people.

I see; this wasn't very clear to me from your previous comment. Does black have a policy when a change of formatting is admissible? I see method chain formatting was changed in a minor version, and it seems to be comparably important. [Also NB: this isn't my proposed change (just to give credit to the author).]

---

The reason why #824 exists is because black makes code less readable by minimizing punctuation over making the hierarchical structure of the ast easy to read. Refactoring complex expressions might offer a workaround, but this way users need to learn to make their code tailored for black, and that is contrary to black's design idea of "turn on and forget it's even there".

While sometimes it is indeed appropriate, in complex expressions (especially in math) it hurts readability. I understand that there may be multiple reasons why this PR isn't acceptable, but so far it isn't clear to me what those reasons are.

PR #928, opened 2019-07-16
Use nullcontext in case when lock is None. Shutdown pool after code formatting.

uriyyo:
After code formatting ProcessPoolExecutor didn't shutdown, fix it.

Also use `nullcontext` from python3.7 to make code more readable.

For instance this code:
```python
if lock:
    lock.acquire()
try
    ...
finally:
    if lock:
        lock.release()
```
Turns into:
```python
with lock or nullcontext():
    ...
```


zsol:
Thanks!

PR #929, opened 2019-07-18
Set correct return statement for `is_type_comment` function

uriyyo:
Looks like there are mistake in  `is_type_comment` function. 

When  function return statement was updated from:
```python
    return bool(
        (t == token.COMMENT or t == STANDALONE_COMMENT) and (v.startswith("# type:"))
    )
```
to
```python
    return t in {token.COMMENT, t == STANDALONE_COMMENT} and v.startswith("# type:")
```

In set there are comparison:
```python
{token.COMMENT, t == STANDALONE_COMMENT}
```
I think set should be :
```python
{token.COMMENT, STANDALONE_COMMENT}
```

Correct me in case if this isn't bug.
Thanks


zsol:
Thanks! Would you mind adding a test for this?

uriyyo:
I was trying to find case that will cover this, but don't find such case.

Only one "type comment" is `STANDALONE_COMMENT`, I am talking about function signature type comment. Such cases already in `comments6.py`, for instance one of them:
```python

# test type comments
def f(a, b, c, d, e, f, g, h, i):
    # type: (int, int, int, int, int, int, int, int, int) -> None
    pass
```

If you know test that will cover this case, please share it with me.
Thanks

ambv:
Good catch!

PR #935, opened 2019-07-22
Add support for walrus operator

### LINKED ISSUE #711: Python 3.8 — PEP 572 compatibility
Operating system: Fedora 29 Linux
Python version: 3.8.0a1 (build from Python.org sources)
*Black* version: 18.9b0
Does also happen on master: yes, localy. on black.now.sh it fails with `type object 'FileMode' has no attribute 'from_configuration'`

My example for a too long line in a file trying the upcoming [PEP 572](https://www.python.org/dev/peps/pep-0572/) Assignment Expressions failed with 
```
error: cannot format test.py: Cannot parse: 4:23:     while not (0 < (answer := int(input("Please guess a positive number up to 10: "))) <= 10):
All done! 💥 💔 💥
1 file failed to reformat.
```

The example is the following:
```
#!/usr/bin/env python38
# -*- coding: utf-8 -*-

while not (0 < (answer := int(input("Please guess a positive number up to 10: "))) <= 10):
    print("That is not correct input, please retry.")

print(f"Your guess {answer} is not the number I chose :-p")
```

I don't think that this is an urgent issue, because Python 3.8 got its first alpha not long ago. I just wanted to test if it works and because it does not, and I haven't found an issue yet, I'm filing this one.

Thank you for your work!

zsol:
- [x] Add ability to parse `:=`
- [x] Adjust parenthesis logic to never produce syntax errors with `:=`

Fixes #711


tests/data/pep_572.py
```diff
@@ -0,0 +1,41 @@
+(a := 1)
+(a := a)
+if (match := pattern.search(data)) is None:
+    pass
+[y := f(x), y ** 2, y ** 3]
+filtered_data = [y for x in data if (y := f(x)) is None]
+(y := f(x))
+y0 = y1 := f(x)
```
zsol: this is invalid

tests/test_black.py
```diff
@@ -287,8 +287,9 @@ def test_pep_572(self) -> None:
         self.assertFormatEqual(expected, actual)
         black.assert_stable(source, actual, black.FileMode())
         major, minor = sys.version_info[:2]
-        if major >= 3 and minor >= 8:
-            black.assert_equivalent(source, actual)
+        # TODO: typed_ast doesn't support python 3.8
+        # if major >= 3 and minor >= 8:
```
hugovk: Careful, this would be `False` for Python 4.0 - 4.7

black.py
```diff
@@ -3493,20 +3494,32 @@ def __str__(self) -> str:
...
-def parse_ast(src: str) -> Union[ast3.AST, ast27.AST]:
-    for feature_version in (7, 6):
-        try:
-            return ast3.parse(src, feature_version=feature_version)
-        except SyntaxError:
-            continue
+def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
+    major, minor = sys.version_info[:2]
+    # TODO: support Python4+ ;)
+    if major == 3 and minor >= 8:
```
JelleZijlstra: You can just do `if (major, minor) >= (3, 8)` and we'll have Python 4 support :)

zsol: No because the for loop below :(

black.py
```diff
@@ -3040,6 +3049,12 @@ def is_one_tuple(node: LN) -> bool:
     )
 
 
+def is_walrus_assignment(node: LN) -> bool:
+    """Return True iff `node` is of the shape ( test := test )"""
```
funkyfuture: minor typo `iff`

zsol: https://en.wikipedia.org/wiki/Iff :)

tests/data/pep_572.py
```diff
@@ -0,0 +1,41 @@
...
+foo(x := 3, cat="vector")
+foo(cat=(category := "vector"))
+if any(len(longline := l) >= 100 for l in lines):
+    print(longline)
+if env_base := os.environ.get("PYTHONUSERBASE", None):
+    return env_base
+if self._is_special and (ans := self._check_nans(context=context)):
+    return ans
+foo(b := 2, a=1)
+foo(b := 2, a=1)
```
florczakraf: do we need a duplicate?

PR #937, opened 2019-07-24
Use better default venv dir when using neovim

Tyilo:


ambv:
I'm not opposed to this change but I would like to understand how this change is better for NeoVim.

Tyilo:
`~/.local/share/nvim/` is Neovim's data directory.

PR #941, opened 2019-07-25
test_black.py - formatting

Abdur-rahmaanJ:


zsol:
I think the original reads better, I don't think I'll merge this. Sorry.

Abdur-rahmaanJ:
Ok ^^_

PR #948, opened 2019-07-28
Fix unstable format involving backslash + whitespace at beginning of file

### LINKED ISSUE #922: Black produced different code on the second pass of the formatter: Explicit line join in the first line
Operating system: macOS Mojave 10.14.5 (18F132)
Python version: Python 3.7.3
*Black* version: 19.3b0 (f3bb22a828)

Minified reproducer:
```python
\


pass
```

Black diff:
```diff
--- source
+++ first pass
@@ -1,5 +1,3 @@
-\
-

 pass

--- first pass
+++ second pass
@@ -1,3 +1,2 @@
-
 pass
```

Looks like some kind of boundary condition in `EmptyLineTracker`?

Jma353:
This PR fixes #922 - an unstable sort involve Python line continuation with `\`.  The issue describes the problem pretty well.

To my knowledge, there should be no situation where Black outputs a file formatted with empty lines at the beginning.  By performing this check in the format function, we fix the above issue.

NOTE: Open to feedback on this solution, threw it together pretty quickly after doing some basic debugging of how Black parses lines and whatnot.  Open to feedback on it.  Also, I'm pretty sure it doesn't break any invariants of the formatter, but would like feedback on that as well.


black.py
```diff
@@ -736,8 +736,11 @@ def format_str(src_contents: str, *, mode: FileMode) -> FileContent:
         for _ in range(after):
             dst_contents.append(str(empty_line))
         before, after = elt.maybe_empty_lines(current_line)
-        for _ in range(before):
-            dst_contents.append(str(empty_line))
+        # Black should not insert empty lines at the beginning
+        # of the file
+        if len(dst_contents) > 0:
```
zsol: I think this logic would be better encapsulated inside `EmptyLineTracker`.

Jma353: Will refactor so I can move it!

Jma353:
Moved the check to `EmptyLineTracker` @zsol, this should be good for another review.

zsol:
Thanks!

Issue #68, opened 2018-03-23
Black changes '#:' to '# :'


sfermigier:
Operating system: Mac OS
Python version: 3.6.4
Black version: 18.3a3
Does also happen on master: yes

---

'#:' are used by Sphinx to generate documentation on object attributes.

Black erroneously changes them to '# :', like in this example:

```
-    #: Environment variable used to locate a config file to load last (after
-    #: instance config file). Use this if you want to override some settings
-    #: on a configured instance.
+    # : Environment variable used to locate a config file to load last (after
+    # : instance config file). Use this if you want to override some settings
+    # : on a configured instance.
```

SethMichaelLarson:
I would argue it's not erroneous. There's supposed to be a space after a comment a-la PEP8.

ambv:
Since Sphinx is so popular and we already don't split shebangs and ##, I will fix that.

ambv:
This is now fixed. Thanks! ✨ 🍰 ✨

PR #960, opened 2019-08-03
Fix comment formatting for slurm/pbs cluster directives.

asford:
This change slightly updates black's comment formatting logic for a non-PEP8 edge case frequently encountered in HPC batch schedulers. 

These tools use block comments that immediately follow the `#!` header to provide default configuration parameters for a batch scheduler of the form [`#SBATCH`](https://slurm.schedmd.com/sbatch.html) or [`#PBS`](http://docs.adaptivecomputing.com/torque/4-0-2/Content/topics/commands/qsub.htm) and, unfortunately, don't support a pep8 compatible `# SBATCH` or `# PBS` syntax.

This change special cases comments beginning with these directives, as there isn't a clear way to resolve these requirements within black's existing syntax. As the directive *must* follow the shebang, `# fmt: off` escape hatches aren't a viable solution and (interestingly) don't prevent block comment reformatting in black's current implementation.

Though this is a low-usage edge case, I hope the precedent from https://github.com/psf/black/issues/68 will apply.

### DIFF
--- a/black.py
+++ b/black.py
@@ -2142,8 +2142,9 @@ def list_comments(prefix: str, *, is_endmarker: bool) -> List[ProtoComment]:
 def make_comment(content: str) -> str:
     """Return a consistently formatted comment from the given `content` string.
 
-    All comments (except for "##", "#!", "#:", '#'", "#%%") should have a single
-    space between the hash sign and the content.
+    All comments (except for "##", "#!", "#:", '#'", "#%%" and certain
+    directives) should have a single space between the hash sign and the
+    content.
 
     If `content` didn't start with a hash sign, one is provided.
     """
@@ -2153,7 +2154,7 @@ def make_comment(content: str) -> str:
 
     if content[0] == "#":
         content = content[1:]
-    if content and content[0] not in " !:#'%":
+    if content and not content.startswith(tuple(" !:#'%") + ("SBATCH", "PBS")):
         content = " " + content
     return "#" + content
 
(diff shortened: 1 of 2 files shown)


asford:
Friendly ping. @JelleZijlstra @zsol, would you mind commenting on this PR?

JelleZijlstra:
I don't feel strongly about this. On the one hand it's a pretty simple change, on the other hand it seems inelegant to hardcode support for ever more kinds of obscure pragma syntax.

asford:
I agree that it's both simple and inelegant. My knee-jerk reaction here would be... *configuration option*, but that seems clearly clearly non-black-ish. Hard-coding a special case list here feels weird, though I firmly believe that black should do-the-right-thing somehow in this case. 

Hopefully this falls under "practicality beats purity", but is there anything I can do in the patch to help smooth your concern? More clearly demarcate the special case list? Add documentation?

I've tried to find other pragma with the strict `#pragma`  form that see use in python and haven't found any. I'm certain more might come up, but I don't think there's a massive list of special cases waiting in the wings. Nearly all comment-pragma I've seen allow (or require) a `# pragma` syntax., and tye strict "#PBS" form seems to be an artifact of the tools somewhat dated lineage.

asford:
@JelleZijlstra I now recognize your point on comment pragma. 🙏 While the case in #1014 doesn't exactly fall under this feature I think adding a special case here opens up a significant slippery slope. 

I've open a meta-issue in #1024.

Issue #836, opened 2019-05-07
"Black produced different code on the second pass of the formatter": unwrapping multiple parentheses


dizzy57:
Operating system: MacOS Mojave 10.14.4 (18E226)
Python version: Python 3.7.3
*Black* version: 19.5b0 (c2c2f7203869)
Does also happen on master: yes

Minified reproducer:
```python
return (("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"))
```

Black diff:
```diff
--- source
+++ first pass
@@ -1,2 +1,4 @@
-return (("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"))
+return (
+    "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+)

--- first pass
+++ second pass
@@ -1,4 +1,2 @@
-return (
-    "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
-)
+return "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```

Jma353:
I believe I have a fix for this.  Will put a PR up by tomorrow likely to address this.

Jma353:
Proposed fix in #961

PR #961, opened 2019-08-03
Fix unstable formatting involving unwrapping multiple parentheses (#836)

Jma353:
This is a fix for #836.  To stabilize the formatting, I introduce a function which dedupes adjacent invisible parentheses, so that downstream formatting logic involving breaking up lines depending on line length get the same input regardless of whether `((({atom})))` or `({atom})` is attempting to be formatted.

Open to feedback on this solution and additional test cases I can throw at it; tried to be conservative w.r.t. how these parens are deduped.

As an aside, perhaps in a future PR we can refactor how `maybe_make_parens_invisible_in_atom` looks.  Currently, it's responsible for returning a boolean *and* causing side-effects, which isn't obvious off the bat.  I added a comment to clarify this for now.


black.py
```diff
@@ -2872,7 +2872,15 @@ def normalize_invisible_parens(node: Node, parens_after: Set[str]) -> None:
...
             if child.type == syms.atom:
-                if maybe_make_parens_invisible_in_atom(child, parent=node):
+                # Determines if the underlying atom should be surrounded with
+                # invisible params - also makes parens invisible recursively
+                # within the atom
+                should_surround_with_parens = maybe_make_parens_invisible_in_atom(
+                    child, parent=node
+                )
+                # Removes repeated invisible parens within in the atom
+                dedupe_invisible_atom_parens(child)
```
JelleZijlstra: Is there a way we could do maybe_make_parens_invisible_in_atom() and dedupe_invisible_atom_parens() in one go, so we don't have to recurse over the expression twice?

JelleZijlstra:
Also, thanks for doing this! It's great to get these unstable formatting issues sorted out.

black.py
Jma353: Yea I can try and tweak this a bit.  Thanks for feedback on the approach.

Jma353:
Addressed your comments @JelleZijlstra and added some more tests.  Should be ready for a second round of review.

Jma353:
Ping @JelleZijlstra for review.

zsol:
lgtm, thanks for working on this

PR #964, opened 2019-08-03
Remove unnecessary if-statement in maybe_make_parens_invisible_in_atom

Jma353:
Noticed this when working on #961.  

This if-statement is superfluous b/c right above, there's a check to see if the first and last element of the array of `node.children` have `token.LPAR` and `token.RPAR` `type`'s respectively.  If this control flow to be hit, it's impossible for the array to have only 1 element.


JelleZijlstra:
Thanks!

PR #965, opened 2019-08-04
Add .svn to default exclusion list

Jma353:
As title and per discussion in #794.  Adds this pattern to the default exclusion list so that files in `.svn` aren't included.


Jma353:
Ping for review of this :)

cc: @zsol @JelleZijlstra

zsol:
Whoops this slipped my radar. Time to make that regex multiline IMO

ambv:
Is SVN still a thing?

Issue #691, opened 2019-01-31
Feature request: add diff output to blackd


simonfxr:
Blackd is currently missing a header to only output diffs. It would be great to add this feature, since blackd really makes a difference in interactive environments.

Jma353:
I agree that this would be pretty useful and nice, especially if the file is big.

I can work on adding this support.

JelleZijlstra:
I'm supportive of adding this since black itself already supports a `--diff` option.

Jma353:
I added this support.  

I believe this can be closed @JelleZijlstra

PR #966, opened 2019-08-05
Fix async blackd tests which won't fail currently

Jma353:
While working on #691, I discovered that I couldn't get `blackd` tests to fail in their current form (e.g. basic sanity check asserting `1 == 2` wouldn't fail the test suite). 

I refactor the tests for `blackd` into their own `AioHTTPTestCase` which uses abstractions documented [here](https://docs.aiohttp.org/en/stable/testing.html?highlight=test#testing-aiohttp-web-servers) from `aiohttp`.  Now, tests fail reliably when they are forced to fail or when there's a legit logical error.

I also fix an exception that is thrown when no event loop is present in `event_loop`.  Additionally, I `raise` in `skip_if_exception` when an exception that shouldn't cause the test to be skipped is caught.

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -3,24 +3,14 @@ import asyncio
 import logging
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
-from functools import partial, wraps
+from functools import partial
 from io import BytesIO, TextIOWrapper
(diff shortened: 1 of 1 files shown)


zsol:
Nice, thanks.

PR #969, opened 2019-08-06
Add diff support to blackd

Jma353:
In this PR, I support to `blackd` equivalent to `--diff` for `black`.  This was requested in #691.

I added a test. I also e2e tested it: 

```
➜  black git:(jma353-add-diff-2-blackd) curl --data "print(   'hello, world'    )" localhost:45484 --header "X-Diff:true"
--- In	2019-08-06 01:01:42.622585 +0000
+++ Out	2019-08-06 01:01:42.630187 +0000
@@ -1 +1,2 @@
-print(   'hello, world'    )
+print("hello, world")
+
```


blackd.py
```diff
@@ -105,10 +108,25 @@ def make_app() -> web.Application:
...
         )
+
+        # Only output the diff in the HTTP response
+        only_diff = bool(request.headers.get(DIFF_HEADER, False))
+        if only_diff:
+            now = datetime.utcnow()
+            src_name = f"In\t{then} +0000"
+            dst_name = f"Out\t{now} +0000"
+            loop = asyncio.get_event_loop()
+            formatted_str = await loop.run_in_executor(
```
zsol: I wonder if we can just run `black.diff` synchronously here instead of delegating to a process pool.

For short files it's going to be probably faster, but if it takes too long it will block the event loop.

zsol: I think I convinced myself that this is probably fine for now, we can be more sophisticated later if it really is too slow

zsol:
Can you update the docs pls and write an entry in the change log?

Jma353:
Thanks for the feedback @zsol, will push a revision later today.

ambv:
If you hurry this will get to the next beta.

Jma353:
I can revise this by EOD.  Sorry for letting this sit, got busy with some other side-projects recently :).

ambv:
Never apologize for voluntary contributions you're making for free in your spare time. We are thankful for any level of commitment you can provide.

Jma353:
Delayed by pip install issues with `pip install regex==2019.08.19`.  Fails on my Mac for some reason.

Jma353:
Should be good to go, added changelog + doc for the new header flag.

ambv:
Thanks! ✨ 🍰 ✨

PR #978, opened 2019-08-13
Restore all cursors, instead of only the current window

ahrex:
If we have the same buffer open in multiple windows/tabs, we'll only restore the current window's cursor.

Iterate through all tabs and windows, and save/restore all cursor positions of windows that contain our buffer.

Addendum to #433.

### DIFF
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -118,12 +118,20 @@ def Black():
   except Exception as exc:
     print(exc)
   else:
-    cursor = vim.current.window.cursor
+    current_buffer = vim.current.window.buffer
+    cursors = []
+    for i, tabpage in enumerate(vim.tabpages):
+      if tabpage.valid:
+        for j, window in enumerate(tabpage.windows):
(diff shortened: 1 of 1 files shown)


ahrex:
@JelleZijlstra, @zsol, could you take a look?

Thanks!

zsol:
Sorry I don't know anything about vim plugins :( I'm definitely not the right person to review this

ahrex:
@ambv, could you see if this change is good?

ambv:
Sorry, I merged another PR which fixes the cursor to be restored to line of code instead of line in file. This conflicts now. Could you rebase your fix?

ahrex:
I took a look at #989, and I don't think this change is compatible with that PR.

Making an additional pass at the file contents gets expensive if you're to scale for `n` buffers.

Feel free to decline this PR, or back the other out and place this one in instead.

ambv:
Thanks! ✨ 🍰 ✨

PR #981, opened 2019-08-13
use versioneer to manage __version__

### LINKED ISSUE #980: 'Symbols' object has no attribute 'namedexpr_test'
Error: ` 'Symbols' object has no attribute 'namedexpr_test'`

Operating system: OSX
Python version: 3.7
*Black* version: master
Does also happen on master:

To answer the last question, you have two options:
1. Use the online formatter at https://black.now.sh/?version=master, which will use the latest master branch.
2. Or run black on your machine:
    * create a new virtualenv (make sure it's the same Python version);
    * clone this repository;
    * run `pip install -e .`;
    * make sure it's sane by running `python setup.py test`; and
    * run `black` like you did last time.

I've run this using pre-commit hooks which gives the following error:

```
hookid: black

error: cannot format /Users/sanyamkhurana/work/mymp/mymp/maps/serializers.py: 'Symbols' object has no attribute 'namedexpr_test'
error: cannot format /Users/sanyamkhurana/work/mymp/mymp/maps/models.py: 'Symbols' object has no attribute 'namedexpr_test'
error: cannot format /Users/sanyamkhurana/work/mymp/mymp/maps/services.py: 'Symbols' object has no attribute 'namedexpr_test'
Oh no! 💥 💔 💥
1 file left unchanged, 3 files failed to reformat.
```

Note that if I install black using pip and try to reformat, these files are reformatted properly without any errors. The version I got from pip install is `19.3b0` which is working fine, but the master is breaking.

Here is my `pre-commit-config.yaml`

```
repos:
- repo: https://github.com/ambv/black
  rev: master
  hooks:
    - id: black
      language_version: python3.7
```

I've trimmed it for brevity. Please note that this was working fine 5 days earlier.

Please let me know if you need any more details.

zsol:
* For releases made from a tagged commit, `__version__` doesn't change.
* For releases made from all other commits, `__version__` changes to `${TAG}+${COMMITS_SINCE}.g${SHA}${DIRTY}` (`TAG` = most recent tag, `COMMITS_SINCE` is the number of commits since `$TAG`, `SHA` is the output of `git rev-parse HEAD`, `DIRTY` is `.dirty` if the working copy has uncommited changes), for example: `19.3b0+77.gf1a148e.dirty`
* This PR also changes the `version` string used for Black's cache to be `$SHA`.

Fixes #980


vezeli:
Alternative to the `versioneer` for dealing with version tracking would be to consider using [setuptools-scm](https://github.com/pypa/setuptools_scm). `setuptools-scm` is a part of the `setuptools` it does everything in the background so that there is no need for polluting the package with `_version.py` and `versioneer.py`.

Maybe I am missing something but was there (or is there still) a strong reason against it `setuptools-scm`?

zsol:
I'll be honest I haven't found `setuptools-scm`, so didn't really evaluate it

vezeli:
Oh I see :)

The final result would be the same, i.e. `black.__version__` will change on every commit in line with the SHA-1 (as needed for #980). The main advantage, however, is that `setuptools-scm` comes together with `setuptools` so there is no need to put a big script such as versioneer.py in the package directory next to the main script. Although one could still keep the _version.py since `setuptools-scm` creates it in the process of package setup (otherwise one could also put it in .gitignore for the purpose of development).

E.g. borg project switched from `versioneer` to `setuptools-scm` borgbackup/borg/pull/161 due to the borgbackup/borg/issues/106.

hugovk:
`setuptools_scm` is great for autoversioning, but comes with an extra big bit of functionality to watch out for. 

It automatically adds all of the SCM-managed files to the sdist. 
Unwanted files must be excluded by discarding them via `MANIFEST.in`. 

My advice is to create an sdist and diff it with an old one, then add excludes to MANIFEST.IN until they match. 

* https://github.com/pypa/setuptools_scm/pull/343
* https://twitter.com/ZaneSelvans/status/1143976645981876224

zsol:
I'm happy to review a PR that switches us over. The main reason versioneer was introduced in a rush is to unbreak people using a non-release version of Black.

vezeli:
Sounds good, I can give it a try :)

Issue #957, opened 2019-08-02
Documentation inconsistent use of  _ and - in cmd line params / pyproject.toml / Vim


skewty:
### Command line arguments
* --line-length
* --target-version
* --skip-string-normalization

### pyproject.toml
* line-length
* target_version
* skip-string-normalization

### Vim
* g:black_linelength
* g:black_skip_string_normalization

Also a `skip-string-normalization = 'true'` value should be included in the example 
`pyproject.toml`.  One might think it should be an `int` because the **Vim** config uses an `int` for this.  Anyhow..there is an opportunity to be more clear about how this should be specified.

hauntsaninja:
It's target-version with hyphen even in pyproject.toml. It's true that pyproject keys are hyphen and vim keys are underscore, but that seems acceptable to me

PR #984, opened 2019-08-14
example for skip-string-normalization in pyproject

artemisart:
See #957


ambv:
Nah, we don't want to endorse that.

artemisart:
Maybe change it to `skip-string-normalization = 'false'` then ?  I made this PR because I spent more time than I should have trying to set this option, and I think it's pretty common to use " for UI strings and ' for internal strings in python projects.

PR #989, opened 2019-08-17
Restore cursor to same line of code, not same line of buffer

chmreid:
This updates the way that the new cursor location in the buffer is determined. It is a two-step process. First, all code ahead of the cursor is run through the Black formatter, and the new length of the code ahead of the cursor is determined. This sets the new location of the cursor. Second, the entire buffer is run through Black, as it was previously.

The end result is that the user can call the Black formatter and have the cursor remain on the same line of code as it was before and after the Black formatting process.

To make sure the code ahead of the cursor will pass through the Black formatter okay, we check if the last line ends with `:` and if so we tack `pass` onto the end.

This pull request is not very robust to corner cases (yet). For example, if the cursor is located in the middle of a long, multi-line list of arguments.

### DIFF
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -110,7 +110,29 @@ def Black():
     string_normalization=not bool(int(vim.eval("g:black_skip_string_normalization"))),
     is_pyi=vim.current.buffer.name.endswith('.pyi'),
   )
-  buffer_str = '\n'.join(vim.current.buffer) + '\n'
+  (cursor_line, cursor_column) = vim.current.window.cursor
+  cb = vim.current.buffer[:]
+  cb_bc = cb[0:cursor_line]
+  # Format all code before the cursor.
+  # Detect unclosed blocks, close them with pass.
+  last_line = cb_bc[-1]
+  if last_line.rstrip().endswith(":"):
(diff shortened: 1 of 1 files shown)


ambv:
Thanks! ✨ 🍰 ✨

ahrex:
Hi @chmreid, I think your PR introduced a few issues:

  * sometimes I receive the message `EOF in multi-line statement` while formatting in the middle of a dictionary
  * other times I receive a lot of errors in the middle of a dictionary as well: https://asciinema.org/a/Rkbr8J2CwC0uyOyAWh99vqIfi

I've given it some thought, and I don't know the best way of restoring to the current line without multiple passes.

I also know that trying to reformat half of a python file could result in errors, as demonstrated above. Maybe we can back this out and give it a second try?

ambv:
Right, I was too hasty in merging this. Let me back this out so @ahrex can land his thing first. Then we can work out the edge cases.

chmreid:
> This pull request is not very robust to corner cases (yet). For example, if the cursor is located in the middle of a long, multi-line list of arguments.

Making this bulletproof/robust to corner cases would, I think, become a quite complicated exercise in parsing Python syntax and is not worth it

PR #999, opened 2019-08-28
#886: Misleading message when black checks only empty directories

Rishi-jha:
This is the output message update for handling empty directories, Also added pycharm .idea folder in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ black.egg-info
 build/
 dist/
 pip-wheel-metadata/
+.idea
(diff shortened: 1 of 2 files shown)


Rishi-jha:
#886

zsol:
I think to reduce confusion we could distinguish between cases when there were genuinely no input files specified and when there were some inputs on the command line but they contain no files to be formatted (due to include/exclude rules)

Rishi-jha:
@zsol , updated, to check if there is no path given or no python file present.

zsol:
Thanks!

PR #1002, opened 2019-08-29
Change variable in README according to the PEP8

vezeli:
_Black_ is constantly aiming to follow PEP 8 and this is nicely put in the README. However, [PEP 8](https://www.python.org/dev/peps/pep-0008/#names-to-avoid) also suggests

> Never use the characters 'l' (lowercase letter el), 'O' (uppercase letter oh), or 'I' (uppercase letter eye) as single character variable names.

For this reason I have made a micro alteration and changed the variable letter `l` in an example to another variable in the spirit of PEP 8.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,14 +178,14 @@ great.
 ```py3
 # in:
 
-l = [1,
+j = [1,
      2,
      3,
 ]
 
 # out:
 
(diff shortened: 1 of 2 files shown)


zsol:
Thanks!

PR #1003, opened 2019-09-03
Make quote style configurable

mark-riley:
This allows the quote style to be configurable.

Why is this necessary?

1. typing double quotes is more work than typing single quotes. I can type a single quote with a single key press. Typing double quotes requires holding shift and pressing another key. That seems like an awful lot of effort that isn't required.

2. Trying to incorporate Black into an existing project is a PITA if you're already established using single quotes. It creates huge diffs that react badly with concurrent changes, resulting in merge conflicts that just aren't necessary.

Default behaviour is, of course, double quotes.

### DIFF
--- a/black.py
+++ b/black.py
@@ -192,6 +192,7 @@ VERSION_TO_FEATURES: Dict[TargetVersion, Set[Feature]] = {
 class FileMode:
     target_versions: Set[TargetVersion] = Factory(set)
     line_length: int = DEFAULT_LINE_LENGTH
+    single_quotes: bool = False
     string_normalization: bool = True
     is_pyi: bool = False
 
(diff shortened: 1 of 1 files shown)


zsol:
Thanks for putting in the work, but this has been discussed at length in #118 and we won't be adding an option to configure the quote style.

hugovk:
> 2\. Trying to incorporate Black into an existing project is a PITA if you're already established using single quotes. It creates huge diffs that react badly with concurrent changes, resulting in merge conflicts that just aren't necessary.

This is the Black option for you:
```
  -S, --skip-string-normalization
                                  Don't normalize string quotes or prefixes.
```

AXGKl:
Thanks for the work @mark-riley , we'll use your branch in the company.

Can't understand why line length - which hits brutal at merges - *is* configurable, but this not.

PR #1008, opened 2019-09-05
Switch from versioneer to setuptools-scm

vezeli:
The idea of version tracking with tools like `versioneer` and [`setuptools-scm`](https://github.com/pypa/setuptools_scm) is to help developers keep track of the code version with respect to SHA-1 and the distance from the last tag. Clean tags are short and useful for users and dirty tags are useful for developers. For detailed examples see [pypa/setuptools_scm](https://github.com/pypa/setuptools_scm) project and the issue [here](https://github.com/pypa/setuptools_scm/issues/349). Starting from #981, _Black_ implemented `versioneer` for version tracking.

This PR further improves the process of version tracking by proposing a switch to `setuptools-scm` package. `setuptools-scm` is a [Python Package Authority (PyPA)](https://www.pypa.io/en/latest/) project, i.e. the organization that maintains many of the relevant projects in Python packaging such as `pip`, `setuptools`, `pipenv` (all of which _Black_ is using) + others.

Some of the benefits of switching to `setuptools-scm` are:
* Unlike `versioneer`, it does not pollute the git project directory with _version.py, versioneer.py and .gitattributes files helping the project clean. Note that a new file _version.py (different from the previous _version.py used for `versioneer`)  is going to be present and contains a single line with the version number once `python setup.py --version` runs. However, it is not going to be a part of the git directory (included in .gitignore).
* As mentioned, the package is maintained by the PyPA, this means that if some changes were to happen in the future with the way packaging works there is a good chance that the `setuptools-scm` will stay compatible with these changes.
* Opens a possibility of extending the `setuptools-scm` to automatically generate MANIFEST.in (which is one of the possibilities since `setuptools-scm` communicates with .git).


hugovk:
Following on from https://github.com/psf/black/pull/981#issuecomment-527738838:

I did `python3 setup.py sdist bdist_wheel` and compared `dist/black-19.3b1.dev84+g0a19a5d.tar.gz` with `black-19.3b0.tar.gz` from https://pypi.org/project/black/#files.

```console
$ ls -l dist/black-19.3b1.dev84+g0a19a5d.tar.gz
-rw-r--r--  1 hugo  staff   1.0M  6 Sep 08:48 dist/black-19.3b1.dev84+g0a19a5d.tar.gz
$ ls -l /tmp/Downloads/black-19.3b0.tar.gz
-rw-r--r--@ 1 hugo  wheel   152K  6 Sep 08:48 /tmp/Downloads/black-19.3b0.tar.gz
```

image

Should some excludes be added to `MANIFEST.in`?

ashwinvis:
Good point, @hugovk! The ideal solution would be for `setuptools_scm` to not hijack the `file_finders` operations, unless explicitly asked, but since that is beyond us we have to change the MANIFEST file like you suggested. I found that you already went through this, so I will leave the links as clues to fix this:

* hugovk/osmviz#30
* hugovk/pypistats#63

Also slides 11-13 in https://blog.ionelmc.ro/presentations/packaging/#slide:11 might be useful.

vezeli:
I think it is good that you pointed it out @hugovk as this is an important difference.

One should bear in mind that this behavior is not a bug in `setuptools-scm` it is an intentional design feature and is discussed in pypa/setuptools_scm#190 and [Python setuptools' MANIFEST.in explained](https://www.remarkablyrestrained.com/python-setuptools-manifest-in/). **tl;dr** In the issue, Jason Coombs explains

> In my opinion the sdist is meant to be more than just a copy of the Python functionality, but is meant to be a distributable copy of the source code. I would expect someone to be able to download the sdist, extract it, and develop on the project much like they would if cloning the repo. These sdists are used by downstream packagers (Debian, RedHat) to run tests and other validation on them.

This frees developers from dealing with MANIFEST.in potentially making errors. In our particular case, the listing of the sdist (that is shown) does not include **docs** which should probably be present there and **.github** contains CODE_OF_CONDUCT.md which I think is nice to have it.

Thanks for the comments @hugovk and @ashwinvis. I prefer to leave it compatible with PyPA but let's see what the maintainers think.

JelleZijlstra:
This is definitely a lot more elegant. @zsol what do you think?

Also, there's a minor merge conflict.

zsol:
I skimmed through it during the weekend and looked good to me. Haven't had time to review in depth yet. I'm happy with the approach though :+1:

zsol:
Thanks!

PR #1009, opened 2019-09-07
Support compilation with mypyc

msullivan:
Initial take on supporting compilation of black with mypyc. Most of the work here was done by @SanjitKal.

In my testing it sped things up by about a factor of 2 (with `--fast`).

Most of my testing was done by running `CC=clang python3 setup.py --use-mypyc build_ext --inplace` and then running black. clang needs to be used as the C compiler currently because gcc (at least the version I tried) barfed on some unicode identifiers (https://github.com/mypyc/mypyc/issues/699).

The bulk of the diff is adding types to blib2to3 so it can be compiled, which is critical to the performance gains here: both because it speeds up access to the data structures from black itself and because the parser is a major hotspot.


black.py
```diff
@@ -188,12 +188,23 @@ class Feature(Enum):
 }
 
 
-@dataclass
 class FileMode:
```
JelleZijlstra: Is dataclass support planned for mypyc?

msullivan: There's a bug open for it (https://github.com/mypyc/mypyc/issues/671) but I'm not sure when I'll have a chance to do it

msullivan: I think I'm going to have dataclass support without optimized `__init__` in the next few days, so we'll be able to keep using dataclasses though I'll probably keep the `__init__` implementation on `Line` and anything else that gets created a lot.

ambv:
Sorry, I haven't noticed that merging blackening of blib2to3 will conflict this PR. If you could blacken so we can merge, we will.

msullivan:
I've rebased and updated this. I dropped the commit that removed the attrs in favor of putting up a separate PR to switch attrs to dataclasses (https://github.com/psf/black/pull/1116) in concert with a mypyc PR to add efficient dataclass support (https://github.com/python/mypy/pull/7817)

blib2to3/pgen2/driver.py
```diff
@@ -25,17 +25,39 @@
 
 # Pgen imports
 from . import grammar, parse, token, tokenize, pgen
+from typing import (
```
JelleZijlstra: These should really be with the "Python imports" above

blib2to3/pgen2/driver.py
```diff
@@ -25,17 +25,39 @@
...
+    Text,
+    Tuple,
+    Union,
+    Sequence,
+)
+from logging import Logger
+from blib2to3.pytree import _Convert, NL
+from blib2to3.pgen2.grammar import Grammar
+
+Path = Union[Text, "os.PathLike"]
```
JelleZijlstra: Missing a type arg to os.PathLike. We should probably turn on the `--disallow-any-generics` option.

blib2to3/pgen2/driver.py
```diff
@@ -25,17 +25,39 @@
...
 # Pgen imports
 from . import grammar, parse, token, tokenize, pgen
+from typing import (
+    Any,
+    Callable,
+    IO,
+    Iterable,
+    List,
+    Optional,
+    Text,
```
JelleZijlstra: You can just use str since this is Python 3-only code.

blib2to3/pytree.py
```diff
@@ -66,9 +93,13 @@ def __eq__(self, other):
             return NotImplemented
         return self._eq(other)
 
-    __hash__ = None  # For Py3 compatibility.
+    # __hash__ = None # For Py3 compatibility.
```
JelleZijlstra: Does mypyc choke on `__hash__ = None`? I'd be fine with just removing the line.

blib2to3/pytree.py
```diff
@@ -592,17 +666,17 @@ def __init__(self, type=None, content=None, name=None):
         if type is not None:
             assert type >= 256, type
         if content is not None:
-            assert not isinstance(content, str), repr(content)
-            content = list(content)
-            for i, item in enumerate(content):
+            # assert not isinstance(content, str), repr(content)
```
JelleZijlstra: What's wrong with this assertion?

blib2to3/pgen2/driver.py
msullivan: I'll leave this to a follow-up, since the lib2to3 stubs used Text pretty heavily

blib2to3/pytree.py
msullivan: It caused mypy to barf because it then couldn't figure out what type `content` (and thus `newcontent`) was. The actual issue is that the type of `content` was wrong, though. I'll fix that, though all this code is dead in black.

blib2to3/pytree.py
msullivan: mypy complains about it with a type error. I can fix it with a `# type: Any` though.

blib2to3/pgen2/driver.py
msullivan: Fixed that and set the flag

PR #1011, opened 2019-09-09
Blacken .py files in blib2to3

msullivan:
This is in preparation for adding type annotations to blib2to3 in
order to compiling it with mypyc (#1009, which I can rebase on top of
this).

To enforce that it stays blackened, I just cargo-culted the existing
test code used for validating formatting. It feels pretty clunky now,
though, so I can abstract the common logic out into a helper if that
seems better. (But error messages might be less clear then?)

### DIFF
--- a/blib2to3/__init__.py
+++ b/blib2to3/__init__.py
@@ -1,1 +1,1 @@
-#empty
+# empty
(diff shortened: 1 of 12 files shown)


JelleZijlstra:
Sorry for the radio silence on this! I'm about to go on a trip for a week, but after that I can give more feedback and hopefully merge this PR in. I'm excited to get Black faster using mypyc.

It would be good to make the tests less repetitive, but that can be a follow-up PR.

msullivan:
I've cleaned up the tests. CI is failing because of an issue found checking under mypy 0.730, which I put up https://github.com/psf/black/pull/1039 to fix.

PR #1015, opened 2019-09-12
Replace unicode empty set identifier with ascii one

mackorone:
Facebook's Python typechecker, [Pyre](https://pyre-check.org/), can't parse unicode source code:
```
 ƛ Could not parse file at black.py:1645:8-1645:8
          Ø: Set[str] = set()
         ^
 ƛ Could not parse 1 file due to syntax errors!
```
According to folks familiar with Pyre's internals, supporting unicode source is nontrivial. For now, let's just rename the variable from `Ø` to `empty_set`.

The irony that our Python typechecker can't handle unicode is not lost on us.

### DIFF
--- a/black.py
+++ b/black.py
@@ -1784,7 +1784,7 @@ class LineGenerator(Visitor[Line]):
     def __attrs_post_init__(self) -> None:
         """You are in a twisty little maze of passages."""
         v = self.visit_stmt
-        Ø: Set[str] = set()
+        empty_set: Set[str] = set()
         self.visit_assert_stmt = partial(v, keywords={"assert"}, parens={"assert", ","})
         self.visit_if_stmt = partial(
             v, keywords={"if", "else", "elif"}, parens={"if", "elif"}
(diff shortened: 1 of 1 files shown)


mackorone:
@JelleZijlstra, @zsol, @ambv, silly question: is there anything I should be doing besides waiting for this to get merged in? Just want to make sure I'm not blocking.

ambv:
Originally the reason I put the cute empty-set identifier there was that Black tests itself by reformatting itself, too. Heavy dogfooding. You may also know that [Unicode issues are very dear to my heart](https://www.youtube.com/watch?v=7m5JA3XaZ4k).

At the risk of sounding petty, accepting this change doesn't make Black better and it doesn't make Pyre better. Have you considered the alternative of using an internal patch for the time it takes you to make Pyre accept valid Python code using Unicode identifiers?

mackorone:
Not petty at all, I understand the value in dogfooding and that there's deeper meaning to this change. I'm not entirely familiar with our patching process - I'm just a developer trying to use Pyre - but I'll follow up with some other folks and get back to you.

mackorone:
Quick update: the short term plan is just to generate a type stub file for `black.py` and use that for type information, rather than trying to parse the source code. (I'm crossing my fingers that unicode doesn't appear at the module level in the near future.) The long term plan is to make Pyre unicode compatible.

mackorone:
Closing this for now.

PR #1018, opened 2019-09-12
enhancement: Add software version to blackd response

jacobbogdanov:
There is a new header in the response that blackd provides,
`X-Black-Version`, which is the same version that is printed when
running `blackd --version` or when blackd is first launched.


zsol:
Looks good, but you'll need to make sure `blackd.py` is blackened. Also might be worth adding a `/version` endpoint that doesn't do anything apart from returning a HTTP 200 with the version header.

ambv:
This is fixed by #1046. Thank you for your contribution!

PR #1027, opened 2019-09-17
Don't allow type comments to be merged behind regular comments

msullivan:
Type comments only apply if they are the first comment on the line,
which means that allowing them to be pushed behind a regular comment
when joining lines is a semantic change (and, indeed, one that black
catches and fails on).

(This issue discovered in the wild at Dropbox by a user who accidentally omitted a colon in a type comment and was confused to find black crashing.)

### DIFF
--- a/black.py
+++ b/black.py
@@ -1296,7 +1296,7 @@ class Line:
                     return True
         return False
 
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Thanks!

PR #1028, opened 2019-09-18
Support PEP 572 in while statements

pablogsal:
Commit d8fa8df0526de9c0968e0a3568008f58eae45364 added support for
parsing and formatting PEP572, but it missed the posibility to add
PEP572 syntax in while statements.


pablogsal:
CC: @zsol

zsol:
Oops. Thanks! :)

PR #1045, opened 2019-10-04
Moved the metadata to setup.cfg

KOLANICH:


ambv:
We won't be doing this because soon enough we want to get rid of setup.cfg entirely. What's your use case for this?

KOLANICH:
I completely don't understand the article you tefer. It says 

>You don't need a `setup.cfg` anymore!  At least not for packaging. `wheel` nowadays collects license files automatically, and we tell `setuptools` to build universal wheels (e.g. attrs-19.3.0-**py2.py3**-none-any.whl instead of one wheel per Python version) in the `setup.py` using the `options` keyword argument.

It looks like the authors think that `setup.cfg` is only for some stuff like license file and indicating that a wheel is an universal one. It is not the case. In this PR all the metadata is moved there.

They also mention some `pep517` package. It is not very clear what does it do.

ambv:
1. You removed the dependency on setuptools-scm which breaks the build.
2. We are planning to depend on mypyc to speed up Black which requires custom setup.py machinery that we cannot put anywhere else.
3. As soon as setuptools supports pyproject.toml for metadata, we will move it there. But moving back to setup.cfg which we don't even intend to support for configuration feels backward.

Thank you for your contribution but we won't be accepting this change.

KOLANICH:
>As soon as setuptools supports pyproject.toml for metadata, we will move it there.

I guess it will be not soon, if ever. I haven't got any messages about someone being working on it, even though I have left an issue about it quite long ago.


>We are planning to depend on mypyc to speed up Black which requires custom setup.py machinery that we cannot put anywhere else.

It doesn't mean the metadata can't and shouldn't be moved there.

>You removed the dependency on setuptools-scm which breaks the build.

Damn, I really have removed. But since you are not going to merge it anyway, there is no point to fix it.

PR #1046, opened 2019-10-06
Add black version header to blackd responses

### LINKED ISSUE #1013: blackd has no API to get server version
I am working on a client to `blackd` and would like to be able to get the version of the `blackd` server from the http endpoint.

This is for the same use-case that `black` itself has; disk caching of already formatted files. It would be advantageous to be able to check the cache and avoid the network IO.
Whether there is a new API endpoint, `/version`, or the response to each POST contains a header containing the server version should be equivalent for this particular use-case

Operating system: macOS
Python version: `python3.6`
*Black* version: `blackd, version 19.3b0`
Does also happen on master: as far as I can tell from the source code

linusg:
This PR will add a `X-Black-Version` header to all blackd responses.

For clarity the internal constant `VERSION_HEADER` (`"X-Protocol-Version"`) has been renamed to `PROTOCOL_VERSION_HEADER`.

Closes #1013.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1054,6 +1054,8 @@ More details can be found in [CONTRIBUTING](CONTRIBUTING.md).
 * *Black* is now able to format Python code that uses positional-only
   arguments (`/` as described in PEP-570) (#946)
 
+* `blackd` now returns the version of *Black* in the response headers (#1013)
+
 
 ### 19.3b0
 
--- a/README.md
+++ b/README.md
@@ -297,7 +297,7 @@ ignore = E501,W503,E203
 ```
 
 You'll find *Black*'s own .flake8 config file is configured like this.
-If you're curious about the reasoning behind B950, 
+If you're curious about the reasoning behind B950,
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -297,7 +297,7 @@ ignore = E501,W503,E203
 ```
 
 You'll find *Black*'s own .flake8 config file is configured like this.
-If you're curious about the reasoning behind B950, 
+If you're curious about the reasoning behind B950,
```
linusg: Sorry, VSCode automatically removed the trailing whitespace here and below. If you want a minimal diff, I'll add a commit reverting this.

JelleZijlstra: It's fine to keep.

JelleZijlstra:
Actually you need to fix the test failures (looks like the constants you renamed is used in a test).

linusg:
> Actually you need to fix the test failures (looks like the constants you renamed is used in a test).

Right, sorry about that! I added a simple test for the new response header as well.

zsol:
Thank you!

Issue #455, opened 2018-08-16
Parse failure on some Unicode identifiers


JelleZijlstra:
Black gives a syntax error on this file:
```python
import unittest

class PEP3131Test(unittest.TestCase):

    def test_valid(self):
        class T:
            ä = 1
            µ = 2 # this is a compatibility character
            蟒 = 3
            x󠄀 = 4
```
Error:
```
error: cannot format unident.py: Cannot parse: 10:13:             x󠄀 = 4
```

This is part of `Lib/test/test_unicode_identifiers.py` from the CPython test suite.

Operating system: OS X
Python version: 3.7.0
*Black* version: master

ambv:
Yeah, this is a bug to be fixed although fortunately low priority.

This only happens for atypical classes of Unicode identifiers. Usual Unicode identifiers work fine and real-world code is unlikely to encounter this (unless malicious code). What's unsupported in lib2to3 is a certain class of legacy identifiers and variant forms. The latter is what we're dealing with in this example.

I happen to have the UTF-8 decoding table on my desk, as you do, so I checked and this example is pretty insidious. It's a base `x` letter followed by a ["variation selector"](https://en.wikipedia.org/wiki/Variant_form_(Unicode)). Variants are for people to be able to choose which *glyph* should be used in rendering the character. It's ambiguous if a glyph variant should be considered a unique identifier since it's rendered with the base glyph if a given font lacks the variant (as is the case here, see example below).
```
# which one is variant-17?
x
x󠄀
```

andersk:
Another example with no variation selectors, reduced from https://github.com/uber/pyflame/blob/v1.6.7/tests/sleeper_ユニコード.py:
```python
មុ = 1
```
That’s U+1798 KHMER LETTER MO + U+17BB KHMER VOWEL SIGN U.

prusse-martin:
Part of our team develops physics simulation and they are used to have Unicode characters in the identifiers to better match the literature formulas (in python and c++ code).
We started to use Black in our projects and one problem we faced is having `Q̇_per_meter` ("Cannot parse" error). This is valid python code and have no malicious intent.

dyus:
I can give a try to replace re module in favor [regex package ](https://pypi.org/project/regex/) which has support for latest unicode standard (unicode 12.1.0)

JelleZijlstra:
@dyus that sounds promising, thanks!

dyus:
[...]

PR #1047, opened 2019-10-07
#455 Fix bug with tricky unicode symbols

dyus:
add test for the bug
add regex lib which supports unicode 12.1.0 standard
replace re usage in project in favor to regex


hugovk:
Also add `regex` to `install_requires`?

https://github.com/psf/black/blob/d9e71a75ccfefa3d9156a64c03313a0d4ad981e5/setup.py#L36-L42

setup.py
```diff
@@ -39,6 +39,7 @@ def get_long_description() -> str:
         "appdirs",
         "toml>=0.9.4",
         "typed-ast>=1.3.1",
+        "regex==2019.8.19",
```
hugovk: I think the idea is that end-user, non-dev dependencies should not be pinned and should be as loose as possible. Does it work like this?

```suggestion
        "regex",
```

dyus: yes it should work, i'll change

dyus:
#455

zsol:
Thank you!

jezdez:
The major downside of this new dependency is that it requires a C compiler during installation. Is this really the only way to fix this?

Issue #144, opened 2018-04-18
Black should have an opinion about doc strings


Lukas0907:
Operating system: Ubuntu 16.04
Python version: 3.6.1
Black version: master
Does also happen on master: yes

Hi,

currently Black doesn't seem to have an opinion about doc strings or rather where the quotes should go.

Black claims for this file for example that it is already formatted:

```python
def test():
    """This is one possibility.

    Very important stuff here.
    """


def test2():
    """This is another one possibility.

    Very important stuff here."""


def test3():
    """
    This is another one possibility.

    Very important stuff here.
    """


def test4():
    """
    What about this?
    """


def test5():
    """What about this?

    Some people also like to have an empty line at the end of the doc string.

    """
```

The tool pydocstyle (in the spirit of PEP-0257) at least complains that the closing quotes should go on a separate line and if the docstring fits one line it should only span one line.

It would be nice if that could be incorporated in Black as well.

Thanks for the great work!

Lukas.

ambv:
So I came here to complain that docstrings are out of scope but reading your comment actually makes me think what you're asking for is reasonable :smile:

Specifically:
* if it fits one line, it should;
* if it doesn't, the closing quotes should be on their own line.

I'll think about possible negative effects some more before deciding whether to do this for sure but it looks like some opinions here might be harmless.

@carljm, what do yo think?

carljm:
I think it's reasonable for Black to be opinionated about this.

I'm not sure why, but I feel like consistent docstring formatting actually has a huge impact on whether code feels consistently formatted. I guess because they tend to be visibly syntax-highlighted large blocks of text.

natecox:
I tend to favor single line where applicable:
```python
def somefunc():
    """This is my docstring"""
``

Multi-line I'd personally prefer the first line be on a new line (i.e., `test3` above) but I wouldn't complain about the `test` format above either.

skapil:
@ambv Not sure, if anyone is looking into it. Can I give a try to this?

willingc:
Agree with the approach @ambv and @carljm put forth. Like Carl, I like the longer docstrings to have """ on their own lines. I often put a blank line before the closing """ as it adds a bit of whitespace but it's totally fine not to have the blank line.

coxley:
+1 to prefer single line if possible, otherwise both opening and closing quotes on their own lines. :)

anowlcalledjosh:
[...]

PR #1053, opened 2019-10-10
Re-indent the contents of docstrings

alexmv:
This cleans up https://github.com/bancek/black/commit/0889797612d14d9dc5954df6a7de73076c7cb353 by @bancek based on the feedback there.


zsol:
Can you help me understand why this change is needed? I don't understand it either from the description of this PR, or the referenced commit, or the test cases (which by the way pass even on master). I'd rather not reverse-engineer the new code.

bancek:
This change is needed to preserve the docstring indent when formatting codebases with 2 spaces indentation to 4 spaces.

Source file:

```py
class MyClass:
  """Multiline
  class docstring
  """

  def method(self):
    """Multiline
    method docstring
    """
    pass
```

After running black:

```py
class MyClass:
    """Multiline
  class docstring
  """

    def method(self):
        """Multiline
    method docstring
    """
        pass
```

Expected:

```py
class MyClass:
    """Multiline
    class docstring
    """

    def method(self):
        """Multiline
        method docstring
        """
        pass
```

zsol:
Nice! @alexmv can we include that example in the test cases, please? And maybe another one that tests the conversion from overindented (more than 4 spaces) docstrings.

bancek:
@alexmv can you split that into two commits so that my original commits stays as is (for contribution)?

zsol:
I'm going to squash it anyway when merging, though

alexmv:
Whoops -- sorry for not wording the commit more carefully to make its need clear.  Tests all fail for me on `master`, now, and include the test case above and an additional deeper indent.

I've added an explicit `Co-authored-by:` trailer which should preserve your attribution more explicitly, @bancek.

alexmv:
Any other comments on this?

ambv:
Can you point me at an issue where this was agreed on?

alexmv:
I wasn't aware that a separate issue needed to be created for what seemed like a reasonable bugfix with code provided.

There seems to be reasonable agreement here on the PR.  What sort of agreement are you looking for in a separate issue?

jboning:
In #144 (where @bancek originally shared his patch) there seems to be some consensus that adjusting docstrings is in scope for `black`, though the discussion there focuses more on leading/trailing wihtespace than on indentation.

jboning:
It might be good to have a testcase that includes lines in the docstring with extra leading whitespace, e.g.:
```
def foo():
    """
    Lorem ipsum dolor sit amet.

    Consectetur adipiscing elit:
     - sed do eiusmod tempor incididunt ut labore
     - dolore magna aliqua
       - enim ad minim veniam
       - quis nostrud exercitation ullamco laboris nisi
     - aliquip ex ea commodo consequat
    """
    pass
```

alexmv:
Thanks for that suggestion, @jboning -- I've added that test, along with one for a too-deep closing `"""`, which this also resolves.

Rebased atop latest `master`.

ambv:
Let's merge this. Don't forget to add it to changelog.

zsol:
Thanks for contributing @alexmv and @bancek !

bancek:
@alexmv thanks for the PR!
@zsol thanks for accepting this into black.

hugovk:
Please see https://github.com/psf/black/issues/1452.

Issue #1073, opened 2019-10-16
Prefer to use local command with pip dependency?


benjaoming:
## How we install black for pre-commit hooks

CC: @asottile

As per current README, a pre-commit hook is shown that will fetch the full `psf/black` git repository and invoke setup.py in order to install the `black` executable for the pre-commit hook.

I would like to recommend that installing the .whl from PyPi is faster and requires less bandwidth.

Moreover, using a locally available `black` command saves the developer double-trouble, i.e. installing black in both the `~/.cache/pre-commit` and in some other dev environments. This might not look good as the default option (mixing pre-commit environment with development/deployment environments), but it's a very nice option for people who prefer having few environments.

## Summary

Using PyPi to install, takes 6 CPU seconds, whereas the repo-based install takes 10 CPU seconds.

According to `nethogs`, using the repo fetches about 3.7MB of data, invoking `/usr/lib/git-core/git-remote-https`. Using `pip` and a completely clean `~/.cache/pip`, 309 KB is fetched (black and dependencies). With pip cache already populated, just 12 KB is fetched from the network.

Some advantages of using PyPi releases with pip:

 * Speedier bootstrapping of linting and development environments
 * Pre-commit hooks are often run in CIs for extra checks (because some developers don't install the hooks!) - this would mean faster CI
 * Invoking `pip` also uses pip's system cache, so installing the same version of black several times won't use network
 * The developer can easily and transparently customize the black command args directly by modifying `entry: black --check`.
 * Less internet traffic means reduced CO2 emissions etc.
 * Specifying black version number for consistent linting environments

## Option 1: Current repo-based configuration

```
repos:
-   repo: https://github.com/psf/black
    rev: stable
    hooks:
    - id: black
      language_version: python3.6
```

## Option 2: A local command with a pip dependency

```
-   repo: local
    hooks:
    -   id: black
        name: black
        entry: black --check
        language: python
        types: [python]
        additional_dependencies: ['black']
        language_version: python3.6
```

## Option 3: Just use black from the current environment

This requires runnning `pip install black` before executing pre-commit hooks.

```
-   repo: local
    hooks:
    -   id: black
        name: black
        entry: black --check
        language: system
        types: [python]
```

### Option 1:

```
time gc         
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
black....................................................................Failed
hookid: black

[...]

PR #874, opened 2019-05-28
Automatic markdown and YAML formatting with Prettier

yhoiseth:
Replaces #870.

PR #1081, opened 2019-10-20
Run pre-commit on Travis CI

hugovk:
https://github.com/psf/black/pull/874#issuecomment-544261947 notes pre-commit is needed on the CI.

This adds a new lint job to run pre-commit.

Also, instead of running local versions of the lint tools, which need installing explicitly on Travis, use pinned repo versions and let pre-commit install them.

This is the standard way of running pre-commit. @asottile wrote in https://github.com/psf/black/issues/1073#issuecomment-542772450:

> `repo: local` and especially `language: system` are meant as escape hatches from the """normal way""" and have significant drawbacks (but are provided for those exceptional cases).

Finally, since pre-commit is also running Black, mypy and Flake8 in a single job, we no longer need them as separate jobs and can reduce the number from 6 to 4 (and Travis gives 5 parallel).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,6 +8,7 @@ repos:
         name: black
         language: system
         entry: black
+        require_serial: true
         types: [python]
 
   - repo: https://gitlab.com/pycqa/flake8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,10 +2,13 @@
 # "Version control integration" in README.md.
 exclude: ^(blib2to3/|profiling/|tests/data/)
 repos:
-  - repo: https://github.com/psf/black
-    rev: 19.3b0
+  - repo: local
     hooks:
       - id: black
+        name: black
+        language: system
+        entry: black
+        types: [python]
 
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.7.8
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -2,21 +2,19 @@
...
-      language: system
-      entry: flake8
-      types: [python]
-    - id: mypy
-      name: mypy
-      language: system
-      entry: mypy
-      types: [python]
-      exclude: ^docs/conf.py
+      - id: black
```
asottile: at least for black, I believe the intention was to dogfood black itself -- the others I agree 👍

JelleZijlstra: We also have unit tests that run Black on itself, so this check would likely conflict with those tests if a PR changes formatting.

hugovk: I think the version pre-commit installs is _only_ available for pre-commit, so wouldn't affect those unit tests.

Anyway, I've reverted this bit so Black dogfoods for pre-commit as well.

.pre-commit-config.yaml
```diff
@@ -2,21 +2,22 @@
...
+        name: black
+        language: system
+        entry: black
+        types: [python]
```
asottile: for `black` you'll want `require_serial: true` as black internally does multiprocessing

hugovk: Added!

ambv:
Thank you for this, it will make our tests tighter.

PR #1086, opened 2019-10-21
Fix fmt on/off when multiple exist in leaf prefix

### LINKED ISSUE #1005: Adjacent # fmt: on+off treated as # fmt: on
Operating system: Linux
Python version: 3.7.4
*Black* version: ae5588cf073be10b51dc48ad2fc513e11c4f3e03 (current master)

When blackening this code:
```
def foo():
    # fmt: off
    x = 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0
    x = 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0
    x = 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0
    # fmt: on

    # fmt: off
    x = 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0
    x = 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0
    x = 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0 + 0
    # fmt: on
```
it seems that the second `# fmt: off` is ignored and the second block is formatted.  If a statement (e.g. `pass`) is inserted in between the adjacent fmt directives, the formatting is left alone, so I think this only happens when `# fmt: on` is immediately followed by a `# fmt: off`.

llchan:
The old behavior would detect the existence of a `# fmt: on` in a leaf
node's comment prefix and immediately mark the node as formatting-on,
even if a subsequent `# fmt: off` in the same comment prefix would turn
it back off. This change modifies that logic to track the state through
the entire prefix and take the final state.

Note that this does not fully solve on/off behavior, since any _comment_
lines between the off/on are still formatted. We may need to add
virtual leaf nodes to truly solve that. I will leave that for a separate
commit/PR.

Fixes #1005

### DIFF
--- a/black.py
+++ b/black.py
@@ -3059,9 +3059,14 @@ def generate_ignored_nodes(leaf: Leaf) -> Iterator[LN]:
     """
     container: Optional[LN] = container_of(leaf)
     while container is not None and container.type != token.ENDMARKER:
+        is_fmt_on = False
         for comment in list_comments(container.prefix, is_endmarker=False):
             if comment.value in FMT_ON:
-                return
+                is_fmt_on = True
+            elif comment.value in FMT_OFF:
+                is_fmt_on = False
+        if is_fmt_on:
+            return
 
         yield container
 
--- a/tests/data/fmtonoff.py
+++ b/tests/data/fmtonoff.py
@@ -108,9 +108,13 @@ def on_and_off_broken():
     # fmt: on
     # fmt: off
     this=should.not_be.formatted()
-    but=it  is  formatted
+    and_=indeed . it  is  not  formatted
     because . the . handling . inside . generate_ignored_nodes()
-    doesnt . consider . ordering . within . one . prefix
+    now . considers . multiple . fmt . directives . within . one . prefix
+    # fmt: on
+    # fmt: off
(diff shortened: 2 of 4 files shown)


ambv:
This is, for lack of a better word, fantastic. Thank you for your contribution, I also enjoyed your tongue-in-cheek test. Awesome job, thanks!

PR #1089, opened 2019-10-24
setup.py: rename _version.py to _black_version.py

mmcclimon:
In #1082, _version.py was renamed to _black_version.py, but the
py_modules line in setup.py wasn't changed, which meant that when
installed from source, running it failed with something like:

```
  File "~/.pyenv/versions/3.6.5/bin/black", line 5, in <module>
    from black import patched_main
  File "~/.pyenv/versions/3.6.5/lib/python3.6/site-packages/black.py", line 55, in <module>
    from _black_version import version as __version__
ModuleNotFoundError: No module named '_black_version'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     author_email="lukasz@langa.pl",
     url="https://github.com/psf/black",
     license="MIT",
-    py_modules=["black", "blackd", "_version"],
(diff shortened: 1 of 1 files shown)


llacroix:
Can reproduce the bug. So +1

PR #1103, opened 2019-10-28
Always move the prefix out when wrapping with parentheses

### LINKED ISSUE #1097: Comment moved when adding tuple parentheses
**To reproduce**

1. Take this file:

```python
class MyClass:
    def my_def(self):
        if self.magic == b"BLP1":
            # Only present for BLP1
            self._blp_encoding, = struct.unpack("<i", self.fp.read(4))
            self._blp_subtype, = struct.unpack("<i", self.fp.read(4))
```

2. Run _Black_ on it with no arguments
3. The comment is moved to the wrong place, and extra parenthesis are added, needing three lines instead of one:

```python
class MyClass:
    def my_def(self):
        if self.magic == b"BLP1":
            (self._blp_encoding,) = struct.unpack(  # Only present for BLP1
                "<i", self.fp.read(4)
            )
            (self._blp_subtype,) = struct.unpack("<i", self.fp.read(4))
```

**Expected behavior**

If I now manually move the comment back to where it belongs:

```python
class MyClass:
    def my_def(self):
        if self.magic == b"BLP1":
            # Only present for BLP1
            (self._blp_encoding,) = struct.unpack(
                "<i", self.fp.read(4)
            )
            (self._blp_subtype,) = struct.unpack("<i", self.fp.read(4))
```

And re-run Black, it's tightened back up:

```python
class MyClass:
    def my_def(self):
        if self.magic == b"BLP1":
            # Only present for BLP1
            (self._blp_encoding,) = struct.unpack("<i", self.fp.read(4))
            (self._blp_subtype,) = struct.unpack("<i", self.fp.read(4))
```

It'd be nice to see this on the first run.

Another (less disruptive) example of a moved comment here:

* https://github.com/hugovk/Pillow/commit/c669a63e1118#diff-b49fc26d17e5d3b539a05c7850abe2b0L3259

**Environment**

- Version: master (version 19.3b1.dev123+g4c4a51c)
- OS and Python version: macOS 10.14, Python 3.7.4

I've only tested master.

ambv:
Fixes #1097

### DIFF
--- a/black.py
+++ b/black.py
@@ -2966,16 +2966,9 @@ def normalize_invisible_parens(node: Node, parens_after: Set[str]) -> None:
 
             if child.type == syms.atom:
                 if maybe_make_parens_invisible_in_atom(child, parent=node):
-                    lpar = Leaf(token.LPAR, "")
-                    rpar = Leaf(token.RPAR, "")
-                    index = child.remove() or 0
-                    node.insert_child(index, Node(syms.atom, [lpar, child, rpar]))
+                    wrap_in_parentheses(node, child, visible=False)
             elif is_one_tuple(child):
-                # wrap child in visible parentheses
-                lpar = Leaf(token.LPAR, "(")
-                rpar = Leaf(token.RPAR, ")")
-                child.remove()
-                node.insert_child(index, Node(syms.atom, [lpar, child, rpar]))
+                wrap_in_parentheses(node, child, visible=True)
             elif node.type == syms.import_from:
                 # "import from" nodes store parentheses directly as part of
                 # the statement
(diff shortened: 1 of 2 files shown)


black.py
```diff
@@ -3158,6 +3143,24 @@ def unwrap_singleton_parenthesis(node: LN) -> Optional[LN]:
...
+
+    If `visible` is False, the leaves will be valueless (and thus invisible).
+    """
+    lpar = Leaf(token.LPAR, "(" if visible else "")
+    rpar = Leaf(token.RPAR, ")" if visible else "")
+    prefix = child.prefix
+    child.prefix = ""
+    index = child.remove() or 0
+    new_child = Node(syms.atom, [lpar, child, rpar])
+    new_child.prefix = prefix
```
ambv: Technically this line is wrong since Nodes don't hold prefixes, Leafs do. But lib2to3 does the correct thing for us fortunately.

ambv:
Haha, better code reuse here lead to falling coverage. Lines which were not covered before either are now a bigger percentage of the file.

PR #1113, opened 2019-10-29
Fix unstable formatting with some `# type: ignore`s

### LINKED ISSUE #1061: Unstable formatting with "# type: ignore" comment
Operating system: Arch Linux
Python version: 3.7
*Black* version: 19.3b0, 19.3b1.dev96+g57ab909

I think I'm seeing a "NTERNAL ERROR: Black produced different code on the second pass of the formatter." which is different from others I find here (#954, #1042, #1044).

With this code:

```python
class Test:
    def _init_host(self, parsed) -> None:
        if (parsed.hostname is None or  # type: ignore
                not parsed.hostname.strip()):
            pass
```

I get this diff:

```diff
--- source
+++ first pass
@@ -1,6 +1,8 @@
 class Test:
     def _init_host(self, parsed) -> None:
-        if (parsed.hostname is None or  # type: ignore
-                not parsed.hostname.strip()):
+        if (
+            parsed.hostname is None
+            or not parsed.hostname.strip()  # type: ignore
+        ):
             pass
 
```

```diff
--- first pass
+++ second pass
@@ -1,8 +1,7 @@
 class Test:
     def _init_host(self, parsed) -> None:
         if (
-            parsed.hostname is None
-            or not parsed.hostname.strip()  # type: ignore
+            parsed.hostname is None or not parsed.hostname.strip()  # type: ignore
         ):
             pass
 
```

msullivan:
`type: ignore` shouldn't block collapsing a line, since it will still
apply fine to the merged line. This prevents an issue where a reformat
causes it to shift lines and then be merged on a subsequent pass.

There is a downside to this, which is that it can cause a `type:
ignore` to apply to more code than was originally intended. There
might be a way to apply this in a more limited situation, but I'm not
sure what it is.

Fixes #1061.

### DIFF
--- a/tests/data/comments7.py
+++ b/tests/data/comments7.py
@@ -93,9 +93,7 @@ result = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 
 
 def func():
-    c = call(
-        0.0123, 0.0456, 0.0789, 0.0123, 0.0789, a[-1],  # type: ignore
-    )
+    c = call(0.0123, 0.0456, 0.0789, 0.0123, 0.0789, a[-1],)  # type: ignore
 
     # The type: ignore exception only applies to line length, not
     # other types of formatting.
--- a/black.py
+++ b/black.py
@@ -1383,7 +1383,10 @@ class Line:
         for leaf_id, comments in self.comments.items():
             for comment in comments:
                 if is_type_comment(comment):
-                    if leaf_id not in ignored_ids or comment_seen:
+                    if comment_seen or (
+                        not is_type_comment(comment, " ignore")
+                        and leaf_id not in ignored_ids
+                    ):
                         return True
 
                 comment_seen = True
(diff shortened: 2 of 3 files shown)


tests/data/comments7.py
```diff
@@ -93,9 +93,7 @@ def func():
 
 
 def func():
-    c = call(
-        0.0123, 0.0456, 0.0789, 0.0123, 0.0789, a[-1],  # type: ignore
-    )
+    c = call(0.0123, 0.0456, 0.0789, 0.0123, 0.0789, a[-1],)  # type: ignore
```
msullivan: As an aside, I really don't like the side-effect of #826 that causes this comma to be preserved. It seems like it won't cause much trouble for an already-formatted codebase but I would find it extremely frustrating to clean up in a project that I'm newly formatting.

JelleZijlstra: @ambv has some exciting ideas about invisible commas that could fix this.

ambv: Yeah, we'll fix it in the next beta.

ambv: @msullivan, while I have you here, if you redo your mypycization of Black for the current state of things I promise to land it very quickly. Sorry the other one went stale.

JelleZijlstra:
Sorry for the delay!

PR #1116, opened 2019-10-29
Switch from attrs to dataclasses

msullivan:
The main motivation here is that mypyc is going to have custom support
for dataclasses but probably not attrs.

### DIFF
--- a/black.py
+++ b/black.py
@@ -39,7 +39,7 @@ from typing import (
 )
 
 from appdirs import user_cache_dir
-from attr import dataclass, evolve, Factory
+from dataclasses import dataclass, field
 import click
 import toml
 from typed_ast import ast3, ast27
@@ -185,7 +185,7 @@ VERSION_TO_FEATURES: Dict[TargetVersion, Set[Feature]] = {
 
 @dataclass
 class FileMode:
-    target_versions: Set[TargetVersion] = Factory(set)
+    target_versions: Set[TargetVersion] = field(default_factory=set)
     line_length: int = DEFAULT_LINE_LENGTH
     string_normalization: bool = True
     is_pyi: bool = False
@@ -629,7 +629,12 @@ def format_file_in_place(
     `mode` and `fast` options are passed to :func:`format_file_contents`.
     """
     if src.suffix == ".pyi":
-        mode = evolve(mode, is_pyi=True)
+        mode = FileMode(
+            mode.target_versions,
+            mode.line_length,
+            mode.string_normalization,
+            is_pyi=True,
+        )
 
     then = datetime.utcfromtimestamp(src.stat().st_mtime)
     with open(src, "rb") as buf:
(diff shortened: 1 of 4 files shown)


black.py
```diff
@@ -629,7 +629,12 @@ def format_file_in_place(
     `mode` and `fast` options are passed to :func:`format_file_contents`.
     """
     if src.suffix == ".pyi":
-        mode = evolve(mode, is_pyi=True)
+        mode = FileMode(
+            mode.target_versions,
+            mode.line_length,
+            mode.string_normalization,
+            is_pyi=True,
+        )
```
ilevkivskyi: I think you can use `dataclasses.replace(mode, is_pyi=True)`, which is kind of more idiomatic.

black.py
```diff
@@ -1028,11 +1033,11 @@ class BracketTracker:
...
     depth: int = 0
-    bracket_match: Dict[Tuple[Depth, NodeType], Leaf] = Factory(dict)
-    delimiters: Dict[LeafID, Priority] = Factory(dict)
+    bracket_match: Dict[Tuple[Depth, NodeType], Leaf] = field(default_factory=dict)
+    delimiters: Dict[LeafID, Priority] = field(default_factory=dict)
     previous: Optional[Leaf] = None
-    _for_loop_depths: List[int] = Factory(list)
-    _lambda_argument_depths: List[int] = Factory(list)
+    _for_loop_depths: List[int] = field(default_factory=list)
+    _lambda_argument_depths: List[int] = field(default_factory=list)
```
ilevkivskyi: _Maybe_ it makes sense to define a helper:
```python
def Factory(fn):
    return field(default_factory=fn)
```
to minimize the size of the diff? Not sure however what looks better.

JelleZijlstra: I think this code is fine. I'd rather use dataclasses idiomatically than minimize diff size.

black.py
```diff
@@ -629,7 +629,7 @@ def format_file_in_place(
     `mode` and `fast` options are passed to :func:`format_file_contents`.
     """
     if src.suffix == ".pyi":
-        mode = evolve(mode, is_pyi=True)
+        mode = mode.replace(is_pyi=True)
```
JelleZijlstra: it's actually `dataclasses.replace`, a global function (presumably so dataclasses don't inject a `.replace` method everywhere)

black.py
```diff
@@ -1160,9 +1160,11 @@ class Line:
     """Holds leaves and comments. Can be printed with `str(line)`."""
 
     depth: int = 0
-    leaves: List[Leaf] = Factory(list)
-    comments: Dict[LeafID, List[Leaf]] = Factory(dict)  # keys ordered like `leaves`
-    bracket_tracker: BracketTracker = Factory(BracketTracker)
+    leaves: List[Leaf] = field(default_factory=list)
+    comments: Dict[LeafID, List[Leaf]] = field(
+        default_factory=dict
+    )  # keys ordered like `leaves`
```
JelleZijlstra: Can you put this comment on the previous line by itself instead? (I'd personally prefer if Black did that automatically in this case.)

setup.py
```diff
@@ -42,7 +42,8 @@ def get_long_description() -> str:
         "typed-ast>=1.4.0",
         "regex",
         "pathspec>=0.6, <1",
-    ],
+    ]
+    + (["dataclasses>=0.6"] if sys.version_info < (3, 7) else []),
```
JelleZijlstra: Prefer to do it like in https://github.com/python/typing/commit/f254d69c56920819b4211f47fdaba6472097dc5c

black.py
msullivan: Wow you'd think I'd have run the tests, huh?

PR #1117, opened 2019-10-29
Split out Change Log

### LINKED ISSUE #1105: Move CHANGES out of README.md
README.md is very long and finding the changes is a long scroll / cmd +f time. Also, some bots are unable to change the CHANGES log with it inside README.md.

Due to this, can we look at making a separate CHANGES* (maybe CHANGES.md) file so bots can see the changes list for future upgrades.
- black uses `pyproject.toml` - This PR needs to ensure the Change log can end up on https://pypi.org/project/black output
- Also need to ensure we don't break documentation building with this PR

cooperlees:
- Move to CHANGES.md to allow bots to see changes
- MANIFEST.in already includes *.md so CHANGES.md will be included
- This maintains format but the change log will now be after acknowledgements
- This also ensures Change Log still gets added to pypi.org via setup.py function

Fixes #1105


setup.py
```diff
@@ -10,9 +10,11 @@
 
 
 def get_long_description() -> str:
+    changes_md = CURRENT_DIR / "CHANGES.md"
     readme_md = CURRENT_DIR / "README.md"
-    with open(readme_md, encoding="utf8") as ld_file:
-        return ld_file.read()
+    with readme_md.open("r", encoding="utf8") as rmd_file:
+        with changes_md.open("r", encoding="utf8") as cmd_file:
+            return rmd_file.read() + "\n\n" + cmd_file.read()
```
graingert: because you've got pathlib you can just do:
```python
def get_long_description() -> str:
    return "\n\n".join(
        (CURRENT_DIR / p).read_text(encoding="utf8")
        + "\n\n"
        + (CURRENT_DIR / "CHANGES.md").read_text(encoding="utf8")
    )
```

graingert: or this if you don't like the duplication:

```python
def get_long_description() -> str:
    return "\n\n".join(
        [
            (CURRENT_DIR / p).read_text(encoding="utf8")
            for p in ["README.md", "CHANGES.md"]
        ]
    )
```

cooperlees:
😱I'd forgot about this.

Issue #1091, opened 2019-10-25
Support unpacking in yield/return


JelleZijlstra:
Python 3.8 made a minor syntax change to allow
```
def g():
    rest = (4, 5, 6)
    return 1, 2, 3, *rest
```
and the equivalent with `yield`. See https://bugs.python.org/issue32117.

Black doesn't support this syntax yet.

vemel:
This looks like a new feature for `plib2to3` as it cannot parse `*` character in this line.

Unfortunately, even in `py38` `plib2to3` return statement does not support startest: https://github.com/python/cpython/blob/3.8/Lib/lib2to3/Grammar.txt#L52

I can prepare a PR though

vemel:
I prepared a [PR](https://github.com/psf/black/pull/1121) above and add a test

vemel:
Added PR to [CPython](https://github.com/python/cpython/pull/16994) as well. I hope `black` can switch to official `lib2to3` soon.

isidentical:
Looks like #1121 merged, guess we can close this?

PR #1121, opened 2019-10-30
Support py38-style starred expressions in return statement

vemel:
### Goal

https://github.com/psf/black/issues/1091

Support `py38` starred expressions in `return` and `yield` statements. [Bug description](https://bugs.python.org/issue32117)

```python
def starred_return():
    my_list = ["value2", "value3"]
    return "value1", *my_list
```

### Investigation

Event the newest [plib2to3 grammar](https://github.com/python/cpython/blob/3.8/Lib/lib2to3/Grammar.txt#L52) does not support starred expressions in return statements, so switching to a new version will not solve the issue.

The idea is to use `return_stmt: 'return' [testlist_star_expr]` in `Grammar.txt` to support both old and new return statements.

### Changes

- Accept starred expressions in `return_stmt` 
- Accept starred expressions in `yield_stmt` 
- Add test `test_python38`

### Not done

- I did not add it as a new `Feature` so it does not require `--target-version=py38`, as it is more of a bug

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -584,6 +584,21 @@ class BlackTestCase(unittest.TestCase):
         self.invokeBlack([str(source_path), "--target-version", "py37"], exit_code=123)
 
     @patch("black.dump_to_file", dump_to_stderr)
+    def test_python37(self) -> None:
+        source_path = (THIS_DIR / "data" / "python37.py").resolve()
+        source, expected = read_data("python37")
+        actual = fs(source)
+        self.assertFormatEqual(expected, actual)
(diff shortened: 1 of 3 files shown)


tests/test_black.py
```diff
@@ -584,19 +584,15 @@ def test_async_as_identifier(self) -> None:
         self.invokeBlack([str(source_path), "--target-version", "py37"], exit_code=123)
 
     @patch("black.dump_to_file", dump_to_stderr)
-    def test_python37(self) -> None:
```
JelleZijlstra: Could you add a new test instead of replacing the old one?

Issue #776, opened 2019-03-22
Concurrency crash


duijf:
Operating system: Ubuntu Xenial (on Travis)
Python version: 3.6.8
*Black* version: 19.3b0
Does also happen on master: Haven't tested but when through the commits since the release. No commits seem to touch the codepath we're seeing here.

`black` crashes on certain system configurations due to multiprocessing. Here is the traceback we're seeing.

```
|Traceback (most recent call last):
|  File "/usr/local/lib/XXX/virtualenv/bin/black", line 11, in <module>
|    sys.exit(patched_main())
|  File "/usr/local/lib/XXX/virtualenv/lib/python3.6/site-packages/black.py", line 3754, in patched_main
|    main()
|  File "/usr/local/lib/XXX/virtualenv/lib/python3.6/site-packages/click/core.py", line 764, in __call__
|    return self.main(*args, **kwargs)
|  File "/usr/local/lib/XXX/virtualenv/lib/python3.6/site-packages/click/core.py", line 717, in main
|    rv = self.invoke(ctx)
|  File "/usr/local/lib/XXX/virtualenv/lib/python3.6/site-packages/click/core.py", line 956, in invoke
|    return ctx.invoke(self.callback, **ctx.params)
|  File "/usr/local/lib/XXX/virtualenv/lib/python3.6/site-packages/click/core.py", line 555, in invoke
|    return callback(*args, **kwargs)
|  File "/usr/local/lib/XXX/virtualenv/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
|    return f(get_current_context(), *args, **kwargs)
|  File "/usr/local/lib/XXX/virtualenv/lib/python3.6/site-packages/black.py", line 435, in main
|    executor = ProcessPoolExecutor(max_workers=os.cpu_count())
|  File "/usr/lib/python3.6/concurrent/futures/process.py", line 402, in __init__
|    EXTRA_QUEUED_CALLS)
|  File "/usr/lib/python3.6/multiprocessing/context.py", line 102, in Queue
|    return Queue(maxsize, ctx=self.get_context())
|  File "/usr/lib/python3.6/multiprocessing/queues.py", line 42, in __init__
|    self._rlock = ctx.Lock()
|  File "/usr/lib/python3.6/multiprocessing/context.py", line 67, in Lock
|    return Lock(ctx=self.get_context())
|  File "/usr/lib/python3.6/multiprocessing/synchronize.py", line 162, in __init__
[...]

PR #1141, opened 2019-11-05
permits black to run in AWS Lambda:

### LINKED ISSUE #1138: Feature request:  permits to disable parallel run for system without support for it
On some virtualized environment you don't have access to `/dev/shm` which prevent the use of multi-processing. The most notable use case being AWS Lambda, where /dev/shm is simply not present.

A quick and dirty patch show that simply replacing https://github.com/psf/black/blob/master/black.py#L544 by `None`
and removing this line https://github.com/psf/black/blob/master/black.py#L559
permits black to run in AWS Lambda, which is a huge win in term of CI (as black fits perfectly the use for serverless, i.e short runtime, low memory consumption, triggered on specific event)

so it could be implemented either (or both) by providing a `--no-parallel` or by detecting that `/dev/shm` is not possible (I need to dig on how to do that) and in these case to not use the ProcessPoolExecutor

I can provide the pull request if it seems acceptable to you

allan-simon:
AWS Lambda and some other virtualized environment may not permit access
to /dev/shm on Linux and as such, trying to use ProcessPoolExecutor will
fail.

As using parallelism is only a 'nice to have' feature of black, if it fails
we gracefully fallback to a monoprocess implementation, which permits black
to finish normally.

should fix #1138 and #776

### DIFF
--- a/black.py
+++ b/black.py
@@ -541,7 +541,14 @@ def reformat_many(
     if sys.platform == "win32":
         # Work around https://bugs.python.org/issue26903
         worker_count = min(worker_count, 61)
-    executor = ProcessPoolExecutor(max_workers=worker_count)
+    try:
+        executor = ProcessPoolExecutor(max_workers=worker_count)
+    except OSError:
+        # we arrive here if the underlying system does not support multi-processing
+        # like in AWS Lambda, in which case we gracefully fallback to the default
+        # mono-process Executor by using None
+        executor = None
+
     try:
         loop.run_until_complete(
             schedule_formatting(
@@ -556,7 +563,8 @@ def reformat_many(
         )
     finally:
         shutdown(loop)
-        executor.shutdown()
+        if executor is not None:
+            executor.shutdown()
 
 
 async def schedule_formatting(
@@ -566,7 +574,7 @@ async def schedule_formatting(
     mode: FileMode,
     report: "Report",
     loop: asyncio.AbstractEventLoop,
-    executor: Executor,
+    executor: Optional[Executor],
 ) -> None:
     """Run formatting of `sources` in parallel using the provided `executor`.
 
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -1159,6 +1159,27 @@ class BlackTestCase(unittest.TestCase):
             self.assertIn(one, cache)
             self.assertIn(two, cache)
 
+    @patch("black.ProcessPoolExecutor", autospec=True)
+    def test_works_in_mono_process_only_environment(self, mock_executor) -> None:
+        mock_executor.side_effect = OSError()
+        mode = black.FileMode()
+        with cache_dir() as workspace:
+            one = (workspace / "one.py").resolve()
+            with one.open("w") as fobj:
+                fobj.write("print('hello')")
+            two = (workspace / "two.py").resolve()
+            with two.open("w") as fobj:
+                fobj.write("print('hello')")
+            black.write_cache({}, [one], mode)
+            self.invokeBlack([str(workspace)])
+            with one.open("r") as fobj:
+                self.assertEqual(fobj.read(), "print('hello')")
+            with two.open("r") as fobj:
+                self.assertEqual(fobj.read(), 'print("hello")\n')
+            cache = black.read_cache(mode)
+            self.assertIn(one, cache)
+            self.assertIn(two, cache)
+
     def test_no_cache_when_writeback_diff(self) -> None:
         mode = black.FileMode()
         with cache_dir() as workspace:


black.py
```diff
@@ -541,7 +541,14 @@ def reformat_many(
     if sys.platform == "win32":
         # Work around https://bugs.python.org/issue26903
         worker_count = min(worker_count, 61)
-    executor = ProcessPoolExecutor(max_workers=worker_count)
+    try
```
zsol: you're missing a `:` here

zsol:
Would be amazing if you could add a test that patches out `ProcessPoolExecutor` and makes it raise so we have coverage of that branch

black.py
allan-simon: yes sorry you've been faster than me to see I forgot to amend before pushing :)

allan-simon:
@zsol  haha I was afraid you would say that, hmmm let me try

allan-simon:
@zsol  I've tried something, but as I'm not that familiar with unittest library and the way "invokeBlack" works, I'm not against a review from you even though the test are failing, to see if i'm missing the elephant in the roomt

hawkins:
I'm not sure if either of you figured this out already or not, but I've found that simply replacing `ProcessPoolExecutor` wherever it occurs in `black.py` with `ThreadPoolExecutor` allows Black to run on AWS Lambda without issue. I can't speak to other virtual environments though.

allan-simon:
yes actually, none default to threadpool :)

allan-simon:
@hawkins  btw, help appreciated on this PR,  I struggle with the test , so if you're also interested in getting it upstream :)

hawkins:
Oh - so by setting `executor = None` we're somehow multiprocessing with `ThreadPoolExecutor`? Very cool. Have you tested this code on Lambda?

And sure, I can try to take a stab at the test too.

allan-simon:
@hawkins  yes our CI is based on my fork currently, running in lambda , so i can assure you it's working :)

hawkins:
Just dove through the asyncio docs and now I see how that works. Thanks! Very cool, nice job with using `None` then :ok_hand:

JelleZijlstra:
CI seems unhappy, could you take a look? Thanks!

allan-simon:
@JelleZijlstra , I tried in the past, but I think I reach the limitation of my understanding on how the tests are made with black. (i.e I think the test I've added is not exactly written the way it should be written, but why  ? :man_shrugging:  )

Terrance:
For what it's worth, Termux on Android also doesn't support `ProcessPoolExecutor`, and fails in the same place, though slightly differently:

```
  File "/data/data/com.termux/files/usr/lib/python3.8/multiprocessing/synchronize.py", line 28, in <module>
    from _multiprocessing import SemLock, sem_unlink
ImportError: cannot import name 'SemLock' from '_multiprocessing' (/data/data/com.termux/files/usr/lib/python3.8/lib-dynload/_multiprocessing.cpython-38.so)

During handling of the above exception, another exception occurred:

  File "/data/data/com.termux/files/usr/bin/black", line 8, in <module>
    sys.exit(patched_main())
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/black.py", line 4135, in patched_main
    main()
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/click/core.py", line 764, in __call__
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/click/core.py", line 956, in invoke
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/click/core.py", line 555, in invoke
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/click/decorators.py", line 17, in new_func
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/black.py", line 463, in main
    reformat_many(
  File "/data/data/com.termux/files/usr/lib/python3.8/site-packages/black.py", line 529, in reformat_many
    executor = ProcessPoolExecutor(max_workers=worker_count)
  File "/data/data/com.termux/files/usr/lib/python3.8/concurrent/futures/process.py", line 555, in __init__
    self._call_queue = _SafeQueue(
  File "/data/data/com.termux/files/usr/lib/python3.8/concurrent/futures/process.py", line 165, in __init__
    super().__init__(max_size, ctx=ctx)
  File "/data/data/com.termux/files/usr/lib/python3.8/multiprocessing/queues.py", line 42, in __init__
    self._rlock = ctx.Lock()
  File "/data/data/com.termux/files/usr/lib/python3.8/multiprocessing/context.py", line 67, in Lock
    from .synchronize import Lock
  File "/data/data/com.termux/files/usr/lib/python3.8/multiprocessing/synchronize.py", line 30, in <module>
    raise ImportError("This platform lacks a functioning sem_open" +
ImportError: This platform lacks a functioning sem_open implementation, therefore, the required synchronization primitives needed will not function, see issue 3770.
```

Replacing the executor creation to fall back to a `ThreadPoolExecutor` also works here.  This pull request only catches `OSError` though, so wouldn't detect this case at present.

ambv:
Thanks! ✨ 🍰 ✨

hugovk:
Looks like this has turned the CI red on master?

JelleZijlstra:
@ambv is telling me he's fixing it.

Terrance:
Looks like the PR didn't get updated to [also handle `ImportError`](#issuecomment-576576943) -- should I raise a separate issue for that, or can that be fixed whilst we're here?

JelleZijlstra:
@Terrance just send a new PR and we'll take a look!

Terrance:
Submitted #1400.

PR #1142, opened 2019-11-06
Issue 957 inconsistent use of _ and -

za:
See #957

### DIFF
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -36,7 +36,7 @@ endif
 if !exists("g:black_linelength")
   let g:black_linelength = 88
 endif
-if !exists("g:black_skip_string_normalization")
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
This actually makes it *more* inconsistent, because some other options used in the same vim plugin still use underscores.

Also, shouldn't we accept both the underscore and the dash version to preserve compatibility for existing users?

ambv:
Accepting both variants is rather tricky because people could have both set and now we'd have no choice but to refuse to start. Plus we'd have to store our defaults in yet a third variable. Sounds more complicated than it's worth.

cooperlees:
This PR has become very stale. Please reopen if you wish to finish it.

PR #1157, opened 2019-11-13
Use vim autoload script

kalekseev:
Python code in vim plugin takes a lot of time to evaluate and it's currently done at startup time, timing `vim --startuptime` on my machine:

```
363.235  155.231  155.231: sourcing /Users/a/.vim/bundle/black/plugin/black.vim
```

Using autoload script technique we can postpone python code evaluation till the first function call.

Startup time improvement:

```
219.125  000.074  000.074: sourcing /Users/a/.vim/bundle/black/plugin/black.vim
```

This changes plugin behaviour, currently it install virtualenv on vim start, with that change env will be installed on first function call. Not sure if we need to change this readme sentence

> On first run, the plugin creates its own virtualenv using the right Python version and automatically installs Black. 

For me "first run" sounds like first time when we called plugin's function.

### DIFF
--- a/autoload/black.vim
+++ b/autoload/black.vim
@@ -1,0 +1,172 @@
+python3 << EndPython3
+import collections
+import os
+import sys
+import vim
+from distutils.util import strtobool
+
+
+class Flag(collections.namedtuple("FlagBase", "name, cast")):
+  @property
+  def var_name(self):
+    return self.name.replace("-", "_")
+
+  @property
+  def vim_rc_name(self):
+    name = self.var_name
+    if name == "line_length":
+      name = name.replace("_", "")
+    return "g:black_" + name
+
+
+FLAGS = [
+  Flag(name="line_length", cast=int),
+  Flag(name="fast", cast=strtobool),
+  Flag(name="string_normalization", cast=strtobool),
+  Flag(name="quiet", cast=strtobool),
+]
+
+
+def _get_python_binary(exec_prefix):
+  try:
+    default = vim.eval("g:pymode_python").strip()
+  except vim.error:
+    default = ""
+  if default and os.path.exists(default):
+    return default
+  if sys.platform[:3] == "win":
+    return exec_prefix / 'python.exe'
+  return exec_prefix / 'bin' / 'python3'
+
+def _get_pip(venv_path):
+  if sys.platform[:3] == "win":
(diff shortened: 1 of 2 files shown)


dimbleby:
cf #416 in which:

- [`dimbleby/black.vim`](https://github.com/dimbleby/black.vim) is introduced, which makes the same move to using `autoload` (and also fixes a couple of other things)
- but then we note that there are many other ways of running Black from Vim, and on the whole it's probably better for this project to point at one of them than to provide its own plugin

I still think that's right: it's a bit weird to embed a Vim plugin in this repository.  I'd suggest that maintainers are better off removing the Vim plugin here altogether, and documenting one or more of the alternatives.

kalekseev:
@dimbleby yeah I use black with AleFix myself, but I like the ability to install/update black via that plugin and use that binary in AleFix. Anyway it need to be addressed somehow, right now anyone who uses that plugin introduce a huge penalty to vim startup time.

cooperlees:
Hi, I’m cleaning up stale PRs. Do you still plan to work on this?

kalekseev:
@cooperlees the pr is ready, as @dimbleby said there were no decision about the future of vim plugin in black at the moment of pr. I'll rebase it if you accept that patch.

ichard26:
I honestly don't know about the state of the vim plugin, but I don't think it's going anytime soon so a rebase would be appreciated! Sorry for taking so long for us to get back you!

kalekseev:
@ichard26 rebased

> Admittingly we should probably have a proper conversation on whether the vim plugin should even exist in this repo (not sure who is qualified to review PRs related to it for one), but that ... seems ... uh .. unlikely :(

I would prefer that plugin removed from that repo too, the current state when plugin impose 150ms startup penalty on each vim start is ... um ... bad? We could at least warn people in docs "THAT PLUGIN WILL MAKE YOUR EDITOR START 100ms-200ms SLOWER"

ichard26:
/cc @JelleZijlstra @cooperlees let's merge this before it gets stale again. Thanks in advance!

I'll open an issue for the "should the vim plugin that exists here... still exist?".

ichard26:
> I'll open an issue for the "should the vim plugin that exists here... still exist?"

done, see GH-2027

PR #1160, opened 2019-11-16
Correct 1% to 1.25%

JonPizza:
Technically, 1/80 is 0.0125, so therefore, it should be 1.25%!


hugovk:
Funny, the CI failed because the line is now too long!

To fix:

```bash
pip install pre-commit
pre-commit run --all-files
```

ala-ableton:
> Technically, 1/80 is 0.0125, so therefore, it should be 1.25%!

Except that _Black_ defaults to 88 characters per line, not 80.

JelleZijlstra:
It's still approximately 1%. I don't think being more precise is valuable here.

PR #1164, opened 2019-11-19
Issue 1155: Handle LHS and RHS values differently

vemel:
### Idea

`black` already has `left_side_split` function, but it is only used for function definitions. The idea is to extract LHS and RHS values from a string, split them separately and join them back to a final result.

I added logic for splitting LHS and RHS and updated `SplitFunc` to always be aware of a target line max length. The downside of this approach is that now if LHS can be rendered to a single line - it always does. I need an advice on how to handle it correctly. Probably, we should explode LHS line if RHS has unsplittable type ignore, but it is not always safe to do this.

### Changes

- Removed `line_str` argument from `is_line_short_enough` function - not used
- Removed `Line.is_collection_with_optional_trailing_comma` method
- Removed `Line.contains_unsplittable_type_ignore` method
- Added `Line.get_optional_trailing_comma_index` method
- Added `Line.get_leaf_index` method
- Added `Line._get_collection_comma_indexes` method - helper for `get_optional_trailing_comma_index`
- Added `Line.get_unsplittable_type_ignore` method
- Added `Line._get_assignment_index` method - helper to split `Line.get_left_hand_side` and `Line.get_right_hand_side` 
- Added `Line._get_top_level_collection_indexes` method - helper for `Line.get_optional_trailing_comma_index`
- Added `Line.get_left_hand_side` method - returns line with leaves and comment for the LHS value
- Added `Line.get_left_right_side`method - returns line with leaves and comment for the RHS value and assignment
- Added `Line.should_be_rendered_as_single_line` method - check if line should not be exploded anf fits desired line length
- Added `Line.is_short_enough` method - potential replacement for `is_line_short_enough` function, but can handle multiline `Line` as well
- Added `right_hand_split_with_omit` function - replacement for `rhs` function that was created dynamically
- Added `Line.get_source_line_numbers` method to check unsplittable type ignore is on the same lines as LHS
- Added `split_line_side` function - `split_line` uses it for LHS and RHS
- All `SplitFunc` now have `line_length` and `first_line_length` arguments
- Updated unit tests

### Output result changes

- 

```python
# LHS 1-item subscript output
Tuple[str,]
# RHS 1-item subscript output
MyType = Tuple[str,]

# 1-arg function definition with trailing comma output
def test(
    a,
):
    pass

# multi-arg function definition with trailing comma output
def test(
    a,
    b,
):
    pass
```


black.py
```diff
@@ -149,8 +149,14 @@ class Feature(Enum):
 
 VERSION_TO_FEATURES: Dict[TargetVersion, Set[Feature]] = {
     TargetVersion.PY27: {Feature.ASYNC_IDENTIFIERS},
-    TargetVersion.PY33: {Feature.UNICODE_LITERALS, Feature.ASYNC_IDENTIFIERS},
-    TargetVersion.PY34: {Feature.UNICODE_LITERALS, Feature.ASYNC_IDENTIFIERS},
+    TargetVersion.PY33: {
```
JelleZijlstra: Is this a result of the changes from your PR? If so, I don't think that's an improvement.

black.py
```diff
@@ -1266,69 +1272,261 @@ def is_stub_class(self) -> bool:
             Leaf(token.DOT, ".") for _ in range(3)
         ]
 
-    @property
-    def is_collection_with_optional_trailing_comma(self) -> bool:
-        """Is this line a collection literal with a trailing comma that's optional?
+    def _get_assignment_index(self) -> Optional[int]:
+        """
+        Get fthe first unbracketed assignment index.
```
JelleZijlstra: the

black.py
```diff
@@ -1266,69 +1272,261 @@ def is_stub_class(self) -> bool:
...
-            # literal) the literal gets wrapped in temporary parens
-            # during parsing. This case is covered by the
-            # collections.py test data.
-            opener = closer.opening_bracket
-            for _open_index, leaf in enumerate(self.leaves):
-                if leaf is opener:
-                    break
+        Can be used to check if line has multiple lines in source.
+        """
+        return set([i.lineno for i in self.leaves if i.lineno])
```
JelleZijlstra: Use a set comprehension

black.py
```diff
@@ -3936,13 +4291,13 @@ def enumerate_with_length(
         yield index, leaf, length
 
 
-def is_line_short_enough(line: Line, *, line_length: int, line_str: str = "") -> bool:
+def is_line_short_enough(line: Line, line_length: int) -> bool:
```
JelleZijlstra: This should be kept keyword-only.

blib2to3/pgen2/driver.py
```diff
@@ -128,7 +128,10 @@ def parse_stream(self, stream: IO[Text], debug: bool = False) -> NL:
         return self.parse_stream_raw(stream, debug)
 
     def parse_file(
-        self, filename: Path, encoding: Optional[Text] = None, debug: bool = False,
+        self,
```
JelleZijlstra: This should be formatted as it was before your change.

vemel: I found that if line is a bit longer, we get

```python
self, filename: Path, encoding: Optional[
    Text
] = None, debug: bool = False,
```

Is it okay? Can easily be fixed, I just should remove `[delimiter_split, standalone_comment_split]` for LHS values in brackets.

black.py
vemel: Reverted

black.py
vemel: Done

black.py
vemel: Fixed

black.py
vemel: Hmm, I added trailing commas for some reason. Reverted this change. With no trailing commas the result is the same as it was.

vemel: Hmm, I added trailing commas for some reason. Reverted this change. With no trailing commas the result is the same as it was.

blib2to3/pgen2/driver.py
vemel: Removed trailing comma, now this line is the same as it was.

vemel:
Makes sense. What do you think about leaving only part that explodes lines on trailing commas in nested collections, e.g. 

```python
def test(
    [
        1,
        2,
    ]
): ...
```

Then we need only changes in `Line` class.

ambv:
@vemel, makes sense to leave only the case you show in your last comment.

vemel:
Okay, I will prepare a new PR then.

lorencarvalho:
Hi @ambv, @JelleZijlstra & @vemel,

I am very interested in this change, as I have been running `black` against some large-ish codebases and I keep running into changes like this:

```
--- test.py	2020-02-21 21:06:16.454911 +0000
+++ test.py	2020-02-21 21:06:17.331524 +0000
@@ -1,13 +1,5 @@
 def __init__(
-    self,
-    an: bool,
-    arbitrarily: str,
-    long_: int,
-    list_: Any,
-    of: str,
-    arguments: bool,
-    *args: Any,
-    **kwargs: Any,
+    self, an: bool, arbitrarily: str, long_: int, list_: Any, of: str, arguments: bool, *args: Any, **kwargs: Any
 ) -> None:
     pass
```
Even if the longer version fits within the line length, I do not want black to collapse the list of arguments, and I don't expect it to since there is a trailing comma (I believe the trailing comma rule only applies to collections that are not nested). This PR appears to solve my problem, but the comments here indicate that it's too radical of a change:

> I noticed that there are significant changes in the formatting output for common cases. We generally avoid making such changes, so that codebases that are already Black-formatted don't see unnecessary change.

While I certainly understand this sentiment, I would argue that the current behavior is sub-optimal. Taking an intentionally broken apart collection (or call signature) and collapsing it seems to be optimizing for fewer vertical lines, but imo harms code readability and _especially_ harms diff readability. In addition, having distinct behavior around how black interprets trailing commas feels counter-intuitive (e.g. if black treats collections with trailing commas one way, why does it treat call signatures with trailing commas differently?). I was also going to mention the bug where trailing commas are preserved after collapsing call signatures, but I believe that has fixed.

Is the implication here that once a style decision has been merged into black's codebase, it can never be changed in order to avoid unexpected changes in existing codebases that are formatted with black? Are there any precedents? Perhaps this sort of thing can be gated behind a major or minor version increment?

Thanks for reading!

ambv:
This is resolved via other pull requests now. Thank you for the comprehensive contribution but we won't be able to merge this.

PR #1165, opened 2019-11-19
Suggest `extend-ignore` over `ignore` for flake8

asottile:
this option was introduced in flake8 3.7.x and is generally preferred over `ignore` (which unsets all default ignores)


asottile:
I guess technically with `extend-ignore` W503 doesn't need to be included (since it's part of the default ignore list)

asottile:
I've updated the patch to suggest a minimal flake8 configuration in addition to the one utilizing flake8-bugbear

zsol:
Sounds good to me, thanks!

PR #1172, opened 2019-11-21
ISSUE 727: optional typed_ast and PyPy support

vemel:
### Changes

- Do not install `typed_ast` automatically, only with `black[typed_ast]` extra requirement
- If `typed_ast` module does not exist - use `ast` as `ast27` and `ast3` as a fallback
- Move all AST node class lists to constants
- Handle `TypeError` for `ast3.parse` in case it is built-in `ast` from Python < 3.8
- Add type `AST` to cover all possible AST options
- Add `TargetVersion.arg_name` property to get argument name
- Add `TargetVersion.get_sys_version` method that gets target version for system info (fallback to py38)
- Add `TargetVersion.has_new_ast` property to check if `ast` is preferred over `typed_ast.ast3`
- Add `TargetVersion.title` property to get human readable python version tiltle
- Add `TargetVersion.get_ast_feature_version` method for Python 3.8+ `ast`
- Add `TargetVersion.get_typed_ast3_feature_version` method for `typed_ast.ast3`
- Add `mode` argument to `parse_ast` - it iterates over `mode.target_versions` from newest to oldest instead of using all versions, system version is used first. Fallback is `py34`, `py35`, `py36`, `py37`, and system version
- Handle `IndexError` in `read_cache` - appears only in `PyPy`

### CI changes
- Updated CI scripts to install `typed_ast` extra
- Added `pypy3` to Travis CI
- Added `py38` with no `typed_ast` to Travis CI

### Unit tests changes

- Added explicit `target_versions` to some test cases
- Removed `PY36_VERSIONS` usage (it is deprecated anyway)
- Added `not TYPED_AST` ignore marker to tests that require `typed_ast`
- Added `IS_NEW_AST` ignore marker to tests that require `typed_ast` or `py38+`
- Added `IS_PYPY` ignore marker to tests that have unicode symbols
- Added `TargetVersionTestCase`
- Added `BlackTestCase.test_lib2to3_parse_mocked` test case
- Added `BlackTestCase.test_lib2to3_parse_mocked` test case
- Added `BlackTestCase.test_parse_ast` test case
- Added `BlackTestCase.test_lib2to3_unparse` test case
- Added `BlackTestCase.test_patched_main` test case
- Added `BlackTestCase.test_patch_click` test case
- Added `BlackTestCase.test_cancel` test case

### Tested

- Python 3.8.0
- Python 3.7.5
- Python 3.6.9
- PyPy3.6 7.2.0

All of the above with no `typed_ast`. Unit tests with explicit target versions set are failing as expected.

### TODO

- probably we need a post-install script that tries to install `typed-ast` and fails silently
- `assert_equivalent` is probably broken for Python < 3.8 with no `typed_ast` as it relies on `ast.TypeIgnore`. However, I could not find any regressions


vemel:
All tests are passing for `PyPy3.6 7.2.0`. I marked some tests that contain unicode symbols to be skipped for `PyPy`. So, we can add it to CI. Unfortunately all `expression.py`-related tests cannot be executed, we should consider removing unicode from it.

black.py
```diff
@@ -96,7 +120,7 @@ class InvalidInput(ValueError):
     """Raised when input source code fails all parse attempts."""
 
 
-class WriteBack(Enum):
+class WriteBack(IntEnum):
```
JelleZijlstra: What is the motivation for this change? I generally prefer Enum to IntEnum where there are no compatibility constraints.

black.py
```diff
@@ -1871,24 +1973,24 @@ def visit_factor(self, node: Node) -> Iterator[Line]:
     def __post_init__(self) -> None:
         """You are in a twisty little maze of passages."""
         v = self.visit_stmt
-        Ø: Set[str] = set()
+        empty_set: Set[str] = set()
```
JelleZijlstra: Please avoid making changes unrelated to the intended effect of your PR.

black.py
```diff
@@ -3691,57 +3793,125 @@ def __str__(self) -> str:
...
+        versions = {
+            TargetVersion.PY34,
+            TargetVersion.PY35,
+            TargetVersion.PY36,
+            TargetVersion.PY37,
+            TargetVersion.get_sys_version(),
+        }
+
+    # try to parse with current version first
+    versions_sorted = list(sorted(versions, reverse=True))
```
JelleZijlstra: sorted() already returns a list. And why do you sort it again on the next line?

tests/test_black.py
```diff
@@ -1614,6 +1793,49 @@ def test_blackd_main(self) -> None:
                 raise result.exception
             self.assertEqual(result.exit_code, 0)
 
+    def test_lib2to3_unparse(self) -> None:
+        node_mock = MagicMock()
+        node_mock.__str__.return_value = "node_mock_str"
+        self.assertEqual(black.lib2to3_unparse(node_mock), "node_mock_str")
+
+    @patch("black.main")
```
JelleZijlstra: I don't think this test tests anything valuable, and again it's unrelated to the purpose of the PR.

vemel: Yes, I added some tests just to raise coverage a bit. I believe I covered all code I added, Black has really good code coverage, I just wanted to raise it a bit more.

vemel: Unrelated test cases:

- `BlackTestCase.test_lib2to3_unparse`
- `BlackTestCase.test_patched_main`
- `BlackTestCase.test_patch_click`
- `BlackTestCase.test_cancel`

black.py
vemel: It is to avoid explicit conversion if `TargetVersion.value` to `int`. For `enum.Enum` `value` could be `Any`, so `mypy` complains if you use it as `int`. As you see, I use `value` in `TargetValue.get_sys_version`. `IntEnum`  is a very simple class, equivalet to `class TargetVersion(int, enum.Enum`): ...

black.py
vemel: `test_black` was failing for `pypy3.6` as it does not support unicode symbols. Alternative is to skip this test for `IS_PYPY` flag.

black.py
vemel: You are right, conversion to lsit is not needed. And I added more meaningful comments.

```python
    # sort versions from the latest to the oldest
    versions_sorted = sorted(versions, reverse=True)

    # put sys version to the first place to parse source with it first
    versions_sorted.sort(key=lambda v: v != system_version)
```

vemel:
@JelleZijlstra I agree that this PR s over engineered. At the same time, changes to `parse_ast` are necessary and they require test updates. Changes to `TargetVersion` are not strictly required, but make life a bit easier. I moved AST class tuples to constants just beause I believe it would make sense to create a module that contains all `ast`/`typed_ast` related logic and exposes only `parse_ast`/`unparse_ast` and types.

New unit tests can be safely removed, as well as `IntEnum` usage.

Please let me know how I can improve my PR.

maxnordlund:
What is left to get this in? Would be really nice to be able to use black with pypy, it's so nice to get all code consistently pretty :)

vemel:
@maxnordlund pretty much everything is done, I will probably create a new PR to include only changes related to PyPy compatibility. There will still be a lot of changes and some of them are potentially dangerous, e.g. changing priority of python version usage, but it is required.

maxnordlund:
Thank you

ambv:
@vemel, thanks for working on this. In general, we can only agree to include PyPy support if it does not make the regular CPython's use case worse. The reality is such that almost all users of Black will be users through CPython.

diefans:
@ambv 
> The reality is such that almost all users of Black will be users through CPython.

I was just trying to use pypy as my neovim python host to test if it speeds up my ui experience, but came across this issue now. But for sure if you take the above fact as an argument, you will conserve the argument as a fact...

ambv:
What are you trying to say, @diefans?

diefans:
@ambv The status quo as an excuse for omission prevents evolution, change, adaptation etc. so
I wanted to say, that not fixing the this issue resp. enabling pypy will keep your argument/statement that only cpython is the target for black alive - it is like "we don't build cars, because most people go on food." It should not be taken offending, but the argument from a developers perspective is weak.

zsol:
> only cpython is the target

I believe his argument above is that we prioritize the cpython use case. So in practice that means any fixes for other platforms must make sure not to degrade the experience of the cpython use case.

cooperlees:
It seems consensus here that this change complicates matters to much for cpython. Please open a new PR to see if we can more safely port the code for PyPy in a simpler fashion. Thanks!

PR #1204, opened 2019-12-15
Set correct type annotations for constants from token.py

uriyyo:
Use `Final[int]` instead of `Final` type annotation for tokens.
It will say that the variable is final and has int type.


JelleZijlstra:
I don't think this is necessary: https://www.python.org/dev/peps/pep-0591/#syntax. Why do you think the `[int]` is needed?

uriyyo:
I am using PyCharm is IDE. Without `[int]` I see errors like this:

image

Looks like this is a bug of PyCharm, but for me, it's more clear to use `Final` with explicit type.
I just want to clarify and ask for your opinion.

JelleZijlstra:
I'd prefer not to add redundant information that the typing PEPs don't require.

uriyyo:
I will close this pull request and open bug for PyCharm.
Thank you for your response.

PR #1206, opened 2019-12-16
Add comment to flake8 configuration explaining line-length mismatch

MichaelAquilina:
We noticed this mismatch in configuration when the configuration to our own repo. Could not see a good reason why this would be any different

Updated to match pyproject.toml configuration https://github.com/psf/black/blob/master/pyproject.toml#L9

UPDATE: changed PR so that it explains why flake8 configuration has a different line length

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,7 @@
 [flake8]
 ignore = E203, E266, E501, W503
+# line length is intentionally set to 80 here because black uses Bugbear
+# See https://github.com/psf/black/blob/master/README.md#line-length for more details
 max-line-length = 80
 max-complexity = 18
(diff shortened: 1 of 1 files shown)


MichaelAquilina:
Not entirely sure if the failure from `Test / build (3.6, windows-latest) (pull_request) ` is related to my change?

hugovk:
This is intentional because Black uses Bugbear.

See https://github.com/psf/black/blob/master/README.md#line-length

MichaelAquilina:
ah right!

Considering how easy it is to miss this, might be worth adding a comment within the flake8 configuration instead?

hugovk:
Good idea!

Issue #1207, opened 2019-12-16
No module named 'regex._regex'; 'regex' is not a package


jonasrk:
Traceback (most recent call last):
  File "/usr/local/bin/black", line 5, in <module>
    from black import patched_main
  File "/usr/local/lib/python3.8/site-packages/black.py", line 15, in <module>
    import regex as re
  File "/usr/local/lib/python3.8/site-packages/regex.py", line 402, in <module>
    import _regex_core
  File "/usr/local/lib/python3.8/site-packages/_regex_core.py", line 21, in <module>
    import regex._regex as _regex
ModuleNotFoundError: No module named 'regex._regex'; 'regex' is not a package

daviddavis:
I'm hitting the same thing. I can reproduce this without black too just by opening a repl:

```
>>> import regex as re
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/pulp/lib64/python3.7/site-packages/regex.py", line 402, in <module>
    import _regex_core
  File "/usr/local/lib/pulp/lib64/python3.7/site-packages/_regex_core.py", line 21, in <module>
    import regex._regex as _regex
ModuleNotFoundError: No module named 'regex._regex'; 'regex' is not a package
```

jonasrk:
@daviddavis So it's the fault of https://pypi.org/project/regex/ ?

daviddavis:
Yea, if I use/pin regex to version 2019.11.1, everything works fine. So I think it's regex 2019.12.9.

bavedarnow:
Looks like they were doing some work recently that may have caused this: https://bitbucket.org/mrabarnett/mrab-regex/commits/all

mikegrima:
Seeing this now with moto: https://travis-ci.org/spulec/moto/jobs/625868515?utm_medium=notification&utm_source=github_status

TomFaulkner:
Will there be a release for this fix? Timeframe?

If not I can pin the requirement in my team's requirements file, just want to know whether I should or just wait a bit.

basnijholt:
Getting the same thing on [Azure](https://dev.azure.com/python-adaptive/adaptive/_build/results?buildId=1098&view=logs&jobId=08e0ed61-b128-5dfb-7579-151be2198b56&j=08e0ed61-b128-5dfb-7579-151be2198b56&t=7266cf22-1be8-5698-35ef-b3176c13f1f2).

I wonder what happened. Anyone knows?

[...]

PR #1208, opened 2019-12-16
Pin regex dependency

daviddavis:
Pin the regex dependency to the last known good release. I would use a y
release but it doesn't appear that regex is using sematic versioning.
Also, it appears that the 2019.12.9 release may have issues (although it
could just be improper packaging of the wheel). See related issue
https://github.com/psf/black/issues/1207.


JelleZijlstra:
Thanks! I decided to go with the alternative approach in #1208 instead.

PR #1209, opened 2019-12-16
Skip the broken version of regex

hugovk:
Fixes https://github.com/psf/black/issues/1207

Hopefully the next release will fix https://bitbucket.org/mrabarnett/mrab-regex/issues/349/no-module-named-regex_regex-regex-is-not-a.


JelleZijlstra:
Thanks! I decided to go with this approach instead of #1208 so that in the future we don't force users who install Black together with other packages into a specific version of `regex`.

PR #1210, opened 2019-12-19
Notify users of missing Python lazily

pganssle:
Currently this message shows up with no context prior to the start of Vim. By changing this to a lazy message, the user will only be notified of a problem with the Black plugin when they are attempting to use the Black plugin.

Another possible option would be to do what YCM does, and warn in the status bar *after* vim has started, but I figured this is the most targeted at the relevant users.


ambv:
Thanks! ✨ 🍰 ✨

PR #1215, opened 2019-12-28
black.vim: Allow user to suppress output from successful reformats

keggsmurph21:
This patch adds the ability for users to `let g:black_quiet = 1` to
suppress output on `Reformatted in ...` and `Already well formatted ...`
messages.


JelleZijlstra:
I don't use the vim plugin myself, but it looks like this may be unnecessary now because the plugin now uses the config file, where people can put `quiet = True`?

keggsmurph21:
Awesome!  That does what I need, thanks :^)

PR #1224, opened 2020-01-08
greedy creation of CACHE_DIR

AlJohri:
`pygram.initialize(CACHE_DIR)` has an explicit dependency on CACHE_DIR existing.

This code gets run the first time the library gets used and it tries to dump the cache files in the cache folder:
https://github.com/psf/black/blob/1ab87a3f67bc80a2e59be2692f2e2f25dd143af7/blib2to3/pgen2/driver.py#L194-L202

This fails since the `CACHE_DIR` only gets generated when the CLI is run for the first time

### DIFF
--- a/black.py
+++ b/black.py
@@ -60,7 +60,7 @@ DEFAULT_LINE_LENGTH = 88
 DEFAULT_EXCLUDES = r"/(\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|\.svn|_build|buck-out|build|dist)/"  # noqa: B950
 DEFAULT_INCLUDES = r"\.pyi?$"
 CACHE_DIR = Path(user_cache_dir("black", version=__version__))
-
+CACHE_DIR.mkdir(parents=True, exist_ok=True)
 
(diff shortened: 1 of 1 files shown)


AlJohri:
Alternatively we can move this code into `pgen2.driver.load_grammar` right before it tries to do `g.dump(gp)`. Let me know how you want to proceed.

JelleZijlstra:
I'd generally prefer if we minimize the side effects of just importing Black, because people may end up importing Black in applications that don't necessarily end up needing the cache directory.

AlJohri:
While I agree with the sentiment, it's already trying to write to disk on import. I don't know too much about the internals of black, but a decent compromise could be just adding the `mkdir` inside of that `if force...` block where it already writes to disk. I chose not to do that for this PR since I wasn't sure whether we wanted to edit a vendored dependency.

JelleZijlstra:
It's OK to edit blib2to3; we've changed it heavily already.

AlJohri:
Great, I will create a new PR with the updated `blib2to3`.

Issue #517, opened 2018-09-18
Non-prerelease black release


lig:
It looks like Black is stable enough at the moment to release it as `stable` to PyPI. There are plenty of tools and integrations already which rely on `black` package and the fact that it is still in `prerelease` state sometime causes inconveniences, i.e. using it with `pipenv` forces `allow_prereleases` flag to be set to `true`.

ambv:
This will happen soon, you're right. We'll be releasing one last (planned) beta, and then an RC of a stable release.

gforcada:
@ambv not to be pushy/disrespectful, but is that *soon*, actually soon? :sweat_smile:

ambv:
Right. Surely before Christmas, likely not in November though.

Aareon:
Looking forward to this as `pipenv` screams if you try to install without the `--pre` flag

Aareon:
I agree. Being able to streamline Black with other development tools is a huge plus. Black is currently a MUST in all of my projects.

lig:
@ambv Merry Christmas!:)

dimaqq:
Please? 🙏

dyve:
See also 
https://github.com/alanhamlett/pip-update-requirements/issues/23#issuecomment-456661501

This means we cannot update black in `requirements.txt` with pur.

A non-beta version would be much appreciated.

Utsav2:
Black at `HEAD` has a fix for issue 282, and we can't beta test it at Dropbox until there's a release with that issue fixed. Please consider creating a new release soon.

Peque:
@ambv Friendly ping. :innocent:

fernandezpablo85:
@ambv friendlier ping

max-wittig:
@ambv even friendlier ping :P

hugovk:
> Give the last beta a serious spin for the remainder of March! I'll tag the first stable release in early April.

https://twitter.com/llanga/status/1106247630370283523

dimaqq:
Cheers @hugovk , it does feel like formatting got pretty stable in `black`:
2 lines changed in 100KLOC codebase with the latest release :)

tribals:
Any changes on this?

zsol:
I hear @ambv loves making stable releases in April, so I have high hopes ;)

ambv:
FYI, we are sprinting on Black at PyCon now with the target stable release at some point right after.

max-wittig:
[...]

PR #1236, opened 2020-01-20
Remove deprecated  --py36 option

hugovk:
The `--py36` option was deprecated in Black 19.3b0 (https://github.com/psf/black/pull/750):

> deprecated `--py36` (use `--target-version=py36` instead) (#724)

https://github.com/psf/black#193b0

It's been a year since it was deprecated, and has been deprecated in two releases (19.3b0 and 19.10b0).

Is it now time to remove the deprecation?

At least I think it would be good to remove before the non-prerelease Black release (https://github.com/psf/black/issues/517), and whilst still in beta too.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
         additional_dependencies: [flake8-bugbear]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.740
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
I agree that it's time for this, but there's a complex merge conflict now (sorry for the wait!).

hugovk:
Merge conflict resolved!

PR #1247, opened 2020-01-26
Update README.md to include appropriate command to install Vim with Python 3 on macOS

morinokami:
The command currently listed in README.md to install Vim with Python 3 on macOS does not work since the `--with-python3` option is removed on [this commit](https://github.com/Homebrew/homebrew-core/commit/1a5a30703721b24198ef165059756ba35813511e). Python 3 support is [enabled by default](https://github.com/Homebrew/homebrew-core/blob/a64140c4640172fea60356137e1e04356c080d72/Formula/vim.rb#L51) on Vim via Homebrew as of 2020, so you don't have to specify the option now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -742,7 +742,7 @@ nnoremap <F9> :Black<CR>
 ```
 
 **How to get Vim with Python 3.6?** On Ubuntu 17.10 Vim comes with Python 3.6 by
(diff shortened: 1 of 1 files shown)


zsol:
Thanks!

PR #1266, opened 2020-02-07
Add option for printing a colored diff

### LINKED ISSUE #994: Coloured `--diff` output
There is an option in [diffutils](https://www.gnu.org/software/diffutils/manual/html_node/diff-Options.html):

> **--color [=_when_]**
> 
> Specify whether to use color for distinguishing different contexts, like header, added or removed lines. _when_ may be omitted, or one of:
>  - `none` Do not use color at all. This is the default when no `--color` option is specified.
>  - `auto` Use color only if standard output is a terminal.
>  - `always` Always use color. 
> 
> Specifying `--color` and no _when_ is equivalent to `--color=auto`.

It would be nice if Black had a similar option, probably with a different name.

dougthor42:
Fixes #994.

This adds a `--color` option to the black command line which adds very basic color support when printing diffs.

+ The coloring mimics `git diff` because why not. Headers are Bold White, line markers Cyan, subtractions Red, and additions Green.
+ There is no change to the default behavior.
+ `--color` has no effect if `--diff` is not supplied.

## Example:

image

## Discussion

+ As is, this will not work with Windows CMD or PowerShell. In order to enable Windows support, the `colorama` package should be included. I'm not sure how you feel about adding packages all willy-nilly, so I figured the first version of this PR would not include it.
+ I opted to add `COLOR_DIFF` to `WriteBack` because that meant I didn't have to change any call signatures.
+ An initial version used an enum for `--color` which let the user choose between `always`, `none`, and `auto`. This is still doable, but will take more effort. Personally I don't much see the need for it - using a boolean is sufficient in my opinion. That said, if you'd like me to pursue it just let me know.

### DIFF
--- a/black.py
+++ b/black.py
@@ -59,6 +59,11 @@ from blib2to3.pgen2.parse import ParseError
 
 from _black_version import version as __version__
 
+# This is needed for MyPy static type analysis.
+# See https://stackoverflow.com/a/38962160/1354930
+if False:
+    import colorama  # noqa: F401
+
 DEFAULT_LINE_LENGTH = 88
 DEFAULT_EXCLUDES = r"/(\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|\.svn|_build|buck-out|build|dist)/"  # noqa: B950
 DEFAULT_INCLUDES = r"\.pyi?$"
@@ -746,6 +751,7 @@ def format_file_in_place(
                 newline=newline,
                 write_through=True,
             )
+            f = wrap_stream_for_windows(f)
             f.write(diff_contents)
             f.detach()
 
@@ -768,6 +774,33 @@ def color_diff(contents: str) -> str:
             line = "\033[31m" + line + "\033[0m"  # red, reset
         lines[i] = line
     return "\n".join(lines)
+
+
+def wrap_stream_for_windows(
+    f: io.TextIOWrapper,
+) -> Union[io.TextIOWrapper, "colorama.AnsiToWin32.AnsiToWin32"]:
+    """
+    Wrap the stream in colorama's wrap_stream so colors are shown on Windows.
+
+    If `colorama` is not found, then no change is made. If `colorama` does
+    exist, then it handles the logic to determine whether or not to change
+    things.
+    """
+    try:
+        from colorama import initialise
+
+        # These are the defaults for colorama.init()
+        f = initialise.wrap_stream(
+            f, convert=None, strip=None, autoreset=False, wrap=True
+        )
+
+        # wrap_stream returns a `colorama.AnsiToWin32.AnsiToWin32` object
+        # which does not have a `detach()` method. So we fake one.
+        f.detach = lambda *args, **kwargs: None  # type: ignore
+    except ImportError:
+        pass
+
+    return f
 
 
 def format_stdin_to_stdout(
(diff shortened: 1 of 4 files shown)


dougthor42:
Whoops, wasn't working off the latest master. Fixed and rebased.

zsol:
FWIW I like this feature. How about dropping the command line flag and always outputting colored diff if we're writing to a terminal (`sys.stdout.isatty()`)?

As for Windows support, I'd be fine with adding `colorama` as an extra (`extras_require` in `setup.py`)

dougthor42:
> dropping the command line flag

I think it would be prudent to keep *some* CLI flag because I'm sure there are cases where a user wants to write to a terminal but not colorize it. Perhaps we just invert it: `--no-color`.

> if we're writing to a terminal (`sys.stdout.isatty()`)

From what I can tell, whenever `WriteBack.DIFF` is used, black is already writing to `io.TextIOWrapper(sys.stdout.buffer, ...)`. Do I even need to check `isatty()`?

black.py
```diff
@@ -657,25 +674,73 @@ def format_file_in_place(
...
+
+        # We set `strip=False` so that we can don't have to modify
+        # test_express_diff_with_color.
+        f = initialise.wrap_stream(
+            f, convert=None, strip=False, autoreset=False, wrap=True
+        )
+
+        # wrap_stream returns a `colorama.AnsiToWin32.AnsiToWin32` object
+        # which does not have a `detach()` method. So we fake one.
+        f.detach = lambda *args, **kwargs: None  # type: ignore
```
dougthor42: I don't like this, but I don't fully understand what `detach` does or why it's needed, so this is the best I could come up with.

black.py
```diff
@@ -56,6 +56,11 @@
 
 from _black_version import version as __version__
 
+# This is needed for MyPy static type analysis.
+# See https://stackoverflow.com/a/38962160/1354930
+if False:
```
zsol, line 61: nit: using `TYPE_CHECKING` is a bit more expressive, so you don't need the above comment.
```suggestion
if typing.TYPE_CHECKING:
```

dougthor42: Ah, good to know! I'll keep that in mind for the future, thanks!

jshin49:
Seems this PR is closed and merged into master. Yet, when I try to use the ```--color``` option, I get an error saying there is no such option. The version I am using is 19.10b0 on python 3.7.7 and python 3.6.10

hugovk:
Yep, it's not yet been released:

This was merged into master in May 2020.

Black 19.10b0 was released in October 2019 (19.10 is yy.mm, see https://calver.org/).

jshin49:
Ah I see.

When is the next release date?

hugovk:
I think it's coming sooner rather than later!

https://github.com/psf/black/issues/517 is the tracking issue, and see https://github.com/psf/black/issues/517#issuecomment-578135675 and https://github.com/psf/black/labels/stable.

cooperlees:
GitHub master or even a branch is basically released these days. Introducing your new friend if you're in a hurry:

pip install git+https://github.com/psf/black.git@master

```console
cooper-mbp1:~ cooper$ /tmp/tb/bin/black --help | grep color
  --color / --no-color            Show colored diff. Only applies when
```

jshin49:
Awesome. Thanks for the help @cooperlees. I'm trying to set up a workflow for a new team project  and I was comparing between Pylint vs Black vs Yapf. Not sure which is the best, but the color coded diffs of Black would definitely make a lot of difference.

PR #1267, opened 2020-02-09
Simplify Line.contains_multiline_strings method

uriyyo:
Simplify Line.contains_multiline_strings method by using `all` function and generator expression instead of for loop.

Replace list appending in loop by using string multiplication.


zsol:
Thanks!

PR #1268, opened 2020-02-10
Include an easier way to integrate black with Pycharm

arnavb:
According to the [documentation for external tools](https://www.jetbrains.com/help/pycharm/configuring-third-party-tools.html), you can use the macro `$PyInterpreterDirectory$` to get the base directory for external tools. This makes it easier to add `black` as a file-watcher.


zsol:
Thanks!

Issue #1255, opened 2020-01-31
No emojis


Makogan:
Emojis, among other things, are not universally supported, are not consistent across systems, and require additional formatting on the part of the terminal. Having support for UTF is great, but forcing non ascii characters onto users for something as frivolous as emojis is perhaps better avoided.

hugovk:
See https://github.com/psf/black/issues/223 and https://github.com/psf/black/issues/1031.

zsol:
You can use `--quiet` to disable non-essential output.

Makogan:
If people are repeatedly requesting to remove a feature, it is, perhaps, symptomatic that that feature should not be your default.

zsol:
Look, nobody is forcing anything onto users. My read on this situation is that the vast majority of people using Black are okay with the emojis being part of the output by default. My opinion is that terminals incapable of displaying these characters are broken, not the tools that output them. Your opinion might be different, and that's completely fine.

Issue #1031, opened 2019-09-24
No-emoji switch for better output


im-n1:
It would be really nice to have a switch which can disable emojis at all but not any other piece of Black output. The reason is emojis are graphical stuff which a lot of terminal emulators are not capable of not speking of pure console (framebuffer). I'm not hating emojis but I think they should be optional.

 An example output from my urxvt:

![image](https://user-images.githubusercontent.com/198654/65497606-cdd3bd80-deba-11e9-85be-2150d7193461.png)

`--quiet` doesn't work here quite good (based on this https://github.com/psf/black/issues/223)

zsol:
Those emojis are unicode characters, which all modern terminals are capable of displaying, provided they're using a decent font.

Try using a font that can display these. There are plenty of free & paid options or you can [patch your own](https://github.com/ryanoasis/nerd-fonts)

im-n1:
Do you think app output should be readable only in certain terminals with certain fonts?

fiendish:
I think you should modernize your terminal. Unicode has been the standard for text in Python for 11 years now.

im-n1:
Try to seek how graphics work in terminals. Maybe you will consider your opinion.

Anyway thank you all for your opinions. I'm out.

Issue #223, opened 2018-05-18
Allow to disable emoji in output


giuscri:
As an environment variable (e.g. `BLACK_HIDE_EMOJIS`) or as an option.

JelleZijlstra:
What's your motivation?

giuscri:
emoji support can be problematic on some platform. https://debbugs.gnu.org/cgi/bugreport.cgi?bug=30045

The reason why I noticed you can't toggle emojis is that my terminal crashed when using `black .` Currently I solved by sending stderr of black to devnull.

ambv:
This sounds like something that should be fixed or worked around in your terminal, not in end-user tools that use Unicode characters. I happen to have a non-Latin 1 character in my first name so [I'm rather sensitive to bad Unicode support](https://www.youtube.com/watch?v=7m5JA3XaZ4k&).

giuscri:
I see your point. Also, I didn't noticed there's a `--quiet` option in black. Closing this.

waldronfaulkner:
Opinion: there should be an option to disable emoji output. There are a lot of people—_old_ or not—who feel emojis are dumb and unprofessional and just lower the tone of everyday work when they're spattered all across our consoles

ArsenArsen:
Hi! Some of my co-developers raised concerns about emojis looking out of place in CI output, so I'd like to propose a `BLACK_NO_FUN` environment variable that allows for them to be disabled on site, if the user hates fun.

I see that it's been proposed to use `--quiet`, the issue with that, though, is that it does not show other information too, so it's not worth it.

PS: I'd be fine with contributing this myself, but I'd like to get prior discussion with the maintainers.

Thanks in advance.

jap:
I just ran across some code making use of emojis, all-caps words and coloring in a script that calls black and some other formatters and linters. 
The reason was to make sure it stands out more than black's output. This is just a race to the bottom for attention :(

TimZaman:
unicode support is fine, but just like jap said - there is no need to hijack attention what is otherwise a calming terminal experience.

draeath:
I'd like to see this as well. Gives it a non-professional feel, which is unfair I know, but there it is.

PR #1270, opened 2020-02-11
Add -n flag to not print emojis

austenadler:
**Is your feature request related to a problem? Please describe.** I want to be able to disable emojis from being printed on success/failure

**Describe the solution you'd like** Either emojis removed entirely, or an option that disables them (this merge request implements the latter in `-n/--no-emojis`)

**Describe alternatives you've considered** Running something like `black ... | sed 's/[^[:alnum:]]//g'` but this is bad design for obvious reasons. Also considered looking at any translation implementations (no emoji could be a 'dialect' of English), but I don't know how to do this in python.

### DIFF
--- a/black.py
+++ b/black.py
@@ -61,12 +61,7 @@ DEFAULT_EXCLUDES = r"/(\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|\.svn|_
 DEFAULT_INCLUDES = r"\.pyi?$"
 CACHE_DIR = Path(user_cache_dir("black", version=__version__))
 
-# Emojis for output strings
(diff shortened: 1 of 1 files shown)


austenadler:
Will add documentation for this. Should I write a test too?

hugovk:
Duplicate of #223, #1031 and #1255.

ambv:
This sounds like something that should be fixed or worked around in your terminal, not in end-user tools that use Unicode characters. I happen to have a non-Latin 1 character in my first name so [I'm rather sensitive to bad Unicode support](https://www.youtube.com/watch?v=7m5JA3XaZ4k&).

PR #1271, opened 2020-02-11
string prefixes: don't normalise capital R-strings, normalise order

### LINKED ISSUE #1244: Preserve r-string prefix capitalization
The [MagicPython](https://github.com/MagicStack/MagicPython) syntax highlighter (and VS Code, Sublime Text and Atom which all use it) interprets upper (`R"foo"`) and lower case r-strings (`r"bar"`) differently. The reason for this is explained in their readme:

> Raw strings are often interpreted as regular expressions. This is a bit of a problem, because depending on the application this may actually not be the most common case. Raw strings can simply be the input to some other processor, in which case regexp-specific highlighting is really hindering the overall readability. MagicPython follows a convention that a lower-case r prefix means a regexp string, but an upper-case R prefix means just a raw string with no special regexp semantics. This convention holds true for all of the legal combinations of prefixes. As always the syntax is biased towards Python 3, thus it will mark Python-2-only prefixes (i.e. variations of ur) as deprecated.

This is a neat feature, especially in cases with more involved string interpolation which are hard to read and easy to mess up without proper highlighting. As a simplified example:

```py
import matplotlib.pyplot as plt

plt.title(R"$y_\mathrm{true}$ vs $y_\mathrm{pred}$")
plt.title(r"$y_\mathrm{true}$ vs $y_\mathrm{pred}$")
```

Note that even GitHub syntax highlighting has this feature. This is what it looks like in VS Code:

upper
lower

Unfortunately, this feature becomes unusable in conjunction with `black` since `black` automatically lower cases all r-strings.

Either add an option to `pyproject.toml` to disable automatic lower casing and/or change `black`'s default behavior to preserve string prefix casing. I prefer the latter.

**Additional context**

This behavior was introduced in #198, apparently without being aware that some syntax highlighters treat the cases differently.

hauntsaninja:
Reverse sort since that causes fewer changes on black's codebase.

Resolves #1244


ambv:
Thanks! ✨ 🍰 ✨

PR #1273, opened 2020-02-14
Teach the Vim plugin to respect pyproject.toml (issue 414)

### LINKED ISSUE #414: Vim plugin should respect pyproject.toml config

TalAmuyal:
Disclaimer: This PR is a continuation of https://github.com/psf/black/pull/589, made by @pappasam, that should address: https://github.com/psf/black/issues/414.

> ...
> Feel free to close this pull request, or use it as a jumping off point in case you're interested in its content. ...

This pull request makes the Vim plugin respect `pyproject.toml` (resolves #414).


TalAmuyal:
Regarding tests, I'd like to add some, but currently there is no infrastructure for testing the plugin.
There is some python code that can be moved the a `.py` file (which can be easily tested) but that requires quite a bit of refactoring (tried it locally).

Are there any suggestions in this regard?

Aside from that, I think the PR is ready for review.

keith:
I'm seeing this locally with this change:

```
Error detected while processing function provider#python3#Call:
line   18:
Error invoking 'python_execute' on channel 3 (python3-script-host):
  File "<string>", line 92, in Black
  File "<string>", line 128, in get_configs
AttributeError: module 'black' has no attribute 'find_pyproject_toml'
```

Any ideas? I can open another issue if needed

j08lue:
> Any ideas? I can open another issue if needed

You need to install black from sources instead of from PyPI. `pip install --upgrade https://github.com/psf/black.git` (make sure to install in the right venv).

Would be nice if these kinds of breaks could be avoided in the future, although I am not sure how. Problem is that `Plug` or `Vundle` pulls the black vim plugin from current master but black gets installed only for releases on pip. Perhaps you should point `Vundle` or whatever at a tagged release, if possible.

erikr:
Did this PR enable anyone's Vim plugin to respect `pyproject.toml`?

My `pyproject.toml` contains:
```
[tool.black]
line-length=79
```
If I call Black from the command line, the specified line length is used as desired.

However, if I call Black within Vim, it uses the default line length.

If I change line length in `~/.vimrc` via `let g:black_linelength = 79`, the specified line length is used as desired:

I installed Black for Vim using Vundle.

To address `AttributeError: module 'black' has no attribute 'find_pyproject_toml'`, I went to `~/.vim/bundle/black` and checked out the pypi release version: `git checkout 19.10b0` (see https://github.com/psf/black/issues/1304).

Any advice is much appreciated!

sbellem:
Adding to https://github.com/psf/black/pull/1273#issuecomment-594114067:

> ```shell
> source ~/.vim/black/bin/activate  # make sure to install in the right venv
> pip install --upgrade git+https://github.com/psf/black.git
> ```

If you are using `neovim`, the default location of the virtual environment is:

```shell
~/.local/share/nvim/black
```

henryiii:
If you need to pin Plug in Vim while waiting for a new release:

```vim
" Support Black code formatting
Plug 'psf/black', { 'tag': '19.10b0' }
```

PR #1279, opened 2020-02-19
Example for using pre-commit args configuration

benjaoming:
I don't think it's sound advice to ask people to avoid using a configuration specifically targeting the pre-commit invocation.

For instance, pre-commit modifies files. Other invocations of black will use `--check`.

What would you think about adding an example for an expanded configuration of pre-commit?

CC: @asottile @ambv 

Refs https://github.com/psf/black/commit/489d00ed8f4fb90d5788609a474632ab5a16591f#diff-04c6e90faac2675aa89e2176d2eec7d8R733-R740

### DIFF
--- a/README.md
+++ b/README.md
@@ -898,15 +898,40 @@ repos:
 
 Then run `pre-commit install` and you're ready to go.
 
-Avoid using `args` in the hook. Instead, store necessary configuration in
-`pyproject.toml` so that editors and command-line usage of Black all behave consistently
-for your project. See _Black_'s own
+### Configuring pre-commit
+
+If you want a project-wide setting, you can store the necessary configuration in
+`pyproject.toml` so that pre-commit, editors and command-line invocation of Black all
+behave consistently for your project. See _Black_'s own
 [pyproject.toml](https://github.com/psf/black/blob/master/pyproject.toml) for an
 example.
+
+You can use `args` in the hook to separately customize how pre-commit should invoke
+_Black_.
 
 If you're already using Python 3.7, switch the `language_version` accordingly. Finally,
 `stable` is a tag that is pinned to the latest release on PyPI. If you'd rather run on
 master, this is also an option.
+
+`pre-commit` has
+[settings for all hooks](https://pre-commit.com/#adding-pre-commit-plugins-to-your-project)
+and [settings for each hook](https://pre-commit.com/#pre-commit-configyaml---hooks).
+
+```yaml
+repos:
+  - repo: https://github.com/psf/black
+    rev: stable
+    hooks:
+      - id: black
+        # Use Python 3.7
+        language_version: python3.7
+        # Customize options for black command
+        args: [--arg-for-black, "value", --flag-for-black]
+        # Include only these files
+        files: "^src/.*"
+        # Exclude patterns
+        exclude: "tests/.*$"
+```
 
 ## Ignoring unmodified files
 


README.md
```diff
@@ -893,12 +893,14 @@ repos:
...
     hooks:
       - id: black
+        args: [--exclude, '/(\.git|\.tox|\.venv|build|static|dist|node_modules)/']
         language_version: python3.6
 ```
 
 Then run `pre-commit install` and you're ready to go.
 
-Avoid using `args` in the hook. Instead, store necessary configuration in
+You can use `args` in the hook to customize how pre-commit should invoke black. If you
```
hugovk: For consistency:
```suggestion
You can use `args` in the hook to customize how pre-commit should invoke _Black_. If you
```

README.md
```diff
@@ -893,12 +893,14 @@ repos:
     rev: stable
     hooks:
       - id: black
+        args: [--exclude, '/(\.git|\.tox|\.venv|build|static|dist|node_modules)/']
```
asottile: well for one this arg is completely unnecessary -- pre-commit only runs on checked in files
for two, this should not be in the default configuration -- people copy paste this blindly and we certainly don't want to cargo-cult this everywhere

benjaoming: That's a good point, thanks! Will come up with an alternative.

It's not completely unnecessary though... I initially wrote this because I was writing a pre-commit based Travis CI lint job. It invokes `pre-commit run --all-files`, but without adding a pyproject.toml configuration for the sole purpose of specifying an excluded set of files.

Other people may have other needs, such as specifying a target folder `src/`.

I'll try adding this as a separate suggestion.

asottile: those wouldn't be via args, that would be `files:` or `exclude:`

README.md
benjaoming: Thanks, adding that!

README.md
benjaoming: Roger, didn't know pre-commit had those global options.

asottile: the docs are pretty thorough in showing all the options: https://pre-commit.com/#adding-pre-commit-plugins-to-your-project

benjaoming: Thanks for the advice, I have now referenced the pre-commit docs.

Re: "thorough". It doesn't matter how thorough another project's docs are if the reader doesn't get the idea to look there ;)

README.md
```diff
@@ -898,16 +898,41 @@ repos:
...
 example.
 
+You can use `args` in the hook to separately customize how pre-commit should invoke
+_Black_.
+
 If you're already using Python 3.7, switch the `language_version` accordingly. Finally,
 `stable` is a tag that is pinned to the latest release on PyPI. If you'd rather run on
 master, this is also an option.
 
+Pre-commit has
```
asottile: `pre-commit` is never capitalized

README.md
```diff
@@ -898,16 +898,41 @@ repos:
...
+  - repo: https://github.com/psf/black
+    rev: stable
+    hooks:
+      - id: black
+        # Use Python 3.7
+        language_version: python3.7
+        # Customize options for black command
+        args: [--arg-for-black, "value", --flag-for-black]
+        # Include only these files
+        files: ['src/']
```
asottile: this is a regex not a list

README.md
```diff
@@ -898,16 +898,41 @@ repos:
...
+        # Exclude patterns
+        exclude: ['tests/*']
```
asottile: this is a regex not a list

benjaoming:
Alright, I expanded the example of using a customized pre-commit configuration and added links to upstream pre-commit docs for these parts.

To me, it makes sense as to where I got lost.

README.md
benjaoming: :+1:

README.md
benjaoming: Right, okay, didn't know that

benjaoming: Thanks

benjaoming:
> This PR also changes the original author's intent which is to strongly discourage args which seems not ok

Why strongly discourage it when there can be a need to have a separate configuration for pre-commit?

Anyways, the original authors words are now firstly mentioned, and then a separate section is present about configuration of pre-commit. That's what I was missing when I firstly read it.

ichard26:
Closing since 1) this has a merge conflict, 2) has had zero activity in a year, 3) may or may not be correct / a good idea, and 4) is a draft PR.

ichard26:
@benjaoming Apologies for the lackluster PR experience. I want to make it clear that the contributing experience here is important here. To be honest, I did not handle this particularly well. I should've pinged you checking in on whether you wanted to continue with this PR or not. 

If you have any more feedback for our PR process, I'd love to hear them. Thanks for understanding!

benjaoming:
It's really okay, I don't know how to identify the exact issue here, I mean the feedback from @asottile was good and it just seemed that the PR was left in a state where it needed more feedback from maintainers in order to progress.

I can solve the merge conflict if someone wants.

For me, this wasn't about solving my own issue -- documentation never is -- it was about helping others who could run into the same questions.

Issue #1290, opened 2020-03-03
[Bug bounty] Black formatting gallery for chosen PyPI packages


ambv:
## Bounty amount
$500 paid on merge to an agreed PayPal account.

## Motivation
We need a way to easily see what formatting would Black apply to given PyPI projects. This is especially important to compare formatting changes between two versions of Black.

### Docker image
Create a Docker image running with Python 3.8 which accepts the following arguments:
- `-p`: name of the PyPI source package to download;
- optionally `-v`: version of the PyPI source package to download; 
- version of Black to apply formatting with;
- optionally: `pyproject.toml` for Black to apply formatting with.

There can be multiple pairs of Black version and configs. If there is more than one, the second formatting should be applied over the first one.

### Result
Running the Docker image will produce a sub-directory in the output directory named $PROJECT-$VERSION. It will be a Git repository with the following branches:
- `master`: a commit with the original contents of the downloaded PyPI source package;
- `black-$N`: a commit with `master` reformatted with Black version $N;
- `black-$N+1`: a commit with `black-$N` reformatted with Black version $N+1;
- `black-$N+2`: a commit with `black-$N+1` reformatted with Black version $N+2;
- and so on.

### Example invocations of the tool
* Django 3.0.3, three versions of Black, no pyproject.toml
```sh
$ docker run -it -v /host/directory/for/output:/output black_gallery:latest -p Django -v 3.0.3 19.3b0 19.10b0 master
```
* latest Twisted, two versions, with pyproject.toml
```sh
$ docker run -it -v /host/directory/for/output:/output black_gallery:latest -p Twisted 19.10b0:pyproject-19.10b0.toml master:pyproject-master.toml
```
Note: no `-v` specified, would download "19.10.0" at the time of writing this issue (3.3.2020).

### Support for CPython standard library tests and PyPy
On top of PyPI packages, the tool needs to accept the following special values of `-p` and `-v`:
* when `-p` is "cpython", the project to download is "https://github.com/python/cpython/archive/VERSION.zip" where VERSION is specified with `-v` (for example `-v v3.9.0a4` or `-v master`);
* when `-p` is "pypy", the project to download is "https://foss.heptapod.net/pypy/pypy/repository/VERSION/archive.tar.bz2" where VERSION is specified with `-v` (for example: `-v release-pypy2.7-v7.3.0rc4` or `-v branch/default`).

### Logs
The tool should log output verbosely to stdout.

### Source code location
[...]

PR #1291, opened 2020-03-03
WIP: Black formatting gallery for chosen PyPI packages

soulshake:
Addresses https://github.com/psf/black/issues/1290

- [x] Specific version of target package (`-v`)
- [x]  Multiple versions of Black from pip
- [x] Black master branch
- [x] pyproject.toml
- [x] `-p` cypthon vs pypy

### DIFF
--- a/gallery/entrypoint.sh
+++ b/gallery/entrypoint.sh
@@ -1,55 +1,43 @@
 #!/bin/bash
-# Usage:
-# docker build -t black_gallery .
-# docker run -it \
-#    -v /host/directory/for/output:/output black_gallery:latest \
-#    -p Twisted \
-#    19.10b0:pyproject-19.10b0.toml \
-#    master:pyproject-master.toml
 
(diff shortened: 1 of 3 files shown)


gallery/entrypoint.sh
```diff
@@ -0,0 +1,166 @@
...
+        if [ ${#PAIR[@]} -gt 1 ]; then
+            bv=${PAIR[0]}
+            toml=${PAIR[1]}
+            echo "black version: ${bv}"
+            echo "requested toml: ${toml}"
+        fi
+        IFS=' '
+
+        config=$(get_config_str "${toml}")
+        init_git "${TARGET}"
```
ambv: We want one Git repository for all branches. This way we will be able to easily compare branches with `git diff` to see how different versions of Black format the same code.

ambv:
Thanks for participating, we went with #1294.

PR #1292, opened 2020-03-03
Draft version of gallery

imomaliev:
Draft implementation of #1290 

TODO:

- pyproject.toml support
- multiple black versions support(need a some way to figure out versions ordering if passed in not sorted order)
- cleanup(we can install package instead of downloading it)
- test coverage
- maybe use https://github.com/gitpython-developers/GitPython to work with git
- maybe use internal pip functionality to download packages


ambv:
Thanks for participating. We went with #1294.

PR #1294, opened 2020-03-03
Implement Black Version Gallery

isidentical:
This patch implements gallery functionality (described in #1290) to black.

### DIFF
--- a/gallery/README.md
+++ b/gallery/README.md
@@ -1,0 +1,38 @@
+# Gallery
+
+Gallery is a script that automates process of applying different black versions to a
+selected PyPI package and seeing the results between black versions.
+
+## Build
+
+```console
+$ [sudo] docker build -t black_gallery .
+```
+
+## Run
+
+```console
+$ sudo docker run -it -v /host/output:/output -v /host/input:/input black_gallery:latest [args]
+```
+
+```
+usage: gallery.py [-h] -p PYPI_PACKAGE -b BLACK_REPO [-v VERSION] [-i INPUT] [-o OUTPUT] [versions ...]
+
+Black Gallery is a script that automates process of applying different black versions to a selected PyPI package and seeing the results between versions.
+
+positional arguments:
+  versions
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -p PYPI_PACKAGE, --pypi-package PYPI_PACKAGE
+                        PyPI package to download.
+  -b BLACK_REPO, --black-repo BLACK_REPO
+                        Black's git repository.
+  -v VERSION, --version VERSION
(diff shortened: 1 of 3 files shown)


gallery/gallery.py
```diff
@@ -0,0 +1,205 @@
...
+
+def format_repo_with_version(
+    repo: Path,
+    from_branch: Optional[str],
+    black_repo: Path,
+    black_version: BlackVersion,
+    input_directory: Path,
+) -> str:
+    current_branch = f"black-{black_version.version}"
+    git_switch_branch(black_version.version, repo=black_repo)
```
ambv: It would be better to have `pip install -e .` after switching branches. Different branches are bound to have different dependencies. So, ideally, for reproducibility, we'd use different virtual environments for each install, too.

gallery/gallery.py
```diff
@@ -0,0 +1,205 @@
...
+
+ArchiveKind = Union[tarfile.TarFile, zipfile.ZipFile]
+
+
+class BlackVersion(NamedTuple):
+    version: str
+    config: Optional[str] = None
+
+
+def get_pypi_download_url(package: str, version: Optional[str]) -> str:
```
ambv: This is clever. Any idea if this is better than `pip download --no-binary --no-deps`?

ambv:
One more thing. Since you're not checking the result of `subprocess.run()` calls, they should at least be made with `check=True` so the program stops right away if something's wrong instead of plowing through and possibly causing more issues.

gallery/gallery.py
isidentical: I was thinking about extending this with options like `--format-pypi-top n`, and for doing that programatically I implemented this in a form of function.

gallery/README.md
```diff
@@ -0,0 +1,38 @@
+# Gallery
+
+Gallery is a script that automates process of applying different black versions to a
+selected PyPI package and seeing the results between black versions.
+
+## Build
+
+```console
+$ [sudo] docker build -t black_gallery .
```
imomaliev: I suggest not putting `sudo` in the example, this is bad practice to run docker under root. https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user

gallery/README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+## Run
+
+```console
+$ sudo docker run -it -v /host/output:/output -v /host/input:/input black_gallery:latest [args]
```
imomaliev: I suggest not putting `sudo` in the example, this is bad practice to run docker under root. https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user

PR #1309, opened 2020-03-16
Prevent the vim plugin from importing system's black

ProgVal:
This commit adds Black's venv before the system paths, so the plugin
imports from the venv even if Black is already installed elsewhere on the system.


JelleZijlstra:
This change was merged as #1384.

PR #1319, opened 2020-03-27
Allow to pass the FileMode options in the vim plugin

shaoran:
The vim plugin takes now into consideration the `target-version` value of
the `pyproject.toml` file.

You can also specify a default target version with the configuration
variable `g:black_target_version`.

Vim's `:Black` command accepts now the following optional arguments:

- `line_length`
- `string_normalization`
- `black_target_version`

Those allow you to override any settings that come either from vim's
configuration or the `pyproject.toml` file.

It also supports supports a simple tab completion to allow to pass the
arguments quicky.


JelleZijlstra:
Sorry for leaving this waiting for so long! There are a few merge conflicts now. Do we still need this change?

shaoran:
Hi

> Sorry for leaving this waiting for so long! There are a few merge conflicts now. Do we still need this change?

I'd say yes, because `:Black` has no options and if you need to pass some command line option like setting python's target version or the line length, you cannot do it directly from vim, you have to do that from the console. In my eyes, that defeats the purpose of having black inside vim, if I have to open a console anyway.

This PR would allow you to do the following in vim: `:Black target_version="py27"`. And because it's something I need all the time, I have this in my vim config:

```vim
Plug 'shaoran/black', { 'branch': 'allow_override_args_19.10b0' }
```

It's so long since I've created this PR that my changed might now work anymore, I'd have to check them, but I'll have time to do it at the weekend.

felix-hilden:
If I may, I'll provide a review posted on our Discord in these messages: [1](https://discord.com/channels/267624335836053506/846434317021741086/852765208942084126), [2](https://discord.com/channels/267624335836053506/846434317021741086/853321176343511071) by Godlygeek:

The core issue seems to be whether or not we'd like to be able to pass arguments as comma-separated `key=value` pairs. Some concerns were that:
- The format is unusual: ordinarily the arguments look more like CLI flags, being space-separated
- Passing varying configuration is an odd design choice since Black is all about consistent formatting, which can be achieved with file configuration or setting `g_*` variables in Vim

Without knowing too much about the plugin system, I do share these concerns. Target version is of course one argument whose value is naturally varied. Could you comment a bit on your view on the matter, @shaoran? Much appreciated!

shaoran:
@felix-hilden I'm sorry, I have been very busy this week and was not able to reply. I cannot at the moment, but soon I will answer your questions.

felix-hilden:
Hi @shaoran, any updates on this? It'd be nice to bring this issue closer to the finish!

shaoran:
@felix-hilden I'm so sorry, I've forgotten about this. I'll try to take a look tomorrow or during the next week.

shaoran:
@felix-hilden I tried merging `upstream/main` in my branch and I got a conflict in `README.md`. Back in `19.10b0` (from which
I originally branched of) the documentation was directly in the `README.md`
file. Now it's been moved to https://black.readthedocs.io

Is the documentation source code for readthedocs still in the repository?

shaoran:
It changed a lot, I basically would have to rewrite the whole thing again. OK, I'll try that.

felix-hilden:
Yep, the documentation source is in the `docs/` folder, particularly `docs/integrations` might be good for this.

shaoran:
Ok. this will take me a little bit longer. Since there are so many changes since I created the pull request, all my changes are useless and basically have to write them again. I hope, tomorrow I can push a new version.

felix-hilden:
No worries! And I'd also like to remind you about the review points I left above 😃

shaoran:
@felix-hilden sorry that it took a while, but I've been very busy at work.

I try to address the concerns here

> - The format is unusual: ordinarily the arguments look more like CLI flags, being space-separated

I made a change so that arguments are space-separated.

> - Passing varying configuration is an odd design choice since Black is all about consistent formatting, which can be achieved with file configuration or setting g_* variables in Vim

Yes, that is true, that's why I removed passing the other options like `fast`, `linelength`, etc. The only option remaining is `target_version` and that because of convenience.

I maintain a python library that targets multiple python version, some modules only work with Py27, some with Py36, etc. Very often I have those files open at the same time and I'd like to format one with py27 as target and the other with py36 as target **without** having to set up variables and special rules in my config. So that's why it is for me more convenient to just pass `target_version=py27` directly to `:Black` when I need it. And thanks to tab completion, I don't even have to type it, just press TAB.

Like I said before, I already use this on my environment and has been a huge time saver.

CI seems to be failing because of *changelog / Changelog Entry Check (pull_request)*. Clicking on *Details* reveals that I should add *Please add '(#1319)' change line to CHANGES.md* but I'm not sure **I** should be the one adding this line, besides I don't even know under which release this pull request would be merged (if accepted). So do I have to add that line, or do you do that?

autoload/black.vim
```diff
@@ -98,13 +98,47 @@ if _initialize_black_env():
   import black
   import time
 
-def Black():
+def get_target_version(tv):
```
felix-hilden, line 101: Should we let Black (or the script) raise on incorrect versions instead of simply filtering? Could be more transparent to users. Also, I'm not entirely sure if this is what's happening here but anyways. Making sure the version is supported by Black is a nice touch though 👌

felix-hilden (changes requested):
Thanks for the updates! Having `target_version` as an argument makes total sense to me as well 👍 I had a look at the code and it looks reasonable, although I don't know vim plugins.

Our changelog is currently at the release, so a new heading needs to be added to the top. So please add a new `## Unreleased` heading and place the change under `### Integrations`. I left one comment as well. Since we're short on vim wizards, I'm willing to take your word on this working just fine 😄

docs/integrations/editors.md
```diff
@@ -116,6 +116,8 @@ Wing supports black via the OS Commands tool, as explained in the Wing documenta
 Commands and shortcuts:
 
 - `:Black` to format the entire file (ranges not supported);
+  - you can optionally pass `target_version=<version>` where `<version>` is one of
+    `[py27, py36, py37, py38, py39]` to set the target version.
```
felix-hilden, line 120: Also: we should probably not advertise specific versions here. I mean it's fine, but seems like another place to eventually get stale as we add support and drop support for older versions. Should be fine just to mention that it's the same argument as in the CLI, right? It would be a bit of an indirection though. Thoughts?

shaoran: > Should be fine just to mention that it's the same argument as in the CLI, right?

Yes, that would be better, and because of TAB completion, you get the correct list anyway. I'll change that.

autoload/black.vim
shaoran: I'll take a closer look at it in the evening. I wrote the code ages ago, so I don't remember without looking at the code.

shaoran: The function is needed because the `target_versions` argument in `black.FileMode` expects a set of the black.TargetVersion` values, so this function maps the string passed by the user to the corresponding `black.TargetVersion` value. You are right, being more transparent to the user is a good thing, hence I decided to print a warning that the default target version is used when Black does not recognize the passed target version.

felix-hilden: Yep, that seems reasonable! Feel free to mark as resolved

felix-hilden (approved):

docs/integrations/editors.md
felix-hilden: Very nice 👌

shaoran:
So, what is the status of this merge request? Is it going to be merged?

felix-hilden:
I'd say yes, but sadly I don't have commit access to the repo. I think if you agree that the conversation is resolved it's only a matter of time 👌

shaoran:
> I'd say yes, but sadly I don't have commit access to the repo. I think if you agree that the conversation is resolved it's only a matter of time ok_hand

I hope so. Yesterday I had to merge it again with upstream because it was showing once again conflicts with upstream (luckily on the changelog). I'm afraid, that if it takes again weeks or even months (like when I created the pull request), my changes will be once again completely outdated and I would have to do it from scratch again.

felix-hilden:
I feel you, sadly sometimes we're busy doing other things. But I'll try to push this to get resolved, and since it is pretty much done and big overhauls aren't planned, I think we're good 👍

PR #1321, opened 2020-03-27
Fix readthedocs build

zsol:
Our doc builds have been failing for a long time because of out of memory issues. Their docs say [we should use `pip` instead of `conda`](https://docs.readthedocs.io/en/stable/guides/build-using-too-many-resources.html#use-pip-when-possible) when possible.

This PR makes that change, while converting the config format to v2.

It also removes the dependency on the `regex` package for the purposes of building docs.


ambv:
We were using Conda originally since RTD at the time didn't have Python 3.6 yet (LOL).

ambv:
Thanks, @zsol, this is great.

PR #1322, opened 2020-03-27
Omit commit hash and date stamp from doc version

### LINKED ISSUE #1104: Documentation on ReadTheDocs is generated in a dirty directory
This causes the PyPI badge to be mangled now. Plus, wherever Sphinx is reporting the version, it's something noisy like "Black 19.10b1.dev0+g6bedb5c.d20191029".

zsol:
This also removes the dependency on setuptools-scm while building the
docs.

Fixes #1104.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,7 +17,7 @@ import re
 import shutil
 import string
 
-from setuptools_scm import get_version
+from pkg_resources import get_distribution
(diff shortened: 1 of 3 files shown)


zsol:
Don't merge this yet, I'd like to test it on top of #1321 too

zsol:
ready now. See https://black.readthedocs.io/en/dirty-docs/ for results

Issue #560, opened 2018-10-12
"fmt: on" does not work when crossing block boundaries and decorators


matejcik:
Operating system: Linux Mint 19.0
Python version: 3.6.6
*Black* version: 18.9b0
Does also happen on master: yes

Simplest reproducer: the following is left unchanged. In fact, it behaves as if `# fmt: off` is true until end of file.
```python
# fmt: off
if (x and y):
# fmt: on
    pass

if (x and y):
    pass
```

The soonest `# fmt: on` takes effect is at end of the indented `if` block.

Ran into this with a series of decorators:
```python
# fmt: off
@click.command()
@click.option("-a", "--align",      help="aligned help texts")
@click.option("-b", "--big-option", help="should stay aligned")
# fmt: on
def foo(align,    big_option):
    pass
```

Simple workaround for me is keep the formatter commands _within_ a "block" - the following works perfectly fine:
```python
@click.command()
# fmt: off
@click.option("-a", "--align",      help="aligned help texts")
@click.option("-b", "--big-option", help="should stay aligned")
# fmt: on
def foo(align, big_option):
    pass
```

matejcik:
I don't know if this is even fixable in the general case.

What I find problematic, though, is that a `# fmt: on` directive is silently ignored. In case black can't recover from a weirdly placed `# fmt: on`, that should be a loud error, IMHO.

ambv:
Thanks for your report, we'll tackle this for the next release.

Carreau:
Got same issue but for a flipped on/off: 

```
# fmt: off
... Whatever
# fmt: on
@dec
# fmt: off
@dec2
def fun():
    g(1,2,3) #don't reformat me

def f2(a,b,c):pass
```

Where I only want to reformat _some_ part of the code., and all the end of the file got reformatted.

cooperlees:
PR merged. Thanks!

PR #1324, opened 2020-03-29
Fix for "# fmt: on" with decorators

otstrel:
This is quick and likely dirty fix for #560. It is based on assumption that indentation has semantic meaning in Python so in case if `# fmt: on` exists in node children at the same column as opening `# fmt: off` we should process each child separately excluding only those before `# fmt: on`.

This seems to work for my use case which is blocking black adoption in my team and doesn't break any tests but it would be great if someone with more experience in codebase could help to polish it because I dived into black source code only two hours ago.

Thanks!

### DIFF
--- a/black.py
+++ b/black.py
@@ -3118,9 +3118,33 @@ def generate_ignored_nodes(leaf: Leaf) -> Iterator[LN]:
         if is_fmt_on:
             return
 
-        yield container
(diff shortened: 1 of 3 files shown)


otstrel:
Submitted this one using wrong email, sorry.

PR #1328, opened 2020-04-02
Fix --diff output when encountering EOF

### LINKED ISSUE #526: --diff output does not `git apply` cleanly when the diff context includes EOF
Thanks for making an awesome tool! I ran across this issue while trying to make a script that asks before applying changes. (Obviously I could reapply the changes by running `black` again so it's not too important, but being able to cache the diff so the apply is faster would be nice, and avoid a race condition if someone edits a file in the middle)

Steps to reproduce issue:

```
# setup
git init black-repro && cd black-repro && echo "x = '1'" > foo.py

black --diff . | git apply
# error: patch failed: foo.py:1
# error: foo.py: patch does not apply
```

The problem is that the diff includes an extra newline at the end compared to what git expects:
```
black --diff . | sed s/1,2/1,1/g | grep -Ev '^ $' | git apply
# works
```

akien-mga:
`split("\n")` includes a final empty element `""` if the final line
ends with `\n` (as it should for POSIX-compliant text files), which
then became an extra `"\n"`.

`splitlines()` solves that, but there's a caveat, as it will split
on other types of line breaks too (like `\r`), which may not be
desired.

Fixes #526.

Thanks to @benkuhn for the well-debugged issue.

### DIFF
--- a/black.py
+++ b/black.py
@@ -3876,8 +3876,8 @@ def diff(a: str, b: str, a_name: str, b_name: str) -> str:
     """Return a unified diff string between strings `a` and `b`."""
     import difflib
 
-    a_lines = [line + "\n" for line in a.split("\n")]
-    b_lines = [line + "\n" for line in b.split("\n")]
+    a_lines = [line + "\n" for line in a.splitlines()]
+    b_lines = [line + "\n" for line in b.splitlines()]
     return "".join(
         difflib.unified_diff(a_lines, b_lines, fromfile=a_name, tofile=b_name, n=5)
     )
--- a/tests/data/blackd_diff.diff
+++ b/tests/data/blackd_diff.diff
@@ -1,6 +1,6 @@
 --- [Deterministic header]
 +++ [Deterministic header]
-@@ -1,7 +1,6 @@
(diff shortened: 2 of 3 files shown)


akien-mga:
As mentioned, `splitlines()` works well but it might be doing more than wanted here. I'm not a Python guru, but I'm sure someone may have an elegant idea for ensuring that we can convert a string with `\n` to a list of lines with trailing `\n`, without having this extra empty line.

akien-mga:
Fixed the test which was also exhibiting this bug (and thus failing after the fix).

akien-mga:
Note that I didn't touch other uses of `split("\n")` which are not linked to the referenced issue, but contributors here might want to check if those occurrences can trigger a similar EOF issue:
```
$ rg --fixed-strings 'split("\n"'
black.py
2379:    for index, line in enumerate(prefix.split("\n")):

tests/data/comments2.py
299:        element.split("\n", 1)[0]

tests/data/comments3.py
10:        element.split("\n", 1)[0]
```

JelleZijlstra:
Thanks! The other cases look harmless (two are in test data and one always ignores an empty line at the end).

PR #1335, opened 2020-04-06
Add Dropbox as a user of Black

utsav-dbx:
We've been using Black on >1 million lines of Python for a few months.

### DIFF
--- a/README.md
+++ b/README.md
@@ -978,6 +978,8 @@ code style: pytest, tox, Pyramid, Django Channels, Hypothesis, attrs, SQLAlchemy
 Poetry, PyPA applications (Warehouse, Pipenv, virtualenv), pandas, Pillow, every Datadog
 Agent Integration, Home Assistant.
 
+The following organizations use _Black_: Dropbox.
+
 Are we missing anyone? Let us know.
 
 ## Testimonials


JelleZijlstra:
Great to hear! If we add this, I feel like we should also mention other companies using Black, which includes at least Quora and Facebook.

ssbarnea:
Maybe it would make sense to make two lists: "organizations and groups" and individual projects, ordering them in alphabetical order. To avoid getting an endless list of updates I suggest adding only those with more than 1k stars.

There is another way to avoid extra maintenance on this is to use a special github topic that can indicate adoption of [black](https://github.com/topics/black), or one like `blacked` or `black-enabled`. Doing this will enable people to advertise its use. We can only link to the github topic, which had default sorting based on popularity. For organizations we may still need to list them manually.

hugovk:
Many repos already display the badge to indicate adoption:

https://github.com/psf/black/blob/master/README.md#show-your-style

How about searching for the badge, something like this?

https://github.com/search?l=&q=%22code+style+black%22+filename%3AREADME&type=Code

The sort options are more limited (no sort by popularity), but it'll have more results.

ssbarnea:
@hugovk If your scope is to do research, you are better off searching for black listed as a dependency. Still, if our goal is to ease discoverability of other projects already using black and interested in exposing it, I think topics are the way to go.

I am sure that once we agree on the topic name to use, it will be very easy for others to update their repository topics. It worked faster on cookiecutter than I was expecting. I expect to be more tricky to decide about topic to use, as we all know naming things is hard ;)

ambv:
We can add other notable companies to the list in followup PRs :-)

ambv:
Thanks! ✨ 🍰 ✨

ambv:
Thanks! ✨ 🍰 ✨

PR #1349, opened 2020-04-17
Fix vim flags processing (#1307)

### LINKED ISSUE #1307: Vim plugin does not reverse the value of let g:black_skip_string_normalization
**Describe the bug**

Setting `let g:black_skip_string_normalization = 1` does string normalization.

Setting `let g:black_skip_string_normalization = 0` skips string normalization.

**To Reproduce**

`let g:black_skip_string_normalization = 1`

String normalization still being applied.

`let g:black_skip_string_normalization = 0`

String normalization not applied.

**Expected behavior**

`let g:black_skip_string_normalization = 1`

String normalization not being applied.

`let g:black_skip_string_normalization = 0`

String normalization being applied.

char101:


ichard26:
It seems like this was superseded by GH-1307, but wasn't formally linked a possible resolving PR? I'll close this under that assumption, but if I'm wrong, someone please prolitely shout :)

@char101 sorry that we missed your solution for GH-1307.

PR #1380, opened 2020-05-04
Make sure sys._base_executable is sane in Vim plugin

### LINKED ISSUE #1379: Installed black via vim-plug on macOS; starting vim throws ModuleNotFoundError: No module named 'black'
**Describe the bug**
I get `ModuleNotFoundError: No module named 'black'` when I run `vim`, and thus cannot use _black_ in `vim` on macOS:

```
$ vim
Error detected while processing /Users/erik/.vim/plugged/black/plugin/black.vim:
line  194:
  File "<string>", line 87, in <module>
ModuleNotFoundError: No module named 'black'
Press ENTER or type command to continue
```

Surely I am not the only macOS + vim user who wants to use _black_!

**To Reproduce**
1. Install `vim-plug` ([link](https://github.com/junegunn/vim-plug#installation)).
1. Add `Plug 'psf/black'` next to the other plugins in `~/.vimrc`.
1. Install via `vim` followed by `:PlugInstall`, or via `vim -c PlugInstall`:

    Screen Shot 2020-05-04 at 4 32 04 PM

3. Run `vim` from terminal.

**Expected behavior**
I expect to not get that error, for `vim` to load, and for me to be able to use _black_ as described in the [README](https://github.com/psf/black#vim).

- Version: whichever is pulled by the default `vim-plug` installation instructions. Is it not master?
- OS and Python version: macOS 10.15.4, Python 3.7.6 managed by Anaconda
- Vim: 8.2 (installed via Homebrew)

**Additional context**
@akselsjogren [suggested](https://github.com/psf/black/issues/1293#issuecomment-621082845) upgrading _black_  used by the vim-plugin to the latest from master:
```
cd ~/.vim
black/bin/pip install --upgrade git+https://github.com/psf/black.git
```

However I do not have `pip` in `bin` and thus cannot try this:
```
$ tree ~/.vim/black
.
├── bin
│   ├── python
│   ├── python3
│   └── vim
├── include
├── lib
│   └── python3.8
└── pyvenv.cfg
```

Also, see #1293 for further discussion and other things I've tried to fix.

dlukes:
Resolves #1379.

At least in Python 3.8, [the `venv` module relies on `sys._base_executable` to determine the Python executable to run](https://github.com/python/cpython/blob/f25fb6ebfec894c01bc927c9aae7924ffc826d11/Lib/venv/__init__.py#L117), but with recent versions of Vim (8.2.654 from Homebrew on macOS), this attribute is set to the *Vim* executable instead of Python: running `:python3 import sys; print(sys._base_executable)` prints `/usr/local/bin/vim`. This makes the Black Vim plugin one-time setup (i.e. virtualenv creation and Black Python package installation) fail at the ensurepip stage (because `vim -Im ensurepip ...` doesn't make any sense).

A possible workaround is to just override the `sys._base_executable` attribute, especially since the Black plugin already overrides `sys.executable` (possibly for similar reasons?) anyway. This is what the present patch does.

More generally though, is this intended behavior or a bug in recent versions of Python and/or Vim? I'm leaning towards a bug in Vim, since running `:python3 import sys; print(sys._base_executable)` in Neovim correctly prints a Python path, even with Python 3.8. Just trying to figure out whether this should be reported somewhere else as well :)

### DIFF
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -117,6 +117,12 @@ def _initialize_black_env(upgrade=False):
       print(f'Creating a virtualenv in {virtualenv_path}...')
       print('(this path can be customized in .vimrc by setting g:black_virtualenv)')
       venv.create(virtualenv_path, with_pip=True)
+    except Exception:
+      print('Encountered exception while creating virtualenv (see traceback below).')
+      print(f'Removing {virtualenv_path}...')
+      import shutil
+      shutil.rmtree(virtualenv_path)
+      raise
     finally:
       sys.executable = _executable
     first_install = True
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -111,7 +111,9 @@ def _initialize_black_env(upgrade=False):
     print('Please wait, one time setup for Black.')
     _executable = sys.executable
     try:
-      sys.executable = str(_get_python_binary(Path(sys.exec_prefix)))
+      executable = str(_get_python_binary(Path(sys.exec_prefix)))
+      sys.executable = executable
+      sys._base_executable = executable
       print(f'Creating a virtualenv in {virtualenv_path}...')
       print('(this path can be customized in .vimrc by setting g:black_virtualenv)')
       venv.create(virtualenv_path, with_pip=True)
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -110,6 +110,7 @@ def _initialize_black_env(upgrade=False):
   if not virtualenv_path.is_dir():
     print('Please wait, one time setup for Black.')
     _executable = sys.executable
+    _base_executable = getattr(sys, "_base_executable", _executable)
     try:
       executable = str(_get_python_binary(Path(sys.exec_prefix)))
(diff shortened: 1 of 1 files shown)


dlukes:
OK, did a bit more digging and:

- Both `sys.executable` and `sys._base_executable` are wrong in Vim on macOS.
- It's a [known issue](https://github.com/vim/vim/issues/3707) on Vim's side.

So my conclusion is that the `sys.executable = ...` hack which was already there previously in `black.vim` was added precisely for this reason, and just stopped working because `venv` in Python 3.8 started relying on `sys._base_executable` instead. So extending the hack to `sys._base_executable` should hopefully be uncontroversial :)

dlukes:
Giving it some more thought, I think it's also a good idea to wipe the virtualenv when virtualenv creation fails part way through (cf. added `except ...` clause). This makes sure the one-time setup will be re-attempted when Vim is restarted, making it more likely the user will actually notice the relevant errors. And hopefully, users will less often need to wipe the virtualenv by hand.

PR #1382, opened 2020-05-06
Add Change Log project URL

scop:


setup.py
```diff
@@ -58,6 +58,7 @@ def get_long_description() -> str:
     author="Łukasz Langa",
     author_email="lukasz@langa.pl",
     url="https://github.com/psf/black",
+    project_urls={"Change Log": "https://github.com/psf/black/blob/master/CHANGES.md"},
```
hugovk, line 61: "Changelog" is more common (28 vs. 2, https://github.com/hugovk/pypi-tools/issues/11#issuecomment-549175199):
```suggestion
    project_urls={"Changelog": "https://github.com/psf/black/blob/master/CHANGES.md"},
```

It'd also be useful to have an explicit link to this repo:

```python
"Source": "https://github.com/psf/black/",
```

scop: No other opimions besides I think the link title should match with the linked doc's title/heading, and the doc currently reads "Change Log".

PR #1384, opened 2020-05-06
Prefer virtualenv packages over global packages (#1383)

### LINKED ISSUE #1383: vim plugin should prefer virtualenv packages over the global packages

The current code appends the virtualenv site directory to `sys.path` after the global and user site directories. If a package that black will use is also installed in global/user namespace, and it has a breaking change, then it will break black because the global/user one will be imported at first

1. Run `pip3 install --user mypy-extensions==0.4.1` to install a package that black will use in the user site dir
2. Run `.vim/black/bin/pip install -U git+https://github.com/psf/black.git` to install the latest black for the vim plugin, it will install `mypy-extensions==0.4.3` into the virtualenv
3. Run `vim`, you'll see errors like this

vim-mypy-extensions

(`mypyc_attr` was introduced in 0.4.2)

- Version: [master]
- OS and Python version: [Centos 7.4/Python 3.6.5, macOS 10.14/Python 3.7.7]

The vim plugin use `sys.path.insert(0, virtualenv_site_packages)` at the first version, but changed to `append` in #804 

I think the problem with #804 is it shares virtualenv with black and other incompatible packages. When using black, it's better to use a separate virtualenv, or ensure other packages in the virtualenv are compatible.

kaiix:
Fixes #1383

### DIFF
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -128,7 +128,7 @@ def _initialize_black_env(upgrade=False):
   if first_install:
     print('Pro-tip: to upgrade Black in the future, use the :BlackUpgrade command and restart Vim.\n')
   if virtualenv_site_packages not in sys.path:
-    sys.path.append(virtualenv_site_packages)
+    sys.path.insert(0, virtualenv_site_packages)
   return True
 
 if _initialize_black_env():


ichard26:
A little heads up, you should probably write "fixes #1383" or "resolves #1383" so if this PR is merged, that issue is automatically closed.

PR #1389, opened 2020-05-08
Add docstrings to fmt checking functions, add to docs

autophagy:
Follow up from #1325

Adds docstrings to the fmt checking functions.
Renames fmt_on to is_fmt_on.
Adds the functions to the autodocs.

### DIFF
--- a/docs/reference/reference_functions.rst
+++ b/docs/reference/reference_functions.rst
@@ -135,6 +135,12 @@ Utilities
 
 .. autofunction:: black.generate_ignored_nodes
 
+.. autofunction:: black.is_fmt_on
+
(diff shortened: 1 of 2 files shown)


docs/reference/reference_functions.rst
```diff
@@ -135,6 +135,12 @@ Utilities
 
 .. autofunction:: black.generate_ignored_nodes
 
+.. autofunction:: black.contains_fmt_on_at_column
+
+.. autofunction:: black.first_leaf_column
+
+.. autofunction:: black.contains
```
ambv, line 142: That should be `is_fmt_on`, right?

Issue #392, opened 2018-07-02
Explain why black avoids using `\` for code organization


zsol:
The README should have a clear, concise explanation for why black avoids using `\`.

[Here](https://github.com/ambv/black/issues/64#issuecomment-375429106)'s a lengthy reasoning :eyes:

sudheer-g:
Hey! I've read @ambv 's explanation on this. Is it okay if I give this a shot? Thanks.

ambv:
Of course, go for it!

ambv:
Wording to use:

> Backslashes are one of the two places in the Python grammar that break significant indentation. The other one is multiline strings. You never need backslashes, they are used to force the grammar to accept breaks that would otherwise be parse errors. That makes them confusing to look at and brittle to modify. This is why Black always gets rid of them.
> 
> If you're reaching for backslashes, that's a clear signal that you can do better if you slightly refactor your code. I hope some of the examples above show you that there are many ways in which you can do it.

However, there will be one exception: with statements with multiple context managers: https://github.com/psf/black/issues/664

Issue #664, opened 2019-01-09
Long with statements are not broken into several lines


mhham:
Long `with` statements are not not broken into several lines:

Something like this should be OK, [according to PEP8](https://www.python.org/dev/peps/pep-0008/#id19)
```python
with averylongnameyoucantsplit as youcheater, \
     averylongnameyoucantsplit as youcheater, \
     averylongnameyoucantsplit as youcheater, \
     averylongnameyoucantsplit as youcheater, \
     averylongnameyoucantsplit as youcheater, \
     averylongnameyoucantsplit as youcheater, \
     averylongnameyoucantsplit as youcheater, \
     averylongnameyoucantsplit as youcheater, \
     averylongnameyoucantsplit as youcheater:
    print("hello")
```

But black formats its as: 

```python
with averylongnameyoucantsplit as youcheater, averylongnameyoucantsplit as youcheater, averylongnameyoucantsplit as youcheater, averylongnameyoucantsplit as youcheater, averylongnameyoucantsplit as youcheater, averylongnameyoucantsplit as youcheater, averylongnameyoucantsplit as youcheater, averylongnameyoucantsplit as youcheater, averylongnameyoucantsplit as youcheater:
    print("hello")
```

Operating system: macOS 10.14.2
Python version: 3.7.2
*Black* version: 18.9b0
Does also happen on master: Yes

JelleZijlstra:
I suppose this is Black strongly dislikes backslashes. I'm not sure there's really a great formatting Black can pick here.

DanCardin:
I for one have a deep hatred of backslashes and would really prefer black never use them. And since python's syntax here doesn't really allow for wrapping, i dont think black should do anything about it.

I would argue repeated `with`s like this to the point where it would want to wrap is an antipattern it shouldn't try to handle and make pretty.

colecrtr:
If developer A wants to format it a different way than developer B in order to prevent wrapping, that's fine. But it shouldn't affect the overall goal of a formatter –– doing the best with that it's given without affecting the AST. In essence, Black shouldn't just ignore and not make lengthy `with` statements as pretty as it can because some disagree with the usage of backslashes. As @mhham said: 

> The backslash usage here is exceptional (as mentioned in PEP 8) and brings a real improvement in readability.

advance512:
+1 for backslashes in this context

rogalski:
Another case to consider, it's fairly common to mock stuff like this:

```python
[...]

PR #558, opened 2018-10-10
Updated README.md with explanation about why black avoids '\'

sudheer-g:
Closes #392

PR #1399, opened 2020-05-09
Add preference of parantheses over backslashes in docs

SanketDG:
Closes https://github.com/psf/black/issues/392

Revived from https://github.com/psf/black/pull/558

Relevant to https://github.com/psf/black/issues/664

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,69 @@ def very_important_function(
         ...
 ```
 
+_Black_ prefers parentheses over backslashes, and will remove backslashes if found.
+
+```py3
+# in:
+
+if some_short_rule1 \
+  and some_short_rule2:
+      ...
+
+# out:
+
+if some_short_rule1 and some_short_rule2:
+  ...
+
+
+# in:
+
+if some_long_rule1 \
+  and some_long_rule2:
+    ...
+
+# out:
+
+if (
+    some_long_rule1
+    and some_long_rule2
+):
+    ...
+
+```
+
+Backslashes and multiline strings are one of the two places in the Python grammar that
+break significant indentation. You never need backslashes, they are used to force the
+grammar to accept breaks that would otherwise be parse errors. That makes them confusing
+to look at and brittle to modify. This is why _Black_ always gets rid of them.
+
+If you're reaching for backslashes, that's a clear signal that you can do better if you
+slightly refactor your code. I hope some of the examples above show you that there are
+many ways in which you can do it.
+
+However there is one exception: `with` statements using multiple context managers.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -238,6 +238,61 @@ def very_important_function(
...
+
+if (
+    some_long_rule1
+    and some_long_rule2
+):
+    ...
+
+```
+
+Backslashes are one of the two places in the Python grammar that break significant indentation. The other one is multiline strings. You never need backslashes, they are used to force the grammar to accept breaks that would otherwise be parse errors. That makes them confusing to look at and brittle to modify. This is why Black always gets rid of them.
```
cooperlees, line 272:
```suggestion
Backslashes and multiline strings are two places that break significant indentation. You never need backslashes, they are used to force the grammar to accept breaks that would otherwise be parse errors. That makes them confusing to look at and brittle to modify. This is why Black always gets rid of them.
```

hugovk, line 272:
```suggestion
Backslashes are one of the two places in the Python grammar that break significant indentation. The other one is multiline strings. You never need backslashes, they are used to force the grammar to accept breaks that would otherwise be parse errors. That makes them confusing to look at and brittle to modify. This is why _Black_ always gets rid of them.
```

README.md
```diff
@@ -238,6 +238,61 @@ def very_important_function(
...
+
+If you're reaching for backslashes, that's a clear signal that you can do better if you slightly refactor your code. I hope some of the examples above show you that there are many ways in which you can do it.
+
+However there is one exception: `with` statements using multiple context managers. Python's grammar does not allow organizing parentheses around the series of context managers.
+
+We don't want formattings like:
```
hugovk, line 278:
```suggestion
We don't want formatting like:
```

README.md
```diff
@@ -238,6 +238,61 @@ def very_important_function(
...
+
+with make_context_manager1() as cm1, make_context_manager2() as cm2, make_context_manager3() as cm3, make_context_manager4() as cm4:
+    ...  # nothing to split on - line too long
```
hugovk, line 281: Add a py3 codeblock around this

README.md
```diff
@@ -238,6 +238,61 @@ def very_important_function(
...
+
+Black will now format it like this:
```
hugovk, line 283:
```suggestion
_Black_ will now format it like this:
```

README.md
SanketDG: Done, thanks!

cooperlees:
Thanks again! Welcome to being a black contributor now too! (@SanketDG has also been a legend helping with bandersnatch!)

PR #1403, opened 2020-05-10
refactor: modularise black

P403n1x87:
Split the monolithic src/black/__init__.py into logical submodules. Care
has been taken to avoid circular dependencies and group related classes 
and procedures.


ambv:
Thanks for the contribution but we'll be taking care of this inside the core team. There's quite a few things we need to do to make sure we end up where we want to be.

P403n1x87:
No problem. Shall I just close this PR then?

Cheers.

cooperlees:
Hi @P403n1x87 !

Thanks for the effort here! But as @ambv said, we’ve got some plans already. One thing is ensuring we get black/__init__.py working with mypyc compiler and then we’ll start the much needed refactor.

We’ll definitely steal some ideas from this refactor for sure. So this is not wasted effort! Once we get that working we might be able to public ally design the refactor and share the work!

PR #1417, opened 2020-05-16
fix crashes on docstring whitespace changes

### LINKED ISSUE #1415: Crash on docstring ending with newlines
Maybe a side effect of #1053?

```
$ cat common/qfuture/nltk2.py 
def word_tokenizer():
    """Returns NLTK's standard word tokenizer, but without a change that was applied in NLTK 3.0.3.

    """
    pass
$ black common/qfuture/nltk2.py
error: cannot format common/qfuture/nltk2.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_0awe_h_l.log
Oh no! 💥 💔 💥
1 file failed to reformat.
$ cat  /tmp/blk_0awe_h_l.log
--- src
+++ dst
@@ -15,11 +15,11 @@
       body=
         Expr(
           value=
             Constant(
               value=
-                "Returns NLTK's standard word tokenizer, but without a change that was applied in NLTK 3.0.3.\n\n ",  # str
+                "Returns NLTK's standard word tokenizer, but without a change that was applied in NLTK 3.0.3.",  # str
             )  # /Constant
         )  # /Expr
         Pass(
         )  # /Pass
       decorator_list=
```

Making this change is probably a good thing but we'll need to enhance the AST equivalence checker to let this through.

JelleZijlstra:
Fixes #1415

### DIFF
--- a/tests/data/docstring.py
+++ b/tests/data/docstring.py
@@ -1,5 +1,5 @@
 class MyClass:
-  """Multiline
+  """ Multiline
   class docstring
   """
 
@@ -11,7 +11,7 @@ class MyClass:
 
 
 def foo():
-  """This is a docstring with
+  """This is a docstring with             
   some lines of text here
   """
   return
@@ -65,6 +65,13 @@ def over_indent():
     - And the closing quote is too deep
     """
   pass
+
+
+def single_line():
+    """But with a newline after it!
+
+    """
+    pass
 
 # output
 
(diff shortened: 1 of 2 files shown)


tests/data/docstring.py
```diff
@@ -136,3 +143,8 @@ def over_indent():
       - And the closing quote is too deep
     """
     pass
+
+
+def single_line():
+    """But with a newline after it!"""
+    pass
```
ichard26, line 150: Is this an accidental duplicate of the one above at lines 69 to 75? It is refining the function...

ichard26: Nevermind, didn't see the output comment.

Issue #1413, opened 2020-05-15
Colon inside a range gets a colon added to it


WhyNotHugo:
The colon before inside a range gets a space added to it:

input:
```python
def a(code: str, pos: int):
    return code[pos: pos + 10]
```

output
```python
def a(code: str, pos: int):
    return code[pos : pos + 10]
    #              ^ Note this space
```

This seems to be the only scenario in which black adds a space **before** a colon, so I'm assuming it's a bug. Flake8 immediately complains since that's, well, wrong.

This doesn't seem to happen if ` + 10` is not there.

**Environment:**
```
$ black --version
black, version 19.10b0
$ python --version
Python 3.8.2
$ uname -a
Linux aphrodite 5.6.12-arch1-1 #1 SMP PREEMPT Sun, 10 May 2020 10:43:42 +0000 x86_64 GNU/Linux
```

**Does this bug also happen on master?:**
Yes

hugovk:
This is documented in the README:

https://github.com/psf/black/blob/master/README.md#slices

WhyNotHugo:
But it would seem that this is not enforced consistently, eg:
```python
def a(code: str, pos: int):
    x = code[pos : pos + 10]  # Spaces here
    y = code[pos:pos]  # No spaces here
```

zsol:
We should probably expand that paragraph a bit in the docs. PEP8 recommends no spaces around `:` for "simple" expressions, and extra space for "complex expressions". Black treats anything more than variable names as "complex".

WhyNotHugo:
Ah, this makes sense. I guess I'm just too used to flake8's syntax.

I think it would be good to mention, in "installation and usage", a basic configuration for flake8 that aligns with black. Would you guys accept a PR that describes this?

WhyNotHugo:
(I realise the individual settings are mentioned in different places, but I'm talking about a single-block copy-pasteable thing :) )

zsol:
FWIW here's the relevant code that implements this behavior: https://github.com/psf/black/blob/45c98cf150674dd49cb51d57f8baede4c287f043/src/black/__init__.py#L1411-L1413

I would turn off flake8 warnings related to subscripts (in fact, related to all code formatting). #1371 added an example flake8 config. If you think that needs tuning, feel free to open a PR :)

FWIW this is a duplicate of #889, #940, #1343, and possibly more. Thanks @hugovk for replying to all of them :)

WhyNotHugo:
[...]

PR #1418, opened 2020-05-16
Expand docs about spaces around ':' operators

ichard26:
As per https://github.com/psf/black/issues/1413#issuecomment-629162592.
> We should probably expand that paragraph a bit in the docs. PEP8 recommends no spaces around : for "simple" expressions, and extra space for "complex expressions". Black treats anything more than variable names as "complex".


JelleZijlstra:
Looks good, will merge this in a bit if nobody else has feedback. Thanks for the PR!

zsol:
LGTM! Thanks

Issue #760, opened 2019-03-15
Please don't move tags


djmitche:
When updating the repo:
```
Failed to download remote objects and refs:  From https://github.com/ambv/black
   ea55ff2..d6db1c1  master     -> origin/master
 * [new tag]         19.3b0     -> 19.3b0
 ! [rejected]        stable     -> stable  (would clobber existing tag)
```

I assume that `stable` is meant to track the latest stable release?  That's a job for a branch, not a tag :)

ambv:
OK, we will remove the "stable" tag and make it a branch with the next release.

ambv:
OK, this is solved. `stable` is no longer a tag, it is a **branch**.

sersorrel:
fwiw, this goes directly against a statement in the pre-commit documentation:

> `pre-commit` assumes that the value of `rev` is an immutable ref (such as a tag or SHA) and will cache based on that. Using a branch name (or `HEAD`) for the value of `rev` is not supported and will only represent the state of that mutable ref at the time of hook installation (and will NOT update automatically).

PR #1421, opened 2020-05-16
GitHub action

WhyNotHugo:
Implement a GitHub action that can be reused across projects that want to run black as part of their CI workflows.

This should facilitate new projects setting up their CI to just use black. I know I'd be happy to drop my custom actions.

Note: I had to do a small refactor in `/docs` because github's action runner gets confused with broken symlinks. However, these changes should be non-intrusive. See 47a83c4d39453124d0711142db00104b42c9ed00.

There's a ~test~ proof of concept usage repository [WhyNotHugo/test-black-action](https://github.com/WhyNotHugo/test-black-action).


README.md
```diff
@@ -1100,6 +1100,24 @@ If you're already using Python 3.7, switch the `language_version` accordingly. F
 `stable` is a tag that is pinned to the latest release on PyPI. If you'd rather run on
 master, this is also an option.
 
+## GitHub actions
+
+Create a file named `.github/workflows/black.yml` inside your repository with:
```
ichard26, line 1105: NIT: Might want to be clearer about 'black.yml' being the file name and the rest is the directory structure.

WhyNotHugo:
Rebased: Hadn't run the pre-commit hooks and `prettier` was unhappy.

README.md
```diff
@@ -1100,6 +1100,24 @@ If you're already using Python 3.7, switch the `language_version` accordingly. F
...
+
+```
```
ichard26, line 1107: NIT: Please add a language annotation for yaml, example:
\```yaml
\```

action.yml
```diff
@@ -0,0 +1,9 @@
+name: "Black"
+description: "The uncompromising Python code formatter."
+author: "Łukasz Langa et al"
```
ichard26, line 3: NIT: Possible messed up copy and paste?

README.md
```diff
@@ -1100,6 +1100,24 @@ If you're already using Python 3.7, switch the `language_version` accordingly. F
 `stable` is a tag that is pinned to the latest release on PyPI. If you'd rather run on
 master, this is also an option.
 
+## GitHub actions
```
ichard26, line 1103: NIT: actions -> Actions

action.yml
WhyNotHugo: [Łukasz Langa](https://github.com/psf/black/blob/master/LICENSE#L3)
et al: ["And others"](https://en.wiktionary.org/wiki/et_al.#English)

README.md
WhyNotHugo: Wasn't sure, since other titles don't capitalise, but this is a proper noun... such hard choices!

action.yml
ichard26: Fair enough, good example of today I learned!

README.md
WhyNotHugo: Uhm... wouldn't know how to make it clearer without it being too verbose.

README.md
ichard26: I based it on how GitHub marketed their service.
https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions

But, in the end, it's just a NIT :)

README.md
ichard26: Maybe this: "In the `.github/workflows` directory of your repository, create a file named `black.yml` with:"

Or is that too verbose in your opinion?

ichard26:
Also, since you're adding a new sub-header, should we add it to the table of contents?. The only sub-header not in the table of contents is the LICENSE section.

README.md
WhyNotHugo: I can't just copy-paste the whole path in one go now, can I?

WhyNotHugo:
Added entry to the TOC.

Note to self: eventually rebase this once it's past review.

README.md
ichard26: Makes sense.

README.md
```diff
@@ -1100,6 +1101,24 @@ If you're already using Python 3.7, switch the `language_version` accordingly. F
...
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-python@v1
+      - uses: psf/black@stable
```
cooperlees, line 1119: Is this a tag we (the maintainers) have to manually do and keep up to date?

ichard26: Yes. Although, it is a branch*, one that will updated on every release anyway. 

*According to this issue, the stable tag is a now a branch: https://github.com/psf/black/issues/760

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.8
+
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+RUN pip install --upgrade --no-cache-dir black
```
hugovk, line 6: Why skip the cache? CIs can account for a lot of downloads from PyPI and it'd be good to minimise their huge hosting bills where possible.

Also, some may prefer to pin to a given version of Black (eg. how pre-commit does it)

action.yml
hugovk: "and contributors to Black" is used in the docs:

https://github.com/psf/black/blob/9ed2542e938c327a53c17f8932ee5fc53776ba31/docs/conf.py#L83

README.md
```diff
@@ -1100,6 +1101,24 @@ If you're already using Python 3.7, switch the `language_version` accordingly. F
...
+```yaml
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
hugovk, line 1117: V2 is now available
https://github.com/actions/checkout

Dockerfile
WhyNotHugo: Because this run inside docker's build system, it runs on a clean `python:3.8` image each time. The cache from previous runs is never available.

The goal of `--no-cache-dir` here  it for docker not to _save_ the cache, since it ends up in the image otherwise, bloating it with a cache that will never be read.

README.md
WhyNotHugo: It is, but it's something that's already being kept up to date for other reasons anyway. I believe `pre-commit` relies on it too.

WhyNotHugo:
Rebased to cleanly solve conflicts.

README.md
```diff
@@ -1263,6 +1264,24 @@ If you're already using Python 3.7, switch the `language_version` accordingly. F
 `stable` is a branch that tracks the latest release on PyPI. If you'd rather run on
 master, this is also an option.
 
+## GitHub Actions
+
```
SanketDG, line 1268: It should be clear that this just lists all the files that are to be formatted, just so that there is an explcit contrast between `black`'s default behaviour.

WhyNotHugo: It doesn't just list them; CI will fail to pass or linting is wrong.

This should be obvious; CI obviously won't modify a commit you've already committed  and pushed.

SanketDG: Well it's not about modifying, it's just that the output is not directly available to the user, one has to jump through atleast three clicks to see the output.

Also, they could be expecting how other linting tools work with GH actions using https://github.com/reviewdog/reviewdog, which is just not possible with _Black_.

WhyNotHugo: The intent behind this action is to allow other projects to very easily run `black`as part of their CI. This is mostly to ease reviewing PRs (eg: has this person black'd their code?), or to avoid slipups. Generally, seeing "oh, you need to run black on this" is enough.

I've added `--diff` to this Action, since that shows the output too, which is definitely an improvement.

I was unfamiliar with reviewdog, and it looks really cool. I does, however, required quite a bit more setup than this which  "just runs black".

WhyNotHugo: (FWIW, running `black` with `reviewdog` should work passing just a diff.

docs/Makefile
```diff
@@ -10,11 +10,24 @@ BUILDDIR      = _build
...
+	ln -s _build/generated/contributing_to_black.md
+	ln -s _build/generated/editor_integration.md
+	ln -s _build/generated/gnoring_unmodified_files.md
+	ln -s _build/generated/installation_and_usage.md
+	ln -s _build/generated/license.md
+	ln -s _build/generated/pyproject_toml.md
+	ln -s _build/generated/show_your_style.md
+	ln -s _build/generated/testimonials.md
+	ln -s _build/generated/the_black_code_style.md
+	ln -s _build/generated/version_control_integration.md
```
ichard26, line 25: `Makefile` is only used for Unix systems. Windows uses the `make.bat` file instead. I'll look into it and comment the Windows patch below once I have some free time.

ichard26: Edit: Actually with #1456, the symlinks were deleted, so all of this symlink nonsense can just be thrown in the garbage.

WhyNotHugo: Oh, I wasn't following up on that. Rebased to remove my no-op changes, thanks.

WhyNotHugo:
Any update on the mergeability of this?

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.8
```
cooperlees, line 1: My only preference is maybe use python3 here ... Avoids another manual thing we’ll forget to upgrade ... Sorry if I missed a reason for pinning 3.8

README.md
```diff
@@ -35,7 +35,8 @@ _Contents:_ **[Installation and usage](#installation-and-usage)** |
 **[Code style](#the-black-code-style)** | **[Pragmatism](#pragmatism)** |
 **[pyproject.toml](#pyprojecttoml)** | **[Editor integration](#editor-integration)** |
 **[blackd](#blackd)** | **[black-primer](#black-primer)** |
-**[Version control integration](#version-control-integration)** |
+**[Version control integration](#version-control-integration)** | |
```
hugovk, line 38:
```suggestion
**[Version control integration](#version-control-integration)** |
```

action.yml
```diff
@@ -0,0 +1,9 @@
+name: "Black"
+description: "The uncompromising Python code formatter."
+author: "Łukasz Langa and contributors to Black"
+branding:
+  color: "black"
```
hugovk, line 5: Could be worth testing in case the docs are outdated, but https://help.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#branding doesn't list black:

> The background color of the badge. Can be one of: white, yellow, blue, green, orange, red, purple, or gray-dark.

WhyNotHugo: Quickly tested this, it works (I also immediately removed that test too).

2020-06-01T08:51:09,257888262+02:00

Dockerfile
WhyNotHugo: What about just `python:latest`? Or do we want to hold back when `4.0` comes out? :sweat_smile:

Ayrx:
Hi, I get the following error when using GitHub Actions with black:

```
Current runner version: '2.263.0'
Operating System
Virtual Environment
Prepare workflow directory
Prepare all required actions
Download action repository 'actions/checkout@v2'
Download action repository 'actions/setup-python@v2'
Download action repository 'psf/black@stable'
##[error]Could not find file '/home/runner/work/_actions/_temp_e6a60df6-7298-4553-9c69-9049a1c24037/_staging/psf-black-6bedb5c/docs/editor_integration.md'.
```

The .yml file is the one suggested in the README as follows:

```
name: Lint

on: [push, pull_request]

jobs:
  lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: psf/black@stable
```

Ayrx:
Ah, I guess I have to use `psf/black@master` until this gets into a released build. Sorry for the noise!

scratchmex:
Following @Ayrx, this should be documented in the README. I will open a PR about it.

WhyNotHugo:
No, the suggestion is to use `stable`, it's just that there hasn't been any stable releases since this was merged.

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3
+
+ENV PYTHONDONTWRITEBYTECODE 1
```
asottile, line 3: this will make black's startup and execution significantly slower

```console
$ time PYTHONDONTWRITEBYTECODE=1 black t.py 
All done! ✨ 🍰 ✨
1 file left unchanged.

real	0m0.233s
user	0m0.221s
sys	0m0.012s

$ time black t.py   # this one will compile and write the bytecode
All done! ✨ 🍰 ✨
1 file left unchanged.

real	0m0.238s
user	0m0.207s
sys	0m0.028s

$ time black t.py 
All done! ✨ 🍰 ✨
1 file left unchanged.

real	0m0.130s
user	0m0.123s
sys	0m0.004s
```
```

File: Dockerfile [hunk 14]
```diff
@@ -0,0 +1,8 @@
...
+ENV PYTHONUNBUFFERED 1
+
+RUN pip install --upgrade --no-cache-dir black
```
asottile, line 6: this will always install the latest version from pypi, should this install from the local repository such that the version used in the github action will match what you get in the docker image? (so `COPY . . && RUN pip install .`)

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+
+ENTRYPOINT /usr/local/bin/black --check --diff  .
```
asottile, line 8: this causes an extra layer of shell indirection which will be slower since it has to start a shell first (it makes the entrypoint `sh -c '/usr/local/bin/black ...'` -- you're better to use the json syntax:

```dockerfile
ENTRYPOINT ["/usr/local/bin/black", "--check", "--diff", "."]
```

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3
+
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
```
asottile, line 4: just a tip for dockerfiles, you usually want to combine adjacent `ENV` statements so they don't create unnecessary extra layers, in this case you'd write

```dockerfile
ENV PYTHONDONTWRITEBYTECODE=1 PYTHONUNBUFFERED=1
``` 

or if you don't like it all one line you can write it with some backslashes

```dockerfile
ENV \
    PYTHONDONTWRITEBYTECODE=1 \
    PYTHONUNBUFFERED=1
```

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+
+RUN pip install --upgrade --no-cache-dir black
+
```
asottile, line 7: an additional improvement to execution here is to run black once such that it puts the grammar cache into the image, from my experiments this improves execution of black by about ~50ms per run

one way that I did this was by changing the install line to:

```dockerfile
RUN : \
    && pip install --upgrade --no-cache-dir black \
    # have black run against a file so its cache and pyc files are populated
    && touch /tmp/t.py && black /tmp/t.py && rm /tmp/t.py
```

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3
```
asottile, line 1: this is a pretty chunky image, `python:3-slim` or `python:3-alpine` (though, generally I don't suggest using alpine) are probably better choices

```console
$ docker images | grep python.*3
docker.io/library/python                        3-slim    6cf621cb1327  4 days ago     118 MB
docker.io/library/python                        3         28a4c88cdbbf  4 days ago     905 MB
docker.io/library/python                        3-alpine  0f03316d4a27  6 days ago     44.7 MB
```

asottile (commented):
sorry for the potentially late drive-by, was trying to look at some github actions and then I went down a rabbit hole

rafaellehmkuhl:
How does one specify the source path on the GitHub actions .yml file?

PR #1424, opened 2020-05-17
Update and fix Flake8

hugovk:
Updates Flake8 (and mypy) in pre-commit.

Fixes new Flake8 warnings:

* F541 f-string is missing placeholders
* E741 ambiguous variable name 'l'

Update GHA actions to v2.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,10 +10,10 @@ jobs:
         python-version: [3.7]
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,10 +12,10 @@ jobs:
         os: [ubuntu-latest, macOS-latest, windows-latest]
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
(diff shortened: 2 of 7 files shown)


.github/workflows/lint.yml
```diff
@@ -10,10 +10,10 @@ jobs:
         python-version: [3.7]
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
```
cooperlees, line 13: Is there a Tool we can run to check for action upgrades available?

hugovk:
> Many thanks for cleaning this up - Wonder if we can automate updating Actions + pre-commit or create an issue on CI failure.

# GitHub Actions

I couldn't find an existing tool. A DIY solution: parse the YAML file for the used actions, check their repos to see if they have a newer version/tag. Feels like a bit much seeing as they don't change so often.

# pre-commit

The issues to +1 for Dependabot to support it:

* https://github.com/dependabot/feedback/issues/839
* https://github.com/dependabot/dependabot-core/issues/1524

---

Also, this is part of the pre-commit author's current project:

* https://twitter.com/codewithanthony/status/1260437067701669895

---

Or right now, it's easy to run `pre-commit autoupdate` and the fail if there's an update that modifies the config file: `test $(git status --porcelain| wc -l) = 0`

Here's a quick demo:

* https://github.com/hugovk/black/blob/f520f21fbc9bc85c81ce6752f29086a27b38307d/.github/workflows/pre-commit-update.yml
* https://github.com/hugovk/black/runs/682933819?check_suite_focus=true

It's failing because there's a Prettier update (I intentionally omitted it from this PR because the node version on Travis CI would need updating).

It wouldn't make sense to run as part of normal PRs, we don't want to fail someone's unrelated changes. It could be a weekly cron or only on merges to master.

And then perhaps something like https://github.com/JasonEtco/create-an-issue or https://github.com/peter-evans/create-issue-from-file could be used to conditionally create an issue, but we probably wouldn't want that to spam duplicate issues.

Or perhaps something like https://peterevans.dev/posts/github-actions-how-to-automate-code-formatting-in-pull-requests/ to create a PR.

PR #1451, opened 2020-05-22
CI: Move primer to own workflow

hugovk:
black-primer takes 3.5-5.5 mins on the CI. I suggest moving it to its own workflow:

* make it clearer what fails: black-primer or unit tests
* allow unit tests to finish sooner -> quicker feedback
* make builds more granular -> test in isolation
* make full use of the 20 GHA parallel jobs -> finish sooner

---

Is it desired to run black-primer on the full 3x3 Python/OS matrix? Fine if it is! If not, we can select just those needed (eg. `lint.yml` only needs to run on a single Python/OS).

### DIFF
--- a/.github/workflows/primer.yml
+++ b/.github/workflows/primer.yml
@@ -1,4 +1,4 @@
-name: Test
+name: Primer
 
 on: [push, pull_request]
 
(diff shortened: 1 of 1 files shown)


cooperlees:
I'll just add the badge since i have time right now! Thanks again.

hugovk:
You're welcome!

PR #1453, opened 2020-05-22
Update README.md

gcamargo1:
Adding additional info on where to check box for file watchers configuration


cooperlees:
Thanks. Clarity is always good.

PR #1456, opened 2020-05-23
Refactor docs / Maintenance of docs

ichard26:
Resolves #1430.
Related #990 (please see #1457).
Resolves #1435.

**See the latest version of the refactored docs here:**
https://ichard26-testblackdocs.readthedocs.io/en/refactor_docs/index.html
(also is proof that this won't break the ReadTheDocs builds)

### Refactor

- Split these sections from the README into their own files as per #1430:
  - `blackd` -> `docs/blackd.md`
  - `black-primer` -> `docs/black_primer.md`
  - `Editor integration` -> `docs/editor_integration.md`
  - `The Black code style` as per @zsol https://github.com/psf/black/issues/1430#issuecomment-630208621 -> `docs/the_black_code_style.md`
  - `Pragmatism` as per @zsol https://github.com/psf/black/issues/1430#issuecomment-631729659 -> `docs/the_black_code_style.md`

~~Yes, `Pragmatism` and `The Black code style` were merged into `docs/the_black_code_style_and_pragmatism.md`. **Please suggest a better title and filename for this poor piece of documentation.**~~ 

Thank you @hugovk for the suggestions! `the_black_code_style_and_pragmatism.md` has been renamed to `the_black_code_style.md`.

### Maintenance

- Fixed a few broken or likely to be broken links:
  - `README.md`:
    - `https://poetry.eustace.io/` -> `https://python-poetry.org/`
    - `https://docs.pipenv.org/` -> `https://readthedocs.org/projects/pipenv/`
  - `docs/editor_integration.md`
    - `https://github.com/proofit404/blacken` -> `https://github.com/pythonic-emacs/blacken`
- Deleted all of the symlinks in `docs/` since they serve no use now:
  - *you don't need me to list them... right?*
- Manually merged these changes since the docs that they were changing have been moved:
  - https://github.com/psf/black/commit/bb767863b0cc51897f9cd94da84b8b01f3936607
  - https://github.com/psf/black/commit/a2408b3cb23fe252b3674cee484d742496bb3411
  - https://github.com/psf/black/commit/397b16431e7e2af2869353d9310834f8c80e6cde
- Committed my changes from #1436.

### Features / Improvements

- ~~Made the sidebar navigation scrollable since we have so much packed into that sidebar that some of its contents are clipped. The only sad part of this change is that for users of Firefox, they will have to see an ugly scrollbar since Firefox disabled hiding scrollbar by default~~ (please see #1457)
- Completely reworked the logic in `docs/conf.py`.
[...]

PR #1457, opened 2020-05-23
Scrollable sidebar

### LINKED ISSUE #990: [Docs] Navigation is not scrollable and is cropped
Happens here: https://black.readthedocs.io/en/latest/the_black_code_style.html
Open with some window with limited height. It's happening for me even on fullscreen with a large monitor (27-inch (1920 x 1080)).

Screen Shot 2019-08-20 at 17 23 41

ichard26:
Resolves #990.

**See how the scrollable sidebar works here:**
https://ichard26-testblackdocs.readthedocs.io/en/scrollable_sidebar/

This is necessary since we have so many documentation sections that even
on a desktop screen, the navigation can sometimes be clipped. The only
annoyance is that on Firefox, the scrollbar can't be hidden :(

***Ideally this would be merged before my large refactor so I can deal with the merge conflict ahead of its merge. Or, just @ me and I'll remove the second commit so this can be merged without future conflict causing changes. The second commit is only so you can interact with the updated docs, and becomes useless once #1456 is merged.***

### DIFF
--- a/README.md
+++ b/README.md
@@ -1214,7 +1214,7 @@ cooper-mbp:black cooper$ ~/venvs/b/bin/black-primer
 
 Failed projects:
 
-## flake8-bugbear:
+### flake8-bugbear:
  - Returned 1
  - stdout:
 --- tests/b303_b304.py	2020-05-17 20:04:09.991227 +0000
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,38 @@
+/* Make the sidebar scrollable. Fixes https://github.com/psf/black/issues/990 */
+div.sphinxsidebar {
+  max-height: calc(100% - 18px);
+  overflow-y: auto;
+}
+
+/* Hide scrollbar for Chrome, Safari and Opera */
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -1214,7 +1214,7 @@ cooper-mbp:black cooper$ ~/venvs/b/bin/black-primer
 
 Failed projects:
 
-## flake8-bugbear:
+### flake8-bugbear:
```
cooperlees, line 1217: This now does not match output :D

cooperlees:
Also - Want to unblock the amazing big refactor! 🥰

README.md
ichard26: Yeah, that was causing issue with the old build logic... thankfully #1456 solves that :)

cooperlees: All good - I didn’t really care and would even add a ‘#’ into the output to make things easier. I just wanted something in the outputto skip projects easily in vim / less etc.

ichard26: image

:D

PR #1466, opened 2020-05-31
Convert (most of the) configuration values from pyproject.toml to strings

### LINKED ISSUE #1458: AttributeError: 'int' object has no attribute 'lower'
**Describe the bug** 
When I try to run black on a file I get this error. even a simple file that has only ```print('hello world')```
I am runnning ```black temp.py```

```black temp.py```

python 3.8 macos
black.__version__
'19.10b0'
click.__version__
'7.1.2'

**Error** 

`(venv) $ black temp.py
  File "/venv/bin/black", line 10, in <module>
    sys.exit(patched_main())
  File "/venv/lib/python3.8/site-packages/black.py", line 4135, in patched_main
    main()
  File "/venv/lib/python3.8/site-packages/click/core.py", line 829, in __call__
  File "/venv/lib/python3.8/site-packages/click/core.py", line 781, in main
  File "/venv/lib/python3.8/site-packages/click/core.py", line 700, in make_context
    self.parse_args(ctx, args)
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1048, in parse_args
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1623, in handle_parse_result
    value = self.full_process_value(ctx, value)
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1965, in full_process_value
    return Parameter.full_process_value(self, ctx, value)
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1589, in full_process_value
    value = self.process_value(ctx, value)
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1579, in process_value
    return self.type_cast_value(ctx, value)
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1568, in type_cast_value
    return _convert(value, (self.nargs != 1) + bool(self.multiple))
  File "/venv/lib/python3.8/site-packages/click/core.py", line 1565, in _convert
    return self.type(value, self, ctx)
  File "/venv/lib/python3.8/site-packages/click/types.py", line 46, in __call__
    return self.convert(value, param, ctx)
  File "/venv/lib/python3.8/site-packages/click/types.py", line 412, in convert
    value = value.lower()
AttributeError: 'int' object has no attribute 'lower'`

ichard26:
Resolves #1458.

---

We need to convert all configuration values from the pyproject.toml
file because Click already does value processing and conversion and
it only expects the exact Python type or string. Click doesn't like
the integer `1` as a boolean for example. This also means other
unsupported types like datetime.time will be rejected by Click as
an invalid value instead of throwing an exception.

We only skip converting objects that are an instance of
collections.abc.Iterable because it's almost impossible to get back
the original iterable of a stringified iterable.

---

After this change, these following values will be accepted for each of the configuration options in `pyproject.toml`:

| Option | Accepted values |
| ---------- | ----------------------|
| Boolean type / flag type <ul> <li> `--pyi` </li> <li> `-S`  /  `--skip-string-normalization` </li> <li> `--check`</li> <li> `--diff` </li> <li> `--color` / `--no-color` </li> <li> `--fast` / `--safe` </li> <li> `--quiet` </li> <li> `--verbose` </li> </ul> | <ul> <li> Integers <ul>  <li> `0` </li> <li> `1` </li> </ul> </li> <li> Strings (case insensitive) <ul> <li> `"f"` </li> <li> `"false"`</li> <li> `"n"` </li> <li> `"no"` </li> <li> `"0"` </li> <li> `"t"`</li> <li> `"true"` </li> <li> `"y"` </li> <li> `"yes"` </li> <li> `"1"` </li> </ul> </li> <li> Boolean <ul> <li> `true` </li> <li> `True` (technically a string) </li> <li> `false` </li> <li> `False` (technically a string) </li> </ul> </li> </ul>|
| Integer type <ul> <li> `-l` / `--line-length` </li> </ul> | <ul> <li> Any integer </li> <li> Any stringified integer </li> </ul> |
| Choice type <ul> <li> `-t` / `--target-version`</li> </ul>| *The choices available for that option.* |
| Text type <ul> <li> `-c` / `--code` </li> <li> `--include` </li> <li> `--exclude` </li>  <li> `--force-exclude` </li> </ul> | *Any text still accepted.* |

And `--version`, `--help`, and `--config` are still allowed in `pyproject.toml` :laughing:

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -317,12 +317,14 @@ def read_pyproject_toml(
     default_map: Dict[str, Any] = {}
     if ctx.default_map:
         default_map.update(ctx.default_map)
-    default_map.update(config)
+    default_map.update(
+        {
+            k: str(v) if not isinstance(v, collections.abc.Iterable) else v
+            for k, v in config.items()
+        }
+    )
 
-    ctx.default_map = {
-        k: str(v) if not isinstance(v, collections.abc.Iterable) else v
-        for k, v in default_map.items()
-    }
+    ctx.default_map = default_map
     return value
 
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -2,7 +2,6 @@ import ast
 import asyncio
 from abc import ABC, abstractmethod
 from collections import defaultdict
-import collections.abc
 from concurrent.futures import Executor, ThreadPoolExecutor, ProcessPoolExecutor
 from contextlib import contextmanager
 from datetime import datetime
@@ -282,7 +281,12 @@ def parse_pyproject_toml(path_config: str) -> Dict[str, Any]:
     """
     pyproject_toml = toml.load(path_config)
     config = pyproject_toml.get("tool", {}).get("black", {})
-    return {k.replace("--", "").replace("-", "_"): v for k, v in config.items()}
+    return {
+        k.replace("--", "").replace("-", "_"): str(v)
+        if not isinstance(v, (list, dict))
+        else v
+        for k, v in config.items()
+    }
 
 
 def read_pyproject_toml(
@@ -317,12 +321,7 @@ def read_pyproject_toml(
     default_map: Dict[str, Any] = {}
     if ctx.default_map:
         default_map.update(ctx.default_map)
-    default_map.update(
-        {
-            k: str(v) if not isinstance(v, collections.abc.Iterable) else v
-            for k, v in config.items()
-        }
-    )
+    default_map.update(config)
 
     ctx.default_map = default_map
     return value
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -2,6 +2,7 @@ import ast
 import asyncio
 from abc import ABC, abstractmethod
 from collections import defaultdict
+import collections.abc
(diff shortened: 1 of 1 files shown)


ichard26:
I went with `collections.abc.Iterable` since I don't really know what iterable types the TOML parser might return. If it is overkill, I could just replace it with `list` since from my limited testing, that is the only iterable the TOML parser returns (well other than strings).

ichard26:
Well, I have no experience with JSON, but your argument makes sense. The TOML parser shouldn't change what it returns that much since it does follow the TOML specification, which *shouldn't* change that often. I've changed it to `(list, dict)` just to be safe.

ichard26:
Completely unrelated, but the primer check is still a bit flaky on Windows, this run preferred to fail apparently.
https://github.com/ichard26/black/runs/728070221?check_suite_focus=true

cooperlees:
> EDIT: oh, excuse me?
> 
> > PermissionError: [WinError 32] The process cannot access the file because it is being used by another process

I guess we should open a dedicated issue for this. We can always stop Windows as 99% of value is finding logic area in black, it shouldn't be Windows specific, but, still, always nice to check on all platforms.

Issue #1441, opened 2020-05-21
Test failures with Python 3.9.0b1


hroncok:
**Describe the bug**

With Python 3.9.0b1, there are test failures:

```
======================================================================
FAIL: test_beginning_backslash (tests.test_black.BlackTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib64/python3.9/unittest/mock.py", line 1337, in patched
    return func(*newargs, **newkeywargs)
  File ".../black/tests/test_black.py", line 768, in test_beginning_backslash
    black.assert_equivalent(source, actual)
  File ".../black/src/black/__init__.py", line 6050, in assert_equivalent
    raise AssertionError(
AssertionError: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: 
--- src
+++ dst
@@ -1,17 +1,25 @@
 Module(
   body=
-    Print(
-      dest=
-        None,  # NoneType
-      nl=
-        True,  # bool
-      values=
-        Constant(
-          kind=
-            None,  # NoneType
-          value=
-            b'hello, world',  # bytes
-        )  # /Constant
-    )  # /Print
+    Expr(
+      value=
+        Call(
+          args=
+            Constant(
+              kind=
+                None,  # NoneType
+              value=
+                'hello, world',  # str
+            )  # /Constant
+          func=
+            Name(
+              ctx=
+                Load(
+                )  # /Load
+              id=
+                'print',  # str
+            )  # /Name
+          keywords=
+        )  # /Call
+    )  # /Expr
   type_ignores=
 )  # /Module



======================================================================
FAIL: test_expression (tests.test_black.BlackTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File ".../black/src/black/__init__.py", line 6029, in assert_equivalent
    src_ast = parse_ast(src)
  File ".../black/src/black/__init__.py", line 5948, in parse_ast
    return ast27.parse(src)
  File ".../black/__venv39__/lib64/python3.9/site-packages/typed_ast/ast27.py", line 50, in parse
    return _ast27.parse(source, filename, mode)
  File "<unknown>", line 1
    ...
    ^
SyntaxError: invalid syntax

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.9/unittest/mock.py", line 1337, in patched
    return func(*newargs, **newkeywargs)
[...]

PR #1477, opened 2020-06-03
expression tests: adjust starred expression for Python 3.9 (#1441)

AdamWill:
As discussed in #1441, Python 3.9's new parser will not parse
`(*starred)` even using `compile()` with the `PyCF_ONLY_AST`
flag (as `ast.parse()` does), it raises a `SyntaxError`. This
breaks the four tests that use this file with Python 3.9.
Upstream does not consider this to be a bug - see
https://bugs.python.org/issue40848#msg370643 - so we must
adjust the expression. As suggested by @JelleZijlstra, this just
adds a comma, which makes the new parser happy with it (the old
parser is fine with this form also).

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/tests/data/expression.diff
+++ b/tests/data/expression.diff
@@ -179,7 +179,7 @@
 -((i ** 2) for i, _ in ((1, 'a'), (2, 'b'), (3, 'c')))
 +((i ** 2) for i, _ in ((1, "a"), (2, "b"), (3, "c")))
  (((i ** 2) + j) for i in (1, 2, 3) for j in (1, 2, 3))
- (*starred)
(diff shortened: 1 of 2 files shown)


AdamWill:
so as we can see from the [travis output](https://travis-ci.com/github/psf/black/jobs/343368282), we just have the beginning_backslash problem now. For Fedora's downstream package build I've just done a patch that nerfs that test not to have any empty lines in it, but that's obviously not a correct fix so I won't send a PR, we'll have to wait for upstream to fix the parser I guess.

PR #1486, opened 2020-06-08
Move vim flag cast to outside of get

keith:
With this config:

```toml
[tool.black]
line-length = 79
```

In neovim, this is loaded as a string which later causes an exception to
be thrown. This makes sure the value is always cast to an int. I'm not sure why this wouldn't affect everyone using this behavior.

### DIFF
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -190,7 +190,7 @@ def get_configs():
     toml_config = {}
 
   return {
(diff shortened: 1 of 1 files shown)


keith:
These failures appear to be unrelated to my change 😬

ichard26:
@keith, those failures are to a change in bandersnatch... and I made those changes. Don't worry, I'm at fault for those failtures

keith:
thanks!

PR #1495, opened 2020-06-13
Fix find_pyproject_toml type hint

schmir:
The function is being called with a tuple of strings.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -267,7 +267,7 @@ def supports_feature(target_versions: Set[TargetVersion], feature: Feature) -> b
     return all(feature in VERSION_TO_FEATURES[version] for version in target_versions)
 
 
-def find_pyproject_toml(path_search_start: str) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


ichard26:
Thanks for the PR! Please don't mind the failing primer workflows, that is due a different change in an unrelated project. I might've made those changes in bandersnatch pypa/bandersnatch#561 ... whoops :)

cooperlees:
Pushed the black format fix to stop bandersnatch being evil to our CI: https://github.com/pypa/bandersnatch/commit/0ad7b49fb797a2f0c83be15104183d8a4212d0bc

schmir:
> Makes sense to me - Thanks for the find. I think str is an iterable, thus no error. This is much clearer tho imo.

I'm not sure why mypy doesn't catch that error, it's pretty much obvious if you look at how it's being called:
```
value = find_pyproject_toml(ctx.params.get("src", ()))
```
Maybe you should run mypy with more strict settings.

PR #1517, opened 2020-06-23
Simplify black code by using generator expressions

uriyyo:
I look through `black` code and notice parts which in my opinion can be simplified.

Feel free to add any comments in a case if I missed smth.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -4443,10 +4443,11 @@ def contains_pragma_comment(comment_list: List[Leaf]) -> bool:
         of the more common static analysis tools for python (e.g. mypy, flake8,
         pylint).
     """
-    return any(
-        comment.value.startswith(("# type:", "# noqa", "# pylint:"))
-        for comment in comment_list
-    )
+    for comment in comment_list:
+        if comment.value.startswith(("# type:", "# noqa", "# pylint:")):
+            return True
+
+    return False
 
 
 def insert_str_child_factory(string_leaf: Leaf) -> Callable[[LN], None]:
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1581,10 +1581,11 @@ class Line:
 
     def contains_standalone_comments(self, depth_limit: int = sys.maxsize) -> bool:
         """If so, needs to be split before emitting."""
-        return any(
-            leaf.type == STANDALONE_COMMENT and leaf.bracket_depth <= depth_limit
-            for leaf in self.leaves
-        )
+        for leaf in self.leaves:
+            if leaf.type == STANDALONE_COMMENT and leaf.bracket_depth <= depth_limit:
+                return True
+
+        return False
 
     def contains_uncollapsable_type_comments(self) -> bool:
         ignored_ids = set()
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -619,10 +619,9 @@ def path_empty(
     """
     Exit if there is no `src` provided for formatting
     """
-    if len(src) == 0:
-        if verbose or not quiet:
-            out(msg)
-            ctx.exit(0)
+    if not src and (verbose or not quiet):
+        out(msg)
+        ctx.exit(0)
 
 
 def reformat_one(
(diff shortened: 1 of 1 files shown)


uriyyo:
Hi @cooperlees 
Should we ping someone to review this PR?😊

src/black/__init__.py
```diff
@@ -619,10 +619,9 @@ def path_empty(
     """
     Exit if there is no `src` provided for formatting
     """
-    if len(src) == 0:
-        if verbose or not quiet:
-            out(msg)
-            ctx.exit(0)
+    if not src and (verbose or not quiet):
```
ambv, line 622: It's fine to use a single `if` statement here but keep the `len()` variant. The `src` type is specified as `Sized` so the only safe operation we can do on it is `__len__()`.

src/black/__init__.py
```diff
@@ -1040,7 +1039,7 @@ def get_grammars(target_versions: Set[TargetVersion]) -> List[Grammar]:
 
 def lib2to3_parse(src_txt: str, target_versions: Iterable[TargetVersion] = ()) -> Node:
     """Given a string with source, return the lib2to3 Node."""
-    if src_txt[-1:] != "\n":
+    if not src_txt.endswith("\n"):
```
ambv, line 1042: +1

src/black/__init__.py
```diff
@@ -3991,12 +3989,13 @@ class StringParenWrapper(CustomSplitMapMixin, BaseStringSplitter):
...
-        string_idx = string_idx or self._assert_match(LL)
-        string_idx = string_idx or self._assign_match(LL)
-        string_idx = string_idx or self._dict_match(LL)
+        string_idx = (
+            self._return_match(LL)
+            or self._else_match(LL)
+            or self._assert_match(LL)
+            or self._assign_match(LL)
+            or self._dict_match(LL)
+        )
```
ambv, line 3998: +1

src/black/__init__.py
```diff
@@ -4195,7 +4194,7 @@ def do_transform(self, line: Line, string_idx: int) -> Iterator[TResult[Line]]:
         is_valid_index = is_valid_index_factory(LL)
         insert_str_child = insert_str_child_factory(LL[string_idx])
 
-        comma_idx = len(LL) - 1
+        comma_idx = -1
```
ambv, line 4197: +1

ambv (changes requested):

src/black/__init__.py
uriyyo: I think we can leave this one. 

In case if an object doesn't have an explicit `__bool__` implementation but has `__len__` implemented, then `__len__` result will be used in order to convert an object into bool representation. 

Please correct me if I missed smth.

src/black/__init__.py
```diff
@@ -4443,11 +4442,10 @@ def contains_pragma_comment(comment_list: List[Leaf]) -> bool:
         of the more common static analysis tools for python (e.g. mypy, flake8,
         pylint).
     """
-    for comment in comment_list:
-        if comment.value.startswith(("# type:", "# noqa", "# pylint:")):
-            return True
-
-    return False
+    return any(
```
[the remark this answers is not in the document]
uriyyo: Should I revert those changes?

src/black/__init__.py
```diff
@@ -1582,11 +1581,10 @@ def is_triple_quoted_string(self) -> bool:
 
     def contains_standalone_comments(self, depth_limit: int = sys.maxsize) -> bool:
         """If so, needs to be split before emitting."""
-        for leaf in self.leaves:
-            if leaf.type == STANDALONE_COMMENT and leaf.bracket_depth <= depth_limit:
-                return True
-
-        return False
+        return any(
```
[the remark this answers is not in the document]
uriyyo: Reverted the changes.

src/black/__init__.py
ambv: Yeah, alright.

src/black/__init__.py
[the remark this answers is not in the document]
ambv: For consistency, yes, please.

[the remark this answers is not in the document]
uriyyo: Changes are reverted

ambv:
Thanks! ✨ 🍰 ✨

PR #1527, opened 2020-07-01
Fix attribute error on black module in Vim

logston:
When using Black in Vim, the Vim plugin calls
a function on the black module called "find_pyproject_toml"
and "parse_pyproject_toml". However, these functions are not
yet part of the PyPI release of Black. Thus this Vim plugin
fails with an attribute error. This commit fixes that issue
as well as an issue where "skip_string_normalization" was not
properly being cast to a boolean value.

Sorry, there are no tests for this as I am still learning how
to test vim plugins.

### DIFF
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -51,6 +51,10 @@ import os
 import sys
 import vim
 
+from typing import Any, Dict, Iterable, Optional
+from pathlib import Path
+from functools import lru_cache
+
 
 class Flag(collections.namedtuple("FlagBase", "name, cast")):
   @property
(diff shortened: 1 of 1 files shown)


ichard26:
While I am okay with making the vim plugin more independent of API changes of the Black package, I do not like the duplication of code since that means any changes we do to the main version of those functions will have to be copied over. I can accept having this duplicate code for until we release a new version of Black. I am assuming that you are running the latest version of the vim plugin without updating Black as well. Which is understandable.... But this will keep being a problem going forward until we freeze the APIs (or have an insanely fast release cadence). 

I haven't tested this change yet since I don't use and don't know how to use vim plugins. Ignoring my concerns above, it looks good to me as far looking at a diff can be. I would like to see the bug though since a bug report hasn't been filed for what this PR fixes and I am a little confused about `skip_string_normalization` not being cast properly. (probably due to not knowing how to use vim plugins) I mean, I trust you enough that you know what you doing, but an example of what this is fixing would be nice :)

sak96:
@logston for quick fix you can checkout the tag `19.10b0`  in you `~/.vim/bundle/black` repo. (if you use vundle)
the master vim plugin is in sync with master code. 
though this is bit of pain to manually checkout.

logston:
Thank you both @sak96 and @ichard26. I think I'll close this PR and wait for the next cut of Black. It seems Black is on about a 4-month release cycle. Is that correct?

PR #1532, opened 2020-07-06
add Quora to orgs that use Black

JelleZijlstra:


ichard26:
@JelleZijlstra I'm pretty sure that [this branch](https://github.com/psf/black/tree/JelleZijlstra-patch-1) should be deleted too 🙂

PR #1545, opened 2020-07-11
[vim plugin] Fix g:black_fast and g:black_(skip)_string_normalization opts

bigsmoke:
The was a negation bug in the handling of the `g:black_skip_string_normalization` option in the (n)vim plugin. Also both this option and the `g:black_fast` option were incorrectly casted, from the string `'1'` to `bool()`, so that they were always `True`.

### DIFF
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -41,8 +41,12 @@ endif
 if !exists("g:black_linelength")
   let g:black_linelength = 88
 endif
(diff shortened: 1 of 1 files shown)


rodelrod:
This is related to https://github.com/psf/black/issues/1307
There's a previous PR https://github.com/psf/black/pull/1349

ambv:
Thanks! ✨ 🍰 ✨

PR #1552, opened 2020-07-15
pre-commit: show diff on failure on CI

hugovk:
On the CI, if pre-commit makes changes it fails the build (eg. https://github.com/psf/black/pull/1551), but the reporter cannot see what changes exactly were made, without installing and running pre-commit locally.

But let's show the changes on the CI to save PR authors some hassle.


hugovk:
Example diff in another branch, caused by adding rogue whitespace in the README:

* https://travis-ci.org/github/hugovk/black/jobs/708493625#L300

PR #1559, opened 2020-07-21
Update to accomodate isort 5 release changes.

### LINKED ISSUE #1556: Add isort profiles to compatible configs
As we all know, black enforces certain standards. This can collide with other tools like [isort](https://github.com/timothycrosley/isort).
In the [compatible configurations](https://github.com/psf/black/blob/master/docs/compatible_configs.md) section the necessary changes are already described. This is cumbersome to implement thought.

Luckily isort supports [profiles](https://timothycrosley.github.io/isort/docs/configuration/profiles/) in the config, which works perfectly. a simple `profile=black` in the settings file and it is applied. It would be great to add this notice to the docs and possibly also update the examples accordingly.

I am happy to create a PR if so desired!

mcsitter:
Isort 5 introduced [profiles ](https://timothycrosley.github.io/isort/docs/configuration/profiles/) and ensure_newline_before_comments options. Either one needs to be added to work correctly with black.

Using `ensure_newline_before_comments = True` does no harm and is required for new isort releases, so it is appropriate to be put into the config examples.

Previous isort versions will not be configured with `profile = black` so the note should suffice for now. Accomodating users of previous isort versions is appropriate, especially since there are still isort incompatibilities with pylint (see PyCQA/pylint#3725) and flake8-isort (see gforcada/flake8-isort#92).

Fix #1556

### DIFF
--- a/docs/compatible_configs.md
+++ b/docs/compatible_configs.md
@@ -67,17 +67,11 @@ of 88 characters via `line_length = 88` as well as
 `ensure_newline_before_comments = True` to ensure spacing import sections with comments
 works the same as with _Black_.
 
----
-
-NOTE
-
-`ensure_newline_before_comments = True` only works since isort >= 5 but does not break
-older versions so you can keep it if you are running previous versions. If only isort >=
-5 is used you can add `profile = black` instead of all the options since
-[profiles](https://timothycrosley.github.io/isort/docs/configuration/profiles/) are
-available and do the configuring for you.
-
----
+**Please note** `ensure_newline_before_comments = True` only works since isort >= 5 but
+does not break older versions so you can keep it if you are running previous versions.
+If only isort >= 5 is used you can add `profile = black` instead of all the options
+since [profiles](https://timothycrosley.github.io/isort/docs/configuration/profiles/)
+are available and do the configuring for you.
 
 ### Formats
 
--- a/docs/compatible_configs.md
+++ b/docs/compatible_configs.md
@@ -72,7 +72,7 @@ works the same as with _Black_.
 NOTE
 
 `ensure_newline_before_comments = True` only works since isort >= 5 but does not break
-previous ones so you can keep it if you are running previous versions. If only isort >=
+older versions so you can keep it if you are running previous versions. If only isort >=
 5 is used you can add `profile = black` instead of all the options since
 [profiles](https://timothycrosley.github.io/isort/docs/configuration/profiles/) are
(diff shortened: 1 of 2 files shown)


docs/compatible_configs.md
```diff
@@ -62,7 +63,21 @@ The option `force_grid_wrap = 0` is just to tell isort to only wrap imports that
...
+
+NOTE
+
+`ensure_newline_before_comments = True` only works since isort >= 5 but does not break
+previous ones so you can keep it if you are running previous versions. If only isort >=
+5 is used you can add `profile = black` instead of all the options since
+[profiles](https://timothycrosley.github.io/isort/docs/configuration/profiles/) are
+available and do the configuring for you.
+
+---
```
ichard26, line 80: I personally do not like the use of visible `<br>`s here since they feel more segregating than headers IMO. I recommend either using a level 4 header (e.g. `#### NOTE: with isort 5+`) or do something similar to the note at the top of this doucment. OR, you could change the sectioning with two sections, one for isort 4=< with a note regarding the mostly useless option, and the other section for isort 5+

Your call, but IMO, the current formatting looks a bit ugly right now.

docs/compatible_configs.md
```diff
@@ -62,7 +63,21 @@ The option `force_grid_wrap = 0` is just to tell isort to only wrap imports that
...
+of 88 characters via `line_length = 88` as well as
+`ensure_newline_before_comments = True` to ensure spacing import sections with comments
+works the same as with _Black_.
+
+---
+
+NOTE
+
+`ensure_newline_before_comments = True` only works since isort >= 5 but does not break
+previous ones so you can keep it if you are running previous versions. If only isort >=
```
ichard26, line 75: nit
```suggestion
older versions so you can keep it if you are running previous versions. If only isort >=
```

docs/compatible_configs.md
mcsitter: Since this a black repository I should not come up with new ideas for how to style things and adopted the style already present in the document.

mcsitter:
Thank you for the review @ichard26 !

I hope this PR is now satisfying to be merged. I am of course happy to make changes.

I also updated the isort config example in the code style document. They should match across documents.

ichard26:
> I also updated the isort config example in the code style document. They should match across documents.

I forgot to update the embedded version in the README when writing this document originally and I also forgot when reviewing your PR so thanks for that :)

Issue #1557, opened 2020-07-21
Add test for `A᧚ = 0`


Zac-HD:
Black v19.10b0 fails to parse certain assignments involving unicode identifiers - [playground link here](https://black.now.sh/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ABLADtdAD2IimZxl1N_WjMy7A7oUimP5kl9tNitfjGTgMhZej2xgomiikPHniF7YMrqeF7JYab2JGKtxYQLJtMAAACEQNE3-XEpLQABV0wDcxaqH7bzfQEAAAAABFla).

```python
A᧚
A፩
```

This code is in fact valid Python, as you can confirm by pasting it into a repl or with `compile("A\u19da = 0")`.

Found, as with most of my bugs, via [Hypothesmith](https://github.com/Zac-HD/hypothesmith).  Given that this applies to multiple unicode digit characters, it might be due to unicode digits in identifiers?

ichard26:
Yapf has the same problem, https://github.com/google/yapf/issues/403, `lib2to3` (of which `blib2to3` is built on) doesn't really have full unicode support.

ichard26:
Resolved by https://github.com/psf/black/commit/5e2bb528e09df368ed7dea6b7fb9c53e799a569f first available in 21.12b0 which broadened the acceptable character set for NAMES significantly. Thanks for reporting!

JelleZijlstra:
We should add a regression test to make sure this doesn't regress if we change our tokenizer.

Issue #1012, opened 2019-09-10
Black fails to tokenise files ending with a backslash


Zac-HD:
Given a file containing a backslash preceeded and followed by any number of newlines, Black ae5588 and 19.3b0 throw `blib2to3.pgen2.tokenize.TokenError: 'EOF in multi-line statement', (2, 0)`.

I consider this a bug because Python is perfectly happy to execute such files, doing nothing, and `compile("\\", "<string>", "exec")` also works:

```pycon
>>> code = compile("\\", "<string>", "exec")  # or "\\\n", or "\n\\\n", etc.
>>> import dis; dis.dis(code)
  1           0 LOAD_CONST               0 (None)
              2 RETURN_VALUE
```

Like #970, I found this with [Hypothesmith](https://github.com/Zac-HD/hypothesmith).

Zac-HD:
This is still present in Black 19.10b0 - it's a different bug to #922/#948; Python ignores a trailing backslash but Black chokes on it.

jayaddison:
It looks like Python's built-in `compile` behaviour became stricter between py37 and py38; a trailing line continuation statement is no longer accepted.

```pycon
Python 3.7.7 (default, Apr  1 2020, 13:48:52) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> compile('\\', '<STRING>', 'exec')
<code object <module> at 0x7f60bd565270, file "<STRING>", line 1>
>>>
```

```pycon
Python 3.8.7 (default, Dec 22 2020, 10:37:26) 
[GCC 10.2.1 20201207] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> compile('\\', '<STRING>', 'exec')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<STRING>", line 1
    \
    ^
SyntaxError: unexpected EOF while parsing
>>> 
```

This could still be addressed; there's some work-in-progress included in #1961.  Does anyone have suggestions on how best to proceed?

Zac-HD:
"Ignore this until py37 reaches end of life" seems like a reasonable plan to me, and it's easy enough to adjust the tests accordingly.

jayaddison:
Another example where this has surfaced during fuzzer testing, after merging #1991:

https://github.com/psf/black/pull/1958/checks?check_run_id=1945936278

```
Falsifying example: test_idempotent_any_syntatically_valid_python(
[...]

Issue #970, opened 2019-08-07
Parse error found with Hypothesmith


Zac-HD:
```python
weird = "pass #\r#\n"
exec(weird)                                     # works, as does `compile` etc.
black.format_str(weird, mode=black.FileMode())  # fails with `black.InvalidInput`
```

Operating system: Windows
Python version: 3.7.3
*Black* version: 19.3b0
Does also happen on master: yes (website hangs)

I found this weird little string with [`hypothesmith`](https://pypi.org/project/hypothesmith/), a tool for generating Python source code with Hypothesis that I started at the PyCon NA sprints and published at the PyCon AU sprints.  Currently it works based on an approximate grammar, post-processing, and filtering.

CC @ambv who I know was interested in fuzzing 😄.  Next steps... I think a combination of targeted mutation (open PR upstream) and spending more than a few minutes of CPU time could turn up a lot more weird things!

zsol:
Nice!

zsol:
FWIW this is also a bug in upstream lib2to3. You might want to report it on http://bugs.python.org/

MeGaGiGaGon:
This appears to have been solved at some point. If I set up the file using `Alt-013` for `CR` and `Alt-010` for `LF`
![image](https://github.com/user-attachments/assets/136e7e3a-8042-4225-87d4-3e04648c139d)
And after running it through black
![image](https://github.com/user-attachments/assets/a399c0b9-52e7-49b5-8e5d-5d6925f73004)
I couldn't figure out how to type just `CR`/`LF` on the playground, so no playground link.

This also appears to have been fixed at some point in `lib2to3` 
```
$ py -3.12 -c "from lib2to3.pgen2.tokenize import generate_tokens;from io import StringIO;print(*generate_tokens(StringIO('pass #\r#\n').readline))"
<string>:1: DeprecationWarning: lib2to3 package is deprecated and may not be able to parse Python 3.10+
(1, 'pass', (1, 0), (1, 4), 'pass #\r#\n') (53, '#', (1, 5), (1, 6), 'pass #\r#\n') (58, '\r', (1, 6), (1, 7), 'pass #\r#\n') (53, '#', (1, 7), (1, 8), 'pass #\r#\n') (4, '\n', (1, 8), (1, 9), 'pass #\r#\n') (0, '', (2, 0), (2, 0), '')
```
But I couldn't find the relevant issues/PRs, and it doesn't really matter since `lib2to3` is gone in `3.13`.

[...]

PR #1566, opened 2020-07-30
Add a property-based test using Hypothesis & Hypothesmith

Zac-HD:
At PyCon 2019, I gave a few talks about property-based testing, and also spoke to @ambv about generating test inputs for Black.  There's a long history of this working really well for compilers (e.g. [CSmith](https://embed.cs.utah.edu/csmith/)), and for formatting tools like https://github.com/prettier/prettier/issues/371.

So starting at the sprints I wrote [Hypothesmith](https://pypi.org/project/hypothesmith/) to generate Python source code, [reported a bunch of bugs](https://github.com/psf/black/issues/created_by/Zac-HD), and eventually after some more upgrades thought that it might be useful to test `black` in its own test suite as well as my downstream code.  So here we are!

This test will fail due to #970, #1012, #1557, and possibly other bugs.  While Hypothesis will present (mostly) minimal failing inputs and can report multiple bugs based on exception type and location, you can expect to see mostly trivial or who-would-ever-write-that failures for a while.  However, fixing those edge cases will allow Hypothesmith to explore elsewhere and find more subtle bugs.


JelleZijlstra:
This is a great tool, thank you!

I'm not sure a normal unit test is a good place for this sort of test though. For one, it's currently failing, so we have to first fix the bugs you linked. As I understand it, it's also partly random, which is great for finding new edge cases, but means that people submitting PRs or running tests locally may see tests failing randomly because we happened to find a previously undetected bug.

Zac-HD:
Glad you like it!  There are definitely some questions about how you want to use it, but I thought that opening a PR would give us a more concrete starting point for discussion than an issue :smile: 

> For one, it's currently failing, so we have to first fix the bugs you linked. 

Yeah... the obvious options are (1) don't merge it yet, (2) `@unittest.expectedFailure`, or (3) give known bugs a distinct exception type which can be skipped within the test.  Order of increasing work and also order of increasing chance to find other bugs, for what that's worth.

> It's also partly random, which is great for finding new edge cases, but means that people submitting PRs or running tests locally may see tests failing randomly because we happened to find a previously undetected bug.

Using the `derandomize=True` setting would guarantee non-flakiness... it's even possible to use a settings profile to make it quick-and-deterministic or run many random examples depending on e.g. an environment variable.  [*Two kinds of testing*](https://blog.nelhage.com/post/two-kinds-of-testing/) describes this pretty nicely.

'Sleeper' test failures are pretty rare though - upstream in Hypothesis we might have two or three per year over a few thousand tests.  Once property-based tests are passing they don't tend to discover more failures, but you can easily run them in a loop overnight if that's a concern and better tooling for this is planned.

ichard26:
Thanks for the PR! I recently was introduced to Hypothesis and Hypothesmith while contributing to flake8-bugbear and I really like it! I do have a bit of concerns though, here's my (quickly wrote down) two cents.

As a regular contributor, I second JelleZijlstra's slight apprehension about having Hypothesmith as a normal unit test. Since the flow we somewhat document on creating a new PR on the Black repository includes running all unit tests, quite a few PR contributors would hit the Hypothesmith test failing (from looking how many build jobs failed on this PR). Tests are usually designed to either make sure something works as expected or a bug wasn't reintroduced, i.e. making sure stuff still works. While Hypothesmith pushes the program further and further so it can find unseen bugs.

Regular unit tests and Hypothesmith tests are similar, but have different enough goals that they have different appropriate reactions to when they fail. This would add more complexity for a new contributor or a drive-by minor contributor since they might not be familiar enough with project and/or Hypothesmith to know/understand what to do next. Experienced contributor can deal with this new test once they quickly get up to speed with it, but means contributing to Black has a bit more friction. Although, good documentation on this special test could go a long way on making this easier for newer contributor.

Another problem for new / newer contributor is that even if they submit a simple typo fix PR fully through the web interface, there is a non-small chance that the checks will fail due to luck. And if they aren't familiar with Hypothesmith or CI in general, they will probably end up confused. A red X isn't nice to see, and definitely not for newer contributors. Not the greatest experience for someone joining Black, the PSF org, or even open source (and on this front, Black can definitely do better).

TL;DR is that this addition complicates the road of contributing to Black, mostly for newer and less experienced contributors. 

And no, I still want to see Hypothesis and Hypothesmith bring better testing to Black, but as you have pointed out, this is merely the start of a discussion :)

Zac-HD:
Thanks for your comment @ichard26!  I think you're suggesting that we 

- make this test deterministic with the `derandomize=True` setting
- and/or move it out of the main test suite

which makes sense to me.

cooperlees:
Thanks @Zac-HD! +1 on separate test here. Maybe only on Ubuntu too and I’ll be happy to merge!

Zac-HD:
@cooperlees - how will you want to run this?  A new `tests/test_fuzz.py` using unittest, a standalone script to run with `python fuzztests.py`, or what?

cooperlees:
> @cooperlees - how will you want to run this? A new `tests/test_fuzz.py` using unittest, a standalone script to run with `python fuzztests.py`, or what?

I feel a new fuzz_tests.yml GitHub Action workflow only on ubuntu unless you see value testing in different OS's was what I envisioned
- cp .github/workflows/test.yml .github/workflows/fuzz.yml
- edit
- Win

Maybe even run it with coverage to see how much of the code the fuxx tests hit for some signal too?
- coverage run fuzz.py
- coverage report

https://github.com/psf/black/blob/master/.github/workflows/test.yml#L28

PR #1569, opened 2020-08-03
Prefer single quotes

mccolljr:
TODO: update tests


mccolljr:
This was a test for internal use, please ignore. I fully understand why Black doesn't desire to support this.

PR #1570, opened 2020-08-03
Refactor normalize_string_quotes

mccolljr:
This PR does not affect the behavior of the tool, but introduces semantically named variables (`good_docquote`, `good_quote`, `bad_docquote`, and `bad_quote`) inside the `normalize_string_quotes` method. Initially this was done to aid in my own understanding of the logic (particularly the edge case handling for docstrings, towards the end).

This isn't meant as a tuning knob, no parameter is exposed for users of this tool to control these values. Could a sufficiently clever developer choose to patch this at installation time (at their own risk)? Possibly.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -5048,19 +5048,24 @@ def normalize_string_quotes(leaf: Leaf) -> None:
 
     Note: Mutates its argument.
     """
+    good_docquote = '"""'
+    bad_docquote = "'''"
+    good_quote = '"'
+    bad_quote = "'"
+
     value = leaf.value.lstrip(STRING_PREFIX_CHARS)
-    if value[:3] == '"""':
+    if value[:3] == good_docquote:
         return
 
-    elif value[:3] == "'''":
-        orig_quote = "'''"
-        new_quote = '"""'
-    elif value[0] == '"':
-        orig_quote = '"'
-        new_quote = "'"
+    elif value[:3] == bad_docquote:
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Sorry, but I'm not sure this is an improvement—to me it makes the code harder to read.

mccolljr:
For me, seeing `'"""'`, `"'''"`, `'"'`, and `"'"` repeated made things more difficult to visually scan & grok. Font size, font face, etc can make those a little difficult to tell apart at a glance, whereas giving those strings descriptive names and using those in the appropriate places improved my own understanding, and that of a few others I spoke with on my team. In my experience it has been preferable to avoid magic strings and magic numbers where possible.

That said, I won't be offended if the opinion of the team is that this isn't something you want to merge. I made this change to better understand the code flow and thought that it might benefit others, as well.

cooperlees:
So 1:1 with devs - Can we get a tie breaker other black core devs! :)

src/black/__init__.py
```diff
@@ -5102,15 +5107,15 @@ def normalize_string_quotes(leaf: Leaf) -> None:
                 # Do not introduce backslashes in interpolated expressions
                 return
 
-    if new_quote == '"""' and new_body[-1:] == '"':
+    if new_quote == good_docquote and new_body[-1:] == good_docquote[0]:
```
showell, line 5110: Isn't `good_docquote[0]` just `good_quote`?

PR #1574, opened 2020-08-07
docs(gh_actions): fixes #1573

### LINKED ISSUE #1573: GitHub actions error prior stable release of lint.yml
Hi, I get the following error when using GitHub Actions with black:

```
Current runner version: '2.263.0'
Operating System
Virtual Environment
Prepare workflow directory
Prepare all required actions
Download action repository 'actions/checkout@v2'
Download action repository 'actions/setup-python@v2'
Download action repository 'psf/black@stable'
##[error]Could not find file '/home/runner/work/_actions/_temp_e6a60df6-7298-4553-9c69-9049a1c24037/_staging/psf-black-6bedb5c/docs/editor_integration.md'.
```

The .yml file is the one suggested in the README as follows:

```
name: Lint

on: [push, pull_request]

jobs:
  lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: psf/black@stable
```

_Originally posted by @Ayrx in https://github.com/psf/black/pull/1421#issuecomment-638182443_

Fixed by using `psf/black@master`.

scratchmex:
fixes #1573

### DIFF
--- a/docs/integrations/github_actions.md
+++ b/docs/integrations/github_actions.md
@@ -25,6 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      # you can change the tag @stable with any of the available here: https://github.com/psf/black/tags
       - uses: psf/black@stable
 ```
 
--- a/docs/integrations/github_actions.md
+++ b/docs/integrations/github_actions.md
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      # you can change the tag @stable with any of the available here: https://github.com/psf/black/tags
+      # you can change "stable" to a specific release: https://github.com/psf/black/tags
       - uses: psf/black@stable
 ```
 


WhyNotHugo:
This change isn't a good idea; we don't want to encourage users to use black/master in their pipelines.

The issue you're having is that you're reading a README for an unreleased version; it'll become valid once it's released.

ichard26:
I agree with @WhyNotHugo, we shouldn't encourage folks to use master in CI pipelines since Black is beta software and is subject to small/large formatting changes.

This would be fixed by a new release anyway :) Sorrry for me shooting your PR down, but this is inadvisable for production environments.

scratchmex:
The problem is you are forcing to use the master branch anyway as months ago. The new release should be already published. And finally, black appears to be beta-always software. 

I agree about not encouraging to use master in our ci pipelines and that's why I'm putting an explicit message taking about it. The problem here is the releases are not that frequent.

hugovk:
Maybe only recommend version tags for GitHub Actions? 

All the other actions I've used use a (major) version tag, so you're not surprised by breakages. 

(Similarly, someone in another issue was surprised to get dozens of CI failures when a new release came out, because the CI used unpinned Black.)  

Black uses CalVer not SemVer,  so a full version tag may be best.

ichard26:
The more I think about it, the more I concur with @hugovk. While tracking stable is nice and all with automatic upgrades, it does lead to a lot of sudden CI failures when a new released is pushed. Anyway, for pre-commit even though we suggest stable, pre-commit doesn't automatically upgrade Black. So suggesting version pinning would also help reduce behavioural differences between pre-commit and CI.

scratchmex:
> 
> 
> Maybe only recommend version tags for GitHub Actions?
> 
> All the other actions I've used use a (major) version tag, so you're not surprised by breakages.
> 
> (Similarly, someone in another issue was surprised to get dozens of CI failures when a new release came out, because the CI used unpinned Black.)
> 
> Black uses CalVer not SemVer, so a full version tag may be best.

Feel free to suggest a help text

README.md
```diff
@@ -403,7 +403,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - uses: psf/black@stable
+      # check out if this workflow has been released @stable to avoid getting stuck on @20.8b1
```
hugovk, line 406: Maybe just delete the comment? Or suggest to set to the newest version? 
```suggestion
```

JelleZijlstra:
This has unfortunately been left hanging for a while and there is a merge conflict now.

I don't really like the comment that was added in the current version of the PR; as a user I wouldn't understand what I'm expected to do with it.

felix-hilden:
The branch was forced, so I don't know what was previously there. This seems like a nice, instructive comment though.

WRT the merge conflict: @scratchmex I think the place to put that would be the [integrations documentation](https://black.readthedocs.io/en/stable/integrations/github_actions.html). And actually, this is pretty similar to what's on the [VCS integration docs](https://black.readthedocs.io/en/stable/integrations/source_version_control.html) already. We're so close to merging this! Are you available to make the final changes Ivan?

scratchmex:
You are so right. I didn't know who put that comment but that was my initial attempt xd

scratchmex:
Sorry for the forcing. The PR is now ready to merge.

docs/integrations/github_actions.md
```diff
@@ -25,6 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      # you can change the tag @stable with any of the available here: https://github.com/psf/black/tags
```
JelleZijlstra, line 28:
```suggestion
      # you can change "stable" to a specific release: https://github.com/psf/black/tags
```

felix-hilden (approved):
Seems good to me now that Jelle got his edit in as well! So how bout a merge?

JelleZijlstra (approved):

JelleZijlstra:
After all this back-and-forth we already ended up with very similar wording in the docs: https://black.readthedocs.io/en/stable/integrations/github_actions.html. The extra line from this PR feels redundant, so I'm going to close this. Thanks for the work on getting this clearer!

Issue #1575, opened 2020-08-07
Slicing of bracketed expressions is broken by removal of the brackets


michael-sayapin:
**Describe the bug**

Slicing of bracketed expressions is broken by too eager removal of the brackets.

**To Reproduce**

1. Try to reformat a file containing `return ('%x' % value)[-12:]`
2. The resulting expression is not producing equivalent result: `return '%x' % mac[-12:]`

**Expected behavior**

Expected to not remove the brackets in this case.

**Environment (please complete the following information):**

- Version: master
- OS and Python version: Linux, Python 3.7.4

**Does this bug also happen on master?** To answer this, you have two options:

Yes.

JelleZijlstra:
I can't reproduce this on master:
```
$ echo "('%x' % value)[-12:]" | python3 -m black -
("%x" % value)[-12:]
reformatted -
All done! ✨ 🍰 ✨
1 file reformatted.
```

ichard26:
@JelleZijlstra Interestingly I can reproduce this on master.

| Condition | Command | Output  |
|---|---|---|
| Black with `-c` | `black -c "return ('%x' % value)[-12:]"` | `return "%x" % value[-12:]` |
| Black with a file  | `black test.py` | `error: cannot format test.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.` |
| Black with input read from `stdin` | `echo ('%x' % value)[-12:] \| black -`  | ` error: cannot format -: INTERNAL ERROR: Black produced code that is not equivalent to the source.` |

This was my session:

```console
(black-rHKUX7ap) R:\Programming\black>echo return ('%x' % value)[-12:] > test.py
                                                                                                                                                                                                                                                                
(black-rHKUX7ap) R:\Programming\black>type test.py                                                                                                                      
return ('%x' % value)[-12:]
                                                                                                                                                                                                                                                                                                                     
(black-rHKUX7ap) R:\Programming\black>black test.py --check                                                                                                                                                                                                                    
error: cannot format test.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: C:\Users\User\AppData\Local\Temp\blk_0tsx_tzh.log                                                                                             
Oh no! 💥 💔 💥                                                                                                                                                         
1 file would fail to reformat.
                                                                                                                                                                                                                                                                                                                  
[...]

PR #1590, opened 2020-08-12
don't strip brackets before lsqb (#1575)

davidszotten:
it's not safe to remove brackets that are followed by a `[`


davidszotten:
i may need some help running `black-primer`, not familiar with it and not sure i ran it correctly:

```
$black-primer
[2020-08-12 17:23:13,146] INFO: 16 projects to run Black over (lib.py:311)
[2020-08-12 17:23:16,081] INFO: Finished aioexabgp (lib.py:284)
[2020-08-12 17:23:23,151] INFO: Finished bandersnatch (lib.py:284)
[2020-08-12 17:23:23,621] INFO: Finished attrs (lib.py:284)
[2020-08-12 17:23:23,622] INFO: Skipping django as it's disabled via config (lib.py:254)
[2020-08-12 17:23:27,017] INFO: Finished flake8-bugbear (lib.py:284)
[2020-08-12 17:23:27,127] INFO: Finished channels (lib.py:284)
[2020-08-12 17:23:27,127] INFO: Skipping pandas as it's disabled via config (lib.py:254)
[2020-08-12 17:23:53,564] INFO: Finished poetry (lib.py:284)
[2020-08-12 17:23:58,732] INFO: Finished ptr (lib.py:284)
[2020-08-12 17:24:10,766] INFO: Finished hypothesis (lib.py:284)
[2020-08-12 17:24:10,766] INFO: Skipping pytest as it's disabled via config (lib.py:254)
[2020-08-12 17:24:45,690] INFO: Finished pyramid (lib.py:284)
[2020-08-12 17:25:03,774] INFO: Finished tox (lib.py:284)
[2020-08-12 17:25:18,675] INFO: Finished virtualenv (lib.py:284)
[2020-08-12 17:27:03,179] INFO: Finished warehouse (lib.py:284)
[2020-08-12 17:29:11,824] INFO: Finished sqlalchemy (lib.py:284)
[2020-08-12 17:29:11,825] INFO: Analyzing results (lib.py:327)
-- primer results 📊 --

13 / 16 succeeded (81.25%) ✅
0 / 16 FAILED (0.0%) 💩
 - 3 projects disabled by config
 - 0 projects skipped due to Python version
 - 0 skipped due to long checkout
```

src/black/__init__.py
```diff
@@ -3287,8 +3287,11 @@ def do_match(self, line: Line) -> TMatchResult:
                 and LL[next_idx].type == token.RPAR
                 and not is_empty_rpar(LL[next_idx])
             ):
-                # That RPAR should NOT be followed by a '.' symbol.
-                if is_valid_index(next_idx + 1) and LL[next_idx + 1].type == token.DOT:
+                # That RPAR should NOT be followed by a '.' or a '['
+                if is_valid_index(next_idx + 1) and LL[next_idx + 1].type in {
```
JelleZijlstra, line 3291: Should also update the docstring on line 3241 below.

src/black/__init__.py
```diff
@@ -3287,8 +3287,11 @@ def do_match(self, line: Line) -> TMatchResult:
...
+                    token.DOT,
+                    token.LSQB,
```
JelleZijlstra, line 3293: These similar cases still crash:
- `echo '("x" % 3)(1)' | black -`.
- `echo '("x" % 3) ** 1' | black -`.
Could you fix these too?

(Essentially, we should skip anything lower in https://docs.python.org/3/reference/expressions.html#operator-precedence than % is.)

davidszotten: sure. i don't quite understand why only `%` has this problem. (e.g. why does `'("x" or 3)(1)'` work fine?)

davidszotten: ah, there are more issues. operators in the same group are ok after, because it works left to right, but e.g. `'2*("" % a)` still breaks. investigating

davidszotten: ok, added another commit for that

davidszotten: hm, this might be quite a strong restriction, other stuff is failing now.  maybe we need to detect the percent sign and only conditionally skip rules?  thoughts?

JelleZijlstra: Looks like the reason only % is affected is the code in the StringParser class that this class relies on to decide how big the string is. The code looks right to me now and has thorough tests, so I'll merge the PR. Thanks for your contribution!

davidszotten:
thanks for the review!

(though i think there are still edge cases missing. in particular, i don't think my "unary op" detection is right: `'a + -("" % a)'` still fails)

davidszotten:
i think this is a better way to detect unary ops:

```
@@ -3318,7 +3318,8 @@ class StringParenStripper(StringTransformer):
                     )
                     or (
                         # only unary PLUS/MINUS
-                        not is_valid_index(idx - 3)
+                        LL[idx - 2].parent
+                        and LL[idx - 2].parent.type == syms.factor
                         and (LL[idx - 2].type in {token.PLUS, token.MINUS})
                     )
```

i'd raise a pr but mypy gets upset: `src/black/__init__.py:3322:29: error: Item "None" of "Optional[Node]" has no attribute "type"`

the source position is pointing at `LL[idx - 2].parent[.type]` but i check it just above..

PR #1591, opened 2020-08-13
Make --exclude only apply to recursively found files

### LINKED ISSUE #1572: File excluded even if explicitly passed via the command line when specified via --exclude
**Describe the bug**

`--exclude` is supposed to only apply to the recursive discovery of files. Not files explicitly passed to Black through the command line. This is a regression since #1032 was merged.

**To Reproduce** 

1. Create a directory for testing
2. Create and activate a virtual enviroment
3. Install Black on master using `pip install git+git://github.com/psf/black`
4. Create two empty files:
4.1. `pyproject.toml` so Black doesn't find a different configuration file up your directory structure
4.2. `test.py` - this will be our test file
5. Run `black -v --exclude=test.py test.py`
6. Install Black 19.10b0 using `pip install black==19.10b0`
7. Run `black -v --exclude=test.py test.py`
8. Observe that on master the file was excluded and on stable the file was formatted.

**Expected behavior**

Black formats `test.py` since it was explicitly passed through the command line and `--exclude` should not touch such files like it.  

**Environment**

- Version: master (b59a5246577346e6da2cc2802015f08524abf545) and stable (`19.10b0`)
- OS and Python version: `Ubuntu 18.04.04 LTS` and `Python 3.8.1`

*Yes and also N/A - This is a regression*

**Additional context**

I haven't looked into this bug that much, but the problem seems to be related to how even files passed via the command line are given to the recursive `gen_python_files` function as single Path object lists in `get_sources`.
https://github.com/psf/black/blob/b59a5246577346e6da2cc2802015f08524abf545/src/black/__init__.py#L605-L610

ichard26:
Resolves #1572.

**What this PR does:**

- Makes `--exclude` only apply to files recursively discovered in `gen_python_files()` instead of applying to all files, including files given to Black through the CLI
- Adds a test to make sure this doesn't go unnoticed again
- Also changes the output given when a file is excluded to properly say which regex matched
- Fixes a few misplaced or wrong comments

**How was this achieved?**

Before, `get_sources()` when encountering a file that was passed explicitly through the CLI would pass a single Path object list to
`gen_python_files()`. This causes bad behaviour since that function doesn't treat the `exclude` and `force_exclude` regexes differently.  Now when `get_sources()` iterates through `srcs` and encounters a file, it checks if the `force_exclude` regex matches, if not, then the file will be added to the computed sources set.

**What changes of Black's behaviour does this PR cause?**

`--exclude` will now behave as it did on `stable`.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -583,9 +583,6 @@ def get_sources(
     root = find_project_root(src)
     sources: Set[Path] = set()
     path_empty(src, "No Path provided. Nothing to do 😴", quiet, verbose, ctx)
-    exclude_regexes = [exclude_regex]
-    if force_exclude_regex is not None:
-        exclude_regexes.append(force_exclude_regex)
     gitignore = get_gitignore(root)
 
     for s in src:
@@ -596,7 +593,8 @@ def get_sources(
                     p.iterdir(),
                     root,
                     include_regex,
-                    exclude_regexes,
+                    exclude_regex,
+                    force_exclude_regex,
                     report,
                     gitignore,
                 )
@@ -604,12 +602,12 @@ def get_sources(
         elif s == "-":
             sources.add(p)
         elif p.is_file():
-            # Hard-exclude any files that matches the `--force-exclude` regex.
             normalized_path = normalize_path_maybe_ignore(p, root, report)
             if normalized_path is None:
                 continue
 
             normalized_path = "/" + normalized_path
+            # Hard-exclude any files that matches the `--force-exclude` regex.
             if force_exclude_regex:
                 force_exclude_match = force_exclude_regex.search(normalized_path)
             else:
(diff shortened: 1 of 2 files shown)


src/black/__init__.py
```diff
@@ -597,17 +598,27 @@ def get_sources(
...
             )
         elif s == "-":
             sources.add(p)
         elif p.is_file():
-            sources.update(
-                gen_python_files(
-                    [p], root, None, exclude_regexes, report, get_gitignore(root)
-                )
-            )
+            # Hard-exclude any files that matches the `--force-exclude` regex.
```
JelleZijlstra, line 607: This comment makes more sense on line 613 below

src/black/__init__.py
```diff
@@ -5767,44 +5801,34 @@ def gen_python_files(
...
             continue
 
+        # Then ignore with `exclude` and `--force-exclude` options.
         normalized_path = "/" + normalized_path
         if child.is_dir():
             normalized_path += "/"
 
         is_excluded = False
-        for exclude in exclude_regexes:
+        for opt_name, exclude in zip(("--exclude", "--force-exclude"), exclude_regexes):
```
JelleZijlstra, line 5828: This makes me a bit uneasy because it assumes that `exclude_regexes` exactly matches these two flags, when the type doesn't demand that. It would be cleaner to make exclude_regex and force_exclude_regex separate arguments to this function, or to make `exclude_regexes` an iterable of (option name, pattern) pairs.

ichard26:
Thank you @JelleZijlstra for the review! I've addressed both of your comments. I've gone with moving the `force_exclude` regex into its own argument as it allows way for `--extend-exclude` to be added.

src/black/__init__.py
ichard26: Yeah, this is not my cleanest code ever. I have already thought about moving the force exclude regex into its own argument but I decided not doing so as it would have make more sense when `--extend-exclude` is implemented. No problems doing it now than later!

PR #1599, opened 2020-08-13
Accept src as option in pyproject.toml (fixes #861)

### LINKED ISSUE #861: Allow specifying SRC in pyproject.toml
Currently only options can be passed in pyproject.toml, however it would be convenient to pass arguments, specifically SRC . A use case which I have in mind is a gradual migration to black formatting important especially for big projects. It would also allow to track the migration in VCS through changes in pyproject.toml.

It is much more sensible solution than hacking with regex as proposed in https://github.com/python/black/issues/810

e.g. mypy offers such functionality through the _files_ option as described at https://mypy.readthedocs.io/en/latest/config_file.html#config-file-import-discovery-global

mrijken:
Add option to add src in pyprojcect.toml so black can be run without command line parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -613,6 +613,7 @@ Multiple contributions by:
 - [LukasDrude](mailto:mail@lukas-drude.de)
 - Mahmoud Hossam
 - Mariatta
+- [Marc Rijken](mailto:marc@rijken.org)
 - [Matt VanEseltine](mailto:vaneseltine@gmail.com)
 - [Matthew Clapp](mailto:itsayellow+dev@gmail.com)
 - [Matthew Walster](mailto:matthew@walster.org)
--- a/README.md
+++ b/README.md
@@ -322,7 +322,8 @@ exclude = '''
 '''
 ```
 
-Note: `src`  specifies a list of source files / directories which will be formatted (which can be specified as command line argument)
+Note: `src` specifies a list of source files / directories which will be formatted
+(which can be specified as command line argument)
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -322,8 +322,9 @@ exclude = '''
 '''
 ```
 
-Note: `src` specifies a list of source files / directories which will be formatted
-(which can be specified as command line argument)
+Tip: `src` specifies a list of default source files / directories which will be
+formatted when _Black_ is run with no sources provided. Sources given on the command
+line will override the list of sources provided in the configuration file.
 
 </details>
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -610,6 +614,7 @@ Multiple contributions by:
 - [LukasDrude](mailto:mail@lukas-drude.de)
 - Mahmoud Hossam
 - Mariatta
+- [Marc Rijken](mailto:marc@rijken.org)
```
hugovk, line 617: Alphabetically:

```suggestion
- [Marc Rijken](mailto:marc@rijken.org)
- Mariatta
```

README.md
```diff
@@ -321,6 +322,9 @@ exclude = '''
 '''
 ```
 
+Note: `src` specifies a list of source files / directories which will be formatted
+(which can be specified as command line argument)
```
ichard26, line 326: I'm not so sure I like the wording here, perhaps: 
```suggestion
Tip: `src` specifies a list of default source files / directories which will be formatted
when _Black_ is run with no sources provided. Sources given on the command line
will override the list of sources in `pyproject.toml`.
```

(please wordsmith more to your liking - it might be a bit too long 😉 )

ambv:
The patch itself looks fine but I'm concerned about the usability of this. I commented on the linked issue.

tests/test_black.py
```diff
@@ -1832,6 +1833,42 @@ def test_read_pyproject_toml(self) -> None:
...
+        )
+        config = fake_ctx.default_map
+        self.assertEqual(config["verbose"], "1")
+        self.assertEqual(config["check"], "no")
+        self.assertEqual(config["diff"], "y")
+        self.assertEqual(config["color"], "True")
+        self.assertEqual(config["line_length"], "79")
+        self.assertEqual(config["target_version"], ["py36", "py37", "py38"])
+        self.assertEqual(config["exclude"], r"\.pyi?$")
+        self.assertEqual(config["include"], r"\.py?$")
```
felix-hilden, line 1869: I'm not sure if these tests are needed here, since they are already in other functions. Removing them would simplify the test a lot.

tests/test_src.toml
```diff
@@ -0,0 +1,10 @@
+[tool.black]
+verbose = 1
+--check = "no"
+diff = "y"
+color = true
+line-length = 79
+target-version = ["py36", "py37", "py38"]
+exclude='\.pyi?$'
+include='\.py?$'
```
felix-hilden, line 9: Maybe reduce the configuration here too since it's only for testing the src argument.

felix-hilden:
But I may have jumped the gun, because the change itself is still being discussed. My bad!

mrijken:
Closing because there is not enough support for this PR.

felix-hilden:
Thank you for giving it a go though, and for the great discussion around it. It's good that we really got it figured out exactly why this is not what we want to do. Much appreciated! 🖤

PR #1600, opened 2020-08-14
fix unary op detection

davidszotten:
(also need to refactor slightly to help mypy)

naming `before_lpar`:

```
3302           before_lpar = LL[idx - 2]
3303           LL[idx - 2].parent and reveal_type(LL[idx - 2].parent)
3304           before_lpar.parent and reveal_type(before_lpar.parent)

3303:52: note: Revealed type is 'Union[blib2to3.pytree.Node, None]'
3304:52: note: Revealed type is 'blib2to3.pytree.Node'
```

update: looks like this is https://github.com/python/mypy/issues/2199


JelleZijlstra:
Thanks!

Issue #1598, opened 2020-08-13
Unstable formatting on subscripted assignment + string + comment


JelleZijlstra:
```
$ cat test_ranking_store.py 
value.__dict__[
    key
] = "test"  # set some Thrift field to non-None in the struct aa bb cc dd ee
$ black -v test_ranking_store.py 
Using configuration from /Users/jelle/py/black/pyproject.toml.
error: cannot format test_ranking_store.py: INTERNAL ERROR: Black produced different code on the second pass of the formatter.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /var/folders/wx/zkst6jdx1zg_s32d23gz4_0r0000gn/T/blk_mf5ru27i.log
Oh no! 💥 💔 💥
1 file failed to reformat.
$ cat  /var/folders/wx/zkst6jdx1zg_s32d23gz4_0r0000gn/T/blk_mf5ru27i.log
--- source
+++ first pass
@@ -1,3 +1,5 @@
 value.__dict__[
     key
-] = "test"  # set some Thrift field to non-None in the struct aa bb cc dd ee
+] = (  # set some Thrift field to non-None in the struct aa bb cc dd ee
+    "test"
+)
--- first pass
+++ second pass
@@ -1,5 +1,3 @@
 value.__dict__[
     key
-] = (  # set some Thrift field to non-None in the struct aa bb cc dd ee
-    "test"
-)
+] = "test"  # set some Thrift field to non-None in the struct aa bb cc dd ee
```

hugovk:
`git bisect` points to 544ea9c217cad9459d7b60665db787e94b52f93d as the first bad commit, from PR https://github.com/psf/black/pull/1132 "Improve String Handling".

Issue #1597, opened 2020-08-13
Crash on assert and parenthesized % format


JelleZijlstra:
```
$ cat author.py 
class xxxxxxxxxxxxxxxxxxxxx(xxxx.xxxxxxxxxxxxx):
    def xxxxxxx_xxxxxx(xxxx):
        assert xxxxxxx_xxxx in [
            x.xxxxx.xxxxxx.xxxxx.xxxxxx,
            x.xxxxx.xxxxxx.xxxxx.xxxx,
        ], ("xxxxxxxxxxx xxxxxxx xxxx (xxxxxx xxxx) %x xxx xxxxx" % xxxxxxx_xxxx)
$ black -v author.py 
Using configuration from /Users/jelle/py/black/pyproject.toml.
Traceback (most recent call last):
  File "/Users/jelle/py/black/src/black/__init__.py", line 648, in reformat_one
    if changed is not Changed.CACHED and format_file_in_place(
  File "/Users/jelle/py/black/src/black/__init__.py", line 791, in format_file_in_place
    dst_contents = format_file_contents(src_contents, fast=fast, mode=mode)
  File "/Users/jelle/py/black/src/black/__init__.py", line 912, in format_file_contents
    dst_contents = format_str(src_contents, mode=mode)
  File "/Users/jelle/py/black/src/black/__init__.py", line 977, in format_str
    for line in transform_line(
  File "/Users/jelle/py/black/src/black/__init__.py", line 2731, in transform_line
    result.extend(
  File "/Users/jelle/py/black/src/black/__init__.py", line 2725, in transform_line
    for transformed_line in transform(line, features):
  File "/Users/jelle/py/black/src/black/__init__.py", line 2834, in __call__
    for line_result in self.do_transform(line, string_idx):
  File "/Users/jelle/py/black/src/black/__init__.py", line 3315, in do_transform
    append_leaves(new_line, line, LL[: string_idx - 1])
  File "/Users/jelle/py/black/src/black/__init__.py", line 4601, in append_leaves
    new_line.append(new_leaf)
  File "/Users/jelle/py/black/src/black/__init__.py", line 1425, in append
    self.bracket_tracker.mark(leaf)
  File "/Users/jelle/py/black/src/black/__init__.py", line 1291, in mark
    opening_bracket = self.bracket_match.pop((self.depth, leaf.type))
KeyError: (-1, 10)
error: cannot format author.py: (-1, 10)
Oh no! 💥 💔 💥
1 file failed to reformat.
```

[...]

Issue #1596, opened 2020-08-13
Crash on concatenated string + comment


JelleZijlstra:
Another crash on master
```
$ cat xxx.py 
xxxxxxx_xxxxxx_xxxxxxx = xxx(
    [
        xxxxxxxxxxxx(
            xxxxxx_xxxxxxx=(
                '((x.xxxxxxxxx = "xxxxxx.xxxxxxxxxxxxxxxxxxxxx") || (x.xxxxxxxxx = "xxxxxxxxxxxx")) && '
                # xxxxx xxxxxxxxxxxx xxxx xxx (xxxxxxxxxxxxxxxx) xx x xxxxxxxxx xx xxxxxx.
                "(x.xxxxxxxxxxxx.xxx != "
                '"xxx:xxx:xxx::xxxxxxxxxxxx:xxxxxxx-xxxx/xxxxxxxxxxx/xxxxxxxxxxxxxxxxx") && '
            )
        )
    ]
)
$ black -v xxx.py 
Using configuration from /Users/jelle/py/black/pyproject.toml.
Traceback (most recent call last):
  File "/Users/jelle/py/black/src/black/__init__.py", line 648, in reformat_one
    if changed is not Changed.CACHED and format_file_in_place(
  File "/Users/jelle/py/black/src/black/__init__.py", line 791, in format_file_in_place
    dst_contents = format_file_contents(src_contents, fast=fast, mode=mode)
  File "/Users/jelle/py/black/src/black/__init__.py", line 918, in format_file_contents
    assert_stable(src_contents, dst_contents, mode=mode)
  File "/Users/jelle/py/black/src/black/__init__.py", line 6076, in assert_stable
    newdst = format_str(dst, mode=mode)
  File "/Users/jelle/py/black/src/black/__init__.py", line 977, in format_str
    for line in transform_line(
  File "/Users/jelle/py/black/src/black/__init__.py", line 2731, in transform_line
    result.extend(
  File "/Users/jelle/py/black/src/black/__init__.py", line 2725, in transform_line
    for transformed_line in transform(line, features):
  File "/Users/jelle/py/black/src/black/__init__.py", line 2834, in __call__
    for line_result in self.do_transform(line, string_idx):
  File "/Users/jelle/py/black/src/black/__init__.py", line 2978, in do_transform
    msg_result = self.__merge_string_group(new_line, string_idx)
  File "/Users/jelle/py/black/src/black/__init__.py", line 3164, in __merge_string_group
[...]

Issue #1595, opened 2020-08-13
Unstable formatting on split string + % formatting


JelleZijlstra:
This is on master.
```
$ cat kill_senior_procs.py 
if __name__ == "__main__":
    for i in range(4, 8):
        cmd = (
            r"for pid in $(ps aux | grep paster | grep -v grep | grep '\-%d' | awk '{print $2}'); do kill $pid; done"
            % (i)
        )
$ black kill_senior_procs.py 
error: cannot format kill_senior_procs.py: INTERNAL ERROR: Black produced different code on the second pass of the formatter.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /var/folders/wx/zkst6jdx1zg_s32d23gz4_0r0000gn/T/blk_bgpw4vyu.log
Oh no! 💥 💔 💥
1 file failed to reformat.
$ cat /var/folders/wx/zkst6jdx1zg_s32d23gz4_0r0000gn/T/blk_bgpw4vyu.log
--- source
+++ first pass
@@ -1,6 +1,7 @@
 if __name__ == "__main__":
     for i in range(4, 8):
         cmd = (
-            r"for pid in $(ps aux | grep paster | grep -v grep | grep '\-%d' | awk '{print $2}'); do kill $pid; done"
+            r"for pid in $(ps aux | grep paster | grep -v grep | grep '\-%d' | awk"
+            r" '{print $2}'); do kill $pid; done"
             % (i)
         )
--- first pass
+++ second pass
@@ -1,7 +1,6 @@
 if __name__ == "__main__":
     for i in range(4, 8):
         cmd = (
             r"for pid in $(ps aux | grep paster | grep -v grep | grep '\-%d' | awk"
-            r" '{print $2}'); do kill $pid; done"
-            % (i)
+            r" '{print $2}'); do kill $pid; done" % (i)
         )
```

hugovk:
And a fourth `git bisect` pointing to 544ea9c217cad9459d7b60665db787e94b52f93d from "Improve String Handling" (#1132).

PR #1609, opened 2020-08-18
Disable string splitting/merging by default

JelleZijlstra:
The changes in #1132 fixed a number of long-standing limitations in Black, but unfortunately they also introduced a number of new crashes. As we're preparing for a release, I'm proposing that we disable this new feature for now until we're able to make it more stable. We'll add an undocumented `--experimental-string-processing` flag that enables advanced string processing. Once we've fixed all the related crashes, we can remove the flag and turn on the feature by default.

Fixes:
- #1598
- #1597
- #1596
- #1595


hugovk:
Primer on the CI is failing because it expects changes in attrs, hpyothesis and poetry.

Testing this PR, there are changes with `--experimental-string-processing`, but not by default. So the config can be updated to expect no changes with those three. 

Please see https://github.com/JelleZijlstra/black/pull/1 to update the config.

PR #1612, opened 2020-08-20
Re-implement magic trailing comma handling

### LINKED ISSUE #1288: Finish "magic trailing comma" handling
#826 introduced the concept of a magic trailing comma: if you, the programmer, explicitly put a trailing comma in a collection, it signals to Black that you want a given collection exploded into multiple lines, always. While this flies a bit in the way of "doesn't look at pre-existing formatting" and "non-configurability", it is extremely pragmatic and consistent with "make diffs as small as possible".

However, there's issues with it so far. It doesn't yet work on nested collections. Worse yet, in those cases it leaves trailing commas behind when a collection ends up compressed into one line. We need to fix those edge cases to make it generally dependable.

ambv:
Re-implement magic trailing comma handling:
 
- when a trailing comma is specified in any bracket pair, that signals to Black that this bracket pair needs to be always exploded, e.g. presented as "one item per line";

- this causes some changes to previously formatted code that erroneously left trailing commas embedded into single-line expressions;

- internally, Black needs to be able to identify trailing commas that it put itself compared to pre-existing trailing commas. We do this by using/abusing lib2to3's `was_checked` attribute.  It's True for internally generated trailing commas and False for pre-existing ones (in fact, for all pre-existing leaves and nodes).

Fixes #1288

Thank you, [Sentry](https://sentry.io/), for sponsoring this effort.


ambv:
Primer will obviously fail due to expected formatting changes. We'll be dealing with the other failures here, looks like they're infrastructure failures.

ambv:
LOL, such a rookie mistake. Will fix momentarily.

zsol:
This looks great! Much clearer logic than before. I spot checked some of the primer changes and they look good too

PR #1621, opened 2020-08-25
Address pre-existing trailing commas when not in the rightmost bracket pair

### LINKED ISSUE #1619: Black always explodes the rightmost bracket pair when a trailing comma is present
```
--- /tmp/wat.py	2020-08-24 11:48:49.409868 +0000
+++ /tmp/wat.py	2020-08-24 11:49:16.096574 +0000
@@ -1,5 +1,4 @@
 if False:
-    x = {
-        "a": 1,
-        "b": 2,
-    }["a"]
+    x = {"a": 1, "b": 2,}[
+        "a"
+    ]
```

_Originally posted by @GianlucaFicarelli in https://github.com/psf/black/issues/1288#issuecomment-679061119_

ambv:
This required some hackery.  Long story short, we need to reuse the ability to omit rightmost bracket pairs (which glues them together and splits on something else instead), for use with pre-existing trailing commas.

This form of user-controlled formatting is brittle so we have to be careful not to cause a scenario where Black first formats code without trailing commas in one way, and then looks at the same file with pre-existing trailing commas (that it itself put on the previous run) and decides to format the code again.

One particular ugly edge case here is handling of optional parentheses.  In particular, the long-standing `line_length=1` hack got in the way of pre-existing trailing commas and had to be removed.  Instead, a more intelligent but costly solution was put in place: a "second opinion" if the formatting that omits optional parentheses ended up causing lines to be too long.  Again, for efficiency purposes, Black reuses Leaf objects from blib2to3 and modifies them in place, which was invalid for having two separate formattings.  Line cloning was used to mitigate this.

Fixes #1619


tests/data/cantfit.py
```diff
@@ -67,11 +67,15 @@
 normal_name = (
     but_the_function_name_is_now_ridiculously_long_and_it_is_still_super_annoying()
 )
-normal_name = but_the_function_name_is_now_ridiculously_long_and_it_is_still_super_annoying(
-    arg1, arg2, arg3
+normal_name = (
+    but_the_function_name_is_now_ridiculously_long_and_it_is_still_super_annoying(
+        arg1, arg2, arg3
+    )
```
ambv, line 73: Note: formatting improvements!

tests/data/cantfit.py
```diff
@@ -67,11 +67,15 @@
...
 )
-normal_name = but_the_function_name_is_now_ridiculously_long_and_it_is_still_super_annoying(
-    [1, 2, 3], arg1, [1, 2, 3], arg2, [1, 2, 3], arg3
+normal_name = (
+    but_the_function_name_is_now_ridiculously_long_and_it_is_still_super_annoying(
+        [1, 2, 3], arg1, [1, 2, 3], arg2, [1, 2, 3], arg3
+    )
```
ambv, line 78: Ditto.

tests/data/function_trailing_comma.py
```diff
@@ -9,6 +9,12 @@ def f2(a,b,):
 def f(a:int=1,):
     call(arg={'explode': 'this',})
     call2(arg=[1,2,3],)
+    x = {
+        "a": 1,
+        "b": 2,
+    }["a"]
+    if a == {"a": 1,"b": 2,"c": 3,"d": 4,"e": 5,"f": 6,"g": 7,"h": 8,}["a"]:
+        pass
```
ambv, line 17: This is what this PR is really fixing.

tests/data/function_trailing_comma.py
```diff
@@ -51,6 +57,21 @@ def f(
...
+        "a": 1,
+        "b": 2,
+        "c": 3,
+        "d": 4,
+        "e": 5,
+        "f": 6,
+        "g": 7,
+        "h": 8,
+    }["a"]:
+        pass
```
ambv, line 74: Formats as expected.

tests/data/long_strings_flag_disabled.py
```diff
@@ -133,14 +133,11 @@
...
 
-old_fmt_string3 = (
-    "Whereas only the strings after the percent sign were long in the last example, this example uses a long initial string as well. This is another %s %s %s %s"
-    % (
-        "really really really really really",
-        "old",
-        "way to format strings!",
-        "Use f-strings instead!",
-    )
+old_fmt_string3 = "Whereas only the strings after the percent sign were long in the last example, this example uses a long initial string as well. This is another %s %s %s %s" % (
```
ambv, line 136: Now Black doesn't use optional parentheses that do nothing.

src/black/__init__.py
```diff
@@ -2722,6 +2730,7 @@ class StringTransformer(ABC):
 
     line_length: int
     normalize_strings: bool
+    __name__ = "StringTransformer"
```
JelleZijlstra, line 2733: What does this do?

src/black/__init__.py
```diff
@@ -6371,21 +6405,124 @@ def can_omit_invisible_parens(line: Line, line_length: int) -> bool:
...
+        line_str = line_to_string(line)
+    result: List[Line] = []
+    for transformed_line in transform(line, features):
+        if str(transformed_line).strip("\n") == line_str:
+            raise CannotTransform("Line transformer returned an unchanged result")
+
+        result.extend(transform_line(transformed_line, mode=mode, features=features))
+
+    if not (
+        transform.__name__ == "rhs"
```
JelleZijlstra, line 6503: That's a bit hacky; maybe we should turn the Transformer type into a pair instead to store this extra information.

tests/test_black.py
```diff
@@ -346,11 +346,16 @@ def test_function2(self) -> None:
         black.assert_stable(source, actual, DEFAULT_MODE)
 
     @patch("black.dump_to_file", dump_to_stderr)
-    def test_function_trailing_comma_wip(self) -> None:
-        source, expected = read_data("function_trailing_comma_wip")
-        # sys.settrace(tracefunc)
-        actual = fs(source)
-        # sys.settrace(None)
+    def _test_wip(self) -> None:
```
JelleZijlstra, line 349: Should this be removed?

ambv: I guess it does no harm and I keep recreating this kind of mini-test when poking at a problem, so I just kept it, just disabled. ¯\_(ツ)_/¯

src/black/__init__.py
ambv: Yes, it's hacky but then again redoing all transformers into classes for this one hack felt like over-engineering.

src/black/__init__.py
ambv: This gives a name to class-based transformers which otherwise would have to be queried with `__class__.__name__` instead.

PR #1455, opened 2020-05-22
Be more careful with docstrings ending with quotes + whitespace

alexmv:
The PEP 257 algorithm used in #1053 results in trimming trailing
whitespace in docstrings -- see #1415.

Adjust the algorithm to preserve leading and trailing whitespace; this
reverts the changes in #1417.  This diverges from PEP 257, but better
retains the contents of the docstring.

Fixes #1452 because we no longer can end up with four trailing quotes.

PR #1623, opened 2020-08-25
Improve docstring re-indentation handling

ambv:
This addresses a few crashers, namely:

* producing non-equivalent code due to mangling escaped newlines,

* invalid hugging quote characters in the docstring body to the docstring outer triple quotes (causing a quadruple quote which is a syntax error),

* lack of handling for docstrings that start on the same line as the `def`, and

* invalid stripping of outer triple quotes when the docstring contained a string prefix.

As a bonus, tests now also run when string normalization is disabled.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -2037,13 +2037,20 @@ class LineGenerator(Visitor[Line]):
         yield from self.visit_default(node)
 
     def visit_STRING(self, leaf: Leaf) -> Iterator[Line]:
-        # Check if it's a docstring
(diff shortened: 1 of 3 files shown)


ichard26:
> invalid hugging quote characters in the docstring body to the docstring outer triple quotes (causing a quadruple quote which is a syntax error),

So effectively #1455 is superseded and should be closed once this is merged. I'm not missing something here, right?

ambv:
Right, @ichard26.

PR #1627, opened 2020-08-26
Treat all trailing commas as pre-existing, as they effectively are

ambv:
On a second pass of Black on the same file, inserted trailing commas are now pre-existing.  Doesn't make sense to differentiate between the passes then.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1627,14 +1627,13 @@ class Line:
 
     def maybe_should_explode(self, closing: Leaf) -> bool:
         """Return True if this line should explode (always be split), that is when:
-        - there's a pre-existing trailing comma here; and
+        - there's a trailing comma here; and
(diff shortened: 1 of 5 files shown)


ambv:
Sadly, this doesn't address the known unstable formatting issues because by the time the trailing comma was added on a line during the first pass of Black, lines are already pre-split.

src/black/__init__.py
```diff
@@ -2623,7 +2622,7 @@ def init_st(ST: Type[StringTransformer]) -> StringTransformer:
     transformers: List[Transformer]
     if (
         not line.contains_uncollapsable_type_comments()
-        and not line.should_explode
+        and not line.should_explode  # XXX suspicious
```
ambv, line 2625: This is probably the line that causes instability but it cannot be easily removed.

PR #1637, opened 2020-08-26
Fix multiline docstring quote normalization

### LINKED ISSUE #1634: skip string normalization no longer ignores triple quote strings, like docstrings
**Describe the bug** 

previously running with `--skip-string-normalization` would ignore triple-quoted strings that used single quote character `'`

1. Take this file '...'

```
def test_function():
    '''this is my docstring'''
    pass
```
2. Run _Black_ on it with these arguments '....'
`--skip-string-normalization`
3. See error
it modifies the triple-quoted strings to double-quote character

do the same steps after installing `19.10.0b`

**Expected behavior** 

skip string normalization should ignore triple quoted strings

- Version: 20.8b1
- OS and Python version: linux python3.6

yes

**Additional context** 
n/a

ichard26:
Fix #1634.

The quotes of multiline docstrings are now only normalized when string
normalization is off, instead of the string normalization setting being
ignored and the quotes being *always* normalized.

I had to make a new test case and data file since the current pair for
docstrings only worked when there is no formatting difference between the
formatting results with string normalization on and off. I needed to add
tests for when there *are* differences between the two. So I split
test_docstring's test code when string normalization is disabled into a
new test case along with a new data file.

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -517,11 +517,16 @@ class BlackTestCase(unittest.TestCase):
         self.assertFormatEqual(expected, actual)
         black.assert_equivalent(source, actual)
         black.assert_stable(source, actual, DEFAULT_MODE)
+
+    @patch("black.dump_to_file", dump_to_stderr)
+    def test_docstring_no_string_normalization(self) -> None:
+        """Like test_docstring but with string normalization off."""
+        source, expected = read_data("docstring_no_string_normalization")
         mode = replace(DEFAULT_MODE, string_normalization=False)
-        not_normalized = fs(source, mode=mode)
-        self.assertFormatEqual(expected, not_normalized)
-        black.assert_equivalent(source, not_normalized)
-        black.assert_stable(source, not_normalized, mode=mode)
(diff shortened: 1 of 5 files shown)


tests/test_black.py
```diff
@@ -517,11 +517,16 @@ def test_docstring(self) -> None:
...
+        source, expected = read_data("docstring_no_string_normalization")
         mode = replace(DEFAULT_MODE, string_normalization=False)
-        not_normalized = fs(source, mode=mode)
-        self.assertFormatEqual(expected, not_normalized)
-        black.assert_equivalent(source, not_normalized)
-        black.assert_stable(source, not_normalized, mode=mode)
+        actual = fs(source, mode=mode)
+        self.assertFormatEqual(expected, actual)
+        black.assert_equivalent(source, actual)
+        black.assert_stable(source, actual, mode)
```
ichard26, line 529: This is a terrible diff.... 🙁

ambv:
Thanks! ✨ 🍰 ✨

PR #1646, opened 2020-08-28
Fix empty line handling when formatting typing stubs

### LINKED ISSUE #1490: black removes empty line between imports and overloads in stubs
Please consider the follow **stub** file (e.g. `foo.pyi`):

```python
import y
  
@overload
def foo(): ...
```

Running `black foo.pyi` (no configuration file) removes the empty line 2.

- Version: 19.10b0
- OS and Python version: Linux/Python 3.8.3

Using the online formatter: yes.

ichard26:
Fixes #1490.

Black used to erroneously remove all empty lines between functions
preceded by decorators and non-function code in pyi mode. Now a single
empty line is enforced.


src/black/__init__.py
```diff
@@ -1847,8 +1847,11 @@ def _maybe_empty_lines_for_class_or_def(
                     newlines = 0
                 else:
                     newlines = 1
-            elif current_line.is_def and not self.previous_line.is_def:
-                # Blank line between a block of functions and a block of non-functions
+            elif (
+                current_line.is_def or current_line.is_decorator
+            ) and not self.previous_line.is_def:
```
ichard26, line 1852: Right now on master Black enforces this formatting:
```pyi
class A: ...

@bar
class B: ...
    def BMethod(self) -> Any: ...

class C: ...
```

While the empty line between class B and C is right, I am not sure if the empty line between class A and class B is right since class A has an empty body. That empty line happens since Black can't tell the difference between the previous class not having an empty body or the current class being preceded by a decorator. Currently this commit keeps this behaviour since as far as I know classes don't have a reason to be decorated in stubs, but no idea whether that's right or not... 

/cc @ambv @JelleZijlstra (I barely have any experience with typing stubs so your knowledge would be appreciated)

JelleZijlstra: I would put a blank line there. Nontrivial classes (with a body that's not `...`) should have an empty line on either side.

JelleZijlstra: And there are some use cases for decorating classes in stubs, like `@runtime_checkable` for protocols.

ichard26: Sorry, I meant for the case of:
```pyi
class A: ...

@bar
class B: ...
class C: ...
```

The empty line between class A and B doesn't seem right as A and B have empty bodies. Master and this commit enforces this formatting, but I feel like it's wrong.

JelleZijlstra: Oh I see, I could go either way on that. I feel like if we do add blank lines, it should have blank lines both above and below it.

JelleZijlstra (approved):

JelleZijlstra:
I fixed the merge conflict, now there are some primer failures which look unrelated to this change; they're in flake8-bugbear in non-stub code.

hugovk:
Flake8-bugbear is now Black-formatted (https://github.com/PyCQA/flake8-bugbear/pull/136 (some familiar names there!)), restarting the Primer run should hopefully be green now.

JelleZijlstra:
Now it's failing on virtualenv, likely because https://github.com/pypa/virtualenv/commit/13c0498a47acf8a0aad6db9ceb1033dfe5f119a9 reformatted it.

hugovk:
PR https://github.com/psf/black/pull/1695 flips the Primer switch for virtualenv.

PR #1650, opened 2020-08-28
Fix typo in comment

mithrandi:
Hope you don't mind the one-char PR.


fuzz.py
```diff
@@ -43,7 +43,7 @@ def test_idempotent_any_syntatically_valid_python(
         dst_contents = black.format_str(src_contents, mode=mode)
     except black.InvalidInput:
         # This is a bug - if it's valid Python code, as above, black should be
```
hugovk, line 45: If you want to make it a two-char PR :) 

```suggestion
        # This is a bug - if it's valid Python code, as above, Black should be
```

PR #1655, opened 2020-08-29
Fix incorrect space before colon in if/while stmts

### LINKED ISSUE #1174: Incorrect Space Added after Assignment Expression in If Statement
**Description**
Black on the [current master@c53853f](https://github.com/psf/black/tree/c53853f5716f2b5100b17b8f8509e026dd7fa4e2) and ([19.10b0](https://github.com/psf/black/tree/19.10b0)) adds a space after [Assignment Expressions](https://www.python.org/dev/peps/pep-0572/) and before the final colon in `if` statements.

For what it's worth, I'm not sure if this is actually intended behavior or not, but it doesn't seem like it. No other parenthesized expressions in `if` statements get this treatment.

**To Reproduce**:

1. Put this example in a file:
    ```python
    if (foo := True):
        print(foo)
    ```
2. Run `black <file> --diff`

**Expected behavior**
Black should not need to act upon this file. There should be no diff output and black should claim `1 file would be left unchanged.`

**Actual behavior**
Black currently inserts a space between the end of the assignment expression and the colon.

Actual `black <file> --diff` output:
```diff
--- blacktest.py        2019-11-23 00:56:09.070688 +0000
+++ blacktest.py        2019-11-23 00:56:15.282759 +0000
@@ -1,3 +1,3 @@
-if (foo := True):
+if (foo := True) :
     print(foo)
 ```

**Environment:**

- Version:  [current master@c53853f
](https://github.com/psf/black/tree/c53853f5716f2b5100b17b8f8509e026dd7fa4e2)
- OS and Python version: Windows 10 and Linux tested with Python 3.8.0

Yes

**Other Information**
This also affects `if` statements that are broken up into several lines.

_The following example comes come a current project, excuse the specificity. The `# pylint: disable=superfluous-parens` should be treated simply as a comment for the scope of this issue. (In reality, pylint is also currently having [a problem with assignment expressions](https://github.com/PyCQA/pylint/issues/3249#issuecomment-557462095))_
```python
class Foo:
    def method(self):
        if (
            invalid_reasons := self.invalid_reasons()
        ):  # pylint: disable=superfluous-parens
            LOGGER.error(
                f"Not saving raid for the following reasons: {', '.join(invalid_reasons)}"
            )
            return

```

```diff
--- blacktest.py        2019-11-23 01:07:27.116369 +0000
+++ blacktest.py        2019-11-23 01:07:30.956945 +0000
@@ -1,10 +1,10 @@
 class Foo:
     def method(self):
         if (
             invalid_reasons := self.invalid_reasons()
-        ):  # pylint: disable=superfluous-parens
+        ) :  # pylint: disable=superfluous-parens
             LOGGER.error(
                 f"Not saving raid for the following reasons: {', '.join(invalid_reasons)}"
             )
             return
```

### LINKED ISSUE #1588: Internal Error ") :"
The formatter notices this bug itself, and politely asked me to file an issue:

```
INTERNAL ERROR: Black produced different code on the second pass of the formatter.  
Please report a bug on https://github.com/psf/black/issues.
```

Here's a repo:
1. take this file
```
very_long_variable_name = 0
very_very_long_variable_name = 1
extraordinarily_long_variable_name = 2

if assignment_expression := very_long_variable_name + very_very_long_variable_name > extraordinarily_long_variable_name:
    print("fubar")
```

2. format it, and get the error with the following log
```
--- source
+++ first pass
@@ -1,8 +1,10 @@
-
 very_long_variable_name = 0
 very_very_long_variable_name = 1
 extraordinarily_long_variable_name = 2
 
-if assignment_expression := very_long_variable_name + very_very_long_variable_name > extraordinarily_long_variable_name:
+if (
+    assignment_expression := very_long_variable_name + very_very_long_variable_name
+    > extraordinarily_long_variable_name
+):
     print("fubar")
 
--- first pass
+++ second pass
@@ -3,8 +3,8 @@
 extraordinarily_long_variable_name = 2
 
 if (
     assignment_expression := very_long_variable_name + very_very_long_variable_name
     > extraordinarily_long_variable_name
-):
+) :
     print("fubar")

```

- Version: black == 19.10b0
- OS and Python version: Linux/Python 3.8.5

ichard26:
Fixes #1174 
Fixes #1588

Previously Black would format this code

```
if (foo := True):
	print(foo)
```

as

```
if (foo := True) :
	print(foo)
```

adding an incorrect space after the RPAR. Buggy code in the
normalize_invisible_parens function caused the colon to be wrapped in
invisible parentheses. The LPAR of that pair was then prefixed with a
single space at the request of the whitespace function.

This commit fixes the accidental skipping of a pre-condition check
which must return True before parenthesis normalization of a specific
child Leaf or Node can happen. The pre-condition check being skipped
was why the colon was wrapped in invisible parentheses.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,9 @@
 - `Black` now respects `--skip-string-normalization` when normalizing multiline
   docstring quotes (#1637)
 
+- `Black` no longer adds an incorrect space after a parenthesized assignment expression
+  in if/while statements (#1655)
+
 - fixed a crash when PWD=/ on POSIX (#1631)
 
 ### 20.8b1
--- a/docs/change_log.md
+++ b/docs/change_log.md
@@ -9,6 +9,11 @@
 - `Black` now respects `--skip-string-normalization` when normalizing multiline
   docstring quotes (#1637)
 
+- `Black` no longer adds an incorrect space after a parenthesized assignment expression
+  in if/while statements (#1655)
+
+- fixed a crash when PWD=/ on POSIX (#1631)
+
 ### 20.8b1
 
 #### _Packaging_
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -5190,9 +5190,9 @@ def normalize_invisible_parens(node: Node, parens_after: Set[str]) -> None:
 
         if check_lpar:
             if is_walrus_assignment(child):
-                continue
+                pass
 
-            if child.type == syms.atom:
+            elif child.type == syms.atom:
                 if maybe_make_parens_invisible_in_atom(child, parent=node):
                     wrap_in_parentheses(node, child, visible=False)
             elif is_one_tuple(child):
--- a/tests/data/pep_572.py
+++ b/tests/data/pep_572.py
@@ -2,6 +2,8 @@
 (a := a)
 if (match := pattern.search(data)) is None:
     pass
+if (match := pattern.search(data)):
+    pass
 [y := f(x), y ** 2, y ** 3]
 filtered_data = [y for x in data if (y := f(x)) is None]
 (y := f(x))
@@ -40,4 +42,6 @@ foo((b := 2), a=1)
 foo(c=(b := 2), a=1)
 
 while x := f(x):
+    pass
+while (x := f(x)):
     pass


ichard26:
I just realized that I forgot to add an entry in `CHANGES.md`, please wait for me to add an entry. Sorry!

JonZeolla:
Any chance this can be included in the [upcoming release](https://github.com/psf/black/projects/4)?

ichard26:
I'm confused, we release what's been merged into master, so unless a commit reintroduces the bug (which will get detected as I added a test case), the next release will include this fix.

JonZeolla:
Great to hear.  I mistook the beta tag and the use of "stable" on read the docs for an expectation of a more immediate non-beta release, but in [reviewing the documentation](https://black.readthedocs.io/en/stable/installation_and_usage.html#note-this-is-a-beta-product) and prior releases, that doesn't appear to be likely.  I will keep an eye on future releases

Issue #1666, opened 2020-09-02
ImportError: cannot import name 'NoReturn'


jonathan-s:
**Describe the bug** A clear and concise description of what the bug is.

Running black in the console generates a traceback. 

**To Reproduce** Steps to reproduce the behavior:
Running black directly in the console gives me the following traceback. 

```
Traceback (most recent call last):
  File "/Users/jonathan/projects/recorded/integrations/splunk/venv/bin/black", line 7, in <module>
    from black import patched_main
  File "/Users/jonathan/projects/recorded/integrations/splunk/venv/lib/python3.6/site-packages/black/__init__.py", line 57, in <module>
    from blib2to3 import pygram, pytree
  File "/Users/jonathan/projects/recorded/integrations/splunk/venv/lib/python3.6/site-packages/blib2to3/pygram.py", line 13, in <module>
    from .pgen2 import driver
  File "/Users/jonathan/projects/recorded/integrations/splunk/venv/lib/python3.6/site-packages/blib2to3/pgen2/driver.py", line 39, in <module>
    from . import grammar, parse, token, tokenize, pgen
  File "/Users/jonathan/projects/recorded/integrations/splunk/venv/lib/python3.6/site-packages/blib2to3/pgen2/pgen.py", line 7, in <module>
    from typing import (
ImportError: cannot import name 'NoReturn'
```

**Expected behavior** A clear and concise description of what you expected to happen.

Shouldn't have any error. 

**Environment (please complete the following information):**

- Version: black (20.8b1)
- OS and Python version: OSX, python 3.6.1

JelleZijlstra:
`typing.NoReturn` was added in 3.6.2 (https://docs.python.org/3/library/typing.html#typing.NoReturn). I'd accept a PR that falls back to `typing_extensions.NoReturn` for 3.6.0 and 3.6.1. Maybe it's time to upgrade though :)

jonathan-s:
Thanks for the heads up. :) Perhaps an update of the documentation will do? Dropping 3.6.1

hugovk:
Please see PR https://github.com/psf/black/pull/1668 to update the README, and the `black.vim` and `setup.py` checks.

JelleZijlstra:
[...]

PR #1668, opened 2020-09-02
Black requires Python 3.6.2+

hugovk:
Fixes https://github.com/psf/black/issues/1666.

Update README to say 3.6.2+ is required, not 3.6.0+.

Plus update the checks in `black.vim` and `setup.py` to check for 3.6.2+ rather than 3.6+ aka 3.6.0+.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ _Contents:_ **[Installation and usage](#installation-and-usage)** |
 
 ### Installation
 
-_Black_ can be installed by running `pip install black`. It requires Python 3.6.0+ to
+_Black_ can be installed by running `pip install black`. It requires Python 3.6.2+ to
 run but you can reformat Python 2 code with it, too.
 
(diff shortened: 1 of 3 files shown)


cooperlees (approved):

ichard26 (approved):
I'm now also on board due to issues on Python 3.6.0 (see https://github.com/psf/black/issues/1666#issuecomment-787505729).

I guess, supporting Python 3.6.1+ is also an option, but I don't care much.

cooperlees:
I agree with this, sorry I forgot about it. I believe for 3.6 and 3.7 we should only support what ever version GitHub actions is up to and people can try other versions at the own risk and run their own CI.

@hugovk - If you could rebase this and push I'll merge once all is green. Sorry I forgot about this and thanks @ichard26 for the nudge.

hugovk:
Rebased!

PR #1670, opened 2020-09-02
Repair colorama wrapping on non-Windows platforms

jpgrayson:
The `wrap_stream_for_windows()` function calls
`colorama.initialise.wrap_stream()` function to apply colorama's magic to
wrapper to the output stream. Except this wrapper is only applied on
Windows platforms that need it, otherwise the original stream is
returned as-is.

The colorama wrapped stream lacks a `detach()` method, so a no-op lambda
was being assigned to the wrapped stream.

The problem is that the no-op lambda was being assigned unconditionally
whether or not colorama actually returns a wrapped stream, thus
replacing the original `TextIOWrapper`'s `detach()` method. Replacing the
`detach()` method with a no-op lambda is the root cause of the problem
observed in #1664.

The solution is to only assign the no-op `detach()` method if the stream
lacks its own `detach()` method.

Repairs #1664

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@
 
 - fixed a crash when PWD=/ on POSIX (#1631)
 
+- fixed "I/O operation on closed file" when using --diff (#1664)
+
 - Prevent coloured diff output being interleaved with multiple files (#1673)
 
 ### 20.8b1
--- a/README.md
+++ b/README.md
@@ -636,6 +636,7 @@ Multiple contributions by:
 - [Paul Ganssle](mailto:p.ganssle@gmail.com)
 - [Paul Meinhardt](mailto:mnhrdt@gmail.com)
 - [Peter Bengtsson](mailto:mail@peterbe.com)
+- [Peter Grayson](mailto:pete@jpgrayson.net)
 - [Peter Stensmyr](mailto:peter.stensmyr@gmail.com)
 - pmacosta
 - [Quentin Pradet](mailto:quentin@pradet.me)
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -861,31 +861,22 @@ def color_diff(contents: str) -> str:
 
 def wrap_stream_for_windows(
     f: io.TextIOWrapper,
-) -> Union[io.TextIOWrapper, "colorama.AnsiToWin32.AnsiToWin32"]:
+) -> Union[io.TextIOWrapper, "colorama.AnsiToWin32"]:
     """
-    Wrap the stream in colorama's wrap_stream so colors are shown on Windows.
+    Wrap stream with colorama's wrap_stream so colors are shown on Windows.
 
-    If `colorama` is not found, then no change is made. If `colorama` does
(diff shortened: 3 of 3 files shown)


ichard26:
Now looking at the colorama code, none of this "let's fake the detach method" is even required in the first place as `<class 'colorama.ansitowin32.StreamWrapper'>` (the type of object colorama.initialise.wrap_stream returns) hands off the attribute search to its underlying wrapped stream when normal attribute search of itself fails. https://github.com/tartley/colorama/blob/master/colorama/ansitowin32.py#L28

`io.TextIOWrapper` will always have the detach method so since it's the only stream type being wrapped by `colorama.ansitowin32.StreamWrapper`, the now wrapped stream will still have the `detach` method.

jpgrayson:
Saw that colorama's `StreamWrapper` implemented the pass-through `__getattr__()`, but since I don't have a Windows environment at-hand, I went with the minimal safe change. The concern would be that the no-op `detach()` is somehow required for this colorama wrapper to work properly on Windows.

Perhaps this PR could go in as-is and then someone with a Windows environment could confirm that ripping-out the `detach()` no-op lambda works in that environment?

ichard26:
>Perhaps this PR could go in as-is and then someone with a Windows environment could confirm that ripping-out the detach() no-op lambda works in that environment?

Perfect, I have a Windows machine right in front of me :-) I'll test the other idea now.

ichard26:
Never faking the `detach` method doesn't seem to cause any problems on Windows as far as I can tell. Although I just have a single Windows machine which of course isn't representative of all of the Windows machines out there. So it's your choice whether to leave the noop method fallback. It seems highly unlikely that `io.TextIOWrapper` wouldn't have the `detach` method , but then again it doesn't hurt to include the fallback `¯\_(ツ)_/¯`.

Regardless, it would be nice if you could update the nearby comments as some will become outdated with your patch or are already incorrect (like `# wrap_stream returns a colorama.AnsiToWin32.AnsiToWin32 [snip]`).

Oh, and if you'd like, feel free to add yourself to the authors list in the README!

ichard26:
Sorry for forgetting about this but could you please add an entry in the change log (`CHANGES.md`)? I would go with something like "fixed a bug where `Black` would sometimes fail to print diff output emitting I/O errors when formatting many files (PR #1670)".

Requesting review from @zsol as he reviewed the original PR that introduced coloured diffs... and the bug you're fixing.

CHANGES.md
```diff
@@ -12,6 +12,8 @@
 
 - fixed a crash when PWD=/ on POSIX (#1631)
 
+- fixed "I/O operation on closed file" on non-Windows OS when using --diff (#1664)
```
ichard26, line 15: Err, the `I/O operation on closed file` errors were also present on Windows...

jpgrayson: It was not clear to me that this bug manifested on Windows. Good to know.

I've updated CHANGES.md to reflect this.

zsol (approved):
LGTM. Thanks for the contribution @jpgrayson and the thorough review @ichard26

PR #1673, opened 2020-09-04
Handle .COLOR_DIFF in the same way as .DIFF

tom-saunders:
Whilst poking around checking how diff output was generated I noticed that the behaviour for these two values seemed to not be handled consistently.

Added a couple of tests to demonstrate and resolved the points where that seems to occur.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@
 
 - fixed a crash when PWD=/ on POSIX (#1631)
 
+- Prevent coloured diff output being interleaved with multiple files (#1673)
+
 ### 20.8b1
 
 #### _Packaging_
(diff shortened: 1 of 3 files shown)


tom-saunders:
well that's interesting - these pass locally. I imagine it's something I am doing locally which breaks them elsewhere.

ichard26:
Don't worry, it's not as bad as you think. Please ignore the primer failures, those are unrelated and I am currently working on a fix.

ichard26 (commented):
The PR itself looks good to me, but can you please add an entry in the changelog (`CHANGES.md`)? 

I was wondering why the coloured diff output was all mixed up while I doing some debugging. I saw that there was a lock being used so I just thought that my environment was broken. I forgot that there is also a `WriteBack.COLOR_DIFF` enum value 😅

Also if you'd like, feel free to add youself to the author list in the README. Congrats on your first PR in the psf repo!

JelleZijlstra (approved):

PR #1674, opened 2020-09-04
Update primer.json to reflect Black version 20.08b1 adoption

ichard26:
- tox recently adopted Black  
https://github.com/tox-dev/tox/commit/a7903508fa07068b327e15cfdbf8ea330ab78765

- attrs already adopted Black but they updated to 20.08b1 + did a format pass and removed some trailing commas
https://github.com/python-attrs/attrs/commit/f680c5b83e65413eeb684c68ece60198015058c3

PR #1679, opened 2020-09-05
Test primer on Pillow

hugovk:
Adding a new large Black codebase, as requested in https://github.com/psf/black/pull/1674#issuecomment-687391973.

Has 272 Python files, already formatted with Black.


cooperlees (approved):
Nice. More projects the better imo.

PR #1681, opened 2020-09-05
Fix crash on assert and parenthesized % format (fixes #1597, fixes #1605)

### LINKED ISSUE #1597: Crash on assert and parenthesized % format

### LINKED ISSUE #1605: error: cannot format Tests/test_file_libtiff.py: (-1, 8)

1. Take this file https://github.com/python-pillow/Pillow/blob/e54909b97c213278accc8176f1c52e403f8afcb6/Tests/test_file_libtiff.py
2. Run _Black_ on it with these arguments `--check Tests/test_file_libtiff.py`
3. See error

```console
$ black --version
black, version 19.10b1.dev141+g820f387
$ black --check Tests/test_file_libtiff.py
error: cannot format Tests/test_file_libtiff.py: (-1, 8)
Oh no! 💥 💔 💥
```

File formats.

- Version: master
- OS and Python version: macOS/Python 3.8.3

Yes, see above.

Works with last release (the codebase is already formatted with that version):

```console
$ black --version
black, version 19.10b0
$ black --check Tests/test_file_libtiff.py
All done! ✨ 🍰 ✨
```

Looks like it's this part:

```python
                            assert (
                                c_float(val[0][0] / val[0][1]).value
                                == c_float(value[0][0] / value[0][1]).value
                            ), ("%s didn't roundtrip" % tag)
```

https://github.com/python-pillow/Pillow/blob/e54909b97c213278accc8176f1c52e403f8afcb6/Tests/test_file_libtiff.py#L171-L174

(We're about to drop Python 3.5 and adopt f-strings, it's replacement is handled fine:

```python
                            assert (
                                c_float(val[0][0] / val[0][1]).value
                                == c_float(value[0][0] / value[0][1]).value
                            ), f"{tag} didn't roundtrip"
```
)

bbugyi200:
Fixes #1597. Fixes #1605.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -112,6 +112,10 @@ class InvalidInput(ValueError):
     """Raised when input source code fails all parse attempts."""
 
 
+class BracketMatchError(KeyError):
+    """Raised when an opening bracket is unable to be matched to a closing bracket."""
+
+
 T = TypeVar("T")
 E = TypeVar("E", bound=Exception)
 
@@ -1308,7 +1312,13 @@ class BracketTracker:
         self.maybe_decrement_after_lambda_arguments(leaf)
         if leaf.type in CLOSING_BRACKETS:
             self.depth -= 1
-            opening_bracket = self.bracket_match.pop((self.depth, leaf.type))
+            try:
+                opening_bracket = self.bracket_match.pop((self.depth, leaf.type))
+            except KeyError as e:
+                raise BracketMatchError(
+                    "Unable to match a closing bracket to the following opening"
+                    f" bracket: {leaf}"
+                ) from e
             leaf.opening_bracket = opening_bracket
             if not leaf.value:
                 self.invisible.append(leaf)
@@ -3324,10 +3334,17 @@ class StringParenStripper(StringTransformer):
                 yield TErr(
                     "Will not strip parentheses which have comments attached to them."
                 )
+                return
 
         new_line = line.clone()
         new_line.comments = line.comments.copy()
-        append_leaves(new_line, line, LL[: string_idx - 1])
+        try:
+            append_leaves(new_line, line, LL[: string_idx - 1])
+        except BracketMatchError:
+            # HACK: I believe there is currently a bug somewhere in
+            # right_hand_split() that is causing brackets to not be tracked
+            # properly by a shared BracketTracker.
+            append_leaves(new_line, line, LL[: string_idx - 1], preformatted=True)
 
         string_leaf = Leaf(token.STRING, LL[string_idx].value)
         LL[string_idx - 1].remove()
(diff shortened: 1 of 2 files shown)


src/black/__init__.py
```diff
@@ -3306,10 +3316,26 @@ def do_transform(self, line: Line, string_idx: int) -> Iterator[TResult[Line]]:
...
                 )
+                return
 
         new_line = line.clone()
         new_line.comments = line.comments.copy()
-        append_leaves(new_line, line, LL[: string_idx - 1])
+        try:
+            append_leaves(new_line, line, LL[: string_idx - 1])
+        except BracketMatchError as e:
+            # HACK: I believe there is currently a bug somewhere in
```
JelleZijlstra, line 3326: This is unfortunate, but it's better than a crash, and the formatting in the newly added test cases looks reasonable.

JelleZijlstra (approved):

src/black/__init__.py
bbugyi200: @JelleZijlstra On second thought, I don't think this will do. The `append_leaves()` function modifies the Node structure and so must be atomic. I can't fix this right now, but I will have to come up with a different solution.

bbugyi200: Ok. This is fixed. The corrected version of this fix, though still not ideal, has the benefit of stripping the parens as is consistent with blacks style after #1182.

JelleZijlstra (approved):

PR #1683, opened 2020-09-06
Exclude venv directory by default

johnthagen:
There hasn't been 100% consensus on what the folder name for a virtual environment should be. Some tools use `.venv`, others use `venv`, and there are likely other constructs.

`venv` is a common name in use.

- [The `virtualenv` docs show using the name `venv` as the first step in the user guide](https://virtualenv.pypa.io/en/latest/user_guide.html)
- [By default, PyCharm creates virtual environments in a `venv` folder](https://www.jetbrains.com/help/pycharm/creating-virtual-environment.html)

[`isort` already includes `venv` in its default list of skipped files](https://pycqa.github.io/isort/docs/configuration/options/#skip).

IMO, this would be a pragmatic addition to `black`.


ichard26 (approved):
Seems like a pretty safe change without any apparent downsides. I personally use `env` as my virtual environment directory name but oh well `¯\_(ツ)_/¯`.

Looking forward to your first commit to Black!

johnthagen:
@ichard26 / @ambv Conflicts have been resolved and changelog updated.

JelleZijlstra (approved):
I'll merge once CI is green

PR #1697, opened 2020-09-11
docs: add mention to isort profile option

MicaelJarniac:
isort has a `black` profile built-in, with those settings already configured, so getting it to work with Black can be as easy as saying `profile = black` on its settings.

Sure, if the user wants to customize maximum line length or something like that, they'd most likely have to take the long approach, but if they're fine with the defaults of Black, that profile setting can be very handy.

I was quite unsure about how to correctly add those references, so I tried doing it in a way I felt was doable, but feedback is appreciated.

I'm adding it here because I've been searching for that myself for a while, and only today have I discovered isort has pre-built config profiles, and that helped me a whole lot.

### DIFF
--- a/docs/compatible_configs.md
+++ b/docs/compatible_configs.md
@@ -16,6 +16,9 @@ tools, using **their** supported file formats.
 _Black_ also formats imports, but in a different way from isort's defaults which leads
 to conflicting changes.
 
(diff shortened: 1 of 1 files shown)


MicaelJarniac:
I'm blind, it's already mentioned in there.

Issue #1568, opened 2020-08-03
Black output junitxml for CI pretty integration


falcorocks:
* Feature: I use black in CI (gitlab) in my organisation to enforce standardised code. I know this sounds ridiculous, but black output (together with the gitlab job output log) can look a bit intimidating to people unfamiliar with these tools.
Gitlab allows for pretty printing of reports from test tools (like pytest or mutmut) within a merge request, other CI tools do as well. The "standard" format for these tests output is junitxml. See for example gitlab section about it https://docs.gitlab.com/ee/ci/junit_test_reports.html.

* Solution: it would be cool if black supported a junitxml output!

MReinisch:
My organization started using black in gitlab ci pipelines, too.

I started working on implementig this feature on a Fork.

JelleZijlstra:
As mentioned on the linked PR, I'm making the call that this is out of scope for Black right now.

dosas:
https://github.com/quantum-0/black-junit

PR #1715, opened 2020-09-17
feat(junit_xml): added junit_xml output for ci logging

MReinisch:
Added feature to create junitXML Report file. https://github.com/psf/black/issues/1568

New console Parameters :
```
-report-type [console: Default, junit]
-junit [Filename]
```
Warning Path to Filename must exist.

Still Todo for V1:
- [x]  Create Path to junit
- [x]  Add new Params to Configuration File

Todo for V2:
- [ ]  Add Diffs to junit XML Message


src/black/__init__.py
```diff
@@ -498,6 +514,25 @@ def target_version_option_callback(
...
+    type=click.Choice([v.name.lower() for v in ReportType]),
+    callback=report_option_callback,
+    default=DEFAULT_REPORT,
+    multiple=False,
+    help=(
+        "Option if the generated Report should be to the Console or a Junit XML File.\n"
+        "[default : console]"
+    ),
+)
+@click.option(
```
JelleZijlstra, line 528: Not sure we need two options. We can just make it so if `--junit-xml` is given, we output XML.

src/black/__init__.py
```diff
@@ -498,6 +514,25 @@ def target_version_option_callback(
...
+    "--junitxml",
+    type=click.File(mode="w"),
+    help=(
+        "Option is needed for --report-type junit it tells where to store the XML"
+        " Output.\nIf directory is not present it will be created"
```
JelleZijlstra, line 533: Suggested wording: "Write junit XML output to FILE. If the directory does not exist, it will be created."

src/black/__init__.py
```diff
@@ -573,7 +621,11 @@ def main(
 
     if verbose or not quiet:
         out("Oh no! 💥 💔 💥" if report.return_code else "All done! ✨ 🍰 ✨")
-        click.secho(str(report), err=True)
+        if isinstance(report, JunitReport):
+            junitxml.write(str(report))  # type: ignore
```
JelleZijlstra, line 625: Why do you need an ignore?

src/black/__init__.py
```diff
@@ -6144,6 +6196,141 @@ def __str__(self) -> str:
         return ", ".join(report) + "."
 
 
+@dataclass
+class JunitReport:
```
JelleZijlstra, line 6200: I feel like it might be cleaner to add a method to the existing Report class that produces XML output, instead of adding a new class. Would that work?

src/black/__init__.py
MReinisch: Maybe you are right.
My intention was if there will be more reports type in the future this could be a selector.

src/black/__init__.py
MReinisch: Will be changed.

src/black/__init__.py
MReinisch: I had to add ignore because:

junitxml is from type ```click.utils.LazyFile``` and ```mypy``` says that ```LazyFile``` has no attribute ```write```.

src/black/__init__.py
MReinisch: I wanted to leave the existing ```Report``` class untouched.

If you need more Report types in the Future, the ```Report``` class will grow unnecessarily in my opinion.

theduke:
Hey @MReinisch , any interest in continuing this work here?

Would love to see this merged.

MReinisch:
@theduke sure i will continue working on this.

Working on V2 will start after this pr is merged.

JelleZijlstra:
This has been open for a long time and I think most of us maintainers aren't too excited about adding a new niche output format.

Also, it seems fairly easy to have a third-party tool that parses black's output and munges it into the desired format.

NorthIsUp:
this would be a wonderful feature, and "niche" isn't quite the right word for junit xml, for better or worse it is more like the "gold standard" of continuous integration data formats. please reconsider this one 🙏

MReinisch:
I still work on this from time to time and rebase my repo.

I forked it here : https://github.com/Auerswald-GmbH/black

PR #1717, opened 2020-09-18
PEP 614 support

### LINKED ISSUE #1711: [python 3.9] black doesn't support PEP 614
**Describe the bug**
[PEP 614](https://www.python.org/dev/peps/pep-0614/) relaxed the syntax for decorators in python 3.9. However, black doesn't (yet ?) have a `py39` target, and doesn't support this syntax.

**To Reproduce**

1. Take this very simple file `test.py`
```python
from dataclasses import dataclass
d = [dataclass]

@d[0]
class A:
    attr: int

print(A)
```
2. Run _Black_ on it : `black test.py`
3. See error `error: cannot format test.py: Cannot parse: 5:2: @d[0]`

**Expected behavior** _Black_ should accept and format code that is valid for python 3.9 (`python test.py` prints `<class '__main__.A'>` as expected).

- Version: master (commit 811decd7f10fb2fb3ae343b9d9d0a3ae53b86a53)
- OS and Python version: Linux, Python 3.9.0rc1

1. Use the online formatter at https://black.now.sh/?version=master
This produce the above error
2. Or run _Black_ on your machine:
   - create a new virtualenv (make sure it's the same Python version);
   - clone this repository;
   - run `pip install -e .`;
   - make sure it's sane by running `python -m unittest` (this fails, see #1441 )
   - run `black` like you did last time.

This also produce the above error

**Additional context** Python 3.9 is in pre-release stage, so I'm not sure if this is a bug report or a feature request - I can change to the latter. [PEP 614](https://www.python.org/dev/peps/pep-0614/) highlights the use-case of the syntax (with examples from PyQt5), the above file is a minimalist file to illustrate the issue

QuentinSoubeyran:
Adds support for PEP 614 to fix #1711. This is still WIP

**Changes**
- updated `typed-ast` to `1.4.1` so that `pipenv install --dev` doesn't fail within a python 3.9 venv due to compile errors
- changed the definition of`decorator` in `blib2to3/Grammar.txt` to reflect the relaxed decorator syntax
- added a `PY39` target
- added a `RELAXED_DECORATORS` feature
- added an eager detection of the above feature in `get_feature_used()`, with helper functions
- moved `PY36_VERSIONS` from `src/black/__init__.py` to `tests/test_black.py`
- added test for the detection of `RELAXED_DECORATORS` feature
- added test for python 3.9 decorator formatting

**Note**
- `--safe` doesn't work on python 3.9 code from black running on python <=3.8.5. This might already be the case with python 3.8 code formatted from black running on python 3.7, because `typed_ast` doesn't support python 3.8+ syntax. So older python version cannot parse the syntax with either the builtin `ast` or `typed_ast`

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -6146,7 +6146,7 @@ class Report:
 
 def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
     filename = "<unknown>"
-    if sys.version_info >= (3, 9):
+    if sys.version_info >= (3, 8):
         # TODO: support Python 4+ ;)
         for minor_version in range(sys.version_info[1], 4, -1):
             try:
@@ -6154,7 +6154,7 @@ def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
             except SyntaxError:
                 continue
     else:
-        for feature_version in (8, 7, 6):
+        for feature_version in (7, 6):
             try:
                 return ast3.parse(src, filename, feature_version=feature_version)
             except SyntaxError:
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -2194,7 +2194,7 @@ def whitespace(leaf: Leaf, *, complex_subscript: bool) -> str:  # noqa: C901
         ):
             # Python 2 print chevron
             return NO
-        elif prevp.type == token.AT and p.parent.type == syms.decorator:
+        elif prevp.type == token.AT and p.parent and p.parent.type == syms.decorator:
             # no space in decorators
             return NO
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -5762,7 +5762,9 @@ def get_features_used(node: Node) -> Set[Feature]:
             features.add(Feature.ASSIGNMENT_EXPRESSIONS)
 
         elif n.type == syms.decorator:
-            if len(n.children) > 1 and is_simple_decorator_expression(n.children[1]):
+            if len(n.children) > 1 and not is_simple_decorator_expression(
+                n.children[1]
+            ):
                 features.add(Feature.RELAXED_DECORATORS)
 
         elif (
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -178,12 +178,10 @@ class TargetVersion(Enum):
     PY36 = 6
     PY37 = 7
     PY38 = 8
+    PY39 = 9
 
     def is_python2(self) -> bool:
         return self is TargetVersion.PY27
-
-
-PY36_VERSIONS = {TargetVersion.PY36, TargetVersion.PY37, TargetVersion.PY38}
 
 
 class Feature(Enum):
@@ -199,6 +197,7 @@ class Feature(Enum):
     ASYNC_KEYWORDS = 7
     ASSIGNMENT_EXPRESSIONS = 8
     POS_ONLY_ARGUMENTS = 9
+    RELAXED_DECORATORS = 10
     FORCE_OPTIONAL_PARENTHESES = 50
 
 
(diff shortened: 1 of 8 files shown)


QuentinSoubeyran:
I'm looking for help on the space after `@` bug, I have no idea where this is added in the code. My hypothesis is that all expression get a space before, I'd like to special-case that so it doesn't happens if the previous character is `@` and we are in a decorator (so that the `@` as `__matmul__` operator is still properly spaced)

src/black/__init__.py
```diff
@@ -5499,6 +5509,42 @@ def is_walrus_assignment(node: LN) -> bool:
...
+            and len(node.children) == 3
+            and node.children[0].type == token.LPAR
+            # and node.children[1].type == syms.argument
+            and node.children[2].type == token.RPAR
+        )
+    )
+
+
+def is_simple_decorator_expression(node: LN) -> bool:
+    """Return True iff `node` is the 'dotted name' in the grammar @ dotted_name [arguments] NEWLINE"""
```
zsol, line 5532: nit
```suggestion
    """Return True iff `node` is the 'namedexpr_test' in the grammar @ namedexpr_test [arguments] NEWLINE"""
```

zsol:
For the whitespace issue, this will need to be handled in `whitespace` around here: https://github.com/psf/black/blob/master/src/black/__init__.py#L2259

Please add a couple of explicit test cases for the feature detection, as well as some more complex decorator examples.

zsol:
Also feel free to put this PR in draft mode and ping me explicitly if you need more help until you're ready to publish it :)

QuentinSoubeyran:
I should have checked here earlier, I just found the `whitespace()` function by using pdb. Thanks anyway :)

src/black/__init__.py
QuentinSoubeyran: Tried to clarify that this function does test against a `dotted_name` grammar. Is this better ?

QuentinSoubeyran:
Hi @zsol , I'm not sure how to handle the following: the `--safe` option doesn't work on python 3.9 code when black run on python 3.8. (tested on 3.8.2 and 3.8.5)
This is because [`parse_ast`](https://github.com/psf/black/blob/811decd7f10fb2fb3ae343b9d9d0a3ae53b86a53/src/black/__init__.py#L6083) uses the builtin `ast` module, which cannot parse 3.9 code in python 3.8.5. Judging from the grammar on the [`ast` doc](https://docs.python.org/3.8/library/ast.html), where decorator list are `expr*`, this might be fixed in python 3.8.6.

Still, python 3.9 code cannot be safely reformatted used python <=3.8.5. What is the best way to handle that ? Should I introduce a call to `ast3.parse(..., feature_version=9)` in `parse_ast` to fix that ?

QuentinSoubeyran:
Also, I tried running `black-primer`, but it fails for some projects (`hypothesis`, `poetry`, `pyramid`, `pytest`, `sqlalchemy`). Should I just skip them as hinted in the contribution guide, or should I do something about that ?

zsol:
The idea behind that `parse_ast` function is to use the builtin ast module on the latest supported python version. That will be 3.9 with this PR, so feel free to change the conditions so that
- on 3.9 we use the builtin `ast` module
- on 3.6 3.7 and 3.8 we use `typed_ast`

zsol:
I'm not sure about primer. What kinds of changes are there? CI looks green

QuentinSoubeyran:
I changed `typed_ast` as described. Curiously, this makes some unit test fail on python3.8, while everything works fine on python3.9. Perhaps `typed_ast` doesn't lke PEP570, PEP572 and stared returns (those are the unit test that fail).

Sorry I forgot the details about `black-primer`. I didn't look at all changes individually, but those I saw were arguments splitted on multiline, or sometime just adding the trailing comma in already-splitted arguments.

QuentinSoubeyran:
Just found out that `typed_ast` does not support python 3.8 syntax and won't support it. So I guess I should revert the above commit ? To parse python 3.8, we need builtin the `ast`, so we need python >=3.8.

I think `--safe` would not work for python 3.8 code formatted from python 3.7 due to that, so black run on python3.8 not formatting python 3.9 might be acceptable

QuentinSoubeyran:
I went through all changes reported by `black-primer`, those are all one of:
- arguments split on multiline due to long line
- arguments split on multiline due to magic trailing comma
- a docstring reformat to remove spaces after `"""`
- in `sqlalchemy`, reformat due to the call chain style

All in all, every formatting I saw is due to a well-defined rule explained in the code style, so I think this PR is ready for review

tests/data/decorators.py
```diff
@@ -0,0 +1,176 @@
+# This file doesn't use the standard decomposition
+# decorator syntax test cases are separated by double # comments
+# Those before the 'output' comment are valid under the old syntax
+# Those after the 'ouput' comment require PEP614 relaxed syntax
+# Do not remove the double # separator before the first test case, it allows
+# to the comment before the test case
```
JelleZijlstra, line 6: remove "to"?

JelleZijlstra (approved):
Looks good apart from one nit

tests/data/decorators.py
QuentinSoubeyran: I think I also forgot the verb. It should be fixed now. Thanks !

zsol (approved):
LGTM. Thanks for taking the time to work on this

QuentinSoubeyran:
My pleasure :) Thanks to you two for your help. And thank everyone for Black, it is awesome !

PR #1727, opened 2020-09-27
End 'force-exclude' help message with a period

finn0:
It would be nice, if like other options help message, force-exclude's
help message also ends with a period punctuation mark.


ichard26 (approved):
Looks good to me, many thanks! 

Now to wait for another maintainer's approval and merge as I do not have merge privileges :-)

edit: oh and don't worry about Primer failing, it's unrelated to your change.

PR #1728, opened 2020-09-27
Fix unnecessary if checks

hakancelik96:


cooperlees (approved):
LGTM - If the line starts with something else, good chance it does not start with the others!

P.s. a rebase should make primer pass due to #1729

PR #1729, opened 2020-09-27
Primer: Hypothesis is now formatted with Black 20.8b1

hugovk:
Fixes the Primer CI.

Hypothesis updated from Black 19.10b0 to 20.8b1 in https://github.com/HypothesisWorks/hypothesis/pull/2610, so we now expect no formatting changes.


ichard26 (approved):
LGTM!

cooperlees (approved):
Agree! Thanks for keeping on top of this!

Issue #1732, opened 2020-10-01
Quiet mode for Vim plugin


noelevans:
**Situation** I use Black with the included Vim plugin. I have a `.vimrc` setting which runs Black when writing:
`autocmd BufWritePost *.py execute ':Black'
`

It is obviously useful to be prompted when Black cannot format a module because it is malformed. But making writes through the day, it can feel noisy to repeatedly get feedback from Black when _successfully_ reformatting. I.e

`Already well formatted, good job... `
and
`Reformatted in 1.2345s`

On the command line Black offers a `--quiet` option. It would be great if something equivalent were available for the Vim plugin. 

**Proposal** I see that in issue 252, this is discussed and @ambv suggests a similar idea: https://github.com/psf/black/issues/252#issuecomment-392284791

In issue 252, the idea of a :BlackQuiet command is proposed. I think the plugin would remain more uniform by making use of a Vim property and toml setting; similar to how `--fast` and `--skip-string-normalization` are already implemented in the plugin. 

**Alternatives considered** Alternatives with detracting aspects are discussed in 252.

noelevans:
Related code has been merged

PR #1733, opened 2020-10-01
Add quiet flag so non-error actions go unreported

noelevans:
This PR addresses the idea discussed in issue https://github.com/psf/black/issues/1732

### DIFF
--- a/docs/editor_integration.md
+++ b/docs/editor_integration.md
@@ -121,6 +121,7 @@ Configuration:
 - `g:black_linelength` (defaults to `88`)
 - `g:black_skip_string_normalization` (defaults to `0`)
 - `g:black_virtualenv` (defaults to `~/.vim/black` or `~/.local/share/nvim/black`)
+- `g:black_quiet` (defaults to `0`)
 
 To install with [vim-plug](https://github.com/junegunn/vim-plug):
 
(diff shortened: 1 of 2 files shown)


noelevans:
I see the primer checks are failing for some of the configured projects. I have tried running primer on master, locally and see this step fails there too. Is there some way to correct this?

noelevans:
When I initially pushed this branch, the build failed - like others' builds - because of inconsistencies when Primer ran. I also noted it failed when running unmodified origin/master so I took that not to be a problem. Subsequent retries passed this build component.

Do I need to do something else to have this PR reviewed?

traviswaelbro:
> Do I need to do something else to have this PR reviewed?

It would probably help to rebase and squash your commits into one commit

ichard26:
I'm sorry, but I think what is holdin' up this PR that we don't have any qualified and experienced vim users **and** Black developers  who have time (or want) to review this PR at this time unfortunately.

JelleZijlstra:
@ichard26 is right.

Also, contrary to what @traviswaelbro said, there's no need to squash your commits. In general, that actually makes it harder for me to review a PR, because it's more difficult to see how the code has changed over time (admittedly, that doesn't matter much when there's been no previous review).

I don't use the vim plugin, but I'll take a look at this PR now and merge if it seems right.

noelevans:
@JelleZijlstra thanks very much

PR #1734, opened 2020-10-02
Added not formatting files in gitignore (psf#1682)

J-Exodus:
Added statement to the docs to highlight that Black will not format files listed in the `.gitignore` of a project.

### DIFF
--- a/docs/pyproject_toml.md
+++ b/docs/pyproject_toml.md
@@ -33,6 +33,8 @@ You can also explicitly specify the path to a particular file that you want with
 If you're running with `--verbose`, you will see a blue message if a file was found and
 used.
 
+Files listed within a projects `.gitignore` file will not be formatted by _Black_.
+
 Please note `blackd` will not use `pyproject.toml` configuration.
 
 ## Configuration format


docs/pyproject_toml.md
```diff
@@ -33,6 +33,8 @@ You can also explicitly specify the path to a particular file that you want with
 If you're running with `--verbose`, you will see a blue message if a file was found and
 used.
 
+Files listed within a projects `.gitignore` file will not be formatted by _Black_.
```
JelleZijlstra, line 36:
```suggestion
Files listed within a project's `.gitignore` file will not be formatted by _Black_.
```

JelleZijlstra: Also this doesn't seem the right place for this text—the section is about finding `pyproject.toml`.

J-Exodus: I see your point @JelleZijlstra. After looking through the docs further, I believe the best place to insert this information may be within this statement, which is mentioning areas that will not be formatted:
Black docs insert point
Please let me know if you agree and I'll fix up this pull request.  Thanks.

JelleZijlstra: I guess that's an option, or we could have a new section on skipping formatting in general.

J-Exodus: As this is my first time contributing to open source, how is the best way to look at doing that? Do we look to get agreement on a new section being the best way forward first?  I'd love the opportunity to write this section. Just want to make sure I'm doing it the right way.  Thanks for your help.

JelleZijlstra: Sorry for not getting back to you on this. Honestly our configuration is a bit of a mess, with a lot of top-level pages for relatively minor things. I feel like a nicer way to do it would be to rename the pyproject.toml section to "Configuration" and cover gitignore as well as other ways to configure or silence Black, like `# format: off` comments.

JelleZijlstra: I filed #1759 to discuss reorganizing the docs more broadly.

JelleZijlstra:
Going to merge this for now; we can put it somewhere else when we reorganize the documentation.

Issue #1635, opened 2020-08-26
Black removes newline from a multiline docstring.


omry:
**Describe the bug** A clear and concise description of what the bug is.
```python
class Foo:
    """foo
    """
```

Is being reformatted to:
```python
class Foo:
    """foo"""
```

This is removing a newline.
arguably this is okay because it's a comment, but such comments are (as far as I know) accessible through object inspection.

**Expected behavior** A clear and concise description of what you expected to happen.
I am expecting black to not remove newlines from strings.

**Environment (please complete the following information):**

- Version: black==20.8b1
- OS and Python version: Linux/Python 3.8.5

**Does this bug also happen on master?**
Yes.

cooperlees:
Hi Omry, thanks for reporting.

Do you explicitly only want this only for docstrings? If so, black respects ‘\’.

I.e.
```python
class Foo:
  “””Hello \
  “””
```

This would not be reformatted.

Black also respects multi lines variables with 20.8b1
```python
foo = “””
Cooper made a string
“””
```

This is also not reformatted on my tests.

omry:
Hi @cooperlees,
This caught me off guard, specifically for my use case it's indeed a comment and don't mind the re-formatting.
I did notice that this works as I expect for variables.

the workaround does not seem to be generating the same string:

```python
s1 = """hello
"""

s2 = """hello\ 
"""
print("s1", repr(s1))
print("s2", repr(s2))
```
```shell
$ python 1.py 
s1 'hello\n'
s2 'hello\\ \n'
```

cooperlees:
As usual, you've missed the point of my reply. Docstrings need the '\' to not join on the \n, like your original post **only** showed. Normal multi-lines strong should not be doing what you say.

I am unable to repro:

If I
```python
echo 's1 = """hello
"""

s2 = """hello\ 
"""
print("s1", repr(s1))
print("s2", repr(s2))
' > /tmp/omry.py
```

```console
cooper-mbp1:bandersnatch cooper$ /tmp/black/bin/black --diff /tmp/omry.py
All done! ✨ 🍰 ✨
1 file would be left unchanged.
```

Neither multiline string here is reformatted for me ... Please provide a non docstring repro with `20.8b1` or master please.

omry:
Since I do not understand you as usual I am going to close this.
I don't see a point in bothering because as usual I will fail.

omry:
The problem is specific to docstrings.
[...]

PR #1740, opened 2020-10-02
Make black remove leading and trailing spaces from one-line docstrings: fixes #1738 #1812

### LINKED ISSUE #1738: Black removes leading and trailing spaces in multiline docstrings but fails to remove them from one-line docstrings
With the file `test.py` containing
```
def foo(a):
    """ This is fine. """

    return a + 1


def bar(a):
    """As is this."""

    return a + 1


def baz(a):
    """ This is not.

    Since it uses more than one line. """

    return a + 1
```
Running `black --diff test.py` results in a diff of:
```
@@ -9,10 +9,10 @@
 
     return a + 1
 
 
 def baz(a):
-    """ This is not.
+    """This is not.
 
-    Since it uses more than one line. """
+    Since it uses more than one line."""
 
     return a + 1
```

**Expected behavior**

Black should also complain that the docstring of foo contains a leading / trailing space and remove these. Note that all one-line examples in PEP 257 https://www.python.org/dev/peps/pep-0257/#one-line-docstrings have no leading / trailing spaces.

- Version: master
- OS: Linux (Ubuntu 20.04)
- Python Version: Python 3.8.2

This also happens on master.

MarkCBell:
Resolves #1738 Black removes leading and trailing spaces in multiline docstrings but fails to remove them from one-line docstrings


MarkCBell:
It appears that the primer check are checking how black affects a number of different projects. Since this changes how single line docstrings are processed a number of these now change.

felix-hilden:
I've seen the Black team cautious about other than formatting changes in the source code elsewhere, but given that such processing is already done with other docstrings, this seems warranted. Also, the code is simple enough.

The test run logs seem to have timed out. I don't know if they can be rerun manually, or does the branch need to be updated, but having fresh results would be nice. Also there are conflicts with the current master, so fixing that would also update the branch and let you run the tests once more!

I'm not a maintainer though, so you'll sadly have to wait for someone else. Hopefully someone is able to do a quick review!

ichard26:
I think the main issue is that it's not clear how we want to handle docstring formating still. IRCC the current behaviour was a mistake added by a somewhat unrelated feature addition, so we might just rollback to the previous formatting, IDK and I haven't looked into it.

felix-hilden:
Oh, I see! Thanks for the information!

MarkCBell:
> Also there are conflicts with the current master, so fixing that would also update the branch and let you run the tests once more!

I have resolved the conflicts (and opened issue 2070 to deal with problems in the PR checking process).

As expected, the issues about the Primer checks into how black affects a number of different projects still remain.

CHANGES.md
```diff
@@ -4,6 +4,8 @@
 
 #### _Black_
 
+- `Black` now requires one-line docstrings to be stripped (PR #1740)
```
```suggestion
- `Black` now strips leading and trailing spaces from one-line docstrings (PR #1740)
```
Black doesn't require anything, it does it all for you

JelleZijlstra (changes requested):
Looks good to me except for a suggested improvement in wording

JelleZijlstra:
> IRCC the current behaviour was a mistake added by a somewhat unrelated feature addition

In fact it was added on purpose in #1053.

JelleZijlstra:
Also (sorry for the stream of consciousness), you'll have to update the black-primer configuration to reflect any additional projects with expected changes, assuming the changes all look right.

I checked out a few of them and did notice one special case:
```
 class DummyContext:
-    """ """
+    """"""
```
I don't really know why they're doing this, but it looks a lot weirder with no space in the middle. Maybe we should avoid completely emptying a docstring that consists only of spaces, instead replace it with a single space? I don't feel strongly about this; if others here prefer the completely empty string we can just keep things as is.

ichard26:
Sorry for not submitting my review earlier, to be honest, I was extremely confused by the docstring handling Black has right now. I thought Black modified docstrings much more than what is actually true. I really don't care about the indention or space cleanup Black does, the only thing that isn't immediately clear to me is the behaviour for docstring quotes, in particular the behaviour described in https://github.com/psf/black/issues/1635. 

Putting this file through Black on master d960d5d238bc8c73010a3ca5a9f316678ad91e6f:

```python3
def test():
    """This is one possibility.

    Very important stuff here.
    """


def test2():
    """This is another one possibility.

    Very important stuff here."""


def test3():
    """
    This is another one possibility.

    Very important stuff here.
    """


def test4():
    """
    What about this?
    """


def test5():
    """What about this?

    Some people also like to have an empty line at the end of the doc string.

    """


def test6():
    """Single line docstrings also exist."""


def test7():
    """
    Some people like them like this.
    """


def test8():
    """Or like this.
    """


def test9():
    """
    Or even like this."""


def test10():
    """Plus this.....

    style"""

```

returns this diff:

```diff
--- test.py	2021-04-11 22:07:57.113291 +0000
+++ test.py	2021-04-11 22:08:00.916690 +0000
@@ -42,12 +42,11 @@
     Some people like them like this.
     """
 
 
 def test8():
-    """Or like this.
-    """
+    """Or like this."""
 
 
 def test9():
     """
     Or even like this."""
```

which rubs me in the wrong way, either Black becomes more consistent with the "if can be collapsed into a single line docstring, it will" rule (i.e. collapse test6-10 not just test8), OR just remove this handling to let the user have full choice with where the quotes go.

P.S. perhaps we could document the handling (indention fixing, whitespace removal, etc.) Black has so it's easy to find and clear?

ichard26 (approved):
Please resolve the merge conflict and other than this looks good to me (edit: also please resolve Jelle's comments!). Thanks!

Regarding `""""""` vs `""" """`, I'd go with `""" """` too, but just like Jelle I don't feel strongly about this, so I'm to some good ol' convincing by others :)

ichard26:
PS. I'm pretty sure https://github.com/psf/black/issues/1812 can also be closed by this PR. Can you link in the description so we don't forget to close it?

MarkCBell:
So I've made the changes to CHANGES.md that were suggested and resolved the merge conflicts. I've also marked a number of the primer projects as "expect_formatting_changes". However in doing so I discovered that pyramid and sqlalchemy have some functions with docstrings that only use single quotes. That is,
```
def foo(self):
    '123'
    pass
```

Since the changes in this PR mean that black is actually processing single line docstrings now, I've had to reworked the `visit_STRING` function so that it no longer assumes that there are exactly three leading & trailing 3 quote characters in the ast leaf's value.

Since the function was already inserting leading / trailing spaces into docstrings that start / end with matching quotes (e.g. converting """"x""" -> """ "x""") I have also followed the suggestion and modified it to insert a space into empty docstrings (e.g. """""" -> """ """). I have also made it convert docstrings wrapped in single quotes to triple quotes (e.g. 'x' -> """x""").

JelleZijlstra:
Thanks! I'm a bit hesitant about changing all docstrings to triple spaces. It's not uncommon to see single-quoted docstrings (I see it in the standard library), so we might get user pushback for changing this. On the other hand, standardizing formatting is what Black is here for. I'm curious to hear what other maintainers think.

A concrete problem with this change is that it will make docstrings more likely to cross the line length limit.

MarkCBell:
> Thanks! I'm a bit hesitant about changing all docstrings to triple spaces. It's not uncommon to see single-quoted docstrings (I see it in the standard library), so we might get user pushback for changing this. On the other hand, standardizing formatting is what Black is here for. I'm curious to hear what other maintainers think.
> 
> A concrete problem with this change is that it will make docstrings more likely to cross the line length limit.

That is a very good point. To keep this PR focused I've just made a small change to the patch so that it remembers whether single or triple quotes are used and then later it installs quotes of the same length. 

We can then open a separate issue to discuss whether it is a good idea to replace single quotes with triple quotes (as recommended by PEP 257). However with the way this patch is now written this would hopefully be a very small patch (changing `quote = quote_char * quote_len` to `quote = quote_char * 3`)

JelleZijlstra (approved):

PR #1743, opened 2020-10-05
Automatically build and upload binaries on release

### LINKED ISSUE #1669: Release 20.8b1 Missing Binary

The most recent release only contains source code packages, no binaries. https://github.com/psf/black/releases/tag/20.8b1

1. Visit (https://github.com/psf/black/releases/tag/20.8b1) and notice there is no binary available for download

I expected a binary to be available, like previous releases.

N/A

N/A

ichard26:
Resolves #1669.

https://github.com/ichard26/black/actions/runs/340608417 and https://github.com/ichard26/black/releases/tag/test_binary_action3 shows the workflow in action.

### DIFF
--- a/.github/workflows/upload_binary.yml
+++ b/.github/workflows/upload_binary.yml
@@ -1,0 +1,51 @@
+name: Upload self-contained binaries
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.7]
+        os: [ubuntu-16.04, windows-2019]
+        include:
+          - os: windows-2019
+            pathsep: ";"
+            executable_suffix: ".exe"
+            executable_mime: "application/vnd.microsoft.portable-executable"
+          - os: ubuntu-16.04
+            pathsep: ":"
+            executable_suffix: ".elf"
+            executable_mime: "application/x-executable"
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip wheel setuptools
+          python -m pip install .
+          python -m pip install pyinstaller
+
+      - name: Build binary
+        run: |
+          python -m PyInstaller -F --name black${{ matrix.executable_suffix }} --add-data 'src/blib2to3${{ matrix.pathsep }}blib2to3' src/black/__main__.py
+
+      - name: Upload binary as release asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ github.event.release.upload_url }}
+          asset_path: dist/black${{ matrix.executable_suffix }}
+          asset_name: black${{ matrix.executable_suffix }}
+          asset_content_type: ${{ matrix.executable_mime }}
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,17 +29,3 @@ matrix:
       python: 3.8
     - name: "3.9"
       python: 3.9
-before_deploy:
-  - pip install pyinstaller
-  - pyinstaller --clean -F --add-data src/blib2to3/:blib2to3 src/black/__init__.py
-deploy:
-  provider: releases
-  api_key:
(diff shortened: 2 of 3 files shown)


.github/workflows/upload_binary.yml
```diff
@@ -0,0 +1,57 @@
...
+            executable_mime: "application/vnd.microsoft.portable-executable"
+          - os: ubuntu-16.04
+            pathsep: ":"
+            executable_suffix: ".elf"
+            executable_mime: "application/x-executable"
+
+    steps:
+      - name: Get release asset upload URL from release tag
+        id: get_release
+        uses: bruceadams/get-release@8a90bac3fc6cb911f53d22ce10868b5831f2d244
```
cooperlees, line 28: No Tag?

.github/workflows/upload_binary.yml
```diff
@@ -0,0 +1,57 @@
...
+          python -m pip install pyinstaller
+
+      - name: Build binary
+        run: |
+          python -m PyInstaller -F --name black${{ matrix.executable_suffix }} --add-data 'src/blib2to3${{ matrix.pathsep }}blib2to3' src/black/__main__.py
+
+      - name: Upload binary as release asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
cooperlees, line 52: Has someone stored this secret for you? Or is it auto magical?

cooperlees (approved):
Seems good to me, if the secret is available.

.github/workflows/upload_binary.yml
ichard26: Honestly I was scared that this PR would get rejected since it uses a third-party Action so I tried upping the changes this wouldn't get rejected by using a commit SHA instead of tags which are a weak point since tags can be modified. In hide-sight, creating non-signed binaries for other people to use is already sketchy so I don't know why I was freaking out over using a commit vs a tag :laughing:

.github/workflows/upload_binary.yml
ichard26: https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

.github/workflows/upload_binary.yml
anders-kiaer: I haven't used `actions/upload-release-asset` myself so I'm not 100% sure how it works, but was wondering if it is possible to instead of
```
upload_url: ${{ steps.get_release.outputs.upload_url }}
```
use the gihub event context directly
```
upload_url: ${{ github.event.release.upload_url }}
```
thereby removing the need of the `bruceadams/get-release` action?

<sup>Documentation on [`github.event`](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context) context, and the [`release` payload](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#release).</sup>

ichard26: Oh thank you @anders-kiaer! I didn't know that the release event payload had such a attribute; didn't think of looking in the webhooks documentation. I'll definitely use your suggestion!

.github/workflows/upload_binary.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip wheel setuptools
+          python -m pip install .
+          python -m pip install pyinstaller
+
+      - name: Build binary
+        run: |
+          python -m PyInstaller -F --name black${{ matrix.executable_suffix }} --add-data 'src/blib2to3${{ matrix.pathsep }}blib2to3' src/black/__main__.py
```
JelleZijlstra, line 41: Maybe I'm misunderstanding  how this works, but why is `blib2to3` in the path twice? (`src/blib2to3${{ matrix.pathsep }}blib2to3`)

ichard26: `src/blib2to3` is the source of the data to include and `blib2to3` is the destination path. PyInstaller creates an one-folder bundle that contains an executable so Pyinstaller needs to know where to put the extra data in the bundle directory. `dist` is `blib2to3` and not `src/blib2to3` because PyInstaller hijacks calls to `pkgutil.get_data`. The one file executable PyInstaller can create is really just a bootloader plus an embedded archive that is decompressed and extracted on execution. 

A better explanation is available here: https://pyinstaller.readthedocs.io/en/stable/spec-files.html#adding-data-files

gaborbernat:
Any chance for a release to have again wheels for black?

ichard26:
We plan on the next release to having wheels. Just have to hope they won't be built in a dirty directory again.

gaborbernat:
Any timeframe planned already for that release? 👍

ichard26:
We're trying to target this month, but ideally within a week. No promises though, I don't make those decisions around here.

Issue #1730, opened 2020-09-29
gitignore only respects the root gitignore


max-sixty:
**Is your feature request related to a problem? Please describe.**

This is fairly narrow, and would be understandable if it weren't worth fixing, but posting regardless.

Currently black only respects the root `.gitignore` to ignore files (https://github.com/psf/black/issues/475). So a path like:

```
project
|-.gitignore <- doesn't contain .terraform
|-python-project
|-infra
  |-.gitignore # <- contains .terraform
  |-.terraform
    |-module.py
```

...will still format `module.py`

**Describe the solution you'd like** 

Apply `.gitignore` rules like git does

---

Thanks as ever for the superb library!

hadialqattan:
Hi, I'm working on that, can anyone assign this issue to me?

JulianVolodia:
"have you got home"?

PR #1750, opened 2020-10-08
respect .gitignore files in all levels.

hadialqattan:
`Black` now respects `.gitignore` files in all levels, not only `root/.gitignore` file (apply `.gitignore` rules like `git` does).

Fix: #1730

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 #### _Black_
 
+- `Black` now respects `.gitignore` files in all levels, not only `root/.gitignore` file
+  (#1750).
+
 - `Black` now respects `--skip-string-normalization` when normalizing multiline
   docstring quotes (#1637)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 #### _Black_
 
 - `Black` now respects `.gitignore` files in all levels, not only `root/.gitignore` file
-  (#1750).
+  (like `git` does) (#1750).
 
 - `Black` now respects `--skip-string-normalization` when normalizing multiline
   docstring quotes (#1637)
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -5915,14 +5915,23 @@ def get_future_imports(node: Node) -> Set[str]:
     return imports
 
 
+def get_gitignore_files(root: Path) -> Iterator[Path]:
+    """Yields list of `.gitignore` files paths."""
+    for path in root.iterdir():
+        gitignore = root / ".gitignore"
+        if gitignore.is_file():
+            yield gitignore
+        if path.is_dir():
+            yield from get_gitignore_files(path)
+
+
 @lru_cache()
 def get_gitignore(root: Path) -> PathSpec:
-    """ Return a PathSpec matching gitignore content if present."""
-    gitignore = root / ".gitignore"
-    lines: List[str] = []
-    if gitignore.is_file():
-        with gitignore.open() as gf:
-            lines = gf.readlines()
+    """Return a PathSpec matching gitignore content if present."""
+    lines: Set[str] = set()
+    for path in get_gitignore_files(root):
+        with path.open() as gf:
+            lines |= set(gf.readlines())
     return PathSpec.from_lines("gitwildmatch", lines)
 
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -5917,10 +5917,10 @@ def get_future_imports(node: Node) -> Set[str]:
 
 def get_gitignore_files(root: Path) -> Iterator[Path]:
     """Yields list of `.gitignore` files paths."""
+    gitignore = root / ".gitignore"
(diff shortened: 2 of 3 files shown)


hadialqattan:
Primer has failed because I did a `Black` behavioral changes!

hadialqattan:
@cooperlees would you mind reviewing this?

ichard26 (commented):
I haven't looked much at the code in particular but  I did notice that there is only one root `.gitignore` file in the eyes of `get_sources`. This is a problem since files outside of the scope (or should I say folder) of a `.gitignore` file would still be subject to its exclude rules.

Let's say we have a directory structure like this: 
```
.
├── a
│   ├── a.py
│   ├── b.py
│   └── .gitignore
├── a.py
├── b.py
├── c.py
└── .gitignore
```

And with `./.gitignore` containing `c.py` and `./a/.gitignore` containing `b.py`. 

I would expect that only `./c.py` and `./a/b.py` would be ignored by Git and Black. This is true with Git, but not with Black. Black also ignores the root level `b.py` file because the nested `.gitignore`'s contents was added to the "root" `.gitignore` that all the files and directories Black finds recursively are subject to.

```console
(black) ichard26@acer-ubuntu:~/programming/oss/black$ gh pr checkout 1750
(black) ichard26@acer-ubuntu:~/programming/testing/gitignore$ git status
On branch master
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	a.py
	a/a.py
	b.py

nothing added to commit but untracked files present (use "git add" to track)
(black) ichard26@acer-ubuntu:~/programming/testing/gitignore$ black .  --check -v
.git ignored: matches the --exclude regular expression
c.py ignored: matches the .gitignore file content
b.py ignored: matches the .gitignore file content
a/b.py ignored: matches the .gitignore file content
/home/ichard26/programming/testing/gitignore/a/a.py already well formatted, good job.
/home/ichard26/programming/testing/gitignore/a.py already well formatted, good job.
All done! ✨ 🍰 ✨
2 files would be left unchanged.
```
---

Perhaps make `gen_python_files` only search through one directory and then return the non-excluded files and subdirectories it found. So similar to `os.walk`. And then on every call to `gen_python_files` in `get_sources` a "`.gitignore` file" is created that makes sense for the current directory it's searching. To make this work you would also have to keep track of the all different `.gitignore` files and their locations.

*I'm sorry if that didn't make any sense at all, I'm not great at explain ideas clearly... Anyway, it's just a suggestion to how you could fix this issue. You probably can think of something better :D Thanks for working on this!

hadialqattan:
> I haven't looked much at the code in particular but I did notice that there is only one root `.gitignore` file in the eyes of the exclusion code. This is a problem since files outside of the scope (or should I say folder) of a `.gitignore` file would still be subject to its exclude rules.
> 
> Let's say we have a directory structure like this:
> 
> ```
> .
> ├── a
> │   ├── a.py
> │   ├── b.py
> │   └── .gitignore
> ├── a.py
> ├── b.py
> ├── c.py
> └── .gitignore
> ```
> 
> And with `./.gitignore` containing `c.py` and `./a/.gitignore` containing `b.py`.
> 
> I would expect that only `./c.py` and `./a/b.py` would be ignored by Git and Black. This is true with Git, but not with Black. Black also ignores the root level `b.py` file because the nested `.gitignore`'s contents was added to the "root" `.gitignore` that all the files and directories Black finds recursively are subject to.
> 
> ```
> (black) ichard26@acer-ubuntu:~/programming/oss/black$ gh pr checkout 1750
> (black) ichard26@acer-ubuntu:~/programming/testing/gitignore$ git status
> On branch master
> Untracked files:
>   (use "git add <file>..." to include in what will be committed)
> 	a.py
> 	a/a.py
> 	b.py
> 
> nothing added to commit but untracked files present (use "git add" to track)
> (black) ichard26@acer-ubuntu:~/programming/testing/gitignore$ black .  --check -v
> .git ignored: matches the --exclude regular expression
> c.py ignored: matches the .gitignore file content
> b.py ignored: matches the .gitignore file content
> a/b.py ignored: matches the .gitignore file content
> /home/ichard26/programming/testing/gitignore/a/a.py already well formatted, good job.
> /home/ichard26/programming/testing/gitignore/a.py already well formatted, good job.
> All done! ✨ 🍰 ✨
> 2 files would be left unchanged.
> ```
> 
> Perhaps make `gen_python_files` only search through one directory and then return the non-excluded files and subdirectories it found. So similar to `os.walk`. And then on every call to `gen_python_files` in `get_sources` a "`.gitignore` file" is created that makes sense for the current directory it's searching. To make this work you would also have to keep track of the all different `.gitignore` files and their locations.
> 
> *I'm sorry if that didn't make any sense at all, I'm not great at explain ideas clearly... Anyway, it's just a suggestion to how you could fix this issue. You probably can think of something better :D Thanks for working on this!

I see your point, you know what, instead of adding every `.gitignore` content to the root one, I'm gonna change the entire searching method.

PR #1752, opened 2020-10-09
Fix GitHub markdown links to work on RTD

dwhswenson:
Links to Markdown files on readthedocs.io are swallowing the `.md` extension, causing 404s. Example: from README.md, the text

```markdown
[compatible_configs](https://github.com/psf/black/blob/master/docs/compatible_configs.md)
```

is creating a link to https://github.com/psf/black/blob/master/docs/compatible_configs, which does not exist. ([Here's the relevant paragraph on RTD.](https://black.readthedocs.io/en/stable/installation_and_usage.html#using-black-with-other-tools)) Pointing to a specific anchor seems to fix it, e.g.,

```markdown
[compatible_configs](https://github.com/psf/black/blob/master/docs/compatible_configs.md#black-compatible-configurations)
```

Note that these still lead to GitHub, instead of staying on RTD, which feels a little odd from the user side (but was needed to make links work on PyPI in #1397.)

EDIT: It looks like this has been fixed in https://github.com/readthedocs/recommonmark/pull/181, so next release of recommonmark would make these changes moot.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ your PR. You may need to change
 configuration for it to pass.
 
 For more `black-primer` information visit the
-[documentation](https://github.com/psf/black/blob/master/docs/black_primer.md#).
+[documentation](https://github.com/psf/black/blob/master/docs/black_primer.md#black-primer).
 
 ## Hygiene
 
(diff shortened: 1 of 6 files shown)


hadialqattan (approved):

cooperlees (approved):
LGTM. Thanks!

Issue #1751, opened 2020-10-08
Allow providing command line options to GitHub action


tburrows13:
Currently the GitHub action (https://github.com/psf/black/blob/master/Dockerfile) is hardcoded to use the arguments `--check --diff`. It would be useful to be able to provide custom options, in particular the target Python version.

I used to use https://github.com/lgeiger/black-action, which allowed any command line arguments to be passed through - unfortunately it stopped working this week, so I found this official action instead.

tburrows13:
Sorry, I only just noticed #1755, which has now been merged - it looks great, but I had a couple more points:

1. That PR updates `docs/github_actions.md` to mention the new functionality, but the same information is repeated in the readme: https://github.com/psf/black/blob/master/README.md#github-actions, which has not been updated.

2. Would it be possible to make this repo appear in the GitHub marketplace, so that when you search https://github.com/marketplace?query=black it shows up. I (and presumably many others) only found out that there was an 'official' black action from a random comment somewhere on GitHub.

sam-hoffman:
I'd like to second the request for the official black action in the marketplace!

PR #1755, opened 2020-10-09
Allow black's Github action params overriding.

hadialqattan:
Fix: #1751


cooperlees (approved):
LGTM! Thanks - much more flexible.

hadialqattan:
> LGTM! Thanks - much more flexible.

That's my pleasure 🌹

docs/github_actions.md
```diff
@@ -15,5 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - uses: psf/black@stable
+      - uses: psf/black@stable # the default is equivalent to `black . --diff --check`.
+        with: # (optional - override the default params).
+          args: [".", "--diff", "--check"]
```
ichard26, line 20: I just copied and pasted this example config and it has a syntactical error: https://github.com/ichard26/testblackaction/actions/runs/312826169

Appears that `with.args` doesn't support arrays containing strings :/ 
> A string that defines the inputs for a Docker container. GitHub passes the args to the container's ENTRYPOINT when the container starts up. An array of strings is not supported by this parameter.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepswithargs

ichard26 (changes requested):
Hey thanks for the PR! Please fix the example though.

docs/github_actions.md
hadialqattan: Sorry about that, updated.

hadialqattan:
> Hey thanks for the PR! Please fix the example though.

That's my pleasure 🌹

Thanks a lot for the review!

ichard26 (changes requested):
While the example configuration is now valid, the action is still not configurable. [Look at this run for example](https://github.com/ichard26/testblackaction/runs/1269727798?check_suite_focus=true), the workflow file contains this:

```yaml
     - uses: psf/black@stable # the default is equivalent to `black . --diff --check`.
       with: # (optional - override the default params).
         args: "--version"
```

Black should just print its version and then quit, but the logs shows otherwise. I have no idea why this is happening as when I copy the entrypoint script locally it works perfectly. Maybe Docker/GHA related?

hadialqattan:
> While the example configuration is now valid, the action is still not configurable. [Look at this run for example](https://github.com/ichard26/testblackaction/runs/1269727798?check_suite_focus=true), the workflow file contains this:
> 
> ```yaml
>      - uses: psf/black@stable # the default is equivalent to `black . --diff --check`.
>        with: # (optional - override the default params).
>          args: "--version"
> ```
> 
> Black should just print its version and then quit, but the logs shows otherwise. I have no idea why this is happening as when I copy the entrypoint script locally it works perfectly. Maybe Docker/GHA related?

I don't think so, it's because we're using the stable version which has no customization option!

ichard26:
Okay, good catch, but why does the action still fails to run after pointing your fork's action branch? https://github.com/ichard26/testblackaction/runs/1271524579?check_suite_focus=true

hadialqattan:
> Okay, good catch, but after pointing to your fork's action branch the action still fails to run ... https://github.com/ichard26/testblackaction/runs/1271524579?check_suite_focus=true

OMG, I don't know why I didn't make it executable!!!

**Sorry** for that, **UPDATED**.

**Please see: https://github.com/hadialqattan/testblackaction/runs/1272182754?check_suite_focus=true**

ichard26 (approved):
It finally works! :tada: 
https://github.com/ichard26/testblackaction/runs/1272184851

Don't worry about us having to go back and forth a few times, it wasn't a big deal. Just take it as a lesson to test out your change before submitting a PR if possible :)

Thanks for being cooperative with me and making this a smooth review experience.

hadialqattan:
> It finally works! 🎉
> https://github.com/ichard26/testblackaction/runs/1272184851
> 
> Don't worry about us having to go back and forth a few times, it wasn't a big deal. Just take it as a lesson to test out your change before submitting a PR if possible :)
> 
> Thanks for being cooperative with me and making this a smooth review experience.

I appreciate the fast response!

Thank you for your patience, greetings.

PR #1763, opened 2020-10-14
Switch to pytest and tox

saroad2:
It became a good practice in python to use [Tox](https://tox.readthedocs.io/en/latest/). In that way we can make sure that the contributors of black test in the same environment as the CI process.

Moreover, I propose moving to use [Pytest](https://docs.pytest.org/en/stable/) as the testing infrastructure because it could make our tests much clearer. Consider the following tests from *test_black.py*:

```python
    ...
    def test_function(self) -> None:
        source, expected = read_data("function")
        actual = fs(source)
        black.assert_equivalent(source, actual)
        black.assert_stable(source, actual, DEFAULT_MODE)

    def test_function2(self) -> None:
        source, expected = read_data("function2")
        actual = fs(source)
        black.assert_equivalent(source, actual)
        black.assert_stable(source, actual, DEFAULT_MODE)
    ...
```

Those are practically the same test. Why do we need to write it twice? with *pytest* we could simply write:

```python
    ...
    @pytest.mark.parametrize("file_name", ["function", "function2"])
    def test_function(self, file_name) -> None:
        source, expected = read_data(file_name)
        actual = fs(source)
        black.assert_equivalent(source, actual)
        black.assert_stable(source, actual, DEFAULT_MODE)
    ...
```

I think both changes will be beneficial for *Black*.

What do you think?

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -21,11 +21,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade coverage
-          python -m pip install --upgrade hypothesmith
-          python -m pip install -e ".[d]"
+          python -m pip install --upgrade tox
 
       - name: Run fuzz tests
         run: |
-          coverage run fuzz.py
-          coverage report
+          tox -e fuzz
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,9 +22,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade coverage
-          python -m pip install -e ".[d]"
+          python -m pip install --upgrade coverage tox
 
       - name: Unit tests
         run: |
-          coverage run -m unittest
+          tox -e py
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade coverage tox
+          python -m pip install --upgrade tox
 
       - name: Unit tests
         run: |
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{36,37,38,39}
+envlist = py{36,37,38,39},fuzz
 
 [testenv]
 setenv = PYTHONPATH = {toxinidir}/src
@@ -10,4 +10,17 @@ commands =
     pip install -e .[d]
     coverage erase
     coverage run -m pytest tests
+    coverage report
+
+[testenv:fuzz]
+skip_install = True
+deps =
+    -r{toxinidir}/test_requirements.txt
+    hypothesmith
+    lark-parser < 0.10.0
+; lark-parser's version is set due to a bug in hypothesis. Once it solved, that would be fixed.
+commands =
+    pip install -e .[d]
+    coverage erase
+    coverage run fuzz.py
     coverage report
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,13 @@
+[tox]
+envlist = py{36,37,38,39}
+
+[testenv]
+setenv = PYTHONPATH = {toxinidir}/src
+skip_install = True
+deps =
+    -r{toxinidir}/test_requirements.txt
+commands =
+    pip install -e .[d]
+    coverage erase
+    coverage run -m pytest tests
+    coverage report
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ src/_black_version.py
 .dmypy.json
 *.swp
 .hypothesis/
+venv/
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,0 +1,4 @@
+pytest >= 6.0.1
+pytest-mock >= 3.2.0
+pytest-cases >= 2.1.2
+coverage >= 5.2.1
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -22,7 +22,6 @@ from typing import (
     Dict,
     Generator,
     List,
-    Tuple,
     Iterator,
     TypeVar,
 )
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -36,18 +36,10 @@ from click.testing import CliRunner
 import black
 from black import Feature, TargetVersion
 
-try:
-    import blackd
-    from aiohttp.test_utils import AioHTTPTestCase, unittest_run_loop
-    from aiohttp import web
-except ImportError:
-    has_blackd_deps = False
-else:
-    has_blackd_deps = True
-
 from pathspec import PathSpec
 
 # Import other test classes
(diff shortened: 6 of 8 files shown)


.github/workflows/fuzz.yml
```diff
@@ -21,11 +21,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade coverage
-          python -m pip install --upgrade hypothesmith
-          python -m pip install -e ".[d]"
+          python -m pip install --upgrade tox
```
cooperlees, line 24: Why not install together?
`python -m pip install --upgrade pip tox`

.github/workflows/test.yml
```diff
@@ -22,9 +22,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade coverage
-          python -m pip install -e ".[d]"
+          python -m pip install --upgrade tox
```
cooperlees, line 25: Same - Install both?

cooperlees (commented):
Thanks for this. I'm all for making our tests runnable from `tox` and `pytest`, but not enforcing the use of `pytest` in our CI. What I'm not a big fan of is pytest's magic and overloading of standard python functionality (e.g. assert) and that it's now another dependency to black to test. But I'll live with it if other black core devs like pytest, I've had to learn it from other projects.

Aside from this, love the refactor and split of blackd tests. Love the clearer deps now too.

.github/workflows/fuzz.yml
hugovk: Maybe the newest pip has something new and exciting, and we want to use the new and exciting pip for the other stuff?

cooperlees: Just seems a waste of an extra fork/exec for now ... If we hit a bug then I'd change back to the two executes ... Not a big deal, this why I didn't request changes ...

saroad2:
I'm not in love with the idea of *Pytest*, but I have a great experience with it in all of my resources.

If I'm not using `pytest.mark.parametrize` may I suggest using [this](https://github.com/wolever/parameterized). Never tried it so I don't know how it is, but I think it worths I try. maybe I can try to move tests to it in another PR after you merge this one.

.github/workflows/fuzz.yml
saroad2: I'm with @hugovk on this.

cooperlees:
> I'm not in love with the idea of _Pytest_, but I have a great experience with it in all of my repositories.
> 
> If I'm cannot use `pytest.mark.parametrize`, may I suggest using [this](https://github.com/wolever/parameterized) instead. Never tried it so I don't know how it is, but I think it worths a try. Maybe I can try to move tests to use it in another PR after you merge this one.

Yes, let's definitely do the test conversions in a separate PR.
- Also, lets wait and get other Core Devs for it before doing the conversion. Support is one thing, going `pytest` only is another.

I just want to state I'm not blocking pytest, but sharing I'm not for it. Mainly due to how much long, due to it's `magic` it's taken me to debug things. But that's primarily cause I don't use it primarily.

.github/workflows/fuzz.yml
hugovk: If it were a single call to pip, then there's no point upgrading pip (unless it is cached for _next time_, and it's not).

A concrete reason:

For Python 3.8 on Windows, pip 19.3+ is needed to install wheels (at least for Pillow). We've added the double command to the [installation instructions](https://pillow.readthedocs.io/en/stable/installation.html#basic-installation), and that came up again [this week](https://github.com/python-pillow/Pillow/issues/4970#issuecomment-707299998).

This isn't _specifically_ needed for Black test right now, but I suggest skipping the pip upgrade altogether and using the CI image's version, or having two calls.

(Right now, pip in the image is the latest version, so the call is really quick; and it's pretty quick for an upgrade as well.)

ichard26:
While on the topic of `In that way we can make sure that the contributors of black test in the same environment as the CI process.`, perhaps instead of doing a regular pip editable install, a pipenv development install should be done. This makes sure that the CI is using the identical modules and packages that developers of Black use and therefore increasing reproducibility . The other benefits I can think of include:

* Good caching can be implemented allowing feedback to contributors be to returned faster
* When a dependency upgrade happens, the change is automatically tested. This makes using Dependabot or other services safer to use. It also protects us from \*me\* messing up a dependency upgrade like in PR #1501 and then PR #1510 :-)

The main downside is that Black incompatibilities with newer versions of modules / packages won't be automatically discovered since we'll be pinning to probably older versions. Using PyUp or other services mostly mitigates this issue though.

Just some ideas to think about.

cooperlees:
> While on the topic of `In that way we can make sure that the contributors of black test in the same environment as the CI process.`, perhaps instead of doing a regular pip editable install, a pipenv development install should be done. This makes sure that the CI is using the identical modules and packages that developers of Black use and therefore increasing reproducibility . The other benefits I can think of include:
> 
> * Good caching can be implemented allowing feedback to contributors be to returned faster
> * When a dependency upgrade happens, the change is automatically tested. This makes using Dependabot or other services safer to use. It also protects us from *me* messing up a dependency upgrade like in PR #1501 and then PR #1510 :-)
> 
> The main downside is that Black incompatibilities with newer versions of modules / packages won't be automatically discovered since we'll be pinning to probably older versions. Using PyUp or other services mostly mitigates this issue though.
> 
> Just some ideas to think about.

If you do the PRs I'm sure we'll merge them :) ... 

I love PyUp method, but I always like to have both a pinned and unpinned options. This is due to having situations where I've had to fork to unpin to install things in some environments. And I've not wanted to inflict that pain on users of libraries I've maintained.

cooperlees (approved):
This is an improvement to allow for Devs to get a CI like env easier. Lets merge it.

cooperlees:
No one has spoke up about not liking this. I'm merging.

tox.ini
```diff
@@ -0,0 +1,26 @@
...
+    coverage run -m pytest tests
+    coverage report
+
+[testenv:fuzz]
+skip_install = True
+deps =
+    -r{toxinidir}/test_requirements.txt
+    hypothesmith
+    lark-parser < 0.10.0
+; lark-parser's version is set due to a bug in hypothesis. Once it solved, that would be fixed.
```
jayaddison, line 21: @saroad2 It looks like `lark-parser`'s been updated in the [test dependencies for `hypothesmith`](https://github.com/Zac-HD/hypothesmith/blob/d34c2947e20385c2d1a337b337027d6d39ada3af/deps/test.txt#L35-L36) - could/should we remove this version pin?

saroad2: I guess we should :) I added this pinning so the tests would pass.

ichard26:
Hi @saroad2, thanks for moving us to pytest, it's been a better experience personally. Anyway I'm working on removing parameterized as a test suite dependency and I noticed pytest-mock + pytest-cases. I'm not familiar with these plugins and AFAICT the test suite doesn't use these plugins ATM. If that's the case I'd wish to remove them. Why were they added? 

Thank you in advance!

saroad2:
Hey @ichard26, I think I added them by mistake.

However, if you take a close look at *pytest-cases*, I think you'll find it as a very helpful plugin to *pytest*.
Using it might cause you to rewrite some tests, but in a much prettier form, in my humble opnion.

PR #1775, opened 2020-10-21
Use lowercase hex numbers fixes #1692

### LINKED ISSUE #1692: Suggestion: use lowercase letters in hex literals
Reading the code, when I saw the commend saying `All letters used in the representation are normalized to lowercase`, I thought it also refers to hex letters. However, a few lines later, I was disappointed to see `Change hex literals to upper case.`.

https://github.com/psf/black/blob/6284953d07060804dcdeacf4626b76aed7a20683/src/black/__init__.py#L5158-L5168

There's one practical issue with use of upper-case hex letters, though, and that's similarity between letter `B` and digit `8`.

I have seen bugs caused by this similarity in security-sensitive code, such as UTF-8/-16 decoders. Personally, I always used uppercase letters in hex numerals prior to that incident, because of the aesthetics of it; but, since then, have been switching to lowercase anywhere possible.

Maybe Black would consider switching to lowercase hex letters, as well?

C0DK:
First PR onto black. Please tell me if i did something wrong according to your ordinary rules.

The issue, #1692, argues whether we/you should "simply" do this, as it might impact source code of a large set of users, however you can either merge it if you like it or wait if not. :) Please write me if I can do anything in that regard.

I created two seperate commits:
- Made hex lower case
- Refactored numeric formatting section

This was done to seperate the two focus points, the first one solves the issue, whereas the second adds a bit of refactoring to that part of the code. Whether my version has a higher code quality or not, is definitely subjective, so please ignore it and cherry-pick if you dislike it :) 

I am looking forward to contributing to black, long time user, and all that, so I hope you like it. I'll try to do something once in a while - Am doing this as part of hacktoberfest though, so if you can either label it `hacktoberfest-accepted` or merge it before the end of the month, that will be great ;) I feel super cheap writing that, but i am trying to pimp my bike with a yearly sticker from hacktoberfest, and this month is running a bit short. If any changes are needed, I'll gladly make those :) 

Thank you!


src/black/__init__.py
```diff
@@ -69,7 +69,8 @@
 DEFAULT_INCLUDES = r"\.pyi?$"
 CACHE_DIR = Path(user_cache_dir("black", version=__version__))
 
-STRING_PREFIX_CHARS: Final = "furbFURB"  # All possible string prefix characters.
+# All possible string prefix characters.
+STRING_PREFIX_CHARS: Final = "furbFURB"
```
ichard26, line 73: NIT: IMO this is unnecessary and just ruins blame information.

src/black/__init__.py
```diff
@@ -5296,7 +5341,8 @@ def convert_one_fmt_off_pair(node: Node) -> bool:
                 if not ignored_nodes:
                     continue
 
-                first = ignored_nodes[0]  # Can be a container node with the `leaf`.
+                # Can be a container node with the `leaf`.
+                first = ignored_nodes[0]
```
ichard26, line 5345: Ditto

src/black/__init__.py
```diff
@@ -1901,7 +1902,8 @@ def line(self, indent: int = 0) -> Iterator[Line]:
         """
         if not self.current_line:
             self.current_line.depth += indent
-            return  # Line is empty, don't emit. Creating a new one unnecessary.
+            # Line is empty, don't emit. Creating a new one unnecessary.
+            return
```
ichard26, line 1906: Ditto

.gitignore
```diff
@@ -1,3 +1,6 @@
+.venv/
+
+
```
ichard26, line 3: I would prefer if the two extra empty lines were removed.

ichard26 (commented):
Didn't do a deep check on the code, but I did notice a few stylistic things.

src/black/__init__.py
C0DK: was a result of autoformatter. I see the point. will revert.

.venv
```diff
@@ -0,0 +1 @@
+black
```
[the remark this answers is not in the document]
C0DK: Was an error. sorry. should have been caught by the gitignore.

.gitignore
C0DK: Will fix. Thank you.

src/black/__init__.py
```diff
@@ -5174,35 +5176,78 @@ def normalize_numeric_literal(leaf: Leaf) -> None:
     in Python 2 long literals).
     """
     text = leaf.value.lower()
-    if text.startswith(("0o", "0b")):
+    if is_oct_or_binary(text):
         # Leave octal and binary literals alone.
         pass
-    elif text.startswith("0x"):
```
JelleZijlstra, line 5180: I don't think we need this refactoring; to me it just creates an excessive number of small functions.

test_requirements.txt
```diff
@@ -1,4 +1,6 @@
 pytest >= 6.0.1
 pytest-mock >= 3.2.0
 pytest-cases >= 2.1.2
-coverage >= 5.2.1
\ No newline at end of file
+coverage >= 5.2.1
+aiohttp >= 3.3.2
+aiohttp-cors
```
hugovk, line 6: Are these two additions needed?

C0DK: Yes - i was not able to run all the tests witout them :) Some tests rely on AIO stuff, so when i installed them i added them as well :)

src/black/__init__.py
C0DK: That is fair - I relied on the Uncle Bob principle of "If you need a comment, then you should probably just have a more descriptive function name", but i might be misquoting or misunderstanding that. It is definitely a subjective matter. I liked the, in my opinion, higher level of readability, but the, already long, file gets longer which is maybe not cool. We can also drop the whole refactoring commit (or only take the formatting parts).

It's subjective :) I'll go with what you guys prefer.

test_requirements.txt
hugovk: Aha, [`CONTRIBUTING.md`](https://github.com/psf/black/blob/master/CONTRIBUTING.md) suggests using pipenv, which includes them.

I'm not so familiar with pipenv, but I see Travis CI runs `pip install -e '.[d]'`, where the `d` is a so-called "extra" listed in `setup.py`:
```python
    extras_require={
        "d": ["aiohttp>=3.3.2", "aiohttp-cors"],
        "colorama": ["colorama>=0.4.3"],
    },
```
So that will install them for you. And GitHub Actions uses tox which also installs with the extra.

Want to try running `pip install -e '.[d]'` instead?

.venv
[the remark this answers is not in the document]
C0DK: Fixed

.gitignore
C0DK: Fixed

src/black/__init__.py
C0DK: Fixed (also all the rest)

test_requirements.txt
ichard26: AFAIK this requirements file is for CI not developers of Black. Doing a Pipenv developement install should install Black with its HTTP web server dependencies. https://github.com/psf/black/blob/407052724fa1c97ee8bcd4e96de650def00be03e/Pipfile#L21 

You get pinned dependencies too so doing a Pipenv development install is recommended

C0DK: oic. I use some...thing.. not actually sure. I simply realized it wasn't exhaustive, not that it was defined elsewhere. Should i remove these two lines?

hugovk: Another popular thing to do is to move tests dependencies from `test_requirements.txt` into a `tests` extra, for example:

* https://github.com/hynek/structlog/blob/69d9e9328d86cdc5e395eadd119c9775ee7343f7/setup.py#L41-L49

But that's for another PR.

hugovk: Yes, I think remove those two lines to avoid duplication and keep things lean 👍

src/black/__init__.py
kamahen: It doesn't hurt to add a comment to some of these if statements. E.g.:

`elif "e" in text:  # scientific notation`

C0DK: The function, in my opinion, does exactly the same without duplication and without revealing implementation details. Is a general consesus that the functions decrease quality? As mentioned, i explicitly added it in a seperate commit to make it easy to drop :)

test_requirements.txt
C0DK: Fixed :)

hugovk:
For the Primer failures:

The CI is running Black on a few other projects to see if the formatting changes. Sometimes we don't want that to happen. In this case, some changes are expected.

I checked [a log](https://github.com/psf/black/pull/1775/checks?check_run_id=1294465012), 5 projects now fail and they all look like hex number changes, as expected.

To fix: flip `expect_formatting_changes": false,` to true in `src/black_primer/primer.json` for attrs, hypothesis, Pillow, pyramid and virtualenv.

C0DK:
@hugovk Thank you :) Should be fixed now.

Regarding my refactoring, show i revert those? What is the general consesus? I can always revert it and then put it in another PR

ichard26:
I completely forgot but please also add an entry in the change log, thank you!

src/black/__init__.py
```diff
@@ -5174,35 +5174,78 @@ def normalize_numeric_literal(leaf: Leaf) -> None:
...
+    can cause security issues.
+    see: https://github.com/psf/black/issues/1692
+    """
+
+    before, after = text[:2], text[2:]
+    return f"{before}{after.lower()}"
+
+
+def is_scientific_notation(text: str) -> bool:
+    """Checks if the supplied string is a number with scientific notation"""
```
kamahen, line 5217: This code matches 0xdeadbeef; it's an incorrect abstraction that depends on is_hex and other test having been applied earlier.

C0DK: Valid. Should i move all the checkers into the original function, while still moving the formatting into sub functions? Would that be better.

C0DK: I moved these now :)

cooperlees (approved):
LGTM - But want other core devs completely happy too.

C0DK:
> LGTM - But want other core devs completely happy too.

I'll update the changelog in less than 12 hours. 

I hope i can get a `hacktoberfest-approved` label on this by the end of the month :p if you are desperately against hacktoberfest, then i understand :) I enjoy it as a reminder to give back to the community, meanwhile sticker-hunting is a desperate hobby of mine.

zsol:
LGTM. To make me _completely_ happy, I'd also revert the refactoring that creates three tiny functions that @JelleZijlstra pointed out.

C0DK:
> LGTM. To make me _completely_ happy, I'd also revert the refactoring that creates three tiny functions that @JelleZijlstra pointed out.
Whichh are these? I did revert the smallest of functions, but kept the formatting functions, as the numeric formatting already was in it's own function. This keeps it on the same level of abstraction, IMO.

src/black/__init__.py
```diff
@@ -5178,31 +5178,52 @@ def normalize_numeric_literal(leaf: Leaf) -> None:
...
-        text = f"{format_float_or_int_string(number)}{suffix}"
+        text = format_long_or_complex_number(text)
     else:
         text = format_float_or_int_string(text)
     leaf.value = text
 
 
+def format_hex(text: str) -> str:
+    """
+    Formats a hexidecimal strign like "0x12b3"
```
zsol, line 5193:
```suggestion
    Formats a hexadecimal string like "0x12b3"
```

zsol:
> I did revert the smallest of functions, but kept the formatting functions, as the numeric formatting already was in it's own function. This keeps it on the same level of abstraction, IMO.

You're right, the formatting functions are fine. I was referring to the functions that only called `startswith("0x")` and such. All good now.

C0DK:
I've mentioned it before, but didn't get any response: Will you add a hacktoberfest label? :P It's fine if you guys don't want to encourage it - I totally respect that. I'm just a bit pressed if i want to partake this year otherwise haha.

hugovk:
(@C0DK Some easy ones over here: https://github.com/hugovk/pypistats/issues/181)

cooperlees:
> I've mentioned it before, but didn't get any response: Will you add a hacktoberfest label? :P It's fine if you guys don't want to encourage it - I totally respect that. I'm just a bit pressed if i want to partake this year otherwise haha.

Is this label all you need?

C0DK:
> Is this label all you need?

`hacktoberfest-accepted` - to be really annoying ;)

PR #1780, opened 2020-10-22
Provide a stdin-filename to allow stdin to respect force-exclude rules

### LINKED ISSUE #1776: Provide a --stdin-display-name option like flake8
I'm trying to use `black` as a fixer on ale on vim. It works fine and my config file is respected, except for the `exclude` and `force-exclude` files/folders.

The problem is that the tool calls black passing the code through stdin, and thus, black doesn't know the filename that is being parsed.

flake8 workas around that issue by providing a `--stdin-display-name` which can be used to identify which is the name of the file being passed through stdin, allowing those exclude rules to be followed.

What do you think?

bellini666:
Fix #1776

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Options:
                                   when they are passed explicitly as arguments.
 
   --stdin-filename TEXT           The name of the file when passing it through
-                                  stdin. Useful to make sure black will respect
+                                  stdin. Useful to make sure Black will respect
                                   --force-exclude option on some editors that
                                   rely on using stdin.
 
--- a/README.md
+++ b/README.md
@@ -135,6 +135,11 @@ Options:
                                   matching this regex will be excluded even
                                   when they are passed explicitly as arguments.
 
+  --stdin-filename TEXT           The name of the file when passing it through
+                                  stdin. Useful to make sure black will respect
+                                  --force-exclude option on some editors that
+                                  rely on using stdin.
+
   -q, --quiet                     Don't emit non-error messages to stderr.
                                   Errors are still emitted; silence those with
                                   2>/dev/null.
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -633,19 +633,7 @@ def get_sources(
             p = Path(s)
             is_stdin = False
 
-        if not is_stdin or p.is_dir():
-            sources.update(
-                gen_python_files(
-                    p.iterdir(),
-                    root,
-                    include_regex,
-                    exclude_regex,
-                    force_exclude_regex,
-                    report,
-                    gitignore,
-                )
-            )
-        elif is_stdin or p.is_file():
+        if is_stdin or p.is_file():
             normalized_path = normalize_path_maybe_ignore(p, root, report)
             if normalized_path is None:
                 continue
@@ -664,6 +652,18 @@ def get_sources(
                 p = Path(f"{STDIN_PLACEHOLDER}{str(p)}")
 
             sources.add(p)
+        elif p.is_dir():
+            sources.update(
+                gen_python_files(
+                    p.iterdir(),
+                    root,
+                    include_regex,
+                    exclude_regex,
+                    force_exclude_regex,
+                    report,
+                    gitignore,
+                )
+            )
         elif s == "-":
             sources.add(p)
         else:
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -692,7 +692,6 @@ def reformat_one(
     """
     try:
         changed = Changed.NO
-        is_stdin = False
 
         if str(src) == "-":
             is_stdin = True
@@ -701,8 +700,10 @@ def reformat_one(
             # Use the original name again in case we want to print something
             # to the user
             src = Path(str(src)[len(STDIN_PLACEHOLDER) :])
+        else:
+            is_stdin = False
 
-        if not src.is_file() and is_stdin:
+        if is_stdin:
             if format_stdin_to_stdout(fast=fast, write_back=write_back, mode=mode):
                 changed = Changed.YES
         else:
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -462,7 +462,7 @@ def target_version_option_callback(
     type=str,
     help=(
         "The name of the file when passing it through stdin. Useful to make "
-        "sure black will respect --exclude/--force-exclude options on some "
+        "sure black will respect --force-exclude option on some "
         "editors that rely on using stdin."
     ),
 )
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -644,7 +644,7 @@ def get_sources(
                     gitignore,
                 )
             )
-        elif p.is_file():
+        elif p.is_file() or is_stdin:
             normalized_path = normalize_path_maybe_ignore(p, root, report)
             if normalized_path is None:
                 continue
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -68,6 +68,7 @@ DEFAULT_LINE_LENGTH = 88
 DEFAULT_EXCLUDES = r"/(\.direnv|\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|\.svn|_build|buck-out|build|dist)/"  # noqa: B950
 DEFAULT_INCLUDES = r"\.pyi?$"
 CACHE_DIR = Path(user_cache_dir("black", version=__version__))
+STDIN_PLACEHOLDER = "__BLACK_STDIN_FILENAME__"
 
 STRING_PREFIX_CHARS: Final = "furbFURB"  # All possible string prefix characters.
 
@@ -660,7 +661,7 @@ def get_sources(
                 continue
 
             if is_stdin:
-                p = Path("__BLACK_STDIN_FILENAME__" / p)
+                p = Path(f"{STDIN_PLACEHOLDER}{str(p)}")
 
             sources.add(p)
         elif s == "-":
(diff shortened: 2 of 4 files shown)


src/black/__init__.py
```diff
@@ -626,8 +644,6 @@ def get_sources(
                     gitignore,
                 )
             )
-        elif s == "-":
-            sources.add(p)
         elif p.is_file():
```
ichard26, line 647:
```suggestion
        elif p.is_file() or is_stdin:
```

With a custom stdin filepath given, Black will try to format it in place as if it was a file:

```console
(black) ichard26@acer-ubuntu:~/programming/oss/black$ echo "print('hello world')" | black -v --stdin-filename="hello.py" -
Using configuration from /home/ichard26/programming/oss/black/pyproject.toml.
  File "/home/ichard26/programming/oss/black/src/black/__init__.py", line 706, in reformat_one
  File "/home/ichard26/programming/oss/black/src/black/__init__.py", line 845, in format_file_in_place
    then = datetime.utcfromtimestamp(src.stat().st_mtime)
  File "/opt/python3.8.5/lib/python3.8/pathlib.py", line 1194, in stat
    return self._accessor.stat(self)
FileNotFoundError: [Errno 2] No such file or directory: 'hello.py'
error: cannot format hello.py: [Errno 2] No such file or directory: 'hello.py'
Oh no! 💥 💔 💥
1 file failed to reformat.
```

This is because `Path("-").is_file` returns `False` because a file doesn't actually exist at that path ([See `pathlib.Path.is_file` docs](https://docs.python.org/3.8/library/pathlib.html#pathlib.Path.is_file)). Therefore the path isn't prefixed with the magic string that tells `reformat_one` to use `format_stdin_to_stdout` and not `format_file_in_place`. Also `force-exclude` wouldn't be checked, but Black trying to format a file that doesn't exist is a bigger problem without this suggestion :-)

ichard26, line 647: BTY `--exclude` isn't checked here as it is only checked against recursively found files and directories so the help for your new option is a bit misleading. Only `--force-exclude` is checked.

```console
(black) ichard26@acer-ubuntu:~/programming/oss/black$ echo "print('hello world')" | black -v --force-exclude="hello.*" --stdin-filename="hello.py" -
Using configuration from /home/ichard26/programming/oss/black/pyproject.toml.
hello.py ignored: matches the --force-exclude regular expression
(black) ichard26@acer-ubuntu:~/programming/oss/black$ echo "print('hello world')" | black -v --exclude="hello.*" --stdin-filename="hello.py" -
Using configuration from /home/ichard26/programming/oss/black/pyproject.toml.
print("hello world")
reformatted __BLACK_STDIN_FILENAME__/hello.py
All done! ✨ 🍰 ✨
1 file reformatted.
```
**NOTE**: the suggestion above must be applied for this demo to work

ichard26 (changes requested):
Thanks for the PR @bellini666! 

Please update the help output that [is in the README](https://github.com/psf/black/blob/master/README.md), [add a test as this is a new feature](https://github.com/psf/black/blob/master/CONTRIBUTING.md#hygiene), and see my other comments below.

Also, personally I find the option name of `stdin-filename` a bit too specific as you can pass a full absolute filepath. That could be useful for cases where the filename isn't excluded, but its containing directory **is excluded**. Passing filepaths does break when they would point outside of the current working directory but I dunno how much of an issue that is.

src/black/__init__.py
bellini666: Yeah, you are right! Since this is a solution that passing the path itself, it should not enforce exclude but only force-exclude.

Will make the changes together with the change on README and tests.

bellini666:
@ichard26 I have fixed your comments and also added some tests for the new functionality!

README.md
```diff
@@ -135,6 +135,11 @@ Options:
                                   matching this regex will be excluded even
                                   when they are passed explicitly as arguments.
 
+  --stdin-filename TEXT           The name of the file when passing it through
+                                  stdin. Useful to make sure black will respect
```
hugovk, line 139:
```suggestion
                                  stdin. Useful to make sure Black will respect
```

src/black/__init__.py
```diff
@@ -462,7 +463,7 @@ def target_version_option_callback(
     type=str,
     help=(
         "The name of the file when passing it through stdin. Useful to make "
-        "sure black will respect --exclude/--force-exclude options on some "
+        "sure black will respect --force-exclude option on some "
```
hugovk, line 466:
```suggestion
        "sure Black will respect --force-exclude option on some "
```

bellini666:
@hugovk I tried requesting your review again by clicking in the `Re-request review` next to you on the reviewers section, but for some reason it is not working. Just pinging you here then to say that your suggestions were applied.

hugovk:
OK, thanks! That wording looks fine now, I'll leave the functionality to the maintainers to review 👍

src/black/__init__.py
```diff
@@ -670,7 +692,17 @@ def reformat_one(
...
+
+        if str(src) == "-":
+            is_stdin = True
+        elif str(src).startswith(STDIN_PLACEHOLDER):
+            is_stdin = True
+            # Use the original name again in case we want to print something
+            # to the user
+            src = Path(str(src)[len(STDIN_PLACEHOLDER) :])
+
+        if not src.is_file() and is_stdin:
```
ichard26, line 705: Be careful, if a file actually exists at the custom `stdin-filename` path, Black will switch to using `format_file_in_place` since `not src.is_file()` would return `False`.

```console
(black) ichard26@acer-ubuntu:~/programming/oss/black$ echo "this_is_a_different_piece_of = 'code'" > temp.py
(black) ichard26@acer-ubuntu:~/programming/oss/black$ echo "print('egg spam')" | black --diff --stdin-filename="temp.py" -
--- temp.py	2020-10-23 20:35:39.055756 +0000
+++ temp.py	2020-10-23 20:36:24.060408 +0000
@@ -1 +1 @@
-this_is_a_different_piece_of = 'code'
+this_is_a_different_piece_of = "code"
would reformat temp.py
All done! ✨ 🍰 ✨
1 file would be reformatted.
```

ichard26 (changes requested):
README looks good and well done on the tests!

Unfortunately I found another edge case where Black incorrectly uses `format_file_in_place`.

ichard26:
I completely forgot but please also add an entry in the change log, thank you!
*<sup>I really need to have a PR review checklist . . .</sup>*

hugovk:
> _I really need to have a PR review checklist . . ._

Or put it in a PR template? 
https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository

bellini666:
@ichard26 ok, everything should be fine now :)

cooperlees (approved):
Thanks! Nice addition. Would love @ichard26 to check it one last time before merge tho since they have the most context.

ichard26:
FYI, I plan on reviewing this soon (as in probably tomorrow).

src/black/__init__.py
```diff
@@ -613,7 +626,13 @@ def get_sources(
...
     for s in src:
-        p = Path(s)
+        if s == "-" and stdin_filename:
+            p = Path(stdin_filename)
+            is_stdin = True
+        else:
+            p = Path(s)
+            is_stdin = False
+
         if p.is_dir():
```
ichard26, line 636:
```suggestion
        if p.is_dir() and not is_stdin:
```

If `--stdin-filename` just happens to point to a directory that exists, Black wrongly switches to in-place file formatting. 
```console
(black) ichard26@acer-ubuntu:~/programming/oss/black$ echo "eggs" | black --stdin-filename="../black" -
All done! ✨ 🍰 ✨
15 files left unchanged.
```

ichard26 (changes requested):
Overall these changes look good other than the additional bug I found.

Many thanks for working on this!

src/black/__init__.py
JelleZijlstra: The `is_stdin` check should be before the `p.is_dir()` to avoid an unnecessary disk read.

src/black/__init__.py
```diff
@@ -626,9 +645,7 @@ def get_sources(
                     gitignore,
                 )
             )
-        elif s == "-":
-            sources.add(p)
-        elif p.is_file():
+        elif p.is_file() or is_stdin:
```
ichard26, line 648:
```suggestion
        elif is_stdin or p.is_file():
```

As per Jelle's good point about reducing disk reads.

ichard26 (changes requested):

bellini666:
@ichard26 after applying the suggestions (https://github.com/psf/black/pull/1780/commits/8d13a9c95bb0fd5160d73fd3b5d3d5326ed46215) some tests broke.

I wasn't able to check it yet, but looking at the commit now, wouldn't that `if not is_stdin or p.is_dir():` force to enter the `p.is_for()` block even if it was supposed to go the the `p.is_file()` when `is_stdin` is `False`? Maybe that line should still be `if p.is_dir()` alone

ichard26:
You're correct, my bad. Uh perhaps this patch should work.

```diff
diff --git a/src/black/__init__.py b/src/black/__init__.py
index 7594cce..0850b11 100644
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -633,19 +633,7 @@ def get_sources(
             p = Path(s)
             is_stdin = False
 
-        if not is_stdin or p.is_dir():
-            sources.update(
-                gen_python_files(
-                    p.iterdir(),
-                    root,
-                    include_regex,
-                    exclude_regex,
-                    force_exclude_regex,
-                    report,
-                    gitignore,
-                )
-            )
-        elif is_stdin or p.is_file():
+        if is_stdin or p.is_file():
             normalized_path = normalize_path_maybe_ignore(p, root, report)
             if normalized_path is None:
                 continue
@@ -664,6 +652,18 @@ def get_sources(
                 p = Path(f"{STDIN_PLACEHOLDER}{str(p)}")
 
             sources.add(p)
+        elif p.is_dir():
+            sources.update(
+                gen_python_files(
+                    p.iterdir(),
+                    root,
+                    include_regex,
+                    exclude_regex,
+                    force_exclude_regex,
+                    report,
+                    gitignore,
+                )
+            )
         elif s == "-":
             sources.add(p)
         else:
```

Didn't give much thought into whether this should work. Tests do pass though. If `p` is an existing file it will go through the right block. Same thing if `--stdin-filename` is used regardless if a file or a directory actually exists at the path. If `p` is an existing directory, well then `is_stdin` and `p.is_file()` will be `False` and the `p.is_dir()` check will pass. If STDIN mode is used but without `--stdin-filename`, the `elif s == "-"` will catch that. Please check this manually before pushing though.

bellini666:
@ichard26 that suggestion worked fine! Just pushed it

cooperlees:
Awesome work all! Thanks for your addition to black!

Issue #1778, opened 2020-10-22
black igoners ex/include in pre-commit run --all-files


Bibo-Joshi:
**Describe the bug** When specifying directories to include via `pyproject.toml`, black will still run on excluded/not included files when running `pre-commit run --all-files`.

**To Reproduce** Steps to reproduce the behavior:

1. clone this MWE-Repo: https://github.com/Bibo-Joshi/black-mwe
2. `cd black-mwe`
3. `pre-commit install`
4. `pre-commit run --all-files` → fails
6. `black . --check` → succeeds as expected

**Expected behavior:** `black` should respect `include` and `ignore` in pre-commit even when running with `--all-files`, because that flag just means "run also on non-changed files" not "ignore all hook settings"

**Environment (please complete the following information):**

- Version: 20.8b1
- OS and Python version: Windows, Py3.8

**Does this bug also happen on master?**

Yes. Installed via `pip install git+https://github.com/psf/black.git` and above steps still fail.

**Additional context** I only recently switched to `black` and really do hope that it's not just some stupid configuration error again like in #1773 😬 I'd be happy for a response :)

MarcoGorelli:
As far as I understand, pre-commit passes each file one-by-one, and so what you end up running is:
```
black file1.py file2.py file3.py ... filen.py
```

If you look at the docs https://black.readthedocs.io/en/stable/installation_and_usage.html?highlight=exclude#command-line-options it says

>   --exclude TEXT                  A regular expression that matches files and
                                  directories that should be excluded on
                                  recursive searches. 

As pre-commit doesn't run `black` recursively, it'll have no effect.

Instead, you should use pre-commit's own `files` and `exclude`.

In your example:
```
repos:
  - repo: https://github.com/psf/black
    rev: 20.8b1
    hooks:
      - id: black
        args:
          - --check
        files: ^(foo|bar)/.*\.py$
        exclude = ^bar/vendor
```

Bibo-Joshi:
Thanks again for the explanation @MarcoGorelli !
Based on your answer another solution would be to `force-exclude` everything *not* in `include`, in this case 

```
force-exclude = '^(?!(foo|bar)/).*\.py$'
```

That allows it to have `pre-commit` and `black` behave the same …
Closing, but this confirms me in PRing to improve the documentation of VC integration

Issue #1773, opened 2020-10-21
black --diff alsways passes pre-commit


Bibo-Joshi:
**Describe the bug**

When running `black` as pre-commit hook with the `--diff` option, the hook always succeeds, even if running `black` manually would reformat files.

**To Reproduce** Steps to reproduce the behavior:

1. clone this MWE-Repo: https://github.com/Bibo-Joshi/black-mwe
2. `cd black-mwe`
3. `pre-commit install`
4. `pre-commit run --all-files` → doesn't fail
5. Remove `diff = true` from `pyproject.toml`
6. `pre-commit run --all-files` → reformats `main.py` and fails

**Expected behavior:** pre-commit hook gives the same result as runnig `black` manually

**Environment:**

- Version: 20.8b1
- OS and Python version: Tried on both Windows and Linux on Py3.8

**Does this bug also happen on master?** To answer this, you have two options:

Using the online formatter at https://black.now.sh/?version=master, it reformats the file `main.py` as expected

**Additional context:** I only recently switched to `black` and really do hope that it's not just some stupid configuration error 😬 But I tried to find solutions online and tested on different machines with the same result. I'd be happy for a response :)

hugobuddel:
I'm also new to black, and pre-commit, and `pyproject.toml`, so the perfect person to answer this :wink:.

It seems that you need to add `check = true` to the black configuration in `pyproject.toml`:
```
$ cat pyproject.toml
[tool.black]
diff = true
check = true
verbose = true

$ pre-commit run --all-files
black....................................................................Failed
- hook id: black
- exit code: 1

Using configuration from [..]/black-mwe/pyproject.toml.
--- main.py     2020-10-22 09:38:34.174300 +0000
+++ main.py     2020-10-22 09:41:17.363695 +0000
@@ -1,11 +1,6 @@
-def print_hi(
-
-
-
-        name):
+def print_hi(name):
     print(f"Hello there! - {name}")
 
 
-if __name__ == '__main__':
-    print_hi(
-        "General Kenobi.")
+if __name__ == "__main__":
+    print_hi("General Kenobi.")
would reformat main.py
Oh no! 💥 💔 💥
1 file would be reformatted.
```

See https://github.com/psf/black/blob/407052724fa1c97ee8bcd4e96de650def00be03e/src/black/__init__.py#L6104

Bibo-Joshi:
Thanks for the solution, @hugobuddel :)
[...]

PR #1783, opened 2020-10-23
Document some culprits with pre-commit

Bibo-Joshi:
This PR adds some documentation about version control integration in hopes of preventing duplicatos of  #1778 and #1773  :)

### DIFF
--- a/docs/version_control_integration.md
+++ b/docs/version_control_integration.md
@@ -24,8 +24,8 @@ for your project. See _Black_'s own
 example.
 
 When using the `--diff` flag with `pre-commit`, you must also use the `--check` flag.
-When you want to run `black` only on specific files in `pre-commit`, either use
-`pre-commits` own `files` and `exclude` or, when using `blacks` `--include`, set
+When you want to run *Black* only on specific files in pre-commit, either use
+pre-commit's own `files` and `exclude` or, when using *Black*'s `--include`, set
 `--force-exclude` to the negated regex of `--include`.
 
 If you're already using Python 3.7, switch the `language_version` accordingly. Finally,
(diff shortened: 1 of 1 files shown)


docs/version_control_integration.md
```diff
@@ -24,7 +24,9 @@ for your project. See _Black_'s own
 example.
 
 When using the `--diff` flag with `pre-commit`, you must also use the `--check` flag.
-When you want to run `black` only on specific files in `pre-commit`, either use `pre-commits` own `files` and `exclude` or, when using `blacks` `--include`, set `--force-exclude` to the negated regex of `--include`.
+When you want to run `black` only on specific files in `pre-commit`, either use
+`pre-commits` own `files` and `exclude` or, when using `blacks` `--include`, set
+`--force-exclude` to the negated regex of `--include`.
```
hugovk, line 29:
```suggestion
When you want to run *Black* only on specific files in pre-commit, either use
pre-commit's own `files` and `exclude` or, when using *Black*'s `--include`, set
`--force-exclude` to the negated regex of `--include`.
```

ichard26 (approved):
Looks good! I personally would like an explanation to why `--exclude/--include` can't be depended on when using pre-commit to be added but I don't care enough to press "request changes" instead.

Thanks for taking some time to work on this @Bibo-Joshi!

Bibo-Joshi:
> Looks good! I personally would like an explanation to why `--exclude/--include` can't be depended on when using pre-commit to be added but I don't care enough to press "request changes" instead.

I thought about that, but trying to give a short explanation might actually be more confusing. One could just link to the two issues, but that's not too nice to have in the documentation …

PR #1785, opened 2020-10-25
Extract formatting tests

saroad2:
A lot of the tests in *Black* are exact duplications of other tests.
In order to remove some of the duplications, added 2 parametrized tests that combine those tests.

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,4 +1,5 @@
-pytest >= 6.0.1
-pytest-mock >= 3.2.0
-pytest-cases >= 2.1.2
-coverage >= 5.2.1
+pytest >= 6.1.1
+pytest-mock >= 3.3.1
+pytest-cases >= 2.3.0
+coverage >= 5.3
+parameterized >= 0.7.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ cache:
   directories:
     - $HOME/.cache/pre-commit
 env:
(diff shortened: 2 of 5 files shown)


test_requirements.txt
```diff
@@ -1,4 +1,5 @@
-pytest >= 6.0.1
-pytest-mock >= 3.2.0
-pytest-cases >= 2.1.2
-coverage >= 5.2.1
\ No newline at end of file
+pytest >= 6.1.1
+pytest-mock >= 3.3.1
+pytest-cases >= 2.3.0
+coverage >= 5.3
+parameterized >= 0.7.4
```
hugovk, line 5: Rather then adding a new dependency, can `@pytest.mark.parametrize` be used to do the same thing?

saroad2: It can, but it has been requested specifically to avoid using `pytest` in tests directly in order to allow people to use the testing platform of their choosing.
I personally prefer using `pytest` at any time, but I respect the design decisions of this project.

saroad2:
@cooperlees, any feedback on this?

test_requirements.txt
cooperlees: My main pytest dislike is it's magical patching of thing so I guess the adding of a dep to allow both testing systems to be used works. Especially when this refactor is a big win imo.

cooperlees (approved):
All looks like a good refactor to me. ~250 less lines and all makes sense! Thanks!

PR #1788, opened 2020-10-27
Update readme.md with current version

bluefish6:
Update suggested pre-commit config with current version of black, so anyone who will copy-paste the config will already have the newest version without having to manually fix the version number.


hugovk:
Good idea, update `docs/version_control_integration.md` too?

bluefish6:
> Good idea, update `docs/version_control_integration.md` too?

Sure, I didn't know that this file exists ;)

PR #1794, opened 2020-10-28
Add `--single-quotes` setting

ThiefMaster:
Yes, I know this has almost no chance of getting merged, but before going through the hassle with a fork I thought I'll give it a try anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ Options:
 
   -S, --skip-string-normalization
                                   Don't normalize string quotes or prefixes.
+  --single-quotes                 Prefer single quotes when normalizing strings.
   --check                         Don't write the files back, just return the
                                   status.  Return code 0 means nothing would
                                   change.  Return code 1 means some files
--- a/docs/blackd.md
+++ b/docs/blackd.md
@@ -54,6 +54,8 @@ The headers controlling how source code is formatted are:
 - `X-Skip-String-Normalization`: corresponds to the `--skip-string-normalization`
   command line flag. If present and its value is not the empty string, no string
   normalization will be performed.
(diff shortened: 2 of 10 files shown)


hugovk:
It's been tried before!

https://github.com/psf/black/pull/633
https://github.com/psf/black/pull/1003
https://github.com/psf/black/pull/1569

ichard26:
One serious point of discussion is what happens when an user specifies `--skip-string-normalization` AND `--single-quotes`. Error out? Fall back to a default? What would be that default?

**But yeah, this will probably be rejected by the core devs.** Just trying to not seem like a total douche :-) Please don't actually do any work because of my comment.

ThiefMaster:
I'd say using both arguments is a case where nothing happens - you indicate that you prefer single quotes but don't want anything normalized, so nothing happens. Could error out, but no harm in allowing it.... especially since you may have a mix of CLI and pyproject.toml args,!

JelleZijlstra:
We're not doing this, so no need to discuss how it interacts with other options.

y4rr:
Is there any explanation as to why this option not being added?

Your main page says,
> Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. **You will save time and mental energy for more important matters.**

Actually, after using your tool, I don't have any time and mental energy left **at all**, because I use single quotes everywhere, as they are easier to type on my keyboard.

felix-hilden:
I use them too, for that very reason! The idea is to let Black make the changes regardless of what you type. It doesn't matter if you write singles or doubles, in the end they will be doubles. Does that make sense?

PR #1799, opened 2020-10-31
Fix bug where black tries to split string on escaped space

### LINKED ISSUE #1505: Black can break strings immediately after an escape
**Describe the bug**
Black does not produce identical code if it unluckily chooses to split a regex pattern right after an escape.

Given black installed via a particular git commit - accomplished via `python3 -m pip install git+git://github.com/psf/black.git@ea03b6c5e1f96aa2569f0080a61577b368b7af0d`
```
black                    19.10b1.dev118+gea03b6c
```

and a long regex string, as found in this example file foo.py:
```python
REGEX_PATTERNS = {
    "asdf_hjkl_jkl": re.compile(
        r"(?<!([0-9]\ ))(?<=(^|\ ))([A-Z]+(\ )?|[0-9](\ )|[a-z](\ )){4,7}([A-Z]|[0-9]|[a-z])($|\b)(?!(\ ?([0-9]\ )|(\.)))"
    ),
}
```

Black will error s follows:
```
python3 -m black --diff foo.py
error: cannot format foo.py: INTERNAL ERROR: Black produced invalid code: EOL while scanning string literal (<unknown>, line 3). Please report a bug on https://github.com/psf/black/issues.  This invalid output might be helpful: /var/folders/1z/64vyvtv136v6mfhfr16hws3c0000gn/T/blk_bbdmeyyx.log
Oh no! 💥 💔 💥
```

The contents of that log are as follows:

```shell
% cat /var/folders/1z/64vyvtv136v6mfhfr16hws3c0000gn/T/blk_bbdmeyyx.log
  File "/usr/local/lib/python3.7/site-packages/black/__init__.py", line 6042, in assert_equivalent
    dst_ast = parse_ast(dst)
  File "/usr/local/lib/python3.7/site-packages/black/__init__.py", line 5953, in parse_ast
    return ast27.parse(src)
  File "/Users/matt/Library/Python/3.7/lib/python/site-packages/typed_ast/ast27.py", line 50, in parse
REGEX_PATTERNS = {
    "asdf_hjkl_jkl": re.compile(
        r"(?<!([0-9]\ ))(?<=(^|\ ))([A-Z]+(\ )?|[0-9](\ )|[a-z](\"
        r" )){4,7}([A-Z]|[0-9]|[a-z])($|\b)(?!(\ ?([0-9]\ )|(\.)))"
    ),
}
```

**Expected behavior**

Don't break strings immediately after an escape. This seems like it requires inspection of the string contents to avoid this.

**Environment:**

- Version: master
- OS and Python version: Mac OS X / python 3.7.7

Yes. In fact, using https://black.now.sh/?version=master it ignores the error and produces invalid python code, which is a UI issue with that site.

bbugyi200:
Closes #1505.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -3590,7 +3590,8 @@ class StringSplitter(CustomSplitMapMixin, BaseStringSplitter):
         MIN_SUBSTR_SIZE characters.
 
         The string will ONLY be split on spaces (i.e. each new substring should
-        start with a space).
+        start with a space). Note that the string will NOT be split on a space
+        which is escaped with a backslash.
 
         If the string is an f-string, it will NOT be split in the middle of an
         f-expression (e.g. in f"FooBar: {foo() if x else bar()}", {foo() if x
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Thanks!

PR #1809, opened 2020-11-06
Fix bug which causes f-expressions to be split

### LINKED ISSUE #1807: Black produced invalid code: invalid syntax with --experimental-string-processing
I hope this helps to make the experimental string processing a bit less experimental :-)

## Environment
Python 3.8.6 
black, version 20.8b1
Separate venv (via pipx) where only black and its' dependencies are installed

## Executed Command
`black myfile.py --experimental-string-processing`

## myfile.py
```python
print(f"Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam. {[f'{i}' for i in range(10)]}")
```

## Error Mesage

```
black myfile.py --experimental-string-processing
error: cannot format myfile.py: INTERNAL ERROR: Black produced invalid code: invalid syntax (<unknown>, line 77). Please report a bug on https://github.com/psf/black/issues.  This invalid output might be helpful: /tmp/blk_tqh1ju35.log
Oh no! 💥 💔 💥
1 file failed to reformat.

Process finished with exit code 123
```

## Log file
```
  File "[..]/python3.8/site-packages/black/__init__.py", line 6141, in assert_equivalent
    dst_ast = parse_ast(dst)
  File "[..]/python3.8/site-packages/black/__init__.py", line 6052, in parse_ast
    return ast27.parse(src)
  File "[..]/python3.8/site-packages/typed_ast/ast27.py", line 50, in parse
print(
    f"Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam {[f'{i}'"
    " for i in range(10)]}"
)
```

bbugyi200:
This fix makes use of the [regex](https://pypi.org/project/regex/) library's ability to specify "recursive patterns" to correctly match f-expressions which contain other f-expressions.

Closes #1807.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -3611,13 +3611,14 @@ class StringSplitter(CustomSplitMapMixin, BaseStringSplitter):
     MIN_SUBSTR_SIZE = 6
     # Matches an "f-expression" (e.g. {var}) that might be found in an f-string.
     RE_FEXPR = r"""
-    (?<!\{)\{
+    (?<!\{) (?:\{\{)* \{ (?!\{)
         (?:
             [^\{\}]
             | \{\{
             | \}\}
+            | (?R)
         )+?
-    (?<!\})(?:\}\})*\}(?!\})
+    (?<!\}) \} (?:\}\})* (?!\})
     """
 
     def do_splitter_match(self, line: Line) -> TMatchResult:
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):
Thanks! Can't say I fully understand how that regex works.

bbugyi200:
> Thanks! Can't say I fully understand how that regex works.

That's probably a good indication that a regex should not be used. ¯\\_(ツ)_/¯

PR #1810, opened 2020-11-08
Correctly handle inline tabs in docstrings

### LINKED ISSUE #1601: Crash on docstring with tabs

1. Take this file (in text)
2. Run _Black_ on it without any arguments
3. See error

**Expected behavior** Work

```bash
error: cannot format /home/user/cilantropy/dist_worker.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_rifw8eif.log
Oh no! 💥 💔 💥
10 files left unchanged, 1 file failed to reformat.
```

- Version: 19.10b0
- OS and Python version: Linux (arch linux)/Python 3.8.3

pip install git+git://github.com/psf/black

**Additional context:**
Log:
```bash
--- src
+++ dst
@@ -62,11 +62,11 @@
       body=
         Expr(
           value=
             Constant(
               value=
-                'main worker for run subproccess and return result or error\n\n:param cmd: full command for run\t\n:return: result string and status error',  # str
+                'main worker for run subproccess and return result or error\n\n:param cmd: full command for run\n:return: result string and status error',  # str
             )  # /Constant
         )  # /Expr
         Try(
           body=
             Assign(
```
and my code:
```python
def main_worker(cmd):
	""" 
	main worker for run subproccess and return result or error

	:param cmd: full command for run	
	:return: result string and status error
	"""	

	try:
		cmd_response = subprocess.run(
			cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True
		)
	except subprocess.CalledProcessError as e:
		return "The pip command did not succeed: {stderr}".format(
				stderr=e.stderr.decode("utf-8")
			), True
		
	return cmd_response.stdout.decode("utf-8").replace(" "*6,"\n"), False
```

density:
The `fix_docstring` function expanded all tabs, which caused a
difference in the AST representation when those tabs were inline and not
leading. This changes the function to only expand leading tabs so inline
tabs are preserved.

Fixes #1601.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -6744,14 +6744,33 @@ def is_docstring(leaf: Leaf) -> bool:
     return False
 
 
+def lines_with_leading_tabs_expanded(s: str) -> List[str]:
+    """
+    Splits string into lines and expands only leading tabs (following the normal
+    Python rules)
+    """
+    lines = []
+    for line in s.splitlines():
+        # Find the index of the first non-whitespace character after a string of
+        # whitespace that includes at least one tab
+        match = re.match(r"\s*\t+\s*(\S)", line)
+        if match:
+            first_non_whitespace_idx = match.start(1)
+
+            lines.append(
+                line[:first_non_whitespace_idx].expandtabs()
+                + line[first_non_whitespace_idx:]
+            )
+        else:
+            lines.append(line)
+    return lines
+
+
 def fix_docstring(docstring: str, prefix: str) -> str:
     # https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation
     if not docstring:
         return ""
-    # Convert leading tabs to spaces (following the normal Python rules)
-    # and split into a list of lines:
-    docstring_expanded = re.sub(r"^\t+", "\t".expandtabs(), docstring, flags=re.M)
-    lines = docstring_expanded.splitlines()
+    lines = lines_with_leading_tabs_expanded(docstring)
     # Determine minimum indentation (first line doesn't count):
     indent = sys.maxsize
     for line in lines[1:]:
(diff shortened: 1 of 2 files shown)


src/black/__init__.py
```diff
@@ -6748,9 +6748,10 @@ def fix_docstring(docstring: str, prefix: str) -> str:
     # https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation
     if not docstring:
         return ""
-    # Convert tabs to spaces (following the normal Python rules)
+    # Convert leading tabs to spaces (following the normal Python rules)
     # and split into a list of lines:
-    lines = docstring.expandtabs().splitlines()
+    docstring_expanded = re.sub(r"^\t+", "\t".expandtabs(), docstring, flags=re.M)
```
JelleZijlstra, line 6753: If there are multiple tabs at the beginning, shouldn't they be replaced by multiple instances of `"\t".expandtabs()`?

density: Oh, duh. Will fix.

src/black/__init__.py
```diff
@@ -6744,13 +6744,33 @@ def is_docstring(leaf: Leaf) -> bool:
...
+def lines_with_leading_tabs_expanded(s: str) -> List[str]:
+    """
+    Splits string into lines and expands only leading tabs (following the normal
+    Python rules)
+    """
+    lines = []
+    for line in s.splitlines():
+        # Find the index of the first non-whitespace character after a string of
+        # whitespace that includes at least one tab
+        match = re.match(r"\s*\t+\s*(\S)", line)
```
JelleZijlstra, line 6756: What if there's something like space + tab + space + tab? It's probably easiest to just grab all leading whitespace and expandtabs() it.

density: Yup this covers that case and I added tests that cover it. It effectively does what you described, but only for lines where there is at least one tab. Removing the requirement that there be at least one tab isn't a big deal since the string is likely to be short anyway, so I'm open to doing it that way if you prefer!

JelleZijlstra: Oh you're right, I forgot that `\s` also matches tabs.

JelleZijlstra (approved):

PR #1814, opened 2020-11-11
Added support for Jupyter notebooks

KOLANICH:


MarcoGorelli:
Hi @KOLANICH 

I think it's a bit more complicated than this - for a start, you'd need to take care of magics, e.g.
```python
if condition:
    %time function()
```
and trailing semi-colons, e.g.
```python
fig, ax = plt.subplots()
ax.plot();
```

For what it's worth, I'd advise using the [jupyterlab autoformatter](jupyterlab_code_formatter) or (sorry for the self-promotion) [nbqa](https://github.com/nbQA-dev/nbQA)

KOLANICH:
@MarcoGorelli, thanks, your tool is cool (and maybe magics handling should be modularized out of ipython instead of reinwenting wheel), but ... it uses `subprocess`.

About magics here ... it will likely greatly complicate the code, unless an external lib is used. So, IMHO we should use `less is better` approach here, and those who need to process notebooks with magics can use your tool.

About `;` ... what's wrong with it? Isn't it a valid python syntax?

MarcoGorelli:
> @MarcoGorelli, thanks, your tool is cool (and maybe magics handling should be modularized out of ipython instead of reinwenting wheel), but ... it uses `subprocess`.

Yup - we use `subprocess.run` and are very intentional about calling with `shell=False`. Anyway, more to the point:

> About `;` ... what's wrong with it? Isn't it a valid python syntax?

`black` removes it though, while in Jupyter notebooks it's often desirable to keep them to suppress output

src/black/__init__.py
```diff
@@ -270,14 +280,21 @@ def get_cache_key(self) -> str:
...
 
 
 # Legacy name, left for integrations.
 FileMode = Mode
 
+def lazy_import_json() -> None:
+    global json
+    if json is None:
+        try:
+            import ujson as json
```
cooperlees, line 295: Maybe add `ujson` as an extra install then and document somewhere that it can be used to speed up ipython usage.

src/black/__init__.py
```diff
@@ -383,9 +400,22 @@ def target_version_option_callback(
...
+        " when piping source on standard input). deprecated, use --source-type"
     ),
 )
+# @click.option(
+#    "--source-type",
+#    click.Choice([v.name.lower() for v in SourceType]),
+#    help=(
+#        "allows you to override source type"
+#    ),
+# )
```
cooperlees, line 412: Delete?

cooperlees (changes requested):
This is a great additon, however at the moment break way to many tests to be seriously reviewed. Please try and not cause any regressions and then we'll consider merging this.

src/black/__init__.py
KOLANICH: Added an extra `notebook_speedup`

src/black/__init__.py
KOLANICH: Fixed

KOLANICH:
>Please try and not cause any regressions and then we'll consider merging this.

Yes, it seems it really does.

ichard26:
Hi @KOLANICH I'm curious to whether you still intend to finish this. It's totally OK if you don't, I just want to know what needs our attention and what doesn't.

ichard26:
There has been 10 days of no activity on this PR and especially since it's no state to be merged soon (all of the CI failed, conflicting files, and other things). Since I am not willing to push this PR forward myself (due to lack of experience with / need for Jupyter notebooks) I will be closing this PR as stale.

If someone would like to push this PR to completion, I would be happy to reopen. Thanks @KOLANICH for the initial work, sorry for being unable to get it in.

PR #1819, opened 2020-11-13
Fixed linters warnings

KOLANICH:


cooperlees (changes requested):
Hi, Thanks and I love to have our code pass lints, but why? Can you please provide more context as to why you've done each thing there? 

You have also done more than fix things. There are:
- converting f strings back to format based strings - Why? I wanted to use f-string ...
- Some elif's change to if statements
- Shouldn't `[pycodestyle]` be `[flake8]`

More, but this was just from my quick look.

And now we have CI failing. Is there an issue I should be looking at?

KOLANICH:
>Some elif's change to if statements

Pylint says they are redundant, and they really are. Eliminating them should simplify CFG.

```
In [1]: def a(): 
   ...:     if c(): 
   ...:         return 1 
   ...:     else: 
   ...:         return 2 
   ...:                                                                                                            

In [2]: import dis                                                                                                 
In [3]: dis.dis(a)                                                                                                 
  2           0 LOAD_GLOBAL              0 (c)
              2 CALL_FUNCTION            0
              4 POP_JUMP_IF_FALSE       10

  3           6 LOAD_CONST               1 (1)
              8 RETURN_VALUE

  5     >>   10 LOAD_CONST               2 (2)
             12 RETURN_VALUE
             14 LOAD_CONST               0 (None)
             16 RETURN_VALUE

In [4]: def b(): 
   ...:     if c(): 
   ...:         return 1 
   ...:     return 2 
   ...:                                                                                                            

In [5]: dis.dis(b)                                                                                                 
  2           0 LOAD_GLOBAL              0 (c)
              2 CALL_FUNCTION            0
              4 POP_JUMP_IF_FALSE       10

  3           6 LOAD_CONST               1 (1)
              8 RETURN_VALUE

  4     >>   10 LOAD_CONST               2 (2)
             12 RETURN_VALUE
```

>converting f strings back to format based strings - Why? I wanted to use f-string ...

Yes another inefficiency detected by pylint. F-strings are always constructed, the new variant is constructed only if the log message matches the level.

>Shouldn't [pycodestyle] be [flake8]

Likely no. pycodestyle cli tool doesn't recognize flake configs.

>And now we have CI failing. Is there an issue I should be looking at?

We have already had it failing. That's why I have intervened.

JelleZijlstra:
I haven't looked at all of this diff but I'm not sure much of it is an improvement. We don't need to obey every overly pedantic pylint rule.

KOLANICH:
>We don't need to obey every overly pedantic pylint rule.

We don't, I have excluded the ones that make no sense (there are some), the ones that contradict the practices used by this project and the ones that require too much redesign (it makes sense to fix them, but it requires larger changes, probably that should be done not by me).

JelleZijlstra:
I don't think we need this change, and there are numerous conflicts.

PR #1824, opened 2020-11-15
Add --skip-magic-trailing-comma

hauntsaninja:
Lukasz indicated he'd accept such a PR in https://github.com/psf/black/issues/1288#issuecomment-593873891

This is a pretty straightforward combination of things from before and after #1612 and #826

Feedback I'm looking for:
1) What the best way to get the MAGIC_TRAILING_COMMA option to the places it's needed is.
Obviously the global variable proof of concept is unmergeable, but I'm not sure what the right solution is. Passing it around seems kind of untenable because `Leaf.append` is used in a million places / `should_split_body_explode` has a daunting reverse call graph too (You could make store an attribute on `Leaf` to make things a little easier, but that's pretty ugly as well). Maybe a context manager to set a global current `Mode`?
2) How best to structure tests for this 
Black doesn't have many options, so there isn't a clear pattern for how to test these. I went with reusing expression.py, but I could do something more akin to `numeric_literals_skip_underscores`. Let me know!

### DIFF
--- a/docs/the_black_code_style.md
+++ b/docs/the_black_code_style.md
@@ -438,6 +438,9 @@ into one item per line.
 How do you make it stop? Just delete that trailing comma and _Black_ will collapse your
 collection into one line if it fits.
 
+If you must, you can recover the behaviour of early versions of Black with the option
+`--skip-magic-trailing-comma` / `-C`.
+
 ### r"strings" and R"strings"
 
 _Black_ normalizes string quotes as well as string prefixes, making them lowercase. One
--- a/README.md
+++ b/README.md
@@ -97,6 +97,8 @@ Options:
 
   -S, --skip-string-normalization
                                   Don't normalize string quotes or prefixes.
+  -C, --skip-magic-trailing-comma
+                                  Don't treat trailing commas magically.
   --check                         Don't write the files back, just return the
                                   status.  Return code 0 means nothing would
                                   change.  Return code 1 means some files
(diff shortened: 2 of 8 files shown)


ambv:
You're right that using global variables won't be acceptable for this. Hiding the global variable by running a context manager is not fundamentally changing this either.

Having an attribute on Leaf is what I would look at.

hauntsaninja:
Thanks! You'll notice from the diff that I meant to say `Line`, not `Leaf` — hope it's still acceptable :-) I also updated blackd and the relevant docs that I could find.

graingert:
@hauntsaninja thanks for following up on this, I'd completely missed the concept ack on the other issue

hauntsaninja:
@ambv if you have time to take a look, much obliged!

hauntsaninja:
Just bumping this!

docs/the_black_code_style.md
```diff
@@ -438,6 +438,9 @@ into one item per line.
 How do you make it stop? Just delete that trailing comma and _Black_ will collapse your
 collection into one line if it fits.
 
+If you must, you can recover the behaviour of early versions of Black with the option
```
JelleZijlstra, line 441: we might need to be consistent here at some point:
```
(py38) mbpt-jelle:black jelle$ git grep behavior | wc -l
       7
(py38) mbpt-jelle:black jelle$ git grep behaviour | wc -l
       7
```

JelleZijlstra (approved):

uriva:
any idea when will this be released?

ichard26:
@uriva unfortunately I can't give you a timeframe, but I do hope it will happen soon! Apologies for the lack of any clear messaging.

uriva:
@ichard26 what is this blocked on? or is releasing itself a difficult process? (not demanding anything, just curious)

ichard26:
@uriva I honestly don't know 100% why a new release is blocked, but the general lack of time and specifically the lack of availability for the ones holding the power to create a release is probably why. Maybe @cooperlees can explain in more detail.

uriva:
who has this power?

graingert:
@uriva you can see the list here https://pypi.org/project/black/#data

indigoviolet:
tangential comment: I found this option `--skip-trailing-magic-comma` in the github documentation, but then didn't find it in the most recent release and it was mildly confusing. It might be useful to either have per-feature metadata like `[Added in 20.0b2dev]` etc or a `develop` branch.

ichard26:
@indigoviolet you're not alone! I find it confusing too. I am planning to just move the documentation to ReadTheDocs where the stable version will be provided by default to avoid this mess in the first place (it's part of my effort to revamp the docs). Thanks for the feedback though!

hauntsaninja:
Opened https://github.com/psf/black/pull/2064 to add this to CHANGES.md

Separately, is there anything blocking a release that I could help with?

PR #1834, opened 2020-11-21
Add formatters-python for atom to editor_integration

utkarshgupta137:


utkarshgupta137:
@ambv

ichard26:
Please be wary of @mentioning people like that. A @mention feels like a personal request for our attention, and while they're usually fine, I wouldn't be happy being @mentioned without any context. My stance is that if you're asking for my attention directly, explaining why my attention is needed is the bare minimum. Yes, I can tell what the @mention is serving here without any extra words, but it stills feels... off to me honestly.

Also, in the case of asking for attention on a PR, a simple ping comment with a short message of like "hey, it's been a while and there has been no activity, I would appreciate a review!" is enough. Letting the people watching the repo or the PR (i.e. the ones who will probably be either interested or has some available time) be notified is IMO way better than guessing whether the person (maybe maintainer) you're @mentioning is actually the best choice (availability or expertise wise).

I know you are probably frustrated with the lack of attention your PR has received, lack of time among maintainers and reviewers is a serious problem around here, but I simply don't want fellow core team members to feel their attention and time has been disrespected (especially since they're helping as volunteers).

cooperlees (approved):
Sure, makes sense to me. Sorry I and others missed it.

Issue #921, opened 2019-07-09
Black produced invalid code: `assert (*x)`


dizzy57:
Operating system: macOS Mojave 10.14.5 (18F132)
Python version: Python 3.7.3
*Black* version: 19.3b0 (f3bb22a828)
Does also happen on master: yes

Minified reproducer:
```python
assert (*x)
```

Stack trace:
```
  File "black.py", line 3547, in assert_equivalent
    dst_ast = parse_ast(dst)
  File "black.py", line 3490, in parse_ast
    return ast27.parse(src)
  File "/Users/apreobrazhen/miniconda3/lib/python3.7/site-packages/typed_ast/ast27.py", line 51, in parse
    return _ast27.parse(source, filename, mode)
assert *x
```

This is invalid python code that gets accepted by `blib2to3` grammar. I think *black* should throw `InvalidInput` exception in this case, not `AssertionError`.

Maybe it's worth raising an upstream bug for `lib2to3`?

zsol:
I agree lib2to3 shouldn't accept this. There's a certain class of these python programs that are technically invalid python to start with, so Black is at a severe disadvantage of dealing with. I don't think this is super important to fix in Black as it will usually generate invalid code (like above) and then fail promptly.

I can't come up with an example that's all of:
* invalid python
* Black accepts it
* Black generates valid python at the end

This would be bad as Black would hide the fact that the original code was not valid Python and thus all bets are off, but as a user of Black it's very hard to tell there was an error in the first place.

dizzy57:
I think it still would be nice to show a "Cannot parse" message instead of "INTERNAL ERROR, please report a bug".

isidentical:
I agree with @zsol, there are numerous of examples where black can parse but CPython can't parse (or rather don't accept as input). The current error message seems to be better though: 

```
 $ black -c "assert (*x)"                                                                                                                              19ms
assert (*x)
error: cannot format <string>: cannot use --safe with this file; failed to parse source file: can't use starred expression here (<unknown>, line 1)
```

JelleZijlstra:
Thanks, closing as the current behavior seems fine.

PR #1835, opened 2020-11-21
Allow same RHS expressions in annotated assignments as in regular assignments

### LINKED ISSUE #1189: Black crashes on annotated tuple assignment
Black crashes on tuple assignment without parentheses, when using type annotations.

**To Reproduce**

With no config or flags, this crashes:

```python
from typing import Tuple

p: Tuple[int, int] = 0, 0
```

```
error: cannot format black-fail2.py: Cannot parse: 3:22: p: Tuple[int, int] = 0, 0
Oh no! 💥 💔 💥
1 file failed to reformat.
```

This doesn't crash:

```python
from typing import Tuple

p: Tuple[int, int]  = (0, 0)
```

And neither does this:

```python
p = 0, 0
```

**Environment:**

- Version: 19.10b0
- OS and Python version: Macos 10.15/Python 3.8.0

**Does this bug also happen on master?**: Yes, tested on black.now.sh

### LINKED ISSUE #1828: Can't format a yield expression with a type annotation and without parentheses
**Describe the bug** 
Can't format a [yield expression](https://www.python.org/dev/peps/pep-0342/#new-syntax-yield-expressions) with a type annotation and without parentheses.

**To Reproduce** 
Paste the following snippet in the [black playground](https://black.now.sh/?version=master&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ACvAHVdAD2IimZxl1N_WlbvK5V_5E5NEzEakNf_hFekK8lLVE09BitvHbpZS6OelPZPMcpkchxrJHL7oA6RntadGfB4FSrVHt9pQ6VdwYXLGeUICNj46Ar-Sy7RERu36z958Pu01O2nxAl4SUukN7ZjjEHLtbpf-ISflwAAAACh7HWJ50i_bAABkQGwAQAAJhy_NLHEZ_sCAAAAAARZWg==):
```python
def talkin_bout_my_generator():
    next_yield: str = yield "first_yield"
    yield next_yield
```

**Expected behavior**
Black is able to validate that the snippet is formatted.

**Actual behavior**
`Cannot parse: 2:22:     next_yield: str = yield "first_yield"`

- Version: master
- OS and Python version: [e.g. MacOS/Python 3.8.0]

**Additional context**
This seems to be a very specific corner case. Black is able to process both:
```python
next_yield = yield "first_yield"
```
and 
```python
next_yield: str = (yield "first_yield")
```
It's the combination of the type annotation and lack of parentheses that seems to be the issue. Probably a low priority thing, but since it seems like `next_yield: str = yield "first_yield"` is valid python, black should be able to process it.

ichard26:
Fixes #1828. 
Fixes #1189. 

@zsol I would appreciate if you could review as this is my first time swimming in these waters, and therefore I have no idea what I am doing :D. Thanks in advance!

### DIFF
--- a/tests/data/python38.py
+++ b/tests/data/python38.py
@@ -11,6 +11,14 @@ def starred_yield():
     yield "value1", *my_list
 
 
+# all right hand side expressions allowed in regular assignments are now also allowed in
+# annotated assignments
+a : Tuple[ str, int] = "1", 2
+a: Tuple[int , ... ] = b,  *c, d
+def t():
+    a : str =  yield "a"
+
+
 # output
 
 
@@ -25,3 +33,13 @@ def starred_return():
 def starred_yield():
     my_list = ["value2", "value3"]
     yield "value1", *my_list
+
+
+# all right hand side expressions allowed in regular assignments are now also allowed in
+# annotated assignments
+a: Tuple[str, int] = "1", 2
+a: Tuple[int, ...] = b, *c, d
+
+
+def t():
+    a: str = yield "a"
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,9 @@
 
 - Added support for PEP 614 relaxed decorator syntax on python 3.9 (#1711)
 
+- Added parsing support for unparenthesized tuples and yield expressions in annotated
+  assignments (#1835)
+
 - use lowercase hex strings (#1692)
 
 #### _Packaging_
--- a/src/blib2to3/Grammar.txt
+++ b/src/blib2to3/Grammar.txt
@@ -77,7 +77,7 @@ small_stmt: (expr_stmt | print_stmt  | del_stmt | pass_stmt | flow_stmt |
              import_stmt | global_stmt | exec_stmt | assert_stmt)
 expr_stmt: testlist_star_expr (annassign | augassign (yield_expr|testlist) |
                      ('=' (yield_expr|testlist_star_expr))*)
-annassign: ':' test ['=' test]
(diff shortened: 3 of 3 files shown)


tests/data/python38.py
```diff
@@ -11,6 +11,14 @@ def starred_yield():
     yield "value1", *my_list
 
 
+# all right hand side expressions allowed in regular assignments are now also allowed in
+# annotated assignments
+a : Tuple[ str, int] = "1", 2
+a: List [ str] = *x
```
zsol, line 17: Looks like technically this is a syntax error, but it's allowed in the cpython grammar. Black should probably not be used to validate syntax though, so it's probably fine :)

zsol (approved):
LGTM

hauntsaninja (approved):
This looks right to me, it's the same change CPython made in https://bugs.python.org/issue35814 / matches what's in the last version of the non PEG parser.

tests/data/python38.py
ichard26: Yeah, it's probably fine. As long we don't add parenthesizes to any of the above, the case of the following won't happen :)

> * invalid python
> * Black accepts it
> * Black generates valid python at the end

https://github.com/psf/black/issues/921#issuecomment-509653493

ichard26:
Thanks @hauntsaninja for reviewing; yep I intended this to be a backport of https://github.com/python/cpython/commit/8565f6b6db0fa9f65449b532a5056a98bad3dc37 and https://github.com/python/cpython/commit/62c35a8a8ff5854ed470b1c16a7a14f3bb80368c.

PR #1838, opened 2020-11-24
[src/black/__init__.py] Drop `typed_ast` on Python 3.10+

SamuelMarks:
Tracking issue for Python 3.10 support in `typed_ast`: https://github.com/python/typed_ast/issues/150

In the meantime, this makes `black` somewhat usable in 3.10

(maybe keep open and not merge? - Your call…)

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -49,7 +49,11 @@ from appdirs import user_cache_dir
 from dataclasses import dataclass, field, replace
 import click
 import toml
-from typed_ast import ast3, ast27
+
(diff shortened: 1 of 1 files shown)


cooperlees:
Cheers for alerting us this exists. Since 3.10 is not released, I don't think we will merge this as I think it will have side effects on formatting.

ambv:
Thanks for writing this, we went with an equivalent PR instead.

SamuelMarks:
@ambv You wouldn't need an equivalent PR, click that tracking issue, they've fixed `typed_ast` support

PR #1848, opened 2020-12-01
Add a GitHub Action to build + Upload black to PyPI

cooperlees:
When a new GitHub Release is made this will cause this action to run and do the following:
- Build a wheel + sdist
- Upload via twine using token stored in GitHub secrets

This will need someone to generate a PyPI token + upload to the psf/black secrets on GitHub. I can not do this.


zsol:
LGTM, but I'll defer to @ambv as I don't actually know what the release process is currently.

ichard26:
in a way off-topic but

> [...] I don't actually know what the release process is currently.

Perhaps it's a good idea to publicly write down what the release process so we (well not me, I shouldn't be trusted with this yet lol) can all understand and maintain the release workflow (and maybe do an emergency manual one when needed). Although the use of @ blackrelease may complicate this...

I suppose making sure everyone with write access knows what the release process is through private means is also fine.

ichard26:
I'm surprised <a href=https://github.com/psf/black/pull/1966>GH-1966's</a> changes aren't conflicting the prettier changes here, though @cooperlees you might wanna get rid of them.

PR #1849, opened 2020-12-01
Mypy update to 0.790

cooperlees:
- Update via pre-commit
- Ignore dataclass typing error unless someone smart knows how to type this
- Fix primer to ensure it always raises CalledProcessError with an int error code


cooperlees:
O, don't think I drove git correctly. I will rebase one we merge #1848

src/black/__init__.py
```diff
@@ -263,7 +263,10 @@ def get_cache_key(self) -> str:
         if self.target_versions:
             version_str = ",".join(
                 str(version.value)
-                for version in sorted(self.target_versions, key=lambda v: v.value)
+                # Returning Any from function declared to return "Set[TargetVersion]"
+                for version in sorted(
+                    self.target_versions, key=lambda v: v.value  # type: ignore
+                )
```
cooperlees, line 269: This is the main change I need reviewed here in case someone knows how to type this - I think field() returns Any?

.github/workflows/pypi_upload.yml
```diff
@@ -0,0 +1,31 @@
+name: pypi_upload
```
JelleZijlstra, line 1: doesn't seem like this belongs in this PR?

src/black/__init__.py
JelleZijlstra: That's pretty weird, not sure what's going on here. Is it complaining about the lambda?

cooperlees: I changed it to a real function and was same error too. It’s saying v.value is Any ...

.github/workflows/pypi_upload.yml
cooperlees: Yeah I’ll rebase before I merge ...

src/black/__init__.py
hauntsaninja: I think mypy just doesn't know the type of `v.value`. In 0.782, the return type of sorted's key was Any (https://github.com/python/typeshed/pull/4192/files). I'm assuming mypy's `--warn-return-any` doesn't complain if you return Any from a function that's meant to declare Any, so this only started getting reported when sorted's key gained a non-Any return type.

fbergroth: See also https://github.com/python/mypy/issues/8722

cooperlees:
Will have another go at this with a fresh PR one day unless someone beats me.

Issue #1473, opened 2020-06-01
Configuration example "exclude" doesn't work as documented


orn688:
**Describe the bug** A clear and concise description of what the bug is.

The example `pyproject.toml` shown in the [Configuration format](https://github.com/psf/black/blob/master/README.md#configuration-format) section of the README has the following `exclude`:

```toml
[tool.black]
line-length = 88
target-version = ['py37']
include = '\.pyi?$'
exclude = '''

(
  /(
      \.eggs         # exclude a few common directories in the
    | \.git          # root of the project
    | \.hg
    | \.mypy_cache
    | \.tox
    | \.venv
    | _build
    | buck-out
    | build
    | dist
  )/
  | foo.py           # also separately exclude a file named foo.py in
                     # the root of the project
)
'''
```

However, the comment next to foo.py seems to be incorrect. That `foo.py` will actually exclude any foo.py found in the project, not just in the root.

**To Reproduce** Steps to reproduce the behavior:

1. `mkdir temp && cd temp`
2. Create `pyproject.toml` containing the example configuration from above
3. `touch foo.py`
4. `mkdir bar`
5. `touch bar/foo.py`
6. `black --verbose .` prints the following:
    ```
    Using configuration from /path/to/temp/pyproject.toml.
    foo.py ignored: matches the --exclude regular expression
    bar/foo.py ignored: matches the --exclude regular expression
    No Python files are present to be formatted. Nothing to do 😴
    ```

**Expected behavior** A clear and concise description of what you expected to happen.

I would expect Black to not exclude bar/foo.py. It would be nice if there were a way to actually accomplish what the example says (i.e., only exclude foo.py in the project root). Short of that, I think the documentation could use rewording to avoid suggesting that the example pyproject.toml only excludes foo.py in the project root.

**Environment (please complete the following information):**

- Version: master
- OS and Python version: Linux/Python 3.7.7

**Does this bug also happen on master?** yes

sittingbool:
Same on MaxOSX and Python 3.8

orn688:
[...]

Issue #1759, opened 2020-10-09
Reorganize documentation


JelleZijlstra:
I feel like the current documentation isn't very well organized. It has a lot of pages and it's not obvious where to document things like `.gitignore` support (#1734). Here's the sections we have right now:

![Captura de Pantalla 2020-10-09 a la(s) 2 25 53 p  m](https://user-images.githubusercontent.com/906600/95632854-6348dd80-0a3b-11eb-9c32-7e914d7d28b0.png)

The line wrapping makes it hard to see individual sections, and a lot of them are about fairly minor issues.

Some ideas:
- "Editor integrations", "Version control integration" and "GitHub Actions" should all go into an "Integrations" section
- black-primer isn't really relevant to users (as opposed to developers), it can just be in "Contributing to Black"
- `pyproject.toml` should be renamed to "Configuration" and also cover other ways of configuring Black, like gitignore files and command-line options
- "Ignoring unmodified files" is all about an implementation detail. Maybe it should just go into the "Contributing" section.

Please comment if you have other ideas (or if you think I'm all wrong here and the current organization is fine).

J-Exodus:
I tend to agree @JelleZijlstra. At a minimum, each section title should make it obvious to the reader what content is contained within that section. 
Perhaps a better layout could fall under a more simple structure.  Something such as:
- Overview/Introduction
- Installation
- Configuration
- Integration
- Contribution/Development
- Credits

Transferring to a structure like this should be relatively simple. A good starting point would be allocating all the current sections to one of the new section titles, and then collating those individual sections.

Just my ideas from what I've seen/read.  Thanks.

ichard26:
Okay, I'm picking this up after what seems like *forever*. Apologies for sorta causing this mess in the first place, I didn't really think through what I was doing last time I reorganized the docs. Now it's time for me to fix my mistakes and do it right.

So, I like @J-Exodus's basic structure and I'll be working on transferring our docs to it. I'll post a draft PR when ready.

ichard26:
This is the hierarchy I've been thinking about (**WARNING: IT WILL CHANGE AS I EDIT**) :

```
Overview
Getting Started
  - Installation
  - Usage
  - Using Black with other tools
  - Migrating your project to Black
Configuration
  - Controlling style
  - Controlling file discovery
  - Controlling writeback and reporting
  - Misc.
Integrations
  - GitHub Actions
  - Source Version Control
  - Editors
Development
  - Getting Started
  - Creating Issues
[...]

PR #1861, opened 2020-12-07
Update example exclude to match only files in root

orn688:
The `exclude` section of the example `pyproject.toml` file didn't work as expected. It claimed to exclude matched files only in the project root, but it actually excluded matched files at any directory level within the project. We can address this by prepending `^/` to the regex to ensure that it only matches files in the project root.

See https://github.com/psf/black/issues/1473#issuecomment-740008873 for explanation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,9 +308,10 @@ line-length = 88
 target-version = ['py37']
 include = '\.pyi?$'
 exclude = '''
-
-(
-  /(
+# A regex preceded with ^/ will apply only to files in the root of
+# the project.
+^/(
+  (
       \.eggs         # exclude a few common directories in the
     | \.git          # root of the project
     | \.hg
--- a/README.md
+++ b/README.md
@@ -308,8 +308,8 @@ line-length = 88
 target-version = ['py37']
 include = '\.pyi?$'
 exclude = '''
-# A regex preceded with ^/ will apply only to files in the root of
-# the project.
+# A regex preceded with ^/ will apply only to files and directories
+# in the root of the project.
 ^/(
   (
       \.eggs         # exclude a few common directories in the


README.md
```diff
@@ -308,9 +308,10 @@ line-length = 88
 target-version = ['py37']
 include = '\.pyi?$'
 exclude = '''
-
-(
-  /(
+# A regex preceded with ^/ will apply only to files in the root of
```
ichard26, line 311:
```suggestion
# A regex preceded with ^/ will apply only to files and directories in the root of
```

Given that the regex this comment is preceding is only excluding directories, it would be a good idea to add that usage in the comment.

ichard26 (changes requested):
Looks mostly good, thanks for submitting the PR!

Just a comment error and this should be good to go!

README.md
orn688: Good point, done!

orn688:
(sorry for the noise, I'm not very familiar with the GitHub UI and seems like I messed things up by force-pushing)

ichard26 (approved):
Thanks!

edit: don’t worry about having messed up things, we all make mistakes, it’s all good :)

CircleOnCircles (approved):

README.md
```diff
@@ -308,9 +308,10 @@ line-length = 88
 target-version = ['py37']
 include = '\.pyi?$'
 exclude = '''
-
-(
-  /(
+# A regex preceded with ^/ will apply only to files and directories
+# in the root of the project.
```
jayaddison, line 312: @orn688 It looks like your explanatory comment here might (accidentally?) be inside the multi-line TOML string that contains the regex.

If that's unintended, it might be worth moving the comment to the line prior to the `exclude` definition.

ichard26: Replying while busy and on a phone so apologies in advance, but include, exclude, and force-exclude regexes are compiled with the verbose flag (`re.compile("your-regex", re.VERBOSE)`) meaning comments are ignored.

jayaddison: Take your time, and when you're ready to review more fully: perhaps it would still be worthwhile moving the comment outside of the string, to reduce the escaping complexity?

I'll also note that the [`sphinx-build`](https://github.com/psf/black/blob/988c686d312986760a42cc7edd9a79a8f31c4769/.github/workflows/doc.yml#L35-L36) output differs as a result of placement inside/outside the string; so changes to this file may have other effects.

orn688: I intentionally put the comment inside the string because, as mentioned, comments are ignored in multilinear regex mode, and because there were other comments within the string already and I find it clearer to keep each comment right next to the bit of the string that it applies to.

jayaddison: Thanks for the explanation @orn688.

My previous concern about output differing based on the string/comment escaping hasn't turned out to be relevant here.  It's a long (and no doubt boring) story, but I was using a different markdown parser to perform some local verification checks, and discovered some peculiarities in that tool - but nothing that should affect the documentation build process here.

jayaddison: If it's OK with you both, I'll open a small PR soon to regenerate the `docs` directory and include the root-only-matching comment.

It'll also address a subtle [repeat occurrence of command-line parameters](https://github.com/psf/black/blame/aebd3c37b28bbc0183a58d13b80e7595db3c09bb/docs/installation_and_usage.md#L98-L106) that crept in when two pull requests individually updated the same documentation file (no merge conflicts occurred; rare but it can happen).

ichard26: > It'll also address a subtle repeat occurrence of command-line parameters that crept in when two pull requests individually updated the same documentation file (no merge conflicts occurred; rare but it can happen).

oooo, first time I've seen something like that! Something to watch out for! I should probably add that to the PR review docs this project will hopefully gain soon (see https://github.com/psf/black/issues/1759#issuecomment-772006088). Thanks!

> If it's OK with you both, I'll open a small PR soon to regenerate the docs directory and include the root-only-matching comment. 

Yeah no issues with me. On my docs overhaul effort, I plan to get rid of the regen step since it's frustrating and annoying without much value given in return, but a regen PR can't hurt. (I guess you _could_ show me more ways to subtlety break a merge)

jayaddison: Cool, thanks @ichard26 :)  Yep, I saw a bit of the discussion around removing the regen, sounds like a good project.

The way that repeat parameter merge happened was.. kinda confusing to me, to be honest.  My theory is that it was some kind of strange loophole in the way that `git` calculates diffs, but it's beyond my brainpower to figure out at the moment.  Good to be aware of though for sure.

PR #1867, opened 2020-12-14
Remove all trace of flake8-mypy

hauntsaninja:
flake8-mypy is long dead and shouldn't be used, see
https://github.com/ambv/flake8-mypy. We appear to use pre-commit to run
mypy now anyway.

I ran `pipenv uninstall flake8-mypy`, which seems to have made several
changes to Pipfile.lock. Let me know if there's a better way to do this.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -9,7 +9,6 @@ coverage = "*"
 docutils = "==0.15"  # not a direct dependency, see https://github.com/pypa/pipenv/issues/3865
 flake8 = "*"
 flake8-bugbear = "*"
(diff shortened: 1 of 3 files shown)


ichard26:
Well I'm pretty sure that you just caused pipenv to update every dependency so yeahhh, hopefully everything is fine. I guess having CI pass is good enough. There isn't a better way of doing this as far as I know, modifying a single dependency is a manual and annoying process.

Also this reminds me that I should setup automatic dependency upgrades.

JelleZijlstra (approved):

Issue #1870, opened 2020-12-15
Vim plugin doesn't find pyproject.toml if it's in a subdirectory


austinglaser:
Given the directory organization:

```
<repository_root>
└── subproject
   ├── pyproject.toml
   └── subproject
      └── foo.py
```

if I run `vim subproject/subproject/foo.py` and then `:Black`, black will not find `subproject/pyproject.toml`, since it searches relative to the current working directory instead of the path to the current buffer.

I think there are a number of other cases like this (for instance, editing two files from different projects in two different buffers).

ichard26:
Fixed by GH-1871 and thank you for the bug report!

PR #1871, opened 2020-12-15
Find pyproject from vim relative to current file

austinglaser:
Candidate fix for #1870. May not be the best way to go about this.


JelleZijlstra:
This has a merge conflict now.

Also, unfortunately most of us maintainers don't know much about vimscript and we'll need some help to review this. Can you explain what `@%` means?

godlygeek:
Rebasing this should be trivial; the function that's being patched has just moved from `plugin/black.vim` to `autoload/black.vim`, but the relevant code is the same.

`@x` as an expression in vimscript evaluates to the contents of the `x` register. The `%` register stores the name of the current file, so `@%` is the name of the current file (which may be an empty string, if the current buffer is not associated with a file).

But... OK, both the existing code and the proposed code are broken. The `vim.eval()` call (whether it's `vim.eval("@%")` or `vim.eval("fnamemodify(getcwd(), ':t')`) returns a string, and it passes that string to `find_pyproject_toml`, which expects a sequence of strings, not a single string, and - since a string _is_ a sequence of single character strings - it gets turned into a list of ridiculous paths. I tested with a file called `foo.py`, and added a `print(path_srcs)` into `find_project_root`, which printed out:
```py
[PosixPath('/home/matt/f'), PosixPath('/home/matt/o'), PosixPath('/home/matt/o'), PosixPath('/home/matt'), PosixPath('/home/matt/p'), PosixPath('/home/matt/y')]                                                                  
```

So... that's wrong.

This seems to do the trick better than both the existing code and the proposed code:
```diff
diff --git a/autoload/black.vim b/autoload/black.vim
index f0357b0..0d93aa8 100644
--- a/autoload/black.vim
+++ b/autoload/black.vim
@@ -139,7 +139,8 @@ def Black():
       print(f'Reformatted in {time.time() - start:.4f}s.')
 
 def get_configs():
-  path_pyproject_toml = black.find_pyproject_toml(vim.eval("fnamemodify(getcwd(), ':t')"))
+  filename = vim.eval("@%")
+  path_pyproject_toml = black.find_pyproject_toml((filename,))
   if path_pyproject_toml:
     toml_config = black.parse_pyproject_toml(path_pyproject_toml)
   else:
```

This does work for an unnamed buffer, too - we wind up calling `black.find_pyproject_toml(("",))`, and that winds up prepending the working directory to any relative paths, so `""` just gets turned into the current working directory.

Note that `find_pyproject_toml` needs to be passed a 1-tuple, not a list, because it requires something hashable (thanks to `functools.lru_cache` being used)

CHANGES.md
```diff
@@ -12,6 +12,11 @@
 - Fix incorrect custom breakpoint indices when string group contains fake f-strings
   (#2311)
 
+### Integrations
+
+- The vim plugin now searches the directory containing the current buffer instead of the
```
godlygeek, line 17: May want to say "searches upwards from"

ichard26: That's way better! Thanks!

godlygeek (approved):
LGTM

ichard26:
Given that no one on the core team (and who's active enough) is qualified to review vim PRs, I'm merging under the pretense godlygeek's approval is A-OK.

PR #1875, opened 2020-12-19
Add pyi file support to .pre-commit-hooks.yaml

### LINKED ISSUE #402: pre-commit hook does not work for .pyi files
Because of this issue https://github.com/chriskuehl/identify/issues/51

For now I can solve this by overriding `files` and `types` in `.pre-commit-config.yaml`:

```yaml
repos:
  - repo: https://github.com/ambv/black
    rev: 18.6b4
    hooks:
      - id: black
        language_version: python3.6
        files: \.pyi?$
        types: []
```

I would recommend that `black` use `files` instead of `types` until this issue is resolved in `identify` and `pre-commit`.

Operating system: osx
Python version: 3.6
*Black* version: 18.6b4
Does also happen on master: yep

jdufresne:
Since pre-commit 2.9.0 (2020-11-21), the types_or key can be used to
match multiple disparate file types. For more upstream details, see:
https://github.com/pre-commit/pre-commit/issues/607

Fixes #402

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,8 +8,9 @@ repos:
         name: black
         language: system
         entry: black
+        minimum_pre_commit_version: 2.9.0
         require_serial: true
-        types: [python]
+        types_or: [python, pyi]
 
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.8.1
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,5 +4,6 @@
   entry: black
   language: python
   language_version: python3
+  minimum_pre_commit_version: 2.9.0
   require_serial: true
-  types: [python]
+  types_or: [python, pyi]


ichard26 (approved):
Looks good, many thanks for informing us of the `types_or` config option! Didn't know that the issue was fixed at last :D

MarcoGorelli:
Perhaps it would be good to also put
```
minimum_pre_commit_version: '2.9.0'
```

JelleZijlstra:
Agree with @MarcoGorelli, I'd be worried this causes issues for people running an earlier version of pre-commit.

.pre-commit-hooks.yaml
```diff
@@ -5,4 +5,4 @@
   language: python
   language_version: python3
```
MarcoGorelli, line 6:
```suggestion
  language_version: python3
  minimum_pre_commit_version: '2.9.2'
```

.pre-commit-hooks.yaml
```diff
@@ -4,5 +4,6 @@
   entry: black
   language: python
   language_version: python3
+  minimum_pre_commit_version: '2.9.0'
```
MarcoGorelli, line 7:
```suggestion
  minimum_pre_commit_version: "2.9.0"
```
my bad, sorry, `prettier` wants double quotes

jdufresne:
> Perhaps it would be good to also put: minimum_pre_commit_version

Thank you for this suggestion. I like it. 

Included in the latest revision. The quotes are not necessary as having multiple "." is an invalid number. YAML automatically interprets it as a string.

cooperlees (approved):
Thanks for this fix.

.pre-commit-config.yaml
```diff
@@ -8,8 +8,9 @@ repos:
         name: black
         language: system
         entry: black
+        minimum_pre_commit_version: 2.9.0
```
asottile, line 11: 2.9.2 would be ever so slightly better -- there was some breaking in 2.9.0-flat related to types_or

jdufresne: Thanks for reviewing and suggesting this. Updated in #1895.

PR #1896, opened 2021-01-01
Use Prettier pre-commit mirror

TBBle:
This fixes Prettier install failures similar to those seen in https://github.com/prettier/prettier/issues/9459, and is the solution recommended there.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,8 +24,8 @@ repos:
       - id: mypy
         exclude: ^docs/conf.py
 
-  - repo: https://github.com/prettier/prettier
-    rev: 1.19.1
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v1.19.1
     hooks:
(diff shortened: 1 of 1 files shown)


cooperlees:
Thanks for this. Lets see how happy CI is.

Just want to say, `pre-commit` sure does flip flops a lot from recommending use of it's mirrors vs. tools official repos. Very confusing.

TBBle:
This is actually batching up two moves. Prettier had already moved their pre-commit support to https://github.com/prettier/pre-commit, and then this split-out was (I assume) handed over to the pre-commit team to maintain.

TBBle:
CI seems happy. Looking at [the logs](https://travis-ci.org/github/psf/black/jobs/752428677), I guess it's been reusing the existing Prettier environment (since it only installed the changed Prettier, not the other environments), so this breakage might have been around silently for a little while.

cooperlees (approved):
Makes sense. Thanks!

Issue #1662, opened 2020-08-31
"--diff" option doesn't add a newline at the end of the file using 20.8b1


kimadeline:
**Describe the bug**

Running `black --diff` on a file without an ending newline will not add it to the diff when using 20.8b1.

**To Reproduce** Steps to reproduce the behavior:

1. Take this file:
```py
monday_temperatures = [9.1, 8.8, 7.5]
print(len(monday_temperatures))
```
2. Run `black --diff myfile.py`
3. Observed output:

```powershell
PS C:\Users\kimiguel\Documents\Sandbox\black-newline> black --diff .\myfile.py
would reformat myfile.py
All done! ✨ 🍰 ✨
1 file would be reformatted.
```

**Expected behavior**

For `black --diff` to add a newline at the end of the file in the diff, which was the behaviour in 19.10b0:

```powershell
PS C:\Users\kimiguel\Documents\Sandbox\black-newline> black --diff .\myfile.py
--- myfile.py   2020-08-31 22:20:31.166889 +0000
+++ myfile.py   2020-08-31 22:34:25.251541 +0000
@@ -1,2 +1,3 @@
 monday_temperatures = [9.1, 8.8, 7.5]
 print(len(monday_temperatures))
+
reformatted myfile.py
All done! ✨ 🍰 ✨
1 file reformatted.
```

**Environment (please complete the following information):**

- Version: black, version 20.8b1
- OS and Python version: Windows, Python 3.8.5

**Does this bug also happen on master?** 

This also happens on black, version 20.8b2.dev10+g1d2d726.

**Additional context** 

Running `black myfile.py` correctly without `--diff` adds a newline at the end of the file.

ichard26:
Just dropping by to add some information (I honestly have no idea why this was changed and how to make everyone happy):
* Apparently including the newline causes the diff to be unusable for `git apply` and `patch`: #526 
* The following commit is responsible for this behaviour: 959848c17639bfc646128f6b582c5858164a5001

TBBle:
Looking at #526, it seems to me that the diff format being generated for "add a final newline" was incorrect (it was somehow counting the final 0-byte additional line) but the applied solution just hid the problem by dropping the added newline entirely, invalidating the output of `--diff` against `black --check`.

[...]

PR #1897, opened 2021-01-01
Indicate that a final newline was added in --diff

TBBle:
Fixes: #1662

This works around [bpo2142](https://bugs.python.org/issue2142) from 2008.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,8 @@
 
 - speed up caching by avoiding pathlib (#1950)
 
+- `--diff` correctly indicates when a file doesn't end in a newline (#1662)
+
 #### _Packaging_
 
 - Self-contained native _Black_ binaries are now provided for releases via GitHub
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -6423,14 +6423,14 @@ def assert_stable(src: str, dst: str, mode: Mode) -> None:
 
 
 @mypyc_attr(patchable=True)
-def dump_to_file(*output: str) -> str:
+def dump_to_file(*output: str, ensure_final_newline: bool = True) -> str:
     """Dump `output` to a temporary file. Return path to the file."""
     with tempfile.NamedTemporaryFile(
         mode="w", prefix="blk_", suffix=".log", delete=False, encoding="utf8"
     ) as f:
         for lines in output:
             f.write(lines)
-            if lines and lines[-1] != "\n":
+            if ensure_final_newline and lines and lines[-1] != "\n":
                 f.write("\n")
     return f.name
 
@@ -6448,11 +6448,20 @@ def diff(a: str, b: str, a_name: str, b_name: str) -> str:
     """Return a unified diff string between strings `a` and `b`."""
     import difflib
 
-    a_lines = [line + "\n" for line in a.splitlines()]
-    b_lines = [line + "\n" for line in b.splitlines()]
-    return "".join(
-        difflib.unified_diff(a_lines, b_lines, fromfile=a_name, tofile=b_name, n=5)
-    )
+    a_lines = [line for line in a.splitlines(keepends=True)]
+    b_lines = [line for line in b.splitlines(keepends=True)]
+    diff_lines = []
+    for line in difflib.unified_diff(
+        a_lines, b_lines, fromfile=a_name, tofile=b_name, n=5
+    ):
+        # Work around https://bugs.python.org/issue2142
+        # See https://www.gnu.org/software/diffutils/manual/html_node/Incomplete-Lines.html
+        if line[-1] == "\n":
+            diff_lines.append(line)
+        else:
+            diff_lines.append(line + "\n")
+            diff_lines.append("\\ No newline at end of file\n")
+    return "".join(diff_lines)
 
 
 def cancel(tasks: Iterable["asyncio.Task[Any]"]) -> None:
--- a/docs/authors.md
+++ b/docs/authors.md
@@ -132,6 +132,7 @@ Multiple contributions by:
 - [Pablo Galindo](mailto:Pablogsal@gmail.com)
 - [Paul Ganssle](mailto:p.ganssle@gmail.com)
 - [Paul Meinhardt](mailto:mnhrdt@gmail.com)
+- [Paul "TBBle" Hampson](mailto:Paul.Hampson@Pobox.com)
 - [Peter Bengtsson](mailto:mail@peterbe.com)
 - [Peter Grayson](mailto:pete@jpgrayson.net)
 - [Peter Stensmyr](mailto:peter.stensmyr@gmail.com)
--- a/tests/data/missing_final_newline.diff
+++ b/tests/data/missing_final_newline.diff
@@ -1,0 +1,8 @@
+--- [Deterministic header]
++++ [Deterministic header]
+@@ -1,3 +1,3 @@
+ # A comment-only file, with no final EOL character
+ # This triggers https://bugs.python.org/issue2142
+-# This is the line without the EOL character
+\ No newline at end of file
++# This is the line without the EOL character
--- a/tests/data/missing_final_newline.py
+++ b/tests/data/missing_final_newline.py
@@ -1,0 +1,3 @@
+# A comment-only file, with no final EOL character
+# This triggers https://bugs.python.org/issue2142
+# This is the line without the EOL character
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -300,7 +300,6 @@ class BlackTestCase(BlackBaseTestCase):
             os.unlink(tmp_file)
         actual = result.output
         actual = diff_header.sub(DETERMINISTIC_HEADER, actual)
-        actual = actual.rstrip() + "\n"  # the diff output has a trailing space
         if expected != actual:
             dump = black.dump_to_file(actual)
             msg = (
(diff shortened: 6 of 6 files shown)


TBBle:
Now that I know it's an existing Python bug (I didn't think to look until after I posted the PR), I could reword this as a workaround for that bug, if that's valuable.

src/black/__init__.py
```diff
@@ -6380,11 +6380,18 @@ def diff(a: str, b: str, a_name: str, b_name: str) -> str:
...
+    diff_lines = []
+    for line in difflib.unified_diff(
+        a_lines, b_lines, fromfile=a_name, tofile=b_name, n=5
+    ):
+        if line[-1] == "\n":
+            diff_lines.append(line)
+        else:
+            diff_lines.append(line + "\n")
+            diff_lines.append("\\ No newline at end of file\n")
+    return "".join(diff_lines)
```
ichard26, line 6394: To someone like me, the `diff_lines.append("\\ No newline at end of file\n")` link just looks plain wrong. I would suggest explaining that this code is using a special feature of the patch specification(?) to work around a Python bug. Also linking the BPO would be the cherry on the cake!

ichard26 (changes requested):
The changes themselves look great and testing locally they do work so +1 from me!

Besides my review comment, please add an entry in `CHANGES.md` so the person verifying and updating the changelog before a new release has less work to do :) And I guess while you're at it, why not add yourself to the authors list in the README?

Also while this isn't a blocker, I can't reproduce the bug where Black doesn't add a newline using the VS Code integration on my machine interestingly 😕 Maybe the fact I'm using Visual Studio Code version 1.52.1 or version v2020.12.424452561 of the Python extension is why? I honestly don't care that much since theoretically this should fix the bug, but from experience even seemingly well tested PRs can have undetected serious holes ...

Thanks for looking into this and submitting a patch anyway! You rock!

src/black/__init__.py
TBBle: Yeah, sounds good. Now that I have a BPO reference, I can do that.

TBBle:
Would it be useful to mention _CHANGES.md_ in _CONTRIBUTING.md_? I wasn't aware I was supposed to add to the former, based only on reading the latter. (And based on the PR I modelled from, #1328.

TBBle:
@ichard26 Should _CHANGES.md_ link to the PR or the issue?

ichard26:
> Would it be useful to mention CHANGES.md in CONTRIBUTING.md? I wasn't aware I was supposed to add to the former, based only on reading the latter. (And based on the PR I modelled from, #1328.)

Yeah it would definitely help! Although I have other ideas to deal this issue (hi [towncrier](https://pypi.org/project/towncrier/)).

> Should CHANGES.md link to the PR or the issue

Honestly I'm not sure of what the actual recommendation is, but I would go with linking to the issue personally even though the status quo is to link to the PR.

src/black/__init__.py
TBBle: Done

CHANGES.md
```diff
@@ -26,6 +26,8 @@
 
 - use lowercase hex strings (#1692)
 
+- `--diff` correctly indicates when a file doens't end in a newline (#1662)
```
ichard26, line 29: Not a blocker but:
```suggestion
- `--diff` correctly indicates when a file doesn't end in a newline (#1662)
```

ichard26 (approved):

TBBle:
@ichard26 On the repro in VS Code, it's possible VS Code's general editor setting for "add missing newline" is set, which will be operating _after_ `black --diff` has been applied. That's the workaround I am using until this fix flows through to a release.

TBBle:
Sorry for the wild batch of force-pushes. I got excited and overlooked a couple of requested changes. -_-

It should be all-good now, the only changes since the last review were [textual](https://github.com/psf/black/compare/05b41c23af52ec4314e08b8553e9abfc7f6eb528..5d148d1cad2502bffd46c1ffda30bdba2fc16da9)

ichard26:
> On the repro in VS Code, it's possible VS Code's general editor setting for "add missing newline" is set

I checked if I had that option set and I'm pretty sure I didn't. When I purposely installed an immediately crashing version of Black, no newline was added still. But it's probably some misconfiguration or misunderstanding on my part, anyway 95% of time I use Vim. Does this patch for you work on your Visual Studio Code setup? If yes, I'll just go with that then :-)

TBBle:
I can reproduce the problem with released Black and Python extension ms-python.python-2020.12.424452561, under VSCode 1.52.1, after I remove `"files.insertFinalNewline": true` from the `[python]` section of settings.json (which is not visible in the settings UI editor).

And using the black.exe built from this PR, the problem is fixed.

ichard26:
Looking at this again, somethings are sticking out to me, I'mma do another review hopefully soon.

src/black/__init__.py
```diff
@@ -6424,11 +6424,20 @@ def diff(a: str, b: str, a_name: str, b_name: str) -> str:
...
+        a_lines, b_lines, fromfile=a_name, tofile=b_name, n=5
+    ):
+        # Work around https://bugs.python.org/issue2142
+        # See https://www.gnu.org/software/diffutils/manual/html_node/Incomplete-Lines.html
+        if line[-1] == "\n":
+            diff_lines.append(line)
+        else:
+            diff_lines.append(line + "\n")
+            diff_lines.append("\\ No newline at end of file\n")
+    return "".join(diff_lines)
```
ichard26, line 6440: Taking a second look at this, this is confusing IMO. `\\ No newline at end of file\n` means to me that the else block should only happen at the end of the diff, but the code is written like this condition can happen wherever in the diff. I would appreciate if this could be clarified. If this extra code is only useful for diff's ending, would be it possible to just check and modify the diff's ending like this (just illustrative code):
```py3
diff_lines = unified_diff(a, b, ...)
last_line = diff_lines[-1]
if last_line[-1] != "\n":
    diff_lines[-1] = last_line + "\n"
    diff_lines.append("\\ No newline at end of file\n")
```

If it's the other way around, could the `\\ No newline at end of file\n` message be tweaked?

ichard26 (commented):
Hey this almost looks good to go, but on a second pass, something doesn't seem right. Not a big deal, just something odd I noticed.

src/black/__init__.py
TBBle: Per [the spec](https://www.gnu.org/software/diffutils/manual/html_node/Incomplete-Lines.html), only the last line of a file can be the _incomplete line_, yes.

However, if you pass in non-final lines without newlines, `difflib` will accept that, and happily produce output lines without newlines.

Since that latter effect is really Garbage-In-Garbage-Out, and we're using `keepends` now to ensure that cannot happen, your suggested change makes sense, yes.

TBBle: Actually, wait. `diff_lines` doesn't necessarily have the last line of the input file as its last line. In our case, since we always add a final newline, it's only the last `-...` line that may hit this case, not the last line in `diff_lines`, which will be the corresponding `+...` line.

See tests/data/missing_final_newline.diff for this exact case.

In the general case, a _context_ line (` ...`) or added line (`+ ...`) could require this marker too.

TBBle: Technically, I could walk backwards in `diff_lines`, checking the final `-`, final `+`, and final ` ` for missing final newline (and if it's the ` ` case, barfing if we also see it on a `+` or `-` case) but if I was going to do that, I should also be checking for a `\` line, which would appear if BPO 2142 is fixed, and hence this workaround would be deactivated in that case.

TBBle: If the text of the `\...` line is confusing, that can be trivially changed, this is just copied from the spec, but in the actual implementation of `diff`, it's localised with gettext. (Although I fear that there are patch-parsers in the wild which are looking for _this text_ rather than the `\` marker.)

We could even look at the first character of the affected line, and have the text vary by whether it was a `-...`, `+...`, or ` ...` line.

I won't do that without some positive direction though, as it feels like it's adding needless complexity for what is a pretty-standard text string, i.e. `git diff` also produces this literal string. Even GitHub's diff preview uses the exact string "No newline at end of file" in its popup. And it's complexity that a future fix for BPO 2142 probably won't have, since the existing proposed fixes didn't try to distinguish this either. So the patch format would revert back to what it is here, at that point.

ichard26: Alright that makes sense. Thanks for describing the specification!

ichard26 (approved):
/cc @cooperlees @JelleZijlstra this should be ready to go, a final lookover and eventual merge would be appreciated! Thanks in advance!

TBBle:
Master build failed on [Documentation build](https://github.com/psf/black/runs/1949719732) but I can't see how it relates to my changes.

```
Warning, treated as error:
autodoc: failed to import function 'should_split_body_explode' from module 'black'; the following exception was raised:
  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/sphinx/util/inspect.py", line 325, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: module 'black' has no attribute 'should_split_body_explode'

  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/sphinx/ext/autodoc/importer.py", line 106, in import_object
    obj = attrgetter(obj, mangled_name)
  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/sphinx/ext/autodoc/__init__.py", line 289, in get_attr
    return autodoc_attrgetter(self.env.app, obj, name, *defargs)
  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/sphinx/ext/autodoc/__init__.py", line 2186, in autodoc_attrgetter
    return safe_getattr(obj, name, *defargs)
  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/sphinx/util/inspect.py", line 341, in safe_getattr
    raise AttributeError(name) from exc
AttributeError: should_split_body_explode
```

`should_split_body_explode` was renamed to `should_split_line` in #1963, and I guess _/docs/reference/reference_functions.rst_ wasn't updated.

jayaddison:
@TBBle Out of interest, did you spot the newline handling at https://github.com/psf/black/blob/cd4295dd9888f491cfd9aae060d7832b7f831b24/src/black/__init__.py#L1116-L1117 while preparing your changes here?  I don't think they're directly related, but wondered if we might be able to remove that special case at some point.

TBBle:
That code looks more like it's related to the parser driver requiring newline-terminated text, which I think is probably part of the API contract.

PR #1899, opened 2021-01-02
Support for top-level user configuration

### LINKED ISSUE #1577: Search for pyproject.toml to user root regardless of .git presence
When I have a project that does not have a pyproject.toml file, such as a project without a style guide.  I would still like my code to use a system-wide default for black config.  Ie I would like to place a config file in my user root directory for a system wide default.  I am able to that with a .flake8, .pylintrc, and .isort.cfg file in this directory for system wide default.  Isort also allows to parse a pyproject.toml in the user root as well.  If black did the same, it would make things very tidy.  If this is undesirable for the maintainers, I understand, but wanted to suggest it in case it is something willing to be considered.

rharish101:
This should resolve #1577. At the user level, a TOML config can now be specified with the following paths (not directories, but paths to the TOML file):
* Windows: `~\.black`
* Unix-like: `$XDG_CONFIG_HOME/black` (`~/.config/black` if `XDG_CONFIG_HOME` is not set)

These locations are searched only when no configuration in the current project directory is found. I chose these defaults because this is how [flake8](https://flake8.pycqa.org/en/latest/user/configuration.html#user-configuration) and [pycodestyle](https://pycodestyle.pycqa.org/en/latest/intro.html#configuration) specify their user-level configs.

TODOs:
- [x] Fix existing tests: Running `python -m unittest` results in failed tests if there's a top-level user config, but all of them pass otherwise.
- [x] Fix `black-primer`: Same as above, but with `black-primer`.
- [x] Add new tests.
- [x] Add documentation: README, website, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,20 @@ parent directories. It stops looking when it finds the file, or a `.git` directo
 If you're formatting standard input, _Black_ will look for configuration starting from
 the current working directory.
 
+You can use a "global" configuration, stored in a specific location in your home
+directory. This will be used as a fallback configuration, that is, it will be used if
+and only if _Black_ doesn't find any configuration as mentioned above. Depending on your
+operating system, this configuration file should be stored as:
+
+- Windows: `~\.black`
+- Unix-like (Linux, MacOS, etc.): `$XDG_CONFIG_HOME/black` (`~/.config/black` if the
+  `XDG_CONFIG_HOME` environment variable is not set)
+
+Note that these are paths to the TOML file itself (meaning that they shouldn't be named
+as `pyproject.toml`), not directories where you store the configuration. Here, `~`
+refers to the path to your home directory. On Windows, this will be something like
+`C:\\Users\UserName`.
+
 You can also explicitly specify the path to a particular file that you want with
 `--config`. In this situation _Black_ will not look for any other file.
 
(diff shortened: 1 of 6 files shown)


rharish101:
Currently I've patched `black-primer` to use the work path as `XDG_CONFIG_HOME`, which basically switches off the top-level user config search. However, the Windows path cannot be configured. This means that `black-primer` will fail when run locally on Windows if the user has setup a top-level user config. Any advice on how I should go about fixing this?

Also, what's up with the Fuzz tests? Given that GitHub actions wouldn't be having a top-level user config for _Black_, my changes shouldn't affect them. However, it seems that they're failing randomly.

rharish101:
I've now cleaned up the git commit history for easy reviewing.

rharish101:
> Currently I've patched `black-primer` to use the work path as `XDG_CONFIG_HOME`, which basically switches off the top-level user config search. However, the Windows path cannot be configured. This means that `black-primer` will fail when run locally on Windows if the user has setup a top-level user config. Any advice on how I should go about fixing this?

Solved this by overriding the `USERPROFILE` environment variable. Python (specifically `os.path` and `pathlib`) uses this to determine the home directory on Windows. Therefore, manipulating this helped me hide the actual home directory and thus a possible top-level user config.

Issue #1696, opened 2020-09-10
Comments that are vertically aligned across 3 or more lines can remain vertically aligned.


earonesty:
**Comments aligned.... before black, but code need black:**

```python
   return {'f_bsize': MAX_DATA_SIZE,
           'f_frsize': MAX_DATA_SIZE,
           'f_blocks': total.value // FRAG_SIZE,      # FRAG_SIZE includes headers, so the number of blocks,
           'f_bfree': free.value // FRAG_SIZE,        # free blocks, and available blocks must account for
           'f_bavail': freeavail.value // FRAG_SIZE,  # the extra space they will take up
           'f_files': MAX_ULLONG,
           'f_ffree': MAX_ULLONG,
           'f_favail': MAX_ULLONG,
           'f_namemax': namemax.value, }
```

**Comments not aligned.... after black**

```python
   return {
       "f_bsize": MAX_DATA_SIZE,
       "f_frsize": MAX_DATA_SIZE,
       "f_blocks": total.value // FRAG_SIZE,  # FRAG_SIZE includes headers, so the number of blocks,
       "f_bfree": free.value // FRAG_SIZE,  # free blocks, and available blocks must account for
       "f_bavail": freeavail.value // FRAG_SIZE,  # the extra space they will take up
       "f_files": MAX_ULLONG,
       "f_ffree": MAX_ULLONG,
       "f_favail": MAX_ULLONG,
       "f_namemax": namemax.value,
   }
```

**Desired style... best of both**

```python
   return {
       "f_bsize": MAX_DATA_SIZE,
       "f_frsize": MAX_DATA_SIZE,
       "f_blocks": total.value // FRAG_SIZE,      # FRAG_SIZE includes headers, so the number of blocks,
       "f_bfree": free.value // FRAG_SIZE,        # free blocks, and available blocks must account for
       "f_bavail": freeavail.value // FRAG_SIZE,  # the extra space they will take up
       "f_files": MAX_ULLONG,
       "f_ffree": MAX_ULLONG,
       "f_favail": MAX_ULLONG,
       "f_namemax": namemax.value,
   }
```

ichard26:
Given that a resolving PR was rejected due to ethos incompatibilities, we won't be accepting this. Sorry about that, but thanks for the suggestion.

Issue #379, opened 2018-06-26
Reflows of inline comments can cause misalignment


quodlibetor:
Operating system: macOS
Python version: 3.6
*Black* version: 18.6b4
Does also happen on master: yes

_edit_: moved original first example to end because it's reasonable behavior

With this (example 2), black moves everything down one so it is aligned incorrectly:

```python
# original
revert("!" +
       one(2) +         # extension
       three(4) +       # length
       one(flapping) +  # flapping info present
       two(duration) +  # duration
       two(0))
```

becomes:

```python
revert(
    "!"
    + one(2)
    + three(4)  # extension
    + one(flapping)  # length
    + two(duration)  # flapping info present
    + two(0)  # duration
)
```

example 1 (reasonable behavior, original text preserved):

```python
# original
a = [1,
     2, # very important
     3, # also important
     4, # less important, but still worth reading
     5] # this is just for the line length
```

Black moves the last comment to a weird place, while handling most other lines correctly:

```python
# formatted
a = [
    1,
    2,  # very important
    3,  # also important
    4,  # less important, but still worth reading
    5,
]  # this is just for the line length
```

carljm:
Thanks for the report! 

My feeling is that your first example is "working as designed" and probably can't be reasonably fixed. The last comment lives outside the brackets, and it's reasonable for Black to keep it outside the brackets; how is Black supposed to know that it "goes with" the argument on that line rather than being a comment on the whole bracketed expression? It's not hard to move after reformatting, and then Black will keep it there.

The second example might be fixable, I'm not sure without looking into it more closely.

quodlibetor:
Thanks that's totally reasonable, I didn't consider that POV but I'm fine with it.

The second example seems like a genuine bug (not an enhancement), though, whether or not it's fixable.

carljm:
Judgment call I guess; I call it an enhancement because all comment placement is kind of best-effort by heuristic, there will never be a solution that everyone finds optimal for all cases. Even in that case it's easy to see what Black is doing: it's keeping the comment on the same line as the same point in the AST. So the comment `# extension` occurs in the middle of `+ three(4)` in the original code, and ends up on the same line as that same spot in the reformatted code. It may be possible to improve the heuristic to do a bit better in this case (e.g. privilege association with name nodes in the AST vs operators, or something), but it'll need some care to see that we don't make other scenarios worse by doing so.

quodlibetor:
[...]

PR #1904, opened 2021-01-04
WIP: Added function for vertically aligning inline comments

rsnk96:
Enforces  vertical alignment of inline comments in consecutive lines

Also addresses https://github.com/psf/black/issues/379 and partially https://github.com/psf/black/issues/1696

Input:
```python
# Inline comments on simple statements
v = 10 #variable 1
my_variable_2 = 20 # variable 2

# Inline comments on multi-line statement
my_dict = {
    "a": 10,  # key a
    "b": 20, # key b
    "c": my_variable_2//2,   # key c
    "d": my_variable_2//10,   # key d
    "e": 50,   # key d
}
```

Output:
```python
# Inline comments on simple statements
v = 10              # variable 1
my_variable_2 = 20  # variable 2

# Inline comments on multi-line statement
my_dict = {
    "a": 10,                   # key a
    "b": 20,                   # key b
    "c": my_variable_2 // 2,   # key c
    "d": my_variable_2 // 10,  # key d
    "e": 50,                   # key d
}
```

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1054,7 +1054,17 @@ def align_comments(dst_contents: List[str]) -> str:
     max_hashtag_position = 0
     num_triple_quotes = 0
     hashtag_positions_to_be_aligned = []
-    
+
+    def find_string_ranges(line):
+        quote_string_regex = re.compile('"(?:[^"\\\]|\\\.)*"')
+        apostrophe_string_regex = re.compile("'(?:[^'\\\]|\\\.)*'")
+        line_safe = line.replace('"""', "   ").replace("'''", "   ")
+        inline_strings = [
+            *(match.span() for match in quote_string_regex.finditer(line_safe)),
+            *(match.span() for match in apostrophe_string_regex.finditer(line_safe)),
+        ]
+        return inline_strings
+
     def align_inline_comments(
         lines, hash_position_to_set, start_idx, end_idx, hash_positions
     ):
@@ -1066,15 +1076,28 @@ def align_comments(dst_contents: List[str]) -> str:
                 + line[curr_hash:]
             )
 
-    def search_without_preceeding_backslash(pattern, string):
+    def search_without_preceeding_char(pattern, string):
         return [m.start() for m in re.finditer(f"(?<!\\\){pattern}", string)]
 
+    import time
+
     for line_idx, line in enumerate(dst_lines):
-        # For all the re.finditer() below, ensure there is no preceeding backslash
-        triple_quote_positions = search_without_preceeding_backslash('"""', line)
-        single_quote_positions = search_without_preceeding_backslash('"', line)
-        hashtag_positions = search_without_preceeding_backslash("#", line)
-        hashtag_position = -1
+        inline_string_ranges = find_string_ranges(line)
+
+        # Find all triple quoted strings that are not substrings within inline strings
+        triple_quote_positions = search_without_preceeding_char('"""', line)
+        triple_quote_positions.extend(search_without_preceeding_char("'''", line))
+        invalid_tqps = []
+        for tqp_idx, tqp in enumerate(triple_quote_positions):
+            if any(tqp > low and tqp < high for low, high in inline_string_ranges):
+                invalid_tqps.append(tqp_idx)
+        [
+            triple_quote_positions.pop(posn) for posn in invalid_tqps[::-1]
+        ]  # Safe pop all the triple quotes within an inline string (in reverse)
+
+        # Find the hashtag position in the comment, if any
+        hashtag_positions = search_without_preceeding_char("#", line)  # all hashtags
+        comment_position = -1                                          # hashtag corresponding to comment
         for curr_hash in hashtag_positions:
             closed_triple_quotes = (
                 num_triple_quotes  # multiline strings that are not yet closed
@@ -1082,18 +1105,18 @@ def align_comments(dst_contents: List[str]) -> str:
                     posn < curr_hash for posn in triple_quote_positions
                 )  # triple-quoted strings that begin in the same line, e.g. docstrings
             ) % 2 == 0
-            closed_single_quotes = all(
-                curr_hash > closing_single_quote
-                for closing_single_quote in single_quote_positions[1::2]
+            curr_hash_inside_inline_string = any(
+                curr_hash > low and curr_hash < high
+                for low, high in inline_string_ranges
             )
-            if closed_triple_quotes and closed_single_quotes:
-                hashtag_position = curr_hash
+            if closed_triple_quotes and not curr_hash_inside_inline_string:
+                comment_position = curr_hash
                 break
 
-        if line.strip().find("#") != 0 and hashtag_position != -1:
+        if line.strip().find("#") != 0 and comment_position != -1:
             inline_comment_line_counter += 1
-            hashtag_positions_to_be_aligned.append(hashtag_position)
-            max_hashtag_position = max(max_hashtag_position, hashtag_position)
+            hashtag_positions_to_be_aligned.append(comment_position)
+            max_hashtag_position = max(max_hashtag_position, comment_position)
         elif inline_comment_line_counter:
             align_inline_comments(
                 dst_lines,
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1043,7 +1043,29 @@ def format_str(src_contents: str, *, mode: Mode) -> FileContent:
             current_line, mode=mode, features=split_line_features
         ):
             dst_contents.append(str(line))
-    return "".join(dst_contents)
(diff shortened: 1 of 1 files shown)


rsnk96:
Can someone help me understand why it is bugging out and raising the `not equivalent to the source` error for the following input (from `__init__.py`)

```python
matches = re.findall(
    r"""
    (?:[^{]|^)\{  # start of the string or a non-{ followed by a single {
        ([^{].*?)  # contents of the brackets except if begins with {{
    \}(?:[^}]|$)  # A } followed by end of the string or a non-}
    """
)

```

JelleZijlstra:
I don't think we want to do this. It's going to be disruptive to existing codebases that use Black, and it's not clear that aligning the comments is always better.

JelleZijlstra:
Also, your error is because it's changing things inside a string literal, which is not a (Python) comment. If we do this, it should affect only Python comments.

rsnk96:
Thanks for your review @JelleZijlstra 

There are two options on integration which won't always-enforce this (and hence break codebases as you suggested):
1. Add it as an optional flag in black, to enable vertical alignment of inline comments
2. Make it such that inline comments which were previously vertically aligned remain so after formatting with black. This is currently a major problem with a lot of big codebases that use Python (like CPython), because the devs who write the code keep vertically aligned comments for better readability

Do share your thoughts on these options

rsnk96:
Hey @JelleZijlstra 

Gentle reminder if you have the time, do give your opinion on what would make for a good integration option from the above^ 😄

JelleZijlstra:
My opinion is that we shouldn't do it. The strength of Black is a consistent style with few options, and I don't see a strong reason to go against that principle.

Other maintainers may have different opinions.

rsnk96:
Okay thanks for your take @JelleZijlstra. I'm sorry I haven't contributed to black before, so can you tag some other maintainers who might be interested in reviewing this?

I see @ichard26 who assigned a label to https://github.com/psf/black/issues/1696, and @carljm @zsol  @ambv who gave suggestions on https://github.com/psf/black/issues/1696. Sorry folks, if I dragged your name out of the blue, just wanted to know if I should pursue this PR 😅

cooperlees:
I'll share that I like this since you're asking, but it's not a major win or a must have I feel. I also agree it's another change "black conformant" repos will have CI fail on and I can see it annoying people.

PEP8 is also very loose here:
```quote
Use inline comments sparingly.

An inline comment is a comment on the same line as a statement. Inline comments should be separated by at least two spaces from the statement. They should start with a # and a single space.
```

I think black already ensures there are 2 spaces and a space after the # ... i.e. `  # ` ...

rsnk96:
Thanks for the quick response @cooperlees ! Any thoughts on the alternate modes of integration?
> There are two options on integration which won't always-enforce this (and hence break codebases as you suggested):
> 1. Add it as an optional flag in black, to enable vertical alignment of inline comments
> 2. Make it such that inline comments which were previously vertically aligned remain so after formatting with black

cooperlees:
When @ambv set out to make black, we (as I sat next to him @ work) decided one of the main goals was to really limit the number of flags so that the style is actually a consistent style. Not one of X variations that adding flags creates. This is modeled after `gofmt` and other less configurable autoformatters that have had great success.

Due to this, I vote for 2. If you can detect that they were aligned before formatting and uphold that, then this would truly be a killer feature that I would condone.

rsnk96:
Thanks, that gives me a direction, will give it a shot

JelleZijlstra:
For what it's worth, I wouldn't support that change either. I prefer for Black not to take into account previous formatting as much as possible, so people writing code have to worry about formatting as little as possible.

rsnk96:
Oh, alright. I'll wait for suggestions from other contributors and maintainers in that case! Do tag anyone you might find relevant.

rsnk96:
> so people writing code have to worry about formatting as little as possible

P.S. just to clarify why I opened this PR, I like `black` not just because it takes care of formatting, but also because it makes code a lot more readable. Just thought this feature would be a good addition for the latter reason, I find py code with aligned inline comments very easy to interpret :)

But I do understand that it shouldn't introduce complications since it's already widely adopted. Will await comments from other contributors and maintainers

ichard26:
I wasn't part of this project when its philosophy and design goals were hashed out (I only joined mid-way through 2020!) but here's my thoughts for your consideration (for whatever weight they may or not carry):

Honestly I'm not a big fan of either option. Adding a new option / flag is a big no-no since it goes against the philosophy of Black: to be as non-configurable as possible so style-configuration wars don't break out and waste time. For the second option, I still have an aversion towards it because it still goes against Black's philosophy in a way IMO. Flags directly introduce variants, but previous-formatting-dependent formatting (note-to-self: need a better term) still introduces variants. Except instead of being explicit as you can just look at the Black config and be like "Oh X option is turned, that's why Black does Y formatting", you must read the docs for an explication why Black is doing something unexpected: "Wait what, why did Black format this like this? I know there isn't any config for it..." Really it boils down to that Blackened code loses its consistency.

\*But\*, there's an argument to favour practicality over purity. And I see the value of supporting aligned inline comments, especially given that sometimes they are critical to the understanding of some code. Like keeping aligned comments still aligned after formatting sounds nice; and I do feel you when Black makes my code look (subjectively) worse and less readable. It sucks, but I'm lazy so my feelings of disappointment quickly fade.

I guess a list of (relevant to this discussion) priorities an autoformatter like Black could have are:

1. Consistency
2. Readability
3. Practicality

Personally I hold 1. as the most important thing to work towards. It's what made Black unique, but Black isn't a tiny project. It plays a huge role in the Python ecosystem and maybe keeping 1. as the highest priority isn't worth it anymore. I don't think Black was intended to be \*the\* Python autoformatter, <sup>1</sup> but Black is not getting less popular and influential anytime soon so... hmm 

I'm split but for now consider me a weak -0.5 for this functionality not because it's a bad idea, but because of the slippery slope it can unintentionally kick-start. Feel free to convince me to the other way.

--- 
1. I don't mean to sound ungrateful being part of a huge project like Black. It's a honour that I get to help out and sometimes change the direction the project is going through my participation. I just feel like Black is too opinionated to where it isn't intended to be useful (and maybe shouldn't try to be useful) to all of the general Python developer community at large. It's great a lot of the community finds it worth it though.

felix-hilden:
Just to provide one more opinion to the list, I agree with the previous comments in that I'm 100 % against this being the default or configurable. I'm *not* completely against Black detecting existing aligned comments, but I'd want a stronger go-ahead from other maintainers to actually pursue that.

ambv:
Thanks for your work on Black but as the previous maintainer comments tell you, this is out of scope for inclusion.

Anti-Distinctlyminty:
My comment may be parochial, but I did end up here looking for a solution to this, so with that in mind...
This is one of the very rare times that Black takes my code and makes it less readable.

dseynhae:
It looks like I'm a day late and a dollar short here.
I would love to see Black used by my colleagues, as it makes the diff between different authors working on the same code independent of any configuration options. There aren't any configuration options...


But I'm dealing with people that are against formatters in general, not even Black specifically. The arguments are always about readability... In some cases consistency is NOT the most important thing.

* I agree that we shouldn't make the alignment of configurable (advocated by @cooperlees and others).
* And I think it is reasonable to maintain existing alignment (which I think was the suggestion from @rsnk96).

When I was reading the documentation on how Black deals with comments, I thought that it was already resolved:

> Some types of comments that require specific spacing rules are respected: doc comments (#: comment),

So I tried it:

```python
proc = subprocess.run(
    cmd,
    capture_output=True,  #:PIPE results to stdout and stdin
    shell=True,           #:Command is specified as a string
    text=True,            #:Open stdout and stdin in text mode
    check=False,          #:Raise CalledProcessError exception
)
```

👉 I was hoping that the `#:` syntax had some alignment power... But Black 22.3.0 turned this promptly into an obvious mess:

```python
proc = subprocess.run(
    cmd,
    capture_output=True,  #:PIPE results to stdout and stdin
    shell=True,  #:Command is specified as a string
    text=True,  #:Open stdout and stdin in text mode
    check=False,  #:Raise CalledProcessError exception
)
```

💡 But wouldn't this `#: comment` be a nice compromise?
* No flag needed.
* Existing code is not affected (unless the rare case where the comment starts with `#:`)
* The alignment is not based on the alignment existing in the existing code (taking care of @JelleZijlstra 's objection)
* Alignment is an intent. And this intent can now subtly yet clearly be communicated. And enforced by Black...

aldanor:
Agreed with @dseynhae here, having a clean and easy way of communicating to black "please, don't mess up alignment of these" would be very nice. And something like a `#:` prefix should be good enough indeed.

Are there any serious objections? Could this issue be reopened or is it closed for good and is not open for discussion?

Issue #1656, opened 2020-08-29
Should Black remove optional parentheses that wrap assignment expressions?


ichard26:
**Describe the style change** 

Black would now remove optional parentheses that surround assignment expressions. Also Black will now have to take into account of situations where such removal would be unsafe (e.g. top level unparenthesized assignment expressions are illegal).

**Examples in the current style**

```python
while (temp := True):
    pass

if (temp := True):
    pass
```

**Examples in the new style**:

```python
while temp := True:
    pass

if temp := True:
    pass
```

**Things to discuss**:

* What are the conditions where Black would remove/keep the wrapping parentheses? Perhaps we can use this [specification](https://github.com/psf/black/blob/master/docs/the_black_code_style.md#parentheses) as a starting point?

* Do we want this in the first place?

**Additional context**:

This was originally filed in #1522 by @xrisk. I closed it as a duplicate of another issue and broke out the other topic in it as this new issue.

Related comments in original issue:

1. JelleZijlstra
> I'm not as sure we should remove the parentheses, maybe it's better for clarity to have them? Black is opinionated, but we should make sure we're opinionated in the right way.

2. xrisk (author of original issue)
> [snipped mention] black does convert if (expr): to if expr: so…
>
> And there are multiple examples in the PEP that omit the parentheses when there is no RHS.

JelleZijlstra:
I vote we remove the parentheses if the assignment expr is the whole body of an if/while condition. As someone mentioned on the other issue, that's consistent with Black generally turning `if (expr):` into `if expr:`.

xrisk:
Is there consensus regarding this? If so, I would like to try my hand at attempting a fix :)

JelleZijlstra:
Other maintainers can chime in, but so far nobody has disagreed, so I say go for it.

PR #1906, opened 2021-01-05
Strip redundant parentheses from assignment exprs

xrisk:
#1656 

This converts code of the form:

```
if (foo := 0):
    pass
```

into

```
if foo := 0:
    pass
```

which is consistent with how `if (foo):` is transformed into `if foo:` and also with examples in PEP 572. 

— 

Although all tests are passing, please review particularly this condition:
`and parent.children[1].type == token.EQUAL`

I am unsure if children[1] is always guaranteed to be the operator.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -5540,17 +5540,19 @@ def maybe_make_parens_invisible_in_atom(node: LN, parent: LN) -> bool:
     Returns whether the node should itself be wrapped in invisible parentheses.
 
     """
+
     if (
         node.type != syms.atom
         or is_empty_tuple(node)
         or is_one_tuple(node)
         or (is_yield(node) and parent.type != syms.expr_stmt)
         or max_delimiter_priority_in_atom(node) >= COMMA_PRIORITY
-        or parent.type == syms.expr_stmt
-        and parent.children[1].type == token.EQUAL
-        and is_walrus_assignment(node)
     ):
         return False
+
+    if is_walrus_assignment(node):
+        if parent.type in [syms.annassign, syms.expr_stmt]:
+            return False
 
     first = node.children[0]
     last = node.children[-1]
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -5368,10 +5368,7 @@ def normalize_invisible_parens(node: Node, parens_after: Set[str]) -> None:
             check_lpar = True
 
         if check_lpar:
-            if is_walrus_assignment(child):
-                pass
-
-            elif child.type == syms.atom:
+            if child.type == syms.atom:
                 if maybe_make_parens_invisible_in_atom(child, parent=node):
                     wrap_in_parentheses(node, child, visible=False)
             elif is_one_tuple(child):
@@ -5549,6 +5546,9 @@ def maybe_make_parens_invisible_in_atom(node: LN, parent: LN) -> bool:
         or is_one_tuple(node)
         or (is_yield(node) and parent.type != syms.expr_stmt)
         or max_delimiter_priority_in_atom(node) >= COMMA_PRIORITY
+        or parent.type == syms.expr_stmt
+        and parent.children[1].type == token.EQUAL
+        and is_walrus_assignment(node)
     ):
         return False
 
--- a/tests/data/pep_572.py
+++ b/tests/data/pep_572.py
@@ -2,7 +2,7 @@
 (a := a)
 if (match := pattern.search(data)) is None:
     pass
-if (match := pattern.search(data)):
+if match := pattern.search(data):
     pass
 [y := f(x), y ** 2, y ** 3]
 filtered_data = [y for x in data if (y := f(x)) is None]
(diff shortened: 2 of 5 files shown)


src/black/__init__.py
```diff
@@ -5481,6 +5478,9 @@ def maybe_make_parens_invisible_in_atom(node: LN, parent: LN) -> bool:
         or is_one_tuple(node)
         or (is_yield(node) and parent.type != syms.expr_stmt)
         or max_delimiter_priority_in_atom(node) >= COMMA_PRIORITY
+        or parent.type == syms.expr_stmt
+        and parent.children[1].type == token.EQUAL
```
xrisk, line 5482: I’m not fully sure about this condition

xrisk:
Seems to break on pre-3.8. Will fix.

xrisk:
Alright fuzzer failures seem to be unrelated and they’re failing on HEAD as well.

xrisk:
@JelleZijlstra requesting a review!

ichard26:
Reopening to re-run CI as the fuzz issues *should* be fixed.

ichard26:
Okay we're hitting #1913 by chance now...

xrisk:
Hm the code I've touched seems unrelated to #1913. Do we wait for that to get resolved first?

ichard26:
Sorry, should've made it clearer that the unrelated fuzz error was something we've seen before.  TR;DR: CI passed for your PR :tada:

xrisk:
Hi, please let me know if I need to do anything else here.

ichard26 (changes requested):
Thanks for working on this! This is my first review of a formatting PR (I handle docs usually) so I hope I'm doing this right and carefully :) This should also be reviewed by Jelle since I believe they are our most qualified reviewer for this sort of stuff. But on with my review first!

Ideally this would have more test cases (complex if and while statements would be nice!), but it isn't a blocker.

> I am unsure if children[1] is always guaranteed to be the operator.

The answer is no :(

Here's a few examples of where `parent.children[1].type == token.EQUAL` falls apart:

1. `test += (a := 1)` (who would write this tho?)
2. `a, b = (test := (1, 2))`
3. `test: int = (test2 := 2)`

edit: don't ask me how to workaround this problem, I have no idea

Also, you'll need to edit [this section](https://github.com/psf/black/blob/master/docs/the_black_code_style.md#parentheses) adding a note how Black will not remove parentheses where an assignment expression is used since it's invalid.

xrisk:
Alright, I think it should be correct now. But perhaps we should wait for the refactoring that’s going on in #1958 to finish.

@jayaddison Are you planning to touch this code?

jayaddison:
@xrisk Thanks for checking; I'm not yet fully confident about the changes in #1958, so please don't consider it a blocker.

ichard26 (changes requested):
> Also, you'll need to edit this section adding a note how Black will not remove parentheses where an assignment expression is used since it's invalid.

This was never resolved, please do so.

But overall, great improvement, thanks for your time and effort! If you please, add yourself to the AUTHORS list in the readme.

ichard26:
> This is my first review of a formatting PR (I handle docs usually) so I hope I'm doing this right and carefully :) This should also be reviewed by Jelle since I believe they are our most qualified reviewer for this sort of stuff. But on with my review first

@JelleZijlstra, I would appreciate if you could make sure I haven't missed anything vital! This PR is mostly ready, just a few documentation issues (of which I can handle and review).

JelleZijlstra (approved):
Test cases look good and the change in black itself seems reasonable.

By the way, I'm the main person reviewing formatting changes in Black these days, but even I don't exactly feel like I know the codebase well.

JelleZijlstra:
Oops, I merged without resolving @ichard26's request about the documentation, sorry for that! @xrisk would you be able to send a followup PR? If not I can do it myself.

ichard26:
> By the way, I'm the main person reviewing formatting changes in Black these days, but even I don't exactly feel like I know the codebase well.

You've got some experience at least compared to me :)

> Oops, I merged without resolving @ichard26's request about the documentation, sorry for that! @xrisk would you be able to send a followup PR? If not I can do it myself.

I was about to comment about that, don't fret too much as now I do see how you might have skipped over it since it was poorly worded. I should really reread stuff I write before pressing "comment". Ugh, old habits die hard.

JelleZijlstra:
@ichard26 don't blame yourself, it's my bad! You clearly requested changes, so I should have noticed.

xrisk:
@ichard26 

> Also, you'll need to edit this section adding a note how Black will not remove parentheses where an assignment expression is used since it's invalid.

Isn’t this implicit, since that section in the docs talks about “optional parentheses”? The parentheses as used in something like `x = (y := 5)` aren’t optional.

Also I would expect that behavior from the principle of least surprise, we should never be generating invalid code in the first place :P

ichard26:
@xrisk yeah you're right lol. Apologies for the noise. Thanks for the PR!

PR #1909, opened 2021-01-06
Release gh action

rickstaa:
Makes some changes to the gh-action such that it can be released to the marketplace. I could be wrong, but to my knowledge users do not have access to the `args` docker argument when using the action. I, therefore, added a `black_args` input argument.


README.md
```diff
@@ -411,7 +411,14 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
       - uses: psf/black@stable
+        with:
+          black_args: ". --check"
 ```
+### Inputs
```
cooperlees, line 417:
```suggestion

### Inputs
```
- Make linter happy ...

cooperlees (changes requested):
Please fix the lint - Rest looks good.

Fuzz: libcst errors parsing it seems - I don't understand what but it can't be caused by this PR:
```
 File "<string>", line 6, in __init__
 ```
 - So I'll ignore that and merge and look to fix elsewhere.

**Comment by ichard26** [comment]
> Fuzz: libcst errors parsing it seems - I don't understand what but it can't be caused by this PR:

This PR is hitting https://github.com/psf/black/issues/1012 by chance. Also, LibCST? You mean the obviously better blib2to3 :)

**Comment by cooperlees** [comment]
Ignoring fuzzer changes we need to look into elsewhere and merging this to unblock the release to market place

**Comment by ichard26** [comment]
> but to my knowledge users do not have access to the args docker argument when using the action

Maybe it would be worth investigating https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsargs because IMO this is pretty bad UX even it works. Thanks for working on this though!

**Comment by rickstaa** [comment]
@ichard26 Thanks for your comment! Your correct, I also noticed this some weeks ago on another action. I didn't have the time yet to open a new pull request. I think the code can be simplified to:

```bash
#!/bin/sh
set -e

black_args=(".")
if [ "${INPUT_BLACK_ARGS}" != "" ]; then
  black_args+=(${INPUT_BLACK_ARGS})
else
  # Default (if no args provided).
  black_args+=("--check" "--diff")
fi

sh -c "black . ${black_args[*]}"
```

@ichard26, @cooperlees Let me know what you think.

simaki:
This is a kind reminder. @ichard26 @cooperlees 

Thank you a lot, @rickstaa!  I use this action and find it quite useful.

By the way, that script may give `.` to black twice.  Is it intended?
Also, now that we only have `$INPUT_BLACK_ARGS` which is just a string (rather than an array of args), the script may be able to be further simplified like this.

```sh
#!/bin/sh -e

args=$INPUT_BLACK_ARGS
if [ -z "$args" ]; then
  # Default (If no args provided).
  args='. --check --diff'
fi

sh -c "black $args"
```

Could you make a PR, @rickstaa?  Or I'll handle it if you are busy.

rickstaa:
@simaki Thanks a lot for your comment! I this pull request, I did my best also to support the docker ``args `` option. But as pointed out by @ichard26, this option is only needed during debugger and not useful for the user. I think your last solution is the best one:

```bash
#!/bin/sh -e

sh -c "black $INPUT_BLACK_ARGS"
```

If @cooperlees agrees, I think it would be great if you could create a pull request.

cooperlees:
I see no problem with any of this.

PR #1915, opened 2021-01-10
Update Contributing Docs

### LINKED ISSUE #1907: Update contributing
The correct way to run tests seems to be `tox -e py` (as run on CI) and not `python -m unittest` (that doesn’t even work due to parametrized not being in the Pipfile!) as advised currently. We should also ask contributors to run `tox -e fuzz` too maybe.

cooperlees:
- Update docs with all new tox hotness
- Test running docs build:
  - `sphinx-build -a -b html -W docs/ docs/_build/`

Fixes #1907


CONTRIBUTING.md
```diff
@@ -34,14 +34,39 @@ If you haven't used `pipenv` before but are comfortable with virtualenvs, just r
...
+# Optional CI run to test your changes on many popular python projects
 $ black-primer [-k -w /tmp/black_test_repos]
 ```
 
+### Docs Testing
+
+If you make changes to docs, you can test they still build locally too.
+
+```console
+$ pip install docs/requirements.txt
```
JelleZijlstra, line 66:
```suggestion
$ pip install -r docs/requirements.txt
```
(haven't tested)

CONTRIBUTING.md
```diff
@@ -52,7 +77,7 @@ your PR. You may need to change
 configuration for it to pass.
 
 For more `black-primer` information visit the
-[documentation](https://github.com/psf/black/blob/master/docs/black_primer.md#black-primer).
+[documentation](https://github.com/psf/black/blob/master/docs/black_primer.md#).
```
JelleZijlstra, line 80:
```suggestion
[documentation](https://github.com/psf/black/blob/master/docs/black_primer.md).
```

CONTRIBUTING.md
```diff
@@ -34,14 +34,40 @@ If you haven't used `pipenv` before but are comfortable with virtualenvs, just r
...
+# Optional Fuzz testing
+$ tox -e fuzz
+
+# Optional CI run to test your changes on many popular python projects
 $ black-primer [-k -w /tmp/black_test_repos]
 ```
 
+### Docs Testing
+
+If you make changes to docs, you can test they still build locally too.
```
ichard26, line 63: Technically this step is also needed when you make modifications to documentation files which are duplicated (for reasons I don't really know / understand). I honestly wish and might try to get these duplicated files and the regen step gone for good. It really doesn't matter if the duplicated content is up to date since they will updated before it get deploys on ReadTheDocs as a fresh build always starts, and IMO GitHub isn't the right place for documentation hosting. IIRC one of the reasons why we duplicate the docs is because it improves the GitHub documentation reading experience, but we have ReadTheDocs already, a platform *designed* for documentation.

ichard26 (approved):
I'm not happy with the current state of the documentation workflow and the myriad of different ways of setting up a dev environment and then running the tests, but that's beyond the scope of this PR. Beyond that LGTM.

CONTRIBUTING.md
cooperlees: I +1 this cleanup. I think it's not worth having the duplicates too.

I'm just keeping my PRs to address/solve one main problem with each, as you noted.

PR #1919, opened 2021-01-11
Fix Github action failing due to syntax error in entrypoint.sh

ThomasHagebols:
# Problem

Running `- uses: psf/black@master` Github action step throws the following error:

`/entrypoint.sh: 5: /entrypoint.sh: Syntax error: "(" unexpected`

Bug seems to be introduced in https://github.com/psf/black/pull/1909

# Steps to reproduce

The following examples reproduce the issue + logs

### Example 1

```yaml
name: Lint in python container

on:
  push

jobs:
  lint:
    runs-on: ubuntu-latest
    container: python:3.8
    steps:
      - uses: actions/checkout@v2
      - uses: psf/black@master
```
See [logs](https://github.com/ThomasHagebols/BlackActionFailDemo/runs/1680786909?check_suite_focus=true)

### Example 2

```yaml
name: Lint setup-python

on:
  push

jobs:
  lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: psf/black@master
```
See [logs](https://github.com/ThomasHagebols/BlackActionFailDemo/runs/1680786900?check_suite_focus=true)

# Proposed solution

Currently, entrypoint.sh starts with the `#!/bin/sh` shebang. This causes an error since some parts of the script are non compliant to POSIX. Switching to a bash interpreter fixes the issue.

# Other references
@rickstaa @cooperlees

### DIFF
--- a/action/entrypoint.sh
+++ b/action/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -e
 
 # If no arguments are given use current working directory


rickstaa:
@ThomasHagebols Thanks a lot for the Pull Request! Stupid I missed that my action from which I copied the code doesn't have the shebang and uses bash as the default shell. I will leave it up to @cooperlees to decide if he wants to merge this pull request and switch to using the bash shell or that he wants me to rewrite the code for the dash shell.

cooperlees (approved):
Happy to use bash as default here. GitHub CI should always have bash.

That said will take PR to change it to pure POSIX conformant sh if someone needs it.

PR #1925, opened 2021-01-13
Update link pointing to how-black-wraps-lines

allhailwesttexas:
The doc section about the Black code style has been moved to its own file. Update link on the compatible configs page to point to the right place.


ichard26 (approved):
Thanks mate!

PR #1940, opened 2021-01-22
Add `version` to github action (and rewrite the whole thing while at it)

stefanfoulis:
Allows setting the exact black version use with the github action.
This is useful to be able to avoid needing to deal with black formatting changes introduced in new black versions in unrelated pull-requests.


ichard26:
@stefanfoulis are you still working on this? No worries if you are no longer so, just checking in.

stefanfoulis:
@ichard26 I made some changes to make the tests pass. I suppose we should also add some tests, but I don't know how to test github actions with multiple configurations.

ichard26:
@stefanfoulis thanks! I'll take a look eventually (disclaimer: "eventually" may take a while!). For the testing, I usually just use a spare GitHub repository to test in (e.g. https://github.com/ichard26/testblackaction) but I've heard of https://github.com/nektos/act which allows you to run GHA workflows locally (can't say if it's any good since I haven't used it, but it looks promising at a few quick glances!).

action.yml
```diff
@@ -6,9 +6,20 @@ inputs:
     description: "Black input arguments."
     required: false
     default: ""
+  black_version:
+    description: 'Version of black to use. E.g "==20.8b1"'
```
cooperlees, line 10: Python Version specifier (PEP440) - e.g. "==20.8b1"

action.yml
```diff
@@ -6,9 +6,20 @@ inputs:
...
+    required: false
+    default: ""
 branding:
   color: "black"
   icon: "check-circle"
 runs:
-  using: "docker"
-  image: "action/Dockerfile"
+  using: composite
```
cooperlees, line 17: What base image do we use now?

cooperlees (changes requested):
Sorry about missing this. Lets rebase + fix the help message. I'm happy for this to have no change log since it does not break the main black program itself.

action.yml
ichard26: We aren't using any image here anymore, a composite action runs the steps directly in the GHA runner's environment. I'm a bit concerned this means Python has to be set up or else this will fail. I'll have to do some testing. 

Unfortunately it appears passing build variables during the action build is impossible currently :( https://github.community/t/feature-request-build-args-support-in-docker-container-actions/16846/10

ichard26:
FYI, I'll take over and push this PR to completion @stefanfoulis, thanks for the initial work!

ichard26:
I've hacking on this and I just realized that since now this is a composite action, dependency clashes are possible between Black's dependencies and whatever is installed by the rest of the containing workflow.

While a composite action does mean this action can be used on any OS and doesn't require a [hacky workaround](https://github.community/t/feature-request-build-args-support-in-docker-container-actions/16846/7) (and I don't even know if that   works) the fact build arguments aren't possible using GitHub Actions, the lack of isolation might be an issue (dependencies clashes, requires the host to have Python 3.6+ as the python pip is attached to).

@cooperlees WDYT?

cooperlees:
> I've hacking on this and I just realized that since now this is a composite action, dependency clashes are possible between Black's dependencies and whatever is installed by the rest of the containing workflow.

Make a venv for black so we can never have dependency clashes?

ichard26:
Example runs:
- successful ones (including deprecated `black_args` option): https://github.com/ichard26/testblackaction/actions/runs/888879560
- failing one due to invalid version: https://github.com/ichard26/testblackaction/actions/runs/888903332

action.yml
```diff
@@ -4,21 +4,55 @@ author: "Łukasz Langa and contributors to Black"
...
+  steps:
+    - run: |
+        # Exists since using github.action_path + path to main script doesn't work because bash
+        # interprets the bashslashes in github.action_path (which are used when the runner OS
```
JelleZijlstra, line 31:
```suggestion
        # interprets the backslashes in github.action_path (which are used when the runner OS
```
Or is this some new kind of slashes I don't know about

docs/integrations/github_actions.md
```diff
@@ -17,19 +25,26 @@ jobs:
...
-      - uses: actions/setup-python@v2
       - uses: psf/black@stable
 ```
 
 We recommend the use of the `@stable` tag, but per version tags also exist if you prefer
-that.
+that. Note that the action's version you select is independent of the version of _Black_
+the action will use.
+
+The version of _Black_ the action will use can configured via `version`. The action
```
JelleZijlstra, line 35:
```suggestion
The version of _Black_ the action will use can be configured via `version`. The action
```

action/main.py
```diff
@@ -0,0 +1,52 @@
...
+    stderr=STDOUT,
+    encoding="utf-8",
+)
+if pip_proc.returncode:
+    print(pip_proc.stdout)
+    print("::error::Failed to install Black.")
+    sys.exit(pip_proc.returncode)
+
+# print("::endgroup::")
+# print("::group:: Run Black")
```
cooperlees, line 39: Want to cleanup or maybe comment that this is good for debugging or future use?

action/main.py
```diff
@@ -0,0 +1,52 @@
...
+
+base_cmd = [str(ENV_BIN / "black")]
```
cooperlees, line 41: How does windows not need `.exe` here? It can work without these days? I remember for Windows I've had to add the .exe before ... Just ignore if you're confident it works:
https://github.com/facebookincubator/ptr/blob/master/ptr.py#L545

action/main.py
```diff
@@ -0,0 +1,52 @@
...
+if BLACK_ARGS:
+    # TODO: remove after a while since this is deprecated in favour of SRC + OPTIONS.
+    print(
+        "::warning::Input `with.black_args` is deprecated. Use `with.options` and `with.src` instead."
+    )
```
cooperlees, line 46: Do these have to go to stdout? Should we turn buffering off OR write to stderr?

docs/integrations/github_actions.md
```diff
@@ -3,6 +3,14 @@
 You can use _Black_ within a GitHub Actions workflow without setting your own Python
 environment. Great for enforcing that your code matches the _Black_ code style.
 
+## Compatiblity
+
+This action is known to support all GitHub-hosted runner OSes. In addition, only
+published versions of _Black_ are supported (i.e. whatever is available on PyPI).
+
+Finally, this action installs _Black_ with both the `colorama` and `python2` extras so
+the `--color` flag and formatting Python 2 code are supported.
```
cooperlees, line 12: Nice touch.

cooperlees (approved):
Nice work. Just some nit picks / suggestions for you to take or leave. Thanks for all the work here. Great move going to Python rather than using bash :D

action/main.py
ichard26: Definitely confident this works: https://github.com/ichard26/testblackaction/runs/2702059836?check_suite_focus=true.

docs/integrations/github_actions.md
ichard26: Thank you!

action/main.py
ichard26: Actually this isn't necessary at all since the action YAML file supports configurating a deprecation message directly. Will do that instead.

action/main.py
ichard26: They are pretty nice when the output isn't being mangled by a runner bug :( [example here](https://github.com/ichard26/testblackaction/runs/2606125162?check_suite_focus=true). I'm a fan of adding debug output when it's easy to hide them by default. I'll get rid of them since I honestly doubt the bug will be fixed anytime soon.

ichard26:
Thanks @JelleZijlstra and @cooperlees for the reviews, they are super appreciated :)

Hopefully this is ready to go now @cooperlees!

ichard26:
Thank you @stefanfoulis so much! While I may have been the person to push this to completion, without your initial work, I don't think we would've ever worked on this feature (or atleast for a quite a while!). I'm so happy with the end result, much better usability now. If you have any cycles to spare, it would be great if you could provide some feedback from your experience contributing to psf/black, more details here: #2238.

Thank you once again!

PR #1949, opened 2021-02-01
Consistency: bring default line length in-line with PEP8

### LINKED ISSUE #498: Consider switching line-length default to 79 chars, per PEP8
I've read [the justification](https://github.com/ambv/black#line-length) for using 88 chars as the default max-line-length in Black:

> This number was found to produce significantly shorter files than sticking with 80 (the most popular), or even 79 (used by the standard library). In general, [90-ish seems like the wise choice](https://youtu.be/wf-BqAjZb8M?t=260). If you're paid by the line of code you write, you can pass --line-length with a lower number.

I'd like to suggest that this argument isn't strong enough to justify deviating from a strong standard that exists in the Python community and beyond. Not only does PEP8, and therefore Flake8 and almost every existing Python project use 79 chars, but this is also the default in [Prettier](https://prettier.io/docs/en/options.html) and is [the basis for the 72-character limit in Git commits](https://medium.com/@preslavrachev/what-s-with-the-50-72-rule-8a906f61f09c). (And, conversely to the sarcastic addendum above - who cares if a file is a few lines longer in this day and age?)

I [proposed to Canonical's webteam](https://github.com/canonical-webteam/practices/issues/73#issuecomment-418130616) that we follow Black's standard of 88 chars, but our developers rejected the suggestion primarily based on that it's harder to view in their terminals - which is the reason for the 79 char limit in the first place. And so that we don't need to reconfigure all our Flake8-based CI for our projects immediately to be able to start using Black.

Since Black is meant to use sensible defaults and in most cases remain unconfigured, it seems a real shame that it deviates from PEP8 in just this one case, which will almost inevitably lead to a good number of projects that use it having just one config override for `--line-length=79` (as we're planning to do).

This is, of course, just my (or rather, our) 2 cents. Please feel free to make whatever choice you feel is right for the project.

jayaddison:
Update [`DEFAULT_LINE_LENGTH`](https://github.com/psf/black/blob/71117e730c4f62458b30af820f51890487b458e4/src/black/__init__.py#L71) to 79, bringing it in-line with PEP8.

As far as possible, existing test logic and test data is left as-is; only tests where new failures are introduced by the migration are updated.

Resolves #498

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -68,7 +68,7 @@ else:
 if TYPE_CHECKING:
     import colorama  # noqa: F401
 
-DEFAULT_LINE_LENGTH = 88
+DEFAULT_LINE_LENGTH = 79
 DEFAULT_EXCLUDES = r"/(\.direnv|\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|\.svn|_build|buck-out|build|dist)/"  # noqa: B950
 DEFAULT_INCLUDES = r"\.pyi?$"
 CACHE_DIR = Path(user_cache_dir("black", version=__version__))
--- a/src/black_primer/primer.json
+++ b/src/black_primer/primer.json
@@ -3,7 +3,7 @@
   "projects": {
     "aioexabgp": {
       "cli_arguments": [],
-      "expect_formatting_changes": false,
+      "expect_formatting_changes": true,
       "git_clone_url": "https://github.com/cooperlees/aioexabgp.git",
       "long_checkout": false,
       "py_versions": ["all"]
@@ -17,7 +17,7 @@
     },
     "bandersnatch": {
       "cli_arguments": [],
-      "expect_formatting_changes": false,
+      "expect_formatting_changes": true,
       "git_clone_url": "https://github.com/pypa/bandersnatch.git",
       "long_checkout": false,
       "py_versions": ["all"]
@@ -40,7 +40,7 @@
     },
     "flake8-bugbear": {
       "cli_arguments": [],
-      "expect_formatting_changes": false,
+      "expect_formatting_changes": true,
       "git_clone_url": "https://github.com/PyCQA/flake8-bugbear.git",
       "long_checkout": false,
       "py_versions": ["all"]
@@ -84,7 +84,7 @@
     },
     "ptr": {
       "cli_arguments": [],
-      "expect_formatting_changes": false,
+      "expect_formatting_changes": true,
       "git_clone_url": "https://github.com/facebookincubator/ptr.git",
       "long_checkout": false,
       "py_versions": ["all"]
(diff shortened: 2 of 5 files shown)


jayaddison:
In the `black-primer` output logs, the `hypothesis` and `pytest` projects have encountered failures:

For example ([build ref](https://github.com/psf/black/pull/1949/checks?check_run_id=1807097205)):
```
2021-02-01T14:07:23.1953653Z ## hypothesis:
2021-02-01T14:07:23.1954104Z  - Returned 123
...
2021-02-01T14:07:23.9948689Z ## pytest:
2021-02-01T14:07:23.9949649Z  - Returned 123
```

This didn't occur when running each of those projects individually in a local environment.  They would require reformatting under the updated line length default, but both already have the `expect_formatting_changes` flag set `true` in `primer.json`.

jayaddison:
Eh, actually.. maybe those failures did occur when running `black-primer` locally too.  Checking in more detail..

JelleZijlstra:
Thanks for your effort, but we're not going to do this: 88 is a reasoned choice (explained in the documentation) and changing the default line length would bring a lot of disruption to existing users of Black.

ichard26:
>  a lot of disruption to existing users of Black.

For context, `black-primer` output should (hopefully, I haven't used it personally) put it in perspective *how much* disruption we're talking about. Also, btw most projects don't have configuration on line length and rely on Black's default, _and_ we also state that configuration is not that important in docs so the scale of who will be affected is a bit brutal.

jayaddison:
No problem - I might crack on with it and some related improvements for a while anyway; yep, sure, it's a while since the time of punch-cards, but there's still some validity here while PEP8, the community and a decent number of computing environments could benefit from it.  No need to disrupt anything at the moment, but sometimes disruption for long-term benefit can be necessary.

jayaddison:
Perhaps this'd be considered a microcrime by a future society that's reached a completely console-dimensions-fluid state where everyone can count equally well in arbitrary bases, but for various reasons I think rounding up to 90 might be the way to go for us humans, if this does become a topic of conflict again within the next few years.

That'd hopefully reduce developer time required during code migration, since it'd require fewer clever refactorings and adjustments in order to update to an increased length limit, rather than to fit within a more restrictive one.  Anyway, a PEP seems like the place a suggestion like that should appear, so that's where I'll read up on any history and look out for future proposals.

(NB: was also glad to read that the `black` project has been aware of accessibility issues as well as developer conciseness issues - all good considerations to balance out and include discussion of)

jayaddison:
A softer, perhaps quixotic, approach could be to allocate a (preferred) piano-related alias for `black` during the unveil of a stable, non-beta release.

A piano has eighty-eight keys, of which thirty-six are black and fifty-two are white.  The project plays your code, reinterprets it a little, and provides you and your team with the pleasure of consistent code -- the 'sheet music' to play from.

The published project and binary name might best remain as-is, to avoid causing any disruption to existing users.

I don't think this idea is really very practical; wist rather than engineering going on, here :)

PR #1953, opened 2021-02-01
speed up cache by approximately 42x by avoiding pathlib

### LINKED ISSUE #1950: cache significantly slows down black due to pathlib

With a sufficiently large number of runs of black, the cache accumulates pretty rapidly and causes *significant* slowdown compared to uncached runs:

**uncached**
```console
$ rm -rf t; time XDG_CACHE_HOME=$PWD/t black t.py
reformatted t.py
All done! ✨ 🍰 ✨
1 file reformatted.

real	0m0.295s
user	0m0.237s
sys	0m0.036s
```

**cached**
```console
$ time black t.py
All done! ✨ 🍰 ✨
1 file left unchanged.

real	0m2.115s
user	0m1.999s
sys	0m0.089s
```

The cache in question isn't ~that large:

```console
$ du -hs ~/.cache/black/20.8b1/*.pickle
3.8M	/home/asottile/.cache/black/20.8b1/cache.-.88.1.0.pickle
16K	/home/asottile/.cache/black/20.8b1/Grammar3.8.5.final.0.pickle
4.0K	/home/asottile/.cache/black/20.8b1/PatternGrammar3.8.5.final.0.pickle
```

The reason for all the slowdown appears to be because pathlib is slow -- I've attached a pstats and svg of this:

profile graph

This was generated using [this procedure](https://youtu.be/ey_P64E34g0?list=PLWBKAf81pmOaP9naRiNAqug6EBnkPakvY)

```bash
python -m cProfile -o out.pstats -m black t.py
gprof2dot -z __init__:6604:patched_main out.pstats | dot -Tsvg -o out.svg  # gprof2dot from `yelp-gprof2dot`, `dot` from graphviz
```

[out.zip](https://github.com/psf/black/files/5904859/out.zip) (contains svg and pstats file)

The cache shouldn't make the execution significantly slower

- Version: [e.g. master] 20.8b1
- OS and Python version: [e.g. Linux/Python 3.7.4rc1] 3.8.5 on ubuntu 20.04

yes, if I build up a new cache using the current default branch it reproduces as well

Removing pathlib from the Cache makes the cache serialization cost ~essentially zero -- will follow up with a demonstration patch

asottile:
resolves #1950

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,8 @@
 
 - use lowercase hex strings (#1692)
 
+- speed up caching by avoiding pathlib (#1950)
+
 #### _Packaging_
 
 - Self-contained native _Black_ binaries are now provided for releases via GitHub
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -93,7 +93,7 @@ Transformer = Callable[["Line", Collection["Feature"]], Iterator["Line"]]
 Timestamp = float
 FileSize = int
 CacheInfo = Tuple[Timestamp, FileSize]
-Cache = Dict[Path, CacheInfo]
+Cache = Dict[str, CacheInfo]
 out = partial(click.secho, bold=True, err=True)
 err = partial(click.secho, fg="red", err=True)
 
@@ -724,7 +724,8 @@ def reformat_one(
             if write_back not in (WriteBack.DIFF, WriteBack.COLOR_DIFF):
                 cache = read_cache(mode)
                 res_src = src.resolve()
-                if res_src in cache and cache[res_src] == get_cache_info(res_src):
+                res_src_s = str(res_src)
+                if res_src_s in cache and cache[res_src_s] == get_cache_info(res_src):
                     changed = Changed.CACHED
             if changed is not Changed.CACHED and format_file_in_place(
                 src, fast=fast, write_back=write_back, mode=mode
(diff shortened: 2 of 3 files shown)


cooperlees (approved):
Thanks - Crazy increase. Expensive object this pathlib :D.

(will let some others look before merging)

JelleZijlstra (approved):

JelleZijlstra:
Looks like Travis is just never finishing, do we still need it?

JelleZijlstra:
And thanks for catching and fixing this!

ichard26:
> Looks like Travis is just never finishing, do we still need it?

Travis CI for Open Source is shutting down in a few weeks. I'm 99.99% sure we don't need it as our Test, Lint, Docs, Upload / Package, Primer, and Fuzz workflows are all on GitHub Actions. So even though we *can* migrate to the .com version with its 1000 free Linux minutes(?), I don't think we need to.

([more info here](https://blog.travis-ci.com/oss-announcement), [here](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing), and [also here](https://docs.travis-ci.com/user/migrate/open-source-repository-migration))

ichard26 (commented):
This most likely needs a change log entry, but other than that LGTM, thanks!

edit: maybe when cache operations are done per run rather than per file in another PR, we can add the changelog entry along the lines of "Reduce caching overhead when formatting large quantities of files" that covers both these PRs

asottile:
added the changelog entry

Issue #1900, opened 2021-01-02
 INTERNAL ERROR: Black produced different code on the second pass of the formatter.


ccie18643:
error: cannot format /root/PyTCP/fpp/icmp6.py: INTERNAL ERROR: Black produced different code on the second pass of the formatter.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_74vkg8wd.log

diff

```
(python3.8) root@ubuntu-1:~/PyTCP# cat /tmp/blk_74vkg8wd.log
Mode(target_versions=set(), line_length=160, string_normalization=True, experimental_string_processing=False, is_pyi=False)
--- source
+++ first pass
@@ -275,22 +275,34 @@
     def _read_nd_options(self, optr):
         """ Read ND options """
 
         nd_options = []
         while optr < len(self._frame):
-            nd_options.append({ps.icmp6.ND_OPT_SLLA: NdOptSLLA, ps.icmp6.ND_OPT_TLLA: NdOptTLLA, ps.icmp6.ND_OPT_PI: NdOptPI}.get(self._frame[optr], NdOptUnk)(self._frame, optr))
+            nd_options.append(
+                {ps.icmp6.ND_OPT_SLLA: NdOptSLLA, ps.icmp6.ND_OPT_TLLA: NdOptTLLA, ps.icmp6.ND_OPT_PI: NdOptPI}.get(self._frame[optr], NdOptUnk)(
+                    self._frame, optr
+                )
+            )
             optr += self._frame[optr + 1] << 3
 
         return nd_options
 
     @property
     def nd_options(self):
         """ Read ND options  """
 
         if self.__nd_options is self.__not_cached:
             assert self.type in {ps.icmp6.ROUTER_SOLICITATION, ps.icmp6.ROUTER_ADVERTISEMENT, ps.icmp6.NEIGHBOR_SOLICITATION, ps.icmp6.NEIGHBOR_ADVERTISEMENT}
-            optr = self._hptr + {ps.icmp6.ROUTER_SOLICITATION: 12, ps.icmp6.ROUTER_ADVERTISEMENT: 16, ps.icmp6.NEIGHBOR_SOLICITATION: 24, ps.icmp6.NEIGHBOR_ADVERTISEMENT: 24}[self.type]
+            optr = (
+                self._hptr
+                + {
+                    ps.icmp6.ROUTER_SOLICITATION: 12,
+                    ps.icmp6.ROUTER_ADVERTISEMENT: 16,
[...]

PR #1958, opened 2021-02-03
Stability fixup: address stability issues relating to interaction between magic commas and invisible parentheses

### LINKED ISSUE #1629: "INTERNAL ERROR: Black produced different code on the second pass of the formatter"
- Version: 20.8b0
- OS and Python version: all

This is a rare problem that we're currently investigating. The most common case of it has to do with a combination of magic trailing commas and optional parentheses. Long story short, there's this behavior:

- Black now treats all pre-existing trailing magic commas found in bracket pairs as signal to explode that bracket pair (e.g. to format it "one element per line");
- in other places Black inserts trailing commas itself after reformatting a long expression or statement;
- then on another run, Black finds those trailing commas and treats them as pre-existing trailing commas that signal that the given bracket pair should be exploded.

The expected behavior is that there should be no difference between the first formatting and the second formatting. In practice Black sometimes chooses for or against optional parentheses differently depending on whether the line should be exploded or not. This is what needs fixing.

## Workaround
We're working on fixing this, until then, format the file twice with `--fast`, it will keep its formatting moving forward.

## Call To Action
If you find a case of this, please attach the generated log here so we can investigate. We've already added three identifying examples of this as expected failures to https://github.com/psf/black/pull/1627/commits/25206d8cc6e98143f0b10bcbe9e8b41b8b543abe.

Finally, if you're interested in debugging this yourself, look for `should_explode` in `if` statements in the Black codebase. Those are the decisions that lead to unstable formatting.

### LINKED ISSUE #1954: INTERNAL ERROR: Black produced different code on the second pass of the formatter.
**Describe the bug**
`black` version 20.8b1 encounters difficulty reformatting some code lines that exceed the configured line length limit.

**To Reproduce**
[blk_zqosr6wa.log](https://github.com/psf/black/files/5906912/blk_zqosr6wa.log)

1. Take this Python code and place it in a file

```python
assert function(arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, arg10, arg11, arg12, arg13) != [None]
```

2. Run _Black_ on it with these arguments: `--line-length 79`
3. See error

It seems like `black` should be able to find a valid, stable reformatting for this input.

- Version: 20.8b1
- OS and Python version: Linux, Python 3.9.1

Yes.

jayaddison:
This changeset addresses some stability issues encountered during multi-pass code formatting by `black`.

In particular, the changeset modifies handling of magic trailing commas within code lines.  The most significant change is that the [`generate_trailers_to_omit`](https://github.com/psf/black/blob/aebd3c37b28bbc0183a58d13b80e7595db3c09bb/src/black/__init__.py#L5913-L5922) method will delay or suppress yielding of an empty trailer (leaf set) when a magic comma is present in the line.

May resolve #1629.

**NB**: Other time commitments may mean that I won't be committing much additional code here (or be able to re-investigate the details) within the next few weeks.  I'll do my best to respond to any feedback, suggestions and ideas; and if anyone's eager to, please feel free to take any of the code from this branch and develop / fix it further.

Edit 20210312: added note from https://github.com/psf/black/pull/1958#issuecomment-797605255


jayaddison:
@ambv Apologies to ping directly, you might receive a lot of notifications.  I think there's something of interest to you here though.

It's _almost_ possible to make the `maybe_make_parens_invisible_in_atom` method generalizable by replacing checks for `node.type == syms.atom` with `isinstance(node, Node)`.  At which point it should perhaps be renamed `maybe_make_parens_invisible_in_node`.

This would allow simplifying the `normalize_invisible_parens` loop and conditionals.

There's a catch somewhere, though.  It has to do with some of the `can_omit_invisible_parens` logic, and in particular it seems to have difficulty rewriting a line where a trailing comma has been added in the example test data provided in this pull request.

JelleZijlstra:
I was very excited when I saw the PR title; I think this change may also fix many of the issues mentioned in #1629.

jayaddison:
It seems similar, in that it does relate to the introduction/removal of invisible parentheses.

In the example I've been looking at locally for this branch, [the calls to `ensure_visible`](https://github.com/psf/black/blob/988c686d312986760a42cc7edd9a79a8f31c4769/src/black/__init__.py#L4934-L4935) seem to be where the parentheses are re-surfaced during the second formatting pass.

One question is whether invisible parens should have been introduced at all (they do seem valid for return and assert statements though, I think?); another is whether they could have been [valid for omission](https://github.com/psf/black/blob/988c686d312986760a42cc7edd9a79a8f31c4769/src/black/__init__.py#L6566-L6641).  Is is the latter case where the introduction of a trailing comma to the function argument list (`arg1, arg2` -> `arg1, arg2,`) leads to an unexpected change in the [`last` token here](https://github.com/psf/black/blob/988c686d312986760a42cc7edd9a79a8f31c4769/src/black/__init__.py#L6606-L6612).

jayaddison:
Thanks @JelleZijlstra; there's some useful reading in that issue and it may be that I'll want to close #1954 as a duplicate and attach the logs to the upstream issue.

jayaddison:
Current line of thinking: the magic trailing comma seems like it may be the proximate cause.

tests/data/function_trailing_comma.py
```diff
@@ -61,16 +62,19 @@ def f(
...
+            "a": 1,
+            "b": 2,
+            "c": 3,
+            "d": 4,
+            "e": 5,
+            "f": 6,
+            "g": 7,
+            "h": 8,
+        }["a"]
+    ):
```
jayaddison, line 77: It doesn't seem ideal to update a test expectation, so initially I'm considering this a workaround while reading up on some more of the background and context via #1619 and 586d24236e6b57bc3b5da85fdbe2563835021076.

jayaddison:
With some cautious optimism, and equal amounts of confusion (because I'm not yet clear on _why_ the changes here work, if they do) - this appears to resolve the problem cases tested so far from the linked issue.

Worth noting that there is also currently one non-ideal-looking test expectation update in this changeset.  If possible it would be nice to retain the existing behaviour there; it could be a symptom of a broader problem with the changes.

jayaddison:
Nope, I'm not really sure I've identified the root causes here, and think that this proposed changeset is a naive fix.  Back to reading more of the context behind the linked issue, and the code itself..

jayaddison:
NB: Although the `black-primer` tests are failing for `sqlalchemy`, from build log inspection I think the reformatting of the code that's being applied is more concise and an improvement (it largely relates to `assert` statements, and in particular omission of optional parentheses for those).

jayaddison:
The test case https://github.com/psf/black/blob/ab07b039472b3609fa72a853d441f07a2a4071f3/tests/data/invisible_parens_instability.py#L12-L16 can be solved by adding `syms.if_stmt` to the symbol list at https://github.com/psf/black/pull/1958/commits/21842f22573d8da956faae01835fbf6e6ea26ae0#diff-d23a6917d7262ca9815093bffb7148299e516575251959ed9b405b648e76aa29R5384.  That still doesn't feel like a complete or correct fix.

jayaddison:
Currently this changeset has worked backwards by reverting 586d24236e6b57bc3b5da85fdbe2563835021076.  Reverting that commit allows the suggested test cases in https://github.com/jayaddison/black/blob/ec1b0ab49918c0538bcbd18b0950c72849309248/tests/data/invisible_parens_instability.py to be formatted correctly -- but as a side-effect it re-exposes a separate bracket-matching issue that the commit was written to address.

Understanding the bracket matching issue and how that was resolved may help address the underlying formatter stability issues.

jayaddison:
Most tests may pass again after 58c3cce56550d747415e35a9aca3f35be6317d8e, but we're not quite done.  I'd like to restore some of the test coverage that was added in 586d24236e6b57bc3b5da85fdbe2563835021076 and make sure that the fix suggested here is genuine (partly by explaining the reasoning behind it and documenting that).

jayaddison:
Is the core of the problem that we need to perform a leaf-level split, rather than a line-level split?

Attempting the latter while trying to omit or retain special categories of leaf seems challenging.

jayaddison:
> Is the core of the problem that we need to perform a leaf-level split, rather than a line-level split?

Not necessarily, no.  The [`generate_trailers_to_omit`](https://github.com/psf/black/blob/aebd3c37b28bbc0183a58d13b80e7595db3c09bb/src/black/__init__.py#L5913-L5922) generator method should sequentially emit a series of 'line trailers' of items to be omitted during line split.  That provides leaf-level flexibility despite the fact that we are working with a single (logical) line.

src/black/__init__.py
```diff
@@ -2725,31 +2721,20 @@ def init_st(ST: Type[StringTransformer]) -> StringTransformer:
...
             # This mostly happens to multiline strings that are by definition
-            # reported as not fitting a single line, as well as lines that contain
-            # trailing commas (those have to be exploded).
-            yield from right_hand_split(
-                line, line_length=mode.line_length, features=features
-            )
+            # reported as not fitting a single line.
+            # line_length=1 here was historically a bug that somehow became a feature.
+            # See #762 and #781 for the full story.
+            yield from right_hand_split(line, line_length=1, features=features)
```
jayaddison, line 2737: Note for reviewers: this is one of the less appealing aspects of this changeset at the moment; it re-introduces the `line_length=1` bug/feature.

jayaddison:
(user-contributed tests from #1629 added 62b76ca3939e4b93394dd60b61c8175e9c56ad4 in have been reverted in 62b76ca3939e4b93394dd60b61c8175e9c56ad42; it's probably best to confirm consent and credit authorship of those before including them here)

jayaddison:
Apologies for the comment noise and stream-of-consciousness on this pull request; it's taken me a while to dig through the codebase and consider various approaches to the problem.  If it had been possible to separate the magic comma handling and assert-if-return statement logic, I would have done; they appear intertwined.

This pull request should now be ready for review and builds upon #1963.

src/black/__init__.py
```diff
@@ -6632,7 +6605,7 @@ def can_omit_invisible_parens(
             # unnecessary.
             return True
 
-        if line.should_explode and penultimate.type == token.COMMA:
+        if line.should_split and penultimate.type == token.COMMA:
```
jayaddison, line 6608: Just highlighting that there is something slightly unusual related to this condition.  The left-hand-side of this condition is removed in #1963.

tests/data/composition_no_trailing_comma.py
```diff
@@ -165,20 +165,17 @@ def tricky_asserts(self) -> None:
...
-                key3: value3,
-                key4: value4,
-                key5: value5,
-                key6: value6,
-                key7: value7,
-                key8: value8,
-                key9: value9
-            }
-        )
+        assert expectedexpectedexpectedexpectedexpectedexpectedexpectedexpectedexpect == {
```
jayaddison, line 168: This line exceeds the line length limit, so there is a problem to resolve here.

src/black/__init__.py
```diff
@@ -3425,13 +3429,7 @@ def do_transform(self, line: Line, string_idx: int) -> Iterator[TResult[Line]]:
 
         new_line = line.clone()
         new_line.comments = line.comments.copy()
-        try:
-            append_leaves(new_line, line, LL[: string_idx - 1])
-        except BracketMatchError:
-            # HACK: I believe there is currently a bug somewhere in
-            # right_hand_split() that is causing brackets to not be tracked
-            # properly by a shared BracketTracker.
```
jayaddison, line 3433: NB: This bug still remains, and surfaces when the lines https://github.com/psf/black/pull/1958/files#diff-d23a6917d7262ca9815093bffb7148299e516575251959ed9b405b648e76aa29R5383-R5384 are removed.

src/black/__init__.py
```diff
@@ -5382,6 +5380,8 @@ def normalize_invisible_parens(node: Node, parens_after: Set[str]) -> None:
                     node.insert_child(index, Leaf(token.LPAR, ""))
                     node.append_child(Leaf(token.RPAR, ""))
                 break
+            elif node.type in [syms.assert_stmt, syms.if_stmt]:
+                pass
```
jayaddison, line 5384: Further understanding required here; this `if...pass` path works 'somehow' because it prevents the introduction of unnecessary parentheses.

Open question: are those parentheses being added by the [RHS string-split bug](https://github.com/psf/black/pull/1958/files#r574026249)?  And how does that relate to the [`wrap_in_parentheses`](https://github.com/psf/black/pull/1958/files#diff-d23a6917d7262ca9815093bffb7148299e516575251959ed9b405b648e76aa29R5387) method call?

jayaddison: If we're going to continue to use a specific list of symbols during this conditional check, then I think the relevant Python grammar rule may be 'small statements' ([ref](https://docs.python.org/3/reference/grammar.html) - `small_stmt` rule).

jayaddison:
Although I've found some clues along the way, this is going to need further ideas and insight - I'd appreciate any help :)

tests/data/composition_no_trailing_comma.py
jayaddison: Going by the current [`black` code style](https://github.com/psf/black/blob/b8c1020b526b488a1a216d9a4d58fc8616b61a99/docs/the_black_code_style.md#line-length) documentation, I think it's OK for the line length to be exceeded, under some circumstances.  Whether this qualifies as an acceptable situation for that exception, I don't know and would be glad for feedback on.

src/black/__init__.py
jayaddison: This special case behaviour has been removed.

src/black/__init__.py
jayaddison: The change to re-introduce the `line_length=1` bug/feature has been un-applied.

src/black/__init__.py
jayaddison: (this conditional block has now been removed)

jayaddison:
NB: There have been quite a few reverts and partial reverts during development here.  If this does pass review, then to make future `git bisect` and history inspection easier, it might make sense to squash this down into a smaller set of commits prior to merge (either one commit, or a small number of atomic changes).  That's something I'd be glad to do here, although I think until review it's more transparent to keep the full commit history, even though it's messy.

JelleZijlstra:
@jayaddison that's fine, we'll just squash the commits during the merge process.

jayaddison:
Thanks for your help checking and linking existing issue reports to this PR, @mvolfik!

mvolfik:
You're welcome. However, while triaging, I found one file that doesn't work even after this patch: [whs/runekit@67ce199:runekit/app/view/browser_window.py](https://github.com/whs/runekit/blob/67ce199dd294d443af2e989324be2214f25d91d4/runekit/app/view/browser_window.py)

<details>
<summary>Black log</summary>

```patch
Mode(target_versions=set(), line_length=88, string_normalization=True, experimental_string_processing=False, is_pyi=False)
--- source
+++ first pass
@@ -112,14 +112,14 @@
         ):
             # FIXME: This doesn't really work - PySide2 doesn't have QWebEngineNotification
             self.browser.page().setFeaturePermission(
                 origin, feature, QWebEnginePage.PermissionGrantedByUser
             )
-        elif (
-            feature in (QWebEnginePage.DesktopVideoCapture, QWebEnginePage.DesktopAudioVideoCapture)
-            and self.app.has_permission('pixel')
-        ):
+        elif feature in (
+            QWebEnginePage.DesktopVideoCapture,
+            QWebEnginePage.DesktopAudioVideoCapture,
+        ) and self.app.has_permission("pixel"):
             self.browser.page().setFeaturePermission(
                 origin, feature, QWebEnginePage.PermissionGrantedByUser
             )
         else:
             self.browser.page().setFeaturePermission(
--- first pass
+++ second pass
@@ -112,14 +112,18 @@
         ):
             # FIXME: This doesn't really work - PySide2 doesn't have QWebEngineNotification
             self.browser.page().setFeaturePermission(
                 origin, feature, QWebEnginePage.PermissionGrantedByUser
             )
-        elif feature in (
-            QWebEnginePage.DesktopVideoCapture,
-            QWebEnginePage.DesktopAudioVideoCapture,
-        ) and self.app.has_permission("pixel"):
+        elif (
+            feature
+            in (
+                QWebEnginePage.DesktopVideoCapture,
+                QWebEnginePage.DesktopAudioVideoCapture,
+            )
+            and self.app.has_permission("pixel")
+        ):
             self.browser.page().setFeaturePermission(
                 origin, feature, QWebEnginePage.PermissionGrantedByUser
             )
         else:
             self.browser.page().setFeaturePermission(
```
</details>

mvolfik:
Oh sorry, nevermind, I got lost in the black installation virtualenvs. That file works too

jayaddison:
No worries, thank you again - both successful confirmations, and remaining failure cases are good to know about.

I do have a remaining concern that although the changes appear to work in most cases, they also allow exceeding the line-length in some situations, and that might or might not be acceptable to the project team.  We'll see.

mvolfik:
Ok, this time I really got something:

```python3
def f():
    def f():
        def f():
            optr = self._hptr + {ps.icmp6.ROUTER_SOLICITATION: 12, ps.icmp6.ROUTER_ADVERTISEMENT: 16, ps.icmp6.NEIGHBOR_SOLICITATION: 24, ps.icmp6.NEIGHBOR_ADVERTISEMENT: 24}[self.type]
```

<details>
<summary>Black log, identical on master and this patch</summary>

```patch
Mode(target_versions=set(), line_length=88, string_normalization=True, magic_trailing_comma=True, experimental_string_processing=False, is_pyi=False)
--- source
+++ first pass
@@ -1,4 +1,12 @@
 def f():
     def f():
         def f():
-            optr = self._hptr + {ps.icmp6.ROUTER_SOLICITATION: 12, ps.icmp6.ROUTER_ADVERTISEMENT: 16, ps.icmp6.NEIGHBOR_SOLICITATION: 24, ps.icmp6.NEIGHBOR_ADVERTISEMENT: 24}[self.type]
+            optr = (
+                self._hptr
+                + {
+                    ps.icmp6.ROUTER_SOLICITATION: 12,
+                    ps.icmp6.ROUTER_ADVERTISEMENT: 16,
+                    ps.icmp6.NEIGHBOR_SOLICITATION: 24,
+                    ps.icmp6.NEIGHBOR_ADVERTISEMENT: 24,
+                }[self.type]
+            )
--- first pass
+++ second pass
@@ -1,12 +1,9 @@
 def f():
     def f():
         def f():
-            optr = (
-                self._hptr
-                + {
-                    ps.icmp6.ROUTER_SOLICITATION: 12,
-                    ps.icmp6.ROUTER_ADVERTISEMENT: 16,
-                    ps.icmp6.NEIGHBOR_SOLICITATION: 24,
-                    ps.icmp6.NEIGHBOR_ADVERTISEMENT: 24,
-                }[self.type]
-            )
+            optr = self._hptr + {
+                ps.icmp6.ROUTER_SOLICITATION: 12,
+                ps.icmp6.ROUTER_ADVERTISEMENT: 16,
+                ps.icmp6.NEIGHBOR_SOLICITATION: 24,
+                ps.icmp6.NEIGHBOR_ADVERTISEMENT: 24,
+            }[self.type]
```
</details>

From #1900

jayaddison:
Good find, thanks @mvolfik.  I'll look into that case soon.

jayaddison:
@mvolfik Just a note to let you know that I've taken a brief look at the test case you found - it's definitely valid, and I think it's possible that it exposes some kind of serious/core problem with this branch as it stands.  I've experimented with a few adjustments to try to handle it, but overall I still don't yet fully understand the root problem in the parentheses/magic comma interaction, and I think that will be required for a complete fix.

jayaddison:
Other time commitments may mean that I won't be committing much additional code here (or be able to re-investigate the details) within the next few weeks.  I'll do my best to respond to any feedback, suggestions and ideas; and if anyone's eager to, please feel free to take any of the code from this branch and develop / fix it further.

ambv:
@jayaddison, thanks for the tremendous amount of work you did on this PR. It only goes to show that the optional parentheses + optional trailing commas + magic trailing commas are a killer combination and it's impractical to solve the instability entirely, at least without introducing disruptive formatting changes in the process.

I decided to go with #2126 a simpler, if less elegant, approach to this problem.

jayaddison:
Glad to have helped, @ambv; it was a good learning exercise - I think my favourite finding as a result was discovering the term [`trivia`](https://basarat.gitbook.io/typescript/overview/ast/ast-trivia) in relation to ASTs.

I would definitely concur that something that becomes this puzzling is a difficult combination; if there's a way to update the codebase to simplify some of that in future -- potentially by reverting and backtracking on a few features like the ones you mention -- perhaps that'd help improve long-term maintainability.

PR #1959, opened 2021-02-03
Add gedit integration

Architrixs:
I don't know if anyone else need this, I find it useful.

### DIFF
--- a/docs/editor_integration.md
+++ b/docs/editor_integration.md
@@ -233,6 +233,32 @@ If you later want to update _Black_, you should do it like this:
 $ pip install -U black --no-binary regex,typed-ast
 ```
 
+## Gedit
+gedit is the default text editor of the GNOME, Unix like Operating Systems.
+Open gedit as
+```console
+$ gedit <file_name>
+```
+1. `Go to edit > preferences > plugins`
+2.  Search for `external tools` and activate it.
+3.  In `Tools menu -> Manage external tools`
+4.  Add a new tool using `+` button.
+5.  Copy the below content to the code window.
+```console
+#!/bin/bash
+Name = $GEDIT_CURRENT_DOCUMENT_NAME
+black $Name
+```
+- Set a keyboard shortcut if you like, Ex. `ctrl-B`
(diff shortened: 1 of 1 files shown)


docs/editor_integration.md
```diff
@@ -233,6 +233,32 @@ If you later want to update _Black_, you should do it like this:
...
+```
+1. `Go to edit > preferences > plugins`
+2.  Search for `external tools` and activate it.
+3.  In `Tools menu -> Manage external tools`
+4.  Add a new tool using `+` button.
+5.  Copy the below content to the code window.
+```console
+#!/bin/bash
+Name = $GEDIT_CURRENT_DOCUMENT_NAME
+black $Name
```
ichard26, line 250: I followed your steps (which are easy to follow btw!) and everything worked except for the actual bash script :(
Screenshot from 2021-02-03 19-06-23

Skipping the "saving to a variable" step and directly providing the document path to black worked tho.

docs/editor_integration.md
```diff
@@ -233,6 +233,32 @@ If you later want to update _Black_, you should do it like this:
...
+```
+- Set a keyboard shortcut if you like, Ex. `ctrl-B`
+- Save: `Nothing`
```
ichard26, line 253: I really hope that this doesn't turn into a source of "I JUST LOST A BUNCH OF MY WORK AND I BLAME YOU" type of bug reports, but I don't use gedit so I have no idea.

ichard26 (changes requested):
Please run `pre-commit run prettier -a` locally, stage, commit (or amend), and push since the Prettier pre-commit hook is failing right now. I also have a few concerns about the configuration you've provided.

Thanks for submitting an integrations PR! It really helps out the project. Feel free to add yourself to the Authors list in the README.

docs/editor_integration.md
Architrixs: Alright, thanks for looking into this..
fixed the variable thing... find that I put a space between 'Name' and '=', in the docs while formatting.
Again added a pull request,

Issue #1960, opened 2021-02-03
Fuzzer Discovery: issue relating to multi-line strings and newlines at the EOF


jayaddison:
**Describe the bug**
The [`fuzz.py`](https://github.com/psf/black/blob/988c686d312986760a42cc7edd9a79a8f31c4769/fuzz.py) script discovered the following error during [this build](https://github.com/psf/black/pull/1958/checks?check_run_id=1826290971):

```
Traceback (most recent call last):
  File "/home/runner/work/black/black/fuzz.py", line 60, in <module>
    test_idempotent_any_syntatically_valid_python()
  File "/home/runner/work/black/black/fuzz.py", line 17, in test_idempotent_any_syntatically_valid_python
    max_examples=1000,  # roughly 1k tests/minute, or half that under coverage
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesis/core.py", line 1158, in wrapped_test
    raise the_error_hypothesis_found
  File "/home/runner/work/black/black/fuzz.py", line 43, in test_idempotent_any_syntatically_valid_python
    dst_contents = black.format_str(src_contents, mode=mode)
  File "/home/runner/work/black/black/src/black/__init__.py", line 1025, in format_str
    src_node = lib2to3_parse(src_contents.lstrip(), mode.target_versions)
  File "/home/runner/work/black/black/src/black/__init__.py", line 1121, in lib2to3_parse
    result = drv.parse_string(src_txt, True)
  File "/home/runner/work/black/black/src/blib2to3/pgen2/driver.py", line 142, in parse_string
    return self.parse_tokens(tokens, debug)
  File "/home/runner/work/black/black/src/blib2to3/pgen2/driver.py", line 70, in parse_tokens
    for quintuple in tokens:
  File "/home/runner/work/black/black/src/blib2to3/pgen2/tokenize.py", line 542, in generate_tokens
    raise TokenError("EOF in multi-line statement", (lnum, 0))
blib2to3.pgen2.tokenize.TokenError: ('EOF in multi-line statement', (2, 0))
Highest target scores:
               8  (label='(hypothesmith from_node) number of unique ast node types')
              22  (label='(hypothesmith) number of unique ast node types')
              63  (label='(hypothesmith) instructions in bytecode')
             114  (label='(hypothesmith) total number of ast nodes')
             172  (label='(hypothesmith from_node) instructions in bytecode')
             173  (label='(hypothesmith from_node) total number of ast nodes')

[...]

Issue #1746, opened 2020-10-07
Continue src/black refactoring.


hadialqattan:
The entire `Black` source code has written in `src/black/__init__.py` which leads to poor maintainability!

hadialqattan:
In other words, it's not easy to start contributing to black, because you can't work only with the bug related part!

hugovk:
There was a plan to refactor black.py into a black/ Python package at some point: https://github.com/psf/black/issues/1350.

At least the first bit has been done, to refactor into a src/ directory: https://github.com/psf/black/pull/1376.

ichard26:
Perhaps #1350 should be reopened then?

hadialqattan:
Would you like to rename it?

hadialqattan:
Could anyone put some useful labels like `design`, `enhancement`, and `stable`, please?

felix-hilden:
Hello, I'd like to ask about the status of this issue and the state of the development in general.

- I agree that having such a code base as a single file is suboptimal. It seems that at least ambv agreed in #1350, and as a result a plan was created, and Black was refactored under src/black/...  The plan seems to have been executed fully, as it has no content beyond the basic directory structure. Has there been any concrete discussion about splitting the main source further down? Would now be a time to do start planning it?
- Existing PRs will surely have to be modified, if such a refactoring is done. What do the maintainers think about the priority of existing PRs, however stale they may be?
- What would the refactored structure look like? Do you have any ready ideas or requirements for the split?

I'm interested in contributing, because I feel like having a better structured code base could expedite all other development efforts as well. So I'd like to share my initial thoughts, though I recognise that I have no authority here.

- I see some groups of definitions that could be made into modules, like the ones involving toml and source finding, functions used by `normalize_fmt_off` or `normalize_invisible_parens`
- It seems that there are many constant operators, symbols and priorities that are either used in many places like `STANDALONE_COMMENT`, or in a single function like the various priorities in `is_split_before_delimiter`. They could be grouped into a single file, or the files where the functions that use them will end up.
- There are a couple of functions like `normalize_numeric_literal` (also to the first point) that have been broken down, whose components are only used in that function. They could be separated into submodules, perhaps simultaneously looking for similar functionality like `fix_docstring` that is also used by `LineGenerator`.
- There are bottom-level structures such as `Feature`, `TargetVersion`, `Mode` and `WriteBack` that could similarly be pulled away.
- Exceptions?

In general I think there are two ways of doing this:
[...]

PR #1961, opened 2021-02-04
Tokenization: improve EOF and multi-line statement interaction

### LINKED ISSUE #1012: Black fails to tokenise files ending with a backslash

jayaddison:
This is the kind of change that can cause subtle problems in future, so it's one that should be reviewed and considered carefully.

Currently during fuzz testing, `hypothesmith` is able to find some example cases where apparently-valid Python code that contains a multi-line indicator (`\` character) immediately prior to the EOF indicator (without a newline) fails during `black`'s tokenization stage.

This has been reported in #1012 and more recently #1960.

As per the notes in #1012, using code of this kind as input will successfully compile and run in Python.

This changeset permits the `black` tokenizer to continue when an EOF token is encountered following a multi-line indicator.

Resolves #1012.


JelleZijlstra (approved):

jayaddison:
Maybe worth a little more pause here; looking at the current upstream `blib2to3`, this `TokenError` exception is still raised when an EOF is encountered during a continued statement ([ref](https://github.com/python/cpython/blob/3c3aa4516c70753de06bb142b6793d01330fcf0f/Lib/lib2to3/pgen2/tokenize.py#L441-L444)).

jayaddison:
Eh, ok - I'm going to mark this as ready for review again, having read up a bit more on the status of `blib2to3` via bugs.python.org and understanding that `black`'s vendored version has already diverged a little bit.  Still feels maybe non-ideal, but I don't have any other changes planned here.

ichard26:
> Eh, ok - I'm going to mark this as ready for review again, having read up a bit more on the status of blib2to3 via bugs.python.org and understanding that black's vendored version has already diverged a little bit. Still feels maybe non-ideal, but I don't have any other changes planned here.

We've modified our version of `lib2to3` a lot already so this isn't a big deal.

ichard26 (changes requested):
Actually within a minute I created something that seems to hang Black?

```python
test = (1
        +
        2
```

This is definitely a no-no even if it's invalid code (EOF in multi-line continuation) IMO.

ichard26:
Yeah, this seems to be hanging up tokenize.py generate_tokens... cProfile-ing Black on that file and killing it after 30 seconds of nothing leads to this (partial) call graph:
out

jayaddison:
Thanks @ichard26 - I'll take a look into that.

jayaddison:
The latest pushed changes may appear to work based on the test coverage, but in fact there are code output equivalence issues.  The following input file (with no closing end-of-line) produces a different abstract syntax tree during the second formatter pass:

```
def foo(a):
    pass\
```

(it works with the `--fast` CLI flag provided, since that skips the second pass)

jayaddison:
I think this is closable as a wont-fix; I'll add some more explanation soon in the linked issue.

jayaddison:
(in hindsight, assuming a wont-fix resolution wasn't sensible of me; this could still be addressed if there's appetite to)

jayaddison:
(just cleaning up and closing some old / stale pull requests that I've been working on.  this turned out to be a more-involved fix than it looked like originally; some adjustments were made to the `hyptothesmith` config in #1998 to mitigate the test failures that were occurring as a result)

ichard26:
@jayaddison and we really appreciate it btw! We want to refactor the single-file Black package into a multi-module package to improve maintainability (see GH-1746). Unfortunately such a change would break basically every PR, so we're working on reducing the PR backlog.

PR #1963, opened 2021-02-04
Refactor: distinguish variable responsibilities

jayaddison:
This is largely an investigation aid for #1958.  Separating the concerns of different variables may help discover the cause of the underlying problem(s).

NB: It's not ideal that the `line.magic_trailing_comma` variable name here shadows the `mode.magic_trailing_comma` variable name.  They do have different types however.


src/black/__init__.py
```diff
@@ -6632,7 +6634,7 @@ def can_omit_invisible_parens(
             # unnecessary.
             return True
 
-        if line.should_explode and penultimate.type == token.COMMA:
+        if line.magic_trailing_comma and penultimate.type == token.COMMA:
```
jayaddison, line 6637: It may be possible to remove this conditional check block completely, with only a minor impact on the formatter's behaviour.  See https://github.com/jayaddison/black/commit/1096bc908d56a6013c82f626f08f2915a38ecd32 for a possible change (cc @JelleZijlstra)

I'm not including that commit here yet, since I think it'd be good for this refactoring-focused pull request to maintain near-zero change in formatter behaviour.  In practice I think there are small logical changes here, but I hope they're extremely minimal.

jayaddison: This commit has been included into #1958, so it can be skipped here.

ambv:
Thanks! ✨ 🍰 ✨

jayaddison:
No worries!  I _hope_ this is all fairly safe; there are definitely some small logical differences introduced here, and the scope for potential future impact from any change does worry me a bit.  I'll watch out for any resulting problems, anyway.

TBBle:
Per https://github.com/psf/black/pull/1897#issuecomment-783193898, it looks like an update to _/docs/reference/reference_functions.rst_ which broke the doc-gen job (which is annoyingly not run as part of CI for PRs).

jayaddison:
D'oh, thanks @TBBle - I'll look into fixing that up.

PR #1916, opened 2021-01-11
Add `black` Dockerfile

cooperlees:
- Build a `black` container based on python3-slim
- Test: `docker build --network-host --tag black .`
```console
cooper-mbp1:black cooper$ docker image ls
REPOSITORY                                         TAG            IMAGE ID       CREATED              SIZE
black                                              latest         0c7636402ac2   4 seconds ago        332MB
```

Addresses part of #1914

PR #1986, opened 2021-02-13
Fix duplication of checks on internal PRs

ichard26:
Internal PRs (that come from branches of psf/black) match both the push
and pull_request events. This leads to a duplication of test, lint, etc.
checks on such PRs. Not only is it a waste of resources, it makes the
Status Checks UI more frustrating to go through

Patch taken from:
https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012

--- 

example PR with duplicated workflows: https://github.com/psf/black/pull/1916
example (internal) PR with this patch: https://github.com/ichard26/black/pull/1

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -16,8 +16,14 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


cooperlees (approved):
LGTM

PR #1991, opened 2021-02-17
fuzzer: add special-case for multi-line EOF TokenError

### LINKED ISSUE #1012: Black fails to tokenise files ending with a backslash

jayaddison:
Goal: reduces test failure false-positives relating to #1012.

cc @Zac-HD

### DIFF
--- a/fuzz.py
+++ b/fuzz.py
@@ -5,6 +5,8 @@ generation.  You can run this file with `python`, `pytest`, or (soon)
 a coverage-guided fuzzer I'm working on.
 """
 
+import re
+
 import hypothesmith
 from hypothesis import HealthCheck, given, settings, strategies as st
 
@@ -48,15 +50,15 @@ def test_idempotent_any_syntatically_valid_python(
         # TODO: remove this try-except block when issues are resolved.
         return
     except TokenError as e:
-        # The Python3.7 'compile' method accepts input that terminates with a
-        # multi-line continuation character (\), but black rejects it.
-        # TODO: remove this special-case once py37 reaches end-of-life
+        newlines = r"(\n|\r\n)+"
+        backslash = r"\\"
         exception_message = e.args[0]
-        if (
-            src_contents.endswith("\\")
+        if (  # https://github.com/psf/black/issues/1012
+            re.findall(f"{newlines}{backslash}{newlines}", src_contents)
             and exception_message == "EOF in multi-line statement"
         ):
             return
+        raise
 
     # And check that we got equivalent and stable output.
     black.assert_equivalent(src_contents, dst_contents)
--- a/fuzz.py
+++ b/fuzz.py
@@ -11,9 +11,7 @@ import hypothesmith
 from hypothesis import HealthCheck, given, settings, strategies as st
 
 import black
-from blib2to3.pgen2.tokenize import (  # NB: This intends to (and currently does) import the black-vendored blib2to3
-    TokenError,
-)
+from blib2to3.pgen2.tokenize import TokenError
 
 
 # This test uses the Hypothesis and Hypothesmith libraries to generate random
--- a/fuzz.py
+++ b/fuzz.py
@@ -54,10 +54,11 @@ def test_idempotent_any_syntatically_valid_python(
             e.args[0] == "EOF in multi-line statement"
             and re.search(r"\r?\n\\\r?\n", src_contents) is not None
         ):
-            # A trailing backslash at EOF was accepted in Python <= 3.7, but a
-            # SyntaxError in later versions.  We've decided to just ignore this
-            # until the problem goes away when 3.7 reaches end of life in 2023.
+            # At the time of writing, backslash-related tokenization errors occur
+            # relatively frequently when this fuzzer is run.  Once fixed upstream,
+            # we can remove this special-case test-pass logic.
             # See https://github.com/psf/black/issues/1012
+            # See TODO: link to upstream issue report
             return
         raise
 
--- a/fuzz.py
+++ b/fuzz.py
@@ -50,16 +50,15 @@ def test_idempotent_any_syntatically_valid_python(
         # TODO: remove this try-except block when issues are resolved.
         return
     except TokenError as e:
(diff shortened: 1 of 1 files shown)


fuzz.py
```diff
@@ -46,6 +47,16 @@ def test_idempotent_any_syntatically_valid_python(
         # able to cope with it.  See issues #970, #1012, #1358, and #1557.
         # TODO: remove this try-except block when issues are resolved.
         return
+    except TokenError as e:
+        # The Python3.7 'compile' method accepts input that terminates with a
+        # multi-line continuation character (\), but black rejects it.
+        # TODO: remove this special-case once py37 reaches end-of-life
+        exception_message = e.args[0]
+        if (
+            src_contents.endswith("\\")
```
jayaddison, line 56: Hrm.  Should probably also check that we are currently in a Python 3.7 environment here.

jayaddison, line 56: I don't think this condition (`endswith.("\\")`) is perfect.

This build would have still failed with that condition in place: https://github.com/psf/black/runs/1904676745

jayaddison: As would https://github.com/psf/black/runs/1898084374, which is in fact a Python 3.8 build.

fuzz.py
```diff
@@ -46,6 +49,19 @@ def test_idempotent_any_syntatically_valid_python(
...
+        newlines = r"(\n|\r\n)+"
+        backslash = r"\\"
+        exception_message = e.args[0]
+        if (  # https://github.com/psf/black/issues/1012
+            re.findall(f"{newlines}{backslash}{newlines}", src_contents)
+            and exception_message == "EOF in multi-line statement"
+        ):
+            print('Warning: skipped fuzzer-generated input that contained a backslash')
+            print('See https://github.com/psf/black/issues/1012 for details')
+            print('To replay this fuzzer run, find the PYTHONHASHSEED environment value')
```
jayaddison, line 62: @Zac-HD Could you confirm whether this is reasonable advice?

Zac-HD, line 62:
```suggestion
        if (
            sys.version_info[:2] <= (3, 7)
            and e.args[0] == "EOF in multi-line statement"
            and re.search(r"\r?\n\\\r?\n", src_contents) is not None
        ):
            # A trailing backslash at EOF was accepted in Python <= 3.7, but a
            # SyntaxError in later versions.  We've decided to just ignore this
            # until the problem goes away when 3.7 reaches end of life in 2023.
            # See https://github.com/psf/black/issues/1012
```

I'd simplify the regex logic, add a check for Python version, and have a code comment rather than printing advice - it's very unlikely that anyone would want to reproduce this behaviour, and it would make the logs very noisy if running for a long time.

(and thanks to the `derandomize=True` setting you'd only need to re-run the test)

jayaddison: That sounds good; I'll include most of this soon.

One challenge though: the EOF multi-line exception can occur for other versions of Python after all.

In a previous note I mentioned that Python 3.8+ do reject the "\\" (only a backslash) input - and that's true, but Python3.8 will also compile `'#\n\\\r\n'`, for example, which `black` rejects with the EOF multi-line `TokenError`.

Zac-HD:
> One challenge though: the EOF multi-line exception can occur for other versions of Python after all.
> 
> In a previous note I mentioned that Python 3.8+ do reject the "\" (only a backslash) input - and that's true, but Python3.8 will also compile `'#\n\\\r\n'`, for example, which `black` rejects with the EOF multi-line `TokenError`.

Ah.  If we're not just waiting out the end-of-life for Python 3.7, I think there's an actual bug which should be fixed.

Concretely for this PR though, I'd just remove the version-check and change the comment to explain that we're suppressing a known bug and will fix it later.

jayaddison:
Is there an existing issue that we can link to here, or perhaps we should create one?

(my thinking is that we should add a note/reference upstream to remind us to clean this workaround up once the root cause is resolved)

Zac-HD:
I think the bug is in Black - Hypothesmith is generating source code which is valid according to the compile() builtin, and if Black can't handle it the bug is in Black's tokenisation, not upstream. 

If you think it *is* upstream, where is it and how should it be fixed?

jayaddison:
Ok, yep.  It does follow logically that this is most likely a bug in `black`.  I'll update the explanatory comment accordingly.

fuzz.py
```diff
@@ -5,10 +5,15 @@
 a coverage-guided fuzzer I'm working on.
 """
 
+import re
+
 import hypothesmith
 from hypothesis import HealthCheck, given, settings, strategies as st
 
 import black
+from blib2to3.pgen2.tokenize import (  # NB: This intends to (and currently does) import the black-vendored blib2to3
```
jayaddison, line 14: I added this comment based on a memory of reading some code that variously imported either the `blib2to3` code or Python's built-in `lib2to3` code.  But after checking it again, I realize they use different module names, so I think this comment is redundant and can be removed.

ichard26 (approved):

Zac-HD (approved):
Looks good to me - thanks, @jayaddison!

jayaddison:
Thanks @ichard26, @Zac-HD!

cooperlees:
This is awesome work all. @jayaddison - Appreciate all your efforts, I know I haven't been around much. But appreciate the deep debugging and pulling these PRs together!

jayaddison:
No worries, thanks @cooperlees!

PR #1997, opened 2021-02-22
Fixup: update function name in docs to match source

### LINKED ISSUE #1996: Continuous integration: documentation generation failure due to mismatched reference function name
**Describe the bug**
As [spotted by @TBBle](https://github.com/psf/black/pull/1897#issuecomment-783193898), the 'build documentation' continuous integration step has been [failing with an `AttributeError`](https://github.com/psf/black/runs/1949719732#step:5:68); a mismatched function name in the reference documentation is the cause. 

**To Reproduce**
1. Observe the build output in continuous integration after commit afc8c326bf97ce7a22e8fc0b8cc5ced1586ef022 pending a fix (note that this currently only runs following documentation-related edits)

**Expected behavior**
1. The documentation build should succeed

Continuous Integration via GitHub Actions

Yes

jayaddison:
Considerations to bear in mind when deciding whether to rename this function:

- It has recently been renamed (neither a plus nor minus for renaming again, necessarily)
- It is a public method, by naming convention (makes renaming it slightly less attractive)
- It refers in a strict-ish sense to right-hand-side line splits (so perhaps it could be renamed further; this is mitigated by a variable rename)

Resolves #1996

### DIFF
--- a/docs/reference/reference_functions.rst
+++ b/docs/reference/reference_functions.rst
@@ -171,7 +171,7 @@ Utilities
 
 .. autofunction:: black.re_compile_maybe_verbose
 
-.. autofunction:: black.should_split_body_explode
+.. autofunction:: black.should_split_line
 
 .. autofunction:: black.shutdown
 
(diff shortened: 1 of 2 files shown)


ichard26:
I mean, I'm not approving this, but since we don't provide a stable Python API, modifying internals like this is fair game IMO.

JelleZijlstra:
Sounds like we should at least do this to unbreak the docs build?

ichard26:
> Sounds like we should at least do this to unbreak the docs build?

Yeah I agree on second thought, somehow I ignored that was failing and this is the fix.

Issue #1947, opened 2021-01-28
On lines chaining bitwise operators, Black doesn't try to reduce the line length


tpilewicz:
**Describe the bug**

I use Airflow for scheduling jobs. The syntax for defining dependencies between tasks is the following:

```
task_1 >> task_2 >> task_3
```

If the names of my tasks are very long, and it crosses the line length limit, black will not try to fix it. Worse: actually if I span my dependencies on several lines with backslashes, black will all put them on one line, causing the line length limit to be crossed. The workaround is to span the dependencies on several lines with parenthesis instead of backslashes, and black will not undo it. But I would expect black to even do it for me.

**To Reproduce** Steps to reproduce the behavior:

1. Take these files:

File 1:

```
loooooooooooooooooooooooooooooooooooooong_task_name = 0
looooooooooooooooooooooooooooong_task_name = 1
loooooooooooong_task_name = 2
loooooooooooooooooooooooooooooooooooooong_task_name >> looooooooooooooooooooooooooooong_task_name >> loooooooooooong_task_name
```

and File 2:

```
loooooooooooooooooooooooooooooooooooooong_task_name = 0
looooooooooooooooooooooooooooong_task_name = 1
loooooooooooong_task_name = 2
loooooooooooooooooooooooooooooooooooooong_task_name \
    >> looooooooooooooooooooooooooooong_task_name \
    >> loooooooooooong_task_name
```

2. Run _Black_ on them with no arguments
3. Result:

File 1 is unchanged. File 2 is formatted into exactly File 1.

**Expected behavior**

File 1 should be formatted so as not to cross the line length limit. For example like this:

```
loooooooooooooooooooooooooooooooooooooong_task_name = 0
looooooooooooooooooooooooooooong_task_name = 1
loooooooooooong_task_name = 2
(
    loooooooooooooooooooooooooooooooooooooong_task_name
    >> looooooooooooooooooooooooooooong_task_name
    >> loooooooooooong_task_name
)
```

File 2 should either be left untouched, or if trailing backslashes are bad, it should be formatted into the example given right above.

**Environment (please complete the following information):**

- Version: 20.8b1
- OS and Python version: Mac / python 3.6.8

**Does this bug also happen on master?** To answer this, you have two options:

yes (checked with the online playground)

tpilewicz:
Hello, I would like to add that is this is recognized as an issue, I'd be pleased to try and fix it.

JelleZijlstra:
[...]

PR #2001, opened 2021-02-23
Wrap arithmetic and binary arithmetic expressions in invisible parentheses

tpilewicz:
This addresses issue https://github.com/psf/black/issues/1947

The goal is that this:

```
looooooooooooooooooooooooooooooooooooooooooooooooooooooooong_var + looooooooooooooooooooooooooooooooooooooooooooooooooooooooong_var
looooooooooooooooooooooooooooooooooooooooooooooooooooooooong_var >> looooooooooooooooooooooooooooooooooooooooooooooooooooooooong_var
```

gets formatted to:

```
(
	looooooooooooooooooooooooooooooooooooooooooooooooooooooooong_var
	+ looooooooooooooooooooooooooooooooooooooooooooooooooooooooong_var
)
(
	looooooooooooooooooooooooooooooooooooooooooooooooooooooooong_var
	>> looooooooooooooooooooooooooooooooooooooooooooooooooooooooong_var
)
```

I added a condition in `visit_simple_stmt` to wrap the first child in invisible parentheses in case it is one of `arith_expr`, `shift_expr`, `xor_expr`, or `and_expr`.

I added some tests but I don't know if that's enough? Also they seem to have messed up the expected diff files, should this be done any differently?

Looking forward to your feedback!

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -2046,6 +2046,8 @@ class LineGenerator(Visitor[Line]):
 
     def visit_simple_stmt(self, node: Node) -> Iterator[Line]:
         """Visit a statement without nested statements."""
+        if first_child_is_arith(node):
+            wrap_in_parentheses(node, node.children[0], visible=False)
         is_suite_like = node.parent and node.parent.type in STATEMENT
         if is_suite_like:
             if self.mode.is_pyi and is_stub_body(node):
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
Thanks, this makes sense to me. We do want the tests to pass, of course. The `.diff` tests are a bit difficult to deal with; it might be easier to avoid changes to them and add a new test case that uses a `.py` file with an `# output` section, similar to existing test cases like `test_numeric_literals` (which uses `tests/data/numeric_literals.py`).

tpilewicz:
Just to be clear: the tests do pass already :) But of course I will add a new test case if you think that's cleaner 👍

JelleZijlstra:
Oh, right! I misread the CI output. This should be good to go.

PR #2005, opened 2021-02-24
Add --extend-exclude parameter

thejcannon:
Adding the `--extend-exclude` parameter to add additional exclude regexes on top of the `--exclude` regexes. This is especially useful if you want to specify _additional_ files/directories on top of the very useful default ones (as demonstrated by the change to our own `pyproject.toml`).

The parameter name also matches `flake8`'s parameter of `--extend-exclude`.

--- 

Fixes #1571

PR #2016, opened 2021-03-01
Turn test_regex into a click callback

thejcannon:
This is a cleanup as was suggested by @ichard26 in #2005.


src/black/__init__.py
```diff
@@ -612,39 +627,21 @@ def main(
...
     quiet: bool,
     verbose: bool,
-    include: str,
-    exclude: str,
-    extend_exclude: Optional[str],
-    force_exclude: Optional[str],
+    include: Pattern[str],
+    exclude: Pattern[str],
+    extend_exclude: Optional[Pattern[str]],
+    force_exclude: Optional[Pattern[str]],
```
thejcannon, line 639: This method doesn't start with an underscore (as don't the others) which makes me question the intended visibility of this method (is it public or private? do we expect people to know not to use it if it's private?).

If you'd like I'd be happy to handle str params here for backwards-compatibility and issue a DeprecationWarning.

ichard26: Honestly, I have no idea whether people use this function (I hope not since it certainly wasn't designed for it!), but even though Black technically doesn't have a public API, I bet we're going to hit [Hyrum's Law](https://www.hyrumslaw.com/). My personal opinion is that since it's that since it needs a click context anyway and has some behaviour API consumers wouldn't like (like exiting without a choice in some cases), it's fine to just leave this change unmitigated. **Others may have a different opinion.**

thejcannon: Very good point about the `click` context object!

FWIW you could easily add underscores to everything to imply everything is private. Then if people really want to use it they'll have to fight their linters. You also get some deniability when people inevitably complain "you broke me!".

However, I think you'll have to be choosy about what to initially mark private, as I feel like I've seen people consuming `black` from Python in the wild.

JelleZijlstra: I think it's fine to change the API here. We already don't guarantee a public API, and as @ichard26 says this function is already quite inconvenient to use as an API.

src/black/__init__.py
```diff
@@ -363,6 +363,17 @@ def target_version_option_callback(
     return [TargetVersion[val.upper()] for val in v]
 
 
+def validate_regex(
+    ctx: click.Context,
+    param: Union[click.Option, click.Parameter],
```
ichard26, line 368: `click.Option` is a subclass of `click.Parameter` so I don't think `click.Option` is necessary.
https://github.com/pallets/click/blob/master/src/click/core.py#L2133

```suggestion
    param: click.Parameter,
```

ichard26 (commented):
Just a little nit and I'm approving. Thanks for the PR!

ichard26 (approved):

PR #2021, opened 2021-03-05
Remove unused import statements using Pycln.

hadialqattan:


cooperlees (approved):
Thanks. We must need to tweak our flake8 settings too ... It should find these.

hadialqattan:
> Thanks. We must need to tweak our flake8 settings too ... It should find these.

Or we can add a formatted as `pre-commit` hook such as [Pycln](https://github.com/hadialqattan/pycln), [docs](https://hadialqattan.github.io/pycln)

ichard26:
https://github.com/psf/black/blob/b3ceb293d9e69295a190fed93517cbe1b7372154/.pre-commit-config.yaml#L3

PR #2050, opened 2021-03-18
Recommend B950 + 88 char limit instead of 80

Yajo:
[The section about line length][1] was contradictory.

On one side, it said:

> Black will try to respect that [line length limit]. However, sometimes it won't be able to without breaking other rules. In those rare cases, auto-formatted code will exceed your allotted limit.

So black doesn't guarantee that your code is formatted at 88 chars, even when configured with `--line-length=88` (default). Black uses this limit as a "hint" more than a "rule".

OTOH, it also said:

> If you're using Flake8, you can bump max-line-length to 88 and forget about it. Alternatively, use Bugbear's B950 warning instead of E501 and keep the max line length at 80 which you are probably already using.

But that's not true. You can't "forget about it" because Black sometimes won't respect the limit. Both E501 at 88 and B950 at 80 behave the same: linter error at 89+ length. So, if Black happens to decide that a line of code is better at 90 characters that some other fancy style, you land on a unlucky situation where both tools will fight.

So, AFAICS, the best way to align flake8 and black is to:

1. Use flake8-bugbear
2. Enable B950
3. Disable E501
4. Set `max-line-length = 88`

This way, we also tell flake8 that 88 limit is a "hint" and not a "rule". The real rule will be 88 + 10%. If black decides that a line fits better in 97 characters than in 88 + some formatting, _that_ probably means your code has a real problem.

To avoid further confusion, I propose to change the official recommendation here.

[1]: https://github.com/PyCQA/flake8-bugbear/tree/e82bb8d8b855adbf1f6f9757fb1527e93039e0d9#opinionated-warnings

### DIFF
--- a/docs/the_black_code_style.md
+++ b/docs/the_black_code_style.md
@@ -189,22 +189,22 @@ harder to work with line lengths exceeding 100 characters. It also adversely aff
 side-by-side diff review on typical screen resolutions. Long lines also make it harder
 to present code neatly in documentation or talk slides.
 
-If you're using Flake8, you can bump `max-line-length` to 88 and forget about it.
-Alternatively, use [Bugbear](https://github.com/PyCQA/flake8-bugbear)'s B950 warning
-instead of E501 and keep the max line length at 80 which you are probably already using.
-You'd do it like this:
+If you're using Flake8, you can bump `max-line-length` to 88 and mostly forget about it.
+However, it's better if you use [Bugbear](https://github.com/PyCQA/flake8-bugbear)'s
+B950 warning instead of E501, and bump the max line length to 88 (or the `--line-length`
+you used for black), which will align more with black's _"try to respect
+`--line-length`, but don't become crazy if you can't"_. You'd do it like this:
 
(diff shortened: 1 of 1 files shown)


cooperlees (approved):
SGTM. I can see how it's misleading, but it's a confusing thing to explain. Your use of "mostly" helps a lot! I've never been a fan of hard line limits but love the "around" this limit Bugbear does (I'm also bias as I maintain bugbear etc.)

JelleZijlstra:
Alternatively we could recommend to just disable line length lint warnings (and any other formatting-related lint rules). Black takes care of formatting for you, so you shouldn't need a linter to also check it.

Yajo:
@JelleZijlstra AFAIK black doesn't i.e. wrap comments, right?

If you don't set one line limit somewhere, it's possible that your contributors will write kilometric comments, so I still see it has one use.

Or are there any plans on black to fix that? 🤔 

BTW that doesn't mean we can't add that to the docs, since they're only recommendations after all.

Issue #2044, opened 2021-03-15
Make Python 2 support an "extra" (removed the dependency on typed-ast)


KotlinIsland:
https://github.com/python/mypy/issues/10124
https://github.com/python/mypy/pull/10183/files

typed-ast is a pain and not required for py >= 3.8

Can something similar to mypy be implemented where python 2 support is extra?

ichard26:
I'm personally +1 for this but it would have to quite well managed and well signalled to reduce breakage and breakage duration. Right now we informally tell users that Python 2 support won't be removed for the foreseeable future and this change would break that.

@cooperlees @JelleZijlstra initial thoughts?

JelleZijlstra:
Maybe we can eventually remove Python 2 support? I personally don't use it, but clearly some people still do.

I wonder if making typed-ast an extra could also break some Python 3 use cases, like formatting Python 3.9-only code with Black running under 3.6.

cooperlees:
I feel we just copy mypy here and add the version gate + extra install to `typed-ast`.

E.g.: Adding `; python_version<'3.8'` should fix most things.
https://github.com/psf/black/blob/master/setup.py#L74
- We can also copy the extra install:
```python
extras_require={'python2': 'typed-ast>=1.4.2'},
```

I have no idea with poetry tho. I am sure we'd need changes there too. but I'd totally accept this PR.

ichard26:
Closing as this was implemented by GH-2053. If there's any issues, either comment them here or create a new issue. Thanks @KotlinIsland for the request!

Issue #1038, opened 2019-10-01
Generator comprehension in call with trailing comma spuriously fails


msullivan:
Trying to reformat `list(i for i in range(10), )` fails with "Black produced code that is not equivalent to the source."

Passing a generator comprehension to a function as its sole argument with a trailing comma is valid in Python 2.7 but not in Python 3.7. (It is valid in Python 3.6 as well but typed_ast rejects it in 3.6 mode because it is based on the 3.7 parser... I expected the Python 3.8 parser rejects it as well.)

Black fails on this because the original code parses only in 2.7, so it gets a 2.7 AST. Once the trailing comma is removed, it parses in 3.7 as well, so gets a 3.7 AST, which will never match the 2.7 one.

I think the solution is probably to ensure that the same parser is used for before and after?

JelleZijlstra:
This still reproduces, but only with `--skip-magic-trailing-comma`.

JelleZijlstra:
This no longer crashes because we dropped support for 2.7. 

Technically there's a bug here as this syntax works at runtime in 3.6 and Black rejects it even though we still support formatting Python 3.6, but I don't care.

PR #2053, opened 2021-03-20
don't require typed-ast

KotlinIsland:
#2044

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,6 +40,11 @@
 
 - Lines ending with `fmt: skip` will now be not formatted (#1800)
 
+- PR #2053: Black no longer relies on typed-ast for Python 3.8 and higher
+
+- PR #2053: Python 2 support is now optional, install with
+  `python3 -m pip install black[python2]` to maintain support.
+
 #### _Packaging_
 
 - Self-contained native _Black_ binaries are now provided for releases via GitHub
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -48,7 +48,20 @@ from appdirs import user_cache_dir
 from dataclasses import dataclass, field, replace
 import click
 import toml
-from typed_ast import ast3, ast27
+
+try:
+    from typed_ast import ast3, ast27
+except ImportError:
+    if sys.version_info < (3, 8):
+        print(
+            "The typed_ast package is not installed.\n"
+            "You can install it with `python3 -m pip install typed-ast`.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    else:
+        ast3 = ast27 = ast
+
 from pathspec import PathSpec
 
 # lib2to3 fork
@@ -6336,7 +6349,12 @@ def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
                 return ast3.parse(src, filename, feature_version=feature_version)
             except SyntaxError:
                 continue
-
+    if ast27.__name__ == "ast":
+        raise SyntaxError(
+            "The requested source code has invalid Python 3 syntax.\n"
+            "If you are trying to format Python 2 files please reinstall Black"
+            " with the 'python2' extra: `python3 -m pip install black[python2]`."
+        )
     return ast27.parse(src)
 
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,8 @@ _Contents:_ **[Installation and usage](#installation-and-usage)** |
 ### Installation
 
 _Black_ can be installed by running `pip install black`. It requires Python 3.6.2+ to
-run but you can reformat Python 2 code with it, too.
+run. If you want to format Python 2 code as well, install with
+`pip install black[python2]`.
 
 #### Install from GitHub
 
--- a/setup.py
+++ b/setup.py
@@ -71,7 +71,7 @@ setup(
         "click>=7.1.2",
         "appdirs",
         "toml>=0.10.1",
-        "typed-ast>=1.4.2",
+        "typed-ast>=1.4.2; python_version < '3.8'",
         "regex>=2020.1.8",
         "pathspec>=0.6, <1",
         "dataclasses>=0.6; python_version < '3.7'",
@@ -81,6 +81,7 @@ setup(
     extras_require={
         "d": ["aiohttp>=3.3.2", "aiohttp-cors"],
         "colorama": ["colorama>=0.4.3"],
+        "python2": ["typed-ast>=1.4.2"],
     },
     test_suite="tests.test_black",
     classifiers=[
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,7 @@ skip_install = True
 deps =
     -r{toxinidir}/test_requirements.txt
 commands =
-    pip install -e .[d]
+    pip install -e .[d,python2]
     coverage erase
     coverage run -m pytest tests
     coverage report


cooperlees (approved):
Thanks! Looks good to me and what I’d had planned to do.

Can you please just add a line to CHANGES.md please. I’ll try get some CI added to nudge for this today.

cooperlees (changes requested):
O my bad, I missed CI for >=3.8 is broken. Let’s look into that.

ichard26:
I have some serious concerns about the implementation here and the general side effects of such a change. A review with more detail will come eventually.

KotlinIsland:
If python2 is optional now, does that mean test configuration should be different? ie: don't run python2 tests under normal configuration, run python2 tests under the optional 'python2' configuration?

KotlinIsland:
What on earth is this error, it looks like absolute nonsense.
```
> pip install -e .[d] .[python2]

...

INFO: pip is looking at multiple versions of black[d] to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install black[d]==20.8b2.dev112+gde5d07a.d20210321 and black[python2]==20.8b2.dev112+gde5d07a.d20210321 because these package versions have c
onflicting dependencies.

The conflict is caused by:
    black[d] 20.8b2.dev112+gde5d07a.d20210321 depends on black 20.8b2.dev112+gde5d07a.d20210321 (from C:\Users\Morgan\IdeaProjects\black)
    black[python2] 20.8b2.dev112+gde5d07a.d20210321 depends on black 20.8b2.dev112+gde5d07a.d20210321 (from C:\Users\Morgan\IdeaProjects\black)

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

KotlinIsland:
This looks sus?
```
Mode(target_versions=set(), line_length=88, string_normalization=True, magic_trailing_comma=True, experimental_string_processing=True, is_pyi=False)
--- first pass
+++ second pass
@@ -6349,12 +6349,13 @@
                 return ast3.parse(src, filename, feature_version=feature_version)
             except SyntaxError:
                 continue
     if ast27.__name__ == "ast":
         print(
-            "Black attempted to parse python2 code but the typed_ast package is not installed.\n"
-            "You must install typed_ast with `python3 -m pip install typed-ast`.",
+            "Black attempted to parse python2 code but the typed_ast package is not"
+            " installed.\nYou must install typed_ast with `python3 -m pip install"
+            " typed-ast`.",
             file=sys.stderr,
         )
         sys.exit(1)
     return ast27.parse(src)
```

KotlinIsland:
All passed 🎉

Currently tests are run with `black[python2]` I think additional configurations should be added for python 3.8 with `black` and not `black[python2]`

KotlinIsland:
```py
BlackRunner().invoke(black.main, input=BytesIO('"foo"'.encode("utf8"))).output
```
This outputs `""`, whats up with that?

KotlinIsland:
Wait a second, why is the dependency on typed-ast here in the first place? is it just to parse python 2? if so the dependencies should not be `"typed-ast>=1.4.2; python_version < '3.8'",`, the dependency should instead be removed entirely.

KotlinIsland:
I noticed that ast.parse doesn't include type hints by default.
>If type_comments=True is given, the parser is modified to check and return type comments
-https://docs.python.org/3/library/ast.html#ast.parse

Should the code be updated to include this parameter?

KotlinIsland:
@cooperlees Could you review/address this please.

CHANGES.md
```diff
@@ -40,6 +40,11 @@
 
 - Lines ending with `fmt: skip` will now be not formatted (#1800)
 
+- PR #2053: Black no longer relies on typed-ast for python versions >= 3.8
+
+- PR #2053: Python2 support is now optional, install with
+  `python3 -m pip install black[python2]`
+
```
ichard26, line 47: I like how you split up the two separate but related changes!

NIT: can we be consistent? also wording tweaks and removing unsafe advice
```suggestion
- PR #2053: Black no longer relies on typed-ast for Python 3.8 and higher

- PR #2053: Python 2 support is now optional, install with the `python2` extra to
  maintain support

```

ichard26 (commented):
Hey thanks for the PR! I know the rest of my review is quite negative, but I genuinely appreciate the time and energy you've put into putting this together!

I don't like this change because it provides the wrong feedback to a not an insignificant group of users. If I (as a Python 3 only user) give Black an invalid file, the Python 3.X parse attempts will fail and then Black exits with a "Black attempted to parse python2 code, [...] please install typed_ast" error... which makes no sense to me as a Python 3 only user. Not only is this hiding the true problem: the input file isn't valid Python 3 code, the confused user, might try the advice to install `typed_ast` which 1) might not work or be a pain, and 2) doesn't actually the fix the issue.

Personally I'm open to pushing most if not all of the disruption to the users still using Black's Python 2 support. My idea would be that Black would need `--target-version py27` (configured automatically via magic or configured manually) to even attempt Python 2 parsing. Although this falls apart when Black is formatting multiple files at once (since someone is bound to have a mix of Python 2 and 3 files... of which our target-version logic can't really describe).

There's not much actionable in these two paragraphs of feedback. It's something to just think about, perhaps someone else (maybe you!) has a better idea!

ichard26:
> I noticed that ast.parse doesn't include type comments by default.
> 
> > If type_comments=True is given, the parser is modified to check and return type comments
> > -https://docs.python.org/3/library/ast.html#ast.parse
> 
> Should the code be updated to include this parameter?

I mean, maybe?, but let's do that in another PR since it might have an effect we haven't foreshadowed.

ichard26:
Actually thinking about this more, we could make `parse_ast` use the parsing attempt information from the first parse (IRCC `lib2to3_parse`) to configure whether Python 2 parsing should even be attempted. This would also fix the long-standing issue where our forked AST+CST blib2to3 parser (for formatting) uses a different grammar than the built-in or third-party typed_ast parser (to run safety checks) causing misleading and confusing errors. see also https://github.com/psf/black/issues/1038#issue-501069918

CHANGES.md
KotlinIsland: Is suggesting a pip command unsafe advice? the [readme](https://github.com/psf/black/#installation-and-usage) has similar commands.

KotlinIsland:
@ichard26 Thanks for the review! Could  you please address https://github.com/psf/black/pull/2053#issuecomment-803529392

CHANGES.md
ichard26: A pip command no, but `python3 -m pip install black` with `python{whatever-version}` isn't safe since it may reference a system install of Python and installing stuff into one is... a bad idea. Even `pip install black` isn't entirely safe to just copy and paste depending on how the system is set up, but it's a bit more generic to send a message that you might want to double check what Python you're about to install to. 

Also as an example, my Ubuntu 20.04 LTS installation doesn't have `pip` linked to any executable (and therefore any Python installation) so even a copy+paste would fail and not potentially break my system. I'd have to use a virtual environment to even have `pip install black` work.

ichard26: Honestly I'm not entirely sure whether this is the right stance, ex. see https://snarky.ca/why-you-should-use-python-m-pip/, just that `python3 -m pip install black` looked too copy and pastable and dangerous, it was more a knee jerk reaction than anything really. We can ignore this :)

ichard26:
> Wait a second, why is the dependency on typed-ast here in the first place? is it just to parse python 2? if so the dependencies should not be "typed-ast>=1.4.2; python_version < '3.8'",, the dependency should instead be removed entirely.

Probably something to do with type comments but I'm not sure.

KotlinIsland:
> Probably something to do with type comments but I'm not sure.

This seems to be the only usage of type comments. But this won't work with python >= 3.8 because we aren't using the `type_comments=True` option
__init__.py line 6389
```py
        type_ignore_classes = (ast3.TypeIgnore, ast27.TypeIgnore)
        if sys.version_info >= (3, 8):
            type_ignore_classes += (ast.TypeIgnore,)
```

src/black/__init__.py
```diff
@@ -6336,7 +6349,13 @@ def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
                 return ast3.parse(src, filename, feature_version=feature_version)
             except SyntaxError:
                 continue
-
+    if ast27.__name__ == "ast":
+        raise Exception(
+            f"The requested file: {filename!r} has invalid python 3 syntax.\n"
+            "If you are trying to format Python 2 files please reinstall Black with"
+            " the 'python2' extra.\nSee example #7 here: "
+            "https://pip.pypa.io/en/stable/reference/pip_install/#examples"
```
ichard26, line 6357: please look a bit higher and you should see that `filename` is hard coded to "unknown", not very useful :)

Also, I retracted my feedback about not doing `python3 -m pip install black` since it's questionable at best, so if you feel like it, you can revert to your original suggestion (`python3 -m pip install typed_ast`). The link to the pip docs is worse IMO. I'm sorry for making you flip back and forth on this.

PS. I love your wording here, my wording of anything always uses too many words

CHANGES.md
```diff
@@ -40,6 +40,11 @@
 
 - Lines ending with `fmt: skip` will now be not formatted (#1800)
 
+- PR #2053: Black no longer relies on typed-ast for Python 3.8 and higher
+- PR #2053: Python 2 support is now optional, install with the `python2` extra to
+  maintain support
+  [pip extras #7](https://pip.pypa.io/en/stable/reference/pip_install/#examples)
```
ichard26, line 46: 1) I'm surprised prettier isn't ticked off here by the lack of spacing; and

2) if you'd like, just use your original suggestion of `python3 -m pip install typed_ast` since this reads worse and my reasoning was at best questionable.

KotlinIsland:
Only remaining issue in my opinion is that all tests use `python2` option. I have no idea how to go about setting up a different test configuration and only executing certain tests.

cooperlees:
> Only remaining issue in my opinion is that all tests use `python2` option. I have no idea how to go about setting up a different test configuration and only executing certain tests.

I'd love @ambv be the definitive answer here. But I feel we can move the python2 tests to a separate GitHub action with the plan to eventually deprecate and no longer support. If we were already "stable" I would have said we freeze python2 supported at version X. But we're not there.

Others thoughts?

src/black/__init__.py
```diff
@@ -6336,7 +6349,12 @@ def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
                 return ast3.parse(src, filename, feature_version=feature_version)
             except SyntaxError:
                 continue
-
+    if ast27.__name__ == "ast":
+        raise Exception(
```
ambv, line 6353: Please don't raise bare Exceptions. Make this a SyntaxError as well.

src/black/__init__.py
```diff
@@ -6336,7 +6349,12 @@ def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
...
+            "The requested source code has invalid python 3 syntax.\n"
```
ambv, line 6354: Python should be capitalized here as well.

ambv (approved):
I approve this in general. Good work. There's two nits before we merge this.

Ideally, tests should be running twice: with the python2 extra and without it. We can do this in a separate pull request.

ambv:
Tests with the python2 extra can be a separate environment in tox.ini, just like fuzz tests are. We need to make sure they are ran on CI.

KotlinIsland:
> Tests with the python2 extra can be a separate environment in tox.ini, just like fuzz tests are. We need to make sure they are ran on CI.

I assuming it needs different configuration to not run python 2 tests.

ulgens:
@KotlinIsland @ambv I'm using Black (latest commit from master) with Python 3.9. It still requires typed-ast package and fails 
with 

```
from typed_ast import ast3, ast27
ModuleNotFoundError: No module named 'typed_ast'
```

KotlinIsland:
😳
Oops `ModuleNotFoundError` only catching ImportError`

ichard26:
Isn't `ModuleNotFoundError` supposed to be a subclass of `ImportError`? 

https://docs.python.org/3.9/library/exceptions.html#exception-hierarchy

Issue #2045, opened 2021-03-15
Wrong documentation


rggjan:
On:
https://black.readthedocs.io/en/latest/editor_integration.html

The enumeration is not rendered properly, and it shows:
```
PyCharm/IntelliJ IDEA¶
1. ...
1. ... 
1. ...
1. See it `install_location_from_step_2`
```

bene64:
It seems that this is some error within [sphinx](https://github.com/sphinx-doc/sphinx). The enumeration itself is a valid markdown file, so it should render to 1.,2.,...

The markdown file itself renders nicely on github: https://github.com/psf/black/blob/master/docs/editor_integration.md

PR #2056, opened 2021-03-20
Fix indentation in docs/editor_integration.md

ichard26:
Numbered list entries' bodies need to be indented or else the list won't
render correctly.

Fixes GH-2045.

also \*sigh\* I'm going to have a fun time dealing with the conflicts when trying to merge my documentation reorganization branch /s


cooperlees (approved):
TIL - Thanks!

Issue #1235, opened 2020-01-19
whitespace before ':'  -- `if (ct := await cts.filter(tag=t).first()) :`


waketzheng:
**black version:** `black, version 19.10b1.dev16+g1ab87a3`
**python -V:** `Python 3.8.1`
**pyproject.toml:**
```
[tool.black]
line-length = 88
py38 = true
skip-string-normalization = true
```
**snippet after reformatted:** 
```py
# models/bill.py
class BillCell(BaseModel):
    tag: BillTags = fields.CharEnumField(BillTags)

    async def get_tag_display(self):
        t = self.tag
        if t in CustomTag:
            cts = self.bill.billitem.custom_tags
            if (ct := await cts.filter(tag=t).first()) :
                return ct.name
        return BillTags.get_label(t)
```
**$ flake8 models/bill.py**   # line 100 is: `if (ct := await cts.filter(tag=t).first()) :`
```
models/bill.py:100:55: E203 whitespace before ':'
```
Why the line 100 not reformatted as `if (ct := await cts.filter(tag=t).first()):`?

emakryo:
I saw a similar case;
```
x = [1, 2, 3]
i = 0
print(x[i + 1:])
```

black changes the last line into `print(x[i + 1 :])`.

JelleZijlstra:
@emakryo that's a different (and expected) piece of behavior, for which see https://github.com/psf/black#slices.

The original issue is a duplicate of #1174.

Issue #1234, opened 2020-01-17
Failed to byte-compile debian package


mandeep:
I installed `black` version `18.9b0-1-6` from apt on `Debian 10.2`. During installation I received the following error:

```
The following NEW packages will be installed:
  black
0 upgraded, 1 newly installed, 0 to remove and 65 not upgraded.
Need to get 75.4 kB of archives.
After this operation, 334 kB of additional disk space will be used.
Get:1 http://ftp.us.debian.org/debian buster/main amd64 black all 18.9b0-1-6 [75.4 kB]
Fetched 75.4 kB in 0s (400 kB/s)
Selecting previously unselected package black.
(Reading database ... 495621 files and directories currently installed.)
Preparing to unpack .../black_18.9b0-1-6_all.deb ...
Unpacking black (18.9b0-1-6) ...
Setting up black (18.9b0-1-6) ...
-V is ignored in pypycompile
Failed to byte-compile /usr/lib/python3/dist-packages/black.py:   File "/usr/lib/python3/dist-packages/black.py", line 128
    ) -> "FileMode":
    ^
SyntaxError: invalid syntax

Processing triggers for man-db (2.8.5-2) ...
```

However, `black` seems to work fine afterwards.

zsol:
I'm not familiar with how Black is packaged in debian, but it looks like something is trying to run `pypycompile` on it and it fails. I don't think this is a bug in Black itself. Maybe report this to a debian issue tracker?

PR #2057, opened 2021-03-21
Add a GitHub CHANGELOG/News Check

cooperlees:
- Run grep to see commit has a line mentioning it to CHANGES.md
- Also add a label to disable this being required for PRs that don't need a change entry

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -11,11 +11,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
 
-    - name: Grep CHANGES.md for PR number
-      if: contains(github.event.pull_request.labels.*.name, 'skip news') != true
-      run: |
-        grep -P "PR #${{ github.event.pull_request.number }}[^0-9]" CHANGES.md || \
-        (echo "Please add 'PR #${{ github.event.pull_request.number }}' change line to CHANGES.md" && \
-        exit 1)
+      - name: Grep CHANGES.md for PR number
+        if: contains(github.event.pull_request.labels.*.name, 'skip news') != true
+        run: |
+          grep -P "PR #${{ github.event.pull_request.number }}[^0-9]" CHANGES.md || \
+          (echo "Please add 'PR #${{ github.event.pull_request.number }}' change line to CHANGES.md" && \
+          exit 1)
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,21 @@
+name: changelog
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled, reopened]
+
+jobs:
+  build:
+    name: Changelog Entry Check
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Grep CHANGES.md for PR number
+      if: contains(github.event.pull_request.labels.*.name, 'skip news') != true
+      run: |
+        grep -P "PR #${{ github.event.pull_request.number }}[^0-9]" CHANGES.md || \
+        (echo "Please add 'PR #${{ github.event.pull_request.number }}' change line to CHANGES.md" && \
+        exit 1)


ichard26 (changes requested):
Can you please add some documentation for potential contributors? We don't have to make it harder :)

Honestly I'd like the issue being fixed be referenced too (by ID) for even more context why the change was needed (and not just how it was done) but we can iterate on that later.

cooperlees:
I can, but I feel the documentation will be low signal and just add yet another thing to keep up to date if things change, when this is a relatively simple ask of contributors. The goal here, as I've stated before, is just to get something added as it's much easier for the maintainer to edit CHANGES.md lines before release then go through the commits to workout what to add to this file verbatim.

cooperlees:
> Honestly I'd like the issue being fixed be referenced too (by ID) for even more context why the change was needed (and not just how it was done) but we can iterate on that later.

Well I think the format changing is scope creep to the goal of this imo. I am also a fan of the CHANGES.md to be no more than a 1 line summary. I feel one goes to the PR / issue(s) to get more context.

ichard26:
Good point about the issue ID idea, anyone who _actually_ needs to check the changelog doesn't have enough time to be worth getting more context. I think my sense of who our users are is a bit wonky :(

I'm still of the opinion that this should be referenced somewhere in the contribution docs, but as a compromise let's ignore that for now until this part of the workflow gains some stability.

ichard26 (approved):
Seems good to me as long no else objects!

MarkCBell:
I would like to add just how much more difficult this additional workflow has made submitting PRs to black since I don't know of  any way to know the required PR number before submitting the pull request. 

This means that after forking, cloning, changing and commiting a feature, I have to create a PR and submit this only for this workflow to reject my PR (including sending me an email that "jobs have failed") and tell me that CHANGES.md needs to contain "PR #1234". With this information in hand I can then make another commit to add this information but it means that I have to make a second iteration around the commit / PR loop.

ichard26:
@MarkCBell we are aware of the additional work this places on contributors. It's been decided that it's worth it to reduce the friction when creating a release. Maintainer time is extremely limited and creating a changelog takes a fair bit amount of work. 

Now about the 2-step process, yeah I personally still don't really like it, but this doesn't bother me enough to try convincing the other maintainers to get rid of it. Admittedly I have custom tooling (it's really just a modified data analysis script) on my development machine that can give me the right ID without first submitting a PR so I am certainly biased.

MarkCBell:
@ichard26 Yes I can completely relate to maintainer time being limited and the amount of work that you are putting in. Black is a wonderful tool and does a vey good job.

I think your earlier suggestion that this "should be referenced somewhere in the contribution docs" is a very good idea. Unfortunately there is currently no reference to this in [the contributing documentation](https://github.com/psf/black/blob/master/CONTRIBUTING.md#technicalities) and so this caught me out even though I followed the lints and tests on my local machine. In fact I'm not even sure that there is a way to locally run a github workflow, so it looks contributors encounter this behaviour for the first time just after they submit their pull request.

I'm curious about your "script that can give you the right ID without first submitting a PR", can you share any details of this? I'm guessing it needs to be hooked the PR creation process since if you look at the existing PRs and see that #1234 is the latest then you cant automatically put #1235 in CHANGES.md in case someone else opens a PR before you and so yours gets assigned a different number.

ichard26:
> I think your earlier suggestion that this "should be referenced somewhere in the contribution docs" is a very good idea.

Thank you! but:

> [...] but as a compromise let's ignore that for now until this part of the workflow gains some stability and we won't have to update it 24/7.

Perhaps I could convince Cooper that adding documentation now would a good idea.

> I'm curious about your "script that can give you the right ID without first submitting a PR"

It was never designed to be shared so it's not portable. I like to make graphs with issue and pr data, soooo I have a script that maintains a full clone of all the issue and pr data available from the GitHub API for a certain repository. Then another script imports the data and gets the newest issue / pr and takes its number and adds 1 to it and that's it.

cooperlees:
#2073 

I acknowledge this is not perfect, but the time saved for release is well worth it. This can be added easily from the Web by the contributor or maintainer. I'm all ideas for a better flow that's not super invasive and overly complicated (code / maintenance wise).

ichard26:
@MarkCBell I created https://github.com/ichard26/next-pr-number to make it easier to get the number your pr will be assigned. The following link has this repo already filled in: https://ichard26.github.io/next-pr-number/?owner=psf&name=black

Does this make it easier for you?

Issue #2060, opened 2021-03-24
Issue with --extend-exclude parameter


arefrazavi:
**Describe the bug** 
It seems that `--extend-exclude` doesn't work on the 20.8b1 version.

**To Reproduce** 

1. Run _Black_ on it with these arguments '--extend-exclude ':
```
 black --extend-exclude venv/ .
```
2. See error:
```
Error: no such option: --extend-exclude
```

**Environment (please complete the following information):**

- Version: 20.8b1
- OS and Python version: Linux/Python 3.8.5

**Does this bug also happen on master?** 
No

JelleZijlstra:
It will exist in the next release

PR #2061, opened 2021-03-24
Add entry for `--extend-exclude` in the right place

ichard26:
The PR author added the changelog entry for their `extend-exclude` addition
in `docs/change_log.md` which is understandable but incorrect as it will be
overwritten since it's autogenerated from the readme.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,8 @@
 
 - use lowercase hex strings (#1692)
 
+- added `--extend-exclude` argument (PR #2005)
+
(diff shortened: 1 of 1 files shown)


ichard26:
added `skip news` since this fixing the news file :)

for context, I noticed this error when I was writing a response to GH-2060, although Jelle ended up beating me to it.

cooperlees:
> added `skip news` since this fixing the news file and such a change definitely doesn't need a news entry of its own :)

WE need news about the news!

cooperlees (approved):

PR #2064, opened 2021-03-26
Add `--skip-magic-trailing-comma` to CHANGES.md

hauntsaninja:


hauntsaninja:
Let me know if I should listen the changelog CI check, seems useless in this case :-)

PR #2073, opened 2021-03-29
Add CONTRBUTING info about CHANGES.md requirement

cooperlees:
- Instruct contributors to add the change line to help save maintainer / releaser time

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -16,6 +16,6 @@ jobs:
       - name: Grep CHANGES.md for PR number
         if: contains(github.event.pull_request.labels.*.name, 'skip news') != true
         run: |
-          grep -Pz "\(( |\n\s*)#${{ github.event.pull_request.number }}[^0-9]\)" CHANGES.md || \
+          grep -Pz "\((\n\s*)?#${{ github.event.pull_request.number }}(\n\s*)?\)" CHANGES.md || \
           (echo "Please add '(#${{ github.event.pull_request.number }})' change line to CHANGES.md" && \
           exit 1)
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -16,6 +16,6 @@ jobs:
       - name: Grep CHANGES.md for PR number
         if: contains(github.event.pull_request.labels.*.name, 'skip news') != true
         run: |
-          grep -Pz "PR( |\n\s*)#${{ github.event.pull_request.number }}[^0-9]" CHANGES.md || \
-          (echo "Please add 'PR #${{ github.event.pull_request.number }}' change line to CHANGES.md" && \
+          grep -Pz "\(( |\n\s*)#${{ github.event.pull_request.number }}[^0-9]\)" CHANGES.md || \
+          (echo "Please add '(#${{ github.event.pull_request.number }})' change line to CHANGES.md" && \
           exit 1)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,6 +58,23 @@ $ tox -e fuzz
 $ black-primer [-k -w /tmp/black_test_repos]
 ```
 
+### News / Changelog Requirement
+
+`Black` has CI that will check for an entry corresponding to your PR in `CHANGES.md`. If
+you feel this PR not require a `CHANGES.md` please state that in a comment and a
+maintainer can add a `skip_news` tag to make the CI pass. Otherwise, please ensure you
+have a line in the following format:
+
+```md
+- `Black` is now more awesome (#X)
+```
+
+To workout X, checkout the latest issue and PR number and add 1. This is not perfect but
+saves a lot of release overhead as now the releaser does not need to go back and workout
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -58,6 +58,23 @@ $ tox -e fuzz
...
+
+```md
+- `Black` is now more awesome (#X)
+```
+
+To workout X, checkout the latest issue and PR number and add 1. This is not perfect but
+saves a lot of release overhead as now the releaser does not need to go back and workout
+what to add to the `CHANGES.md` for each release.
+
+_Suggestions welcome on how this could be a better less invasive flow._
```
ichard26, line 76: I may or may not be building a solution to this right now ... :)

CONTRIBUTING.md
```diff
@@ -58,6 +58,23 @@ $ tox -e fuzz
...
 
+### News / Changelog Requirement
+
+`Black` has CI that will check for an entry corresponding to your PR in `CHANGES.md`. If
+you feel this PR not require a `CHANGES.md` please state that in a comment and a
+maintainer can add a `skip_news` tag to make the CI pass. Otherwise, please ensure you
+have a line in the following format:
+
+```md
+- `Black` is now more awesome (#X)
```
ichard26, line 69: watch out the regex will kill you without the PR prefix :D

```suggestion
- `Black` is now more awesome (PR #X)
```

CONTRIBUTING.md
```diff
@@ -58,6 +58,23 @@ $ tox -e fuzz
 $ black-primer [-k -w /tmp/black_test_repos]
 ```
 
+### News / Changelog Requirement
+
+`Black` has CI that will check for an entry corresponding to your PR in `CHANGES.md`. If
+you feel this PR not require a `CHANGES.md` please state that in a comment and a
+maintainer can add a `skip_news` tag to make the CI pass. Otherwise, please ensure you
```
ichard26, line 65: last time I checked, it was `skip news` not `skip_news`, also what's up with the tag terminology?, I first think of Git tags before realizing my mistake, let's use labels instead if you don't mind :)

```suggestion
maintainer can add a `skip news` label to make the CI pass. Otherwise, please ensure you
```

ichard26 (changes requested):
thanks @cooperlees, this is real nice of you!

super duper casual pr review because that's how I'm feeling right now :D

CONTRIBUTING.md
```diff
@@ -58,6 +58,23 @@ $ tox -e fuzz
 $ black-primer [-k -w /tmp/black_test_repos]
 ```
 
+### News / Changelog Requirement
+
+`Black` has CI that will check for an entry corresponding to your PR in `CHANGES.md`. If
+you feel this PR not require a `CHANGES.md` please state that in a comment and a
```
ichard26, line 64: I get what you mean here, but let's be more correct

```suggestion
`Black` has CI that will check for an entry corresponding to your PR in `CHANGES.md`. If
you feel this PR not require a changelog entry please state that in a comment and a
```

ps. I'm sorry if this ticks off prettier

ichard26 (changes requested):
I missed something

CONTRIBUTING.md
cooperlees: > watch out the regex will kill you without the PR prefix :D

Should we fix the regex - My bad, seems there are no 'PR' used in CHANGES.md ... I screwed up there it seems ...

ichard26 (changes requested):
so close to being ready

.github/workflows/changelog.yml
```diff
@@ -16,6 +16,6 @@ jobs:
       - name: Grep CHANGES.md for PR number
         if: contains(github.event.pull_request.labels.*.name, 'skip news') != true
         run: |
-          grep -Pz "PR( |\n\s*)#${{ github.event.pull_request.number }}[^0-9]" CHANGES.md || \
-          (echo "Please add 'PR #${{ github.event.pull_request.number }}' change line to CHANGES.md" && \
+          grep -Pz "\(( |\n\s*)#${{ github.event.pull_request.number }}[^0-9]\)" CHANGES.md || \
```
ichard26, line 19: This regex won't match now because it's expecting `({whitespaceOrNewline}#123{someRandomNonNumericChar}). Here's two that work:
1. `\((\n\s*)?#${{ github.event.pull_request.number }}(\n\s*)?\)` - does support having the brackets on separate lines
2. `\(#${{ github.event.pull_request.number }}\)` - does NOT support having the brackets on separate lines

I have no idea whether prettier will (or how it will) break `(#123)` when it violates the line length limit so IDK which regex is better. Your call, although the first one seems safer?

ichard26 (changes requested):
I totally did not mess up on the previous review ... :(

.github/workflows/changelog.yml
ambv: Let's go with the first suggestion here and land this.

cooperlees: Will fix now.

PR #2080, opened 2021-04-01
Push contributors to use Next PR Number

ichard26:
This is a tool of my own making. Right now our requirement to have the
PR number in the changelog entry is pretty painful / annoying since
the contributor either has to guess or add the # retroactively after
the PR creation. This tool should make it way less painful by making
it simple to get your PR number beforehand.


cooperlees (approved):
We should eventually remove the two contributing .md's

PR #2081, opened 2021-04-01
Run lint/typechecking in latest python + update pre-commit

cooperlees:
- Lets move to latest and greatest of lints
- Update mypy + flake8

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,8 +18,6 @@ jobs:
 
       - name: Set up Python
         uses: actions/setup-python@v2
-        with:
-          python-version: 3.7
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -18,8 +18,6 @@ jobs:
 
       - name: Set up Python
         uses: actions/setup-python@v2
-        with:
-          python-version: 3.7
```
JelleZijlstra, line 22: Would it be useful to specify the version here so things don't break randomly in the future when a new version is released?

ichard26: I think the rational was that making the version unbounded reduces the amount of work we have to do when a new Python version rolls around. Although I feel like we are trading that for the (very annoying although probably uncommon) situations when our CI break randomly. Personally still not fully convinced, but Cooper is the upgrade early and then fix the resulting issues type of guy from what I've heard.

cooperlees: Exactly. I hate keeping things up to date.

JelleZijlstra: OK, let's do it

PR #2085, opened 2021-04-04
Get rid of redundant spaces in docs

programmer04:
This is a minor fix for docs, not worth mentioning in release notes, hence no entry in `CHANGES.md`


ichard26:
Thanks @programmer04!

PR #2092, opened 2021-04-07
Remove NBSP at the beggining of comments

### LINKED ISSUE #2091: Remove non-breaking space at the beggining of comments
Black apply a space after a ``#`` in comment like so:
```diff
- #Comment content 
+ # Comment content 
```
The thing is, often we developers fat fingers an insecable space after creating the ``#`` by not releasing the <kbd>Alt Gr</kbd> key fast enough.

Then black had a space before it resulting in visually two spaces:
```diff
- # Comment content 
+ #  Comment content 
```

I'd like black to remove insecable spaces at the beginning of comment and then replaces them with a real space:
```diff
- # Comment content 
+ # Comment content 
```

Pierre-Sassoulas:
This is a fast attempt to do #2091. I suppose we could remove the while and put a simple ``if`` to not remove multiple consecutive NBSP.

Closes #2091

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 #### _Black_
 
+- `Black` now cleans up leading non-breaking spaces in comments (#2092)
+
 - `Black` now respects `--skip-string-normalization` when normalizing multiline
   docstring quotes (#1637)
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -2709,8 +2709,9 @@ def make_comment(content: str) -> str:
 
     if content[0] == "#":
         content = content[1:]
-    if content and content[0] == " ":  # This is a non-breaking space
-        content = " " + content[1:]  # And this is a simple space
+    NON_BREAKING_SPACE = " "
+    if content and content[0] == NON_BREAKING_SPACE and "type: ignore" not in content:
+        content = " " + content[1:]  # Replace NBSP by a simple space
     if content and content[0] not in " !:#'%":
         content = " " + content
     return "#" + content
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -2709,6 +2709,8 @@ def make_comment(content: str) -> str:
 
     if content[0] == "#":
         content = content[1:]
+    if content and content[0] == " ":  # This is a non-breaking space
+        content = " " + content[1:]  # And this is a simple space
     if content and content[0] not in " !:#'%":
         content = " " + content
     return "#" + content
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -2710,7 +2710,11 @@ def make_comment(content: str) -> str:
     if content[0] == "#":
         content = content[1:]
     NON_BREAKING_SPACE = " "
-    if content and content[0] == NON_BREAKING_SPACE and "type: ignore" not in content:
+    if (
+        content
+        and content[0] == NON_BREAKING_SPACE
+        and not content.lstrip().startswith("type:")
+    ):
         content = " " + content[1:]  # Replace NBSP by a simple space
     if content and content[0] not in " !:#'%":
         content = " " + content
--- a/tests/data/comments7.py
+++ b/tests/data/comments7.py
@@ -129,6 +129,8 @@ class C:
     ):
         ...
 
+square = Square(4) # type: Optional[Square]
+
 # output
 
 from .config import (
@@ -264,3 +266,6 @@ class C:
         self, pyramid_config, db_request, post_data, message
     ):
         ...
+
+
+square = Square(4)  # type: Optional[Square]
--- a/tests/data/comments_non_breaking_space.py
+++ b/tests/data/comments_non_breaking_space.py
@@ -1,0 +1,44 @@
+from .config import (  ConfigTypeAttributes,    Int,    Path,    # String,
+    # DEFAULT_TYPE_ATTRIBUTES,
+)
+
+result = 1  # A simple comment
+result = (    1, ) # Another one
+
+result = 1    # type: ignore
+result = 1# This comment is talking about type: ignore
+square = Square(4) # type: Optional[Square]
+
+def function(a:int=42):
+    """ This docstring is already formatted
+       a
+       b
+    """
+    #    There's a NBSP + 3 spaces before
+    #    And 4 spaces on the next line
+    pass
+
+# output
+from .config import (
+    ConfigTypeAttributes,
+    Int,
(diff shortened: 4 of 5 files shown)


ichard26:
Looking into this, the reason why Python 3.6 and 3.7 are failing is probably because `typed_ast` doesn't recognize `"NUMBER SIGN" "NO-BREAK SPACE" [type: ignore]` as an actual type ignore comment which has a special field in the AST. The built-in parser (`ast`) which is used on Python 3.8 and higher has its type ignore comment specific tracking features disabled which means the "apparently invalid type ignore comment turning into valid one" effect goes unnoticed.

JelleZijlstra:
Thanks for the diagnosis @ichard26! I think for safety we should just leave any nbsps in type comments alone.

Pierre-Sassoulas:
@JelleZijlstra I believe it's now ready for review. Thanks @ichard26 for dong all the leg work :) !

src/black/__init__.py
```diff
@@ -2709,6 +2709,9 @@ def make_comment(content: str) -> str:
 
     if content[0] == "#":
         content = content[1:]
+    NON_BREAKING_SPACE = " "
+    if content and content[0] == NON_BREAKING_SPACE and "type: ignore" not in content:
```
JelleZijlstra, line 2713: Type comments can also be of the form "# type: int". Maybe just check something like `not content.lstrip.startswith("type:")` (assuming that lstrip strips nbsps).

tests/data/comments7.py
```diff
@@ -183,7 +183,7 @@ def func():
         0.0789,
         0.0123,
         0.0789,
-        a[-1],  # type: ignore
+        a[-1],  #  type: ignore
```
JelleZijlstra, line 186: Looks like it turns "#<nbsp>type" into "# <nbsp>type". Let's just leave it as is instead of inserting an additional space.

Pierre-Sassoulas: I replaced the test data from ``# type: ignore`` to ``#[NBSP]type: ignore`` so now it becomes ``# [NBSP]type: ignore``. That's what we're trying to avoid in other comments, but we have to ignore type comment (or the refactor could be huge to take "apparently invalid type ignore comment turning into valid one" into account).

src/black/__init__.py
JelleZijlstra: There is already an `is_type_comment()` function. It's not quite right though, and I just opened psf/black#2097 about it.

tests/data/comments7.py
JelleZijlstra: I actually changed my mind on this, I think it's good to add the space here to alert the user that this is an invalid type comment masquerading as a correct one. So let's just leave this as is.

Pierre-Sassoulas: I think it's a test case we better add or this edge case with ``type: ignore`` comment would not be tested/apparent.

src/black/__init__.py
ichard26, line 2713: FYI, you can also use Unicode name escapes, so instead of a defining an aptly named variable, you could just replace it with `"\N{NO-BREAK SPACE}"`

Now does it matter? Probably not since it's arguable which one is more readable, but I'd like to point it to you as a language you may have not known :)

Pierre-Sassoulas: Thanks I did not know ! 😄

Pierre-Sassoulas: Would a regex be ok ? The exact same code than mypy duplicated ?

CHANGES.md
```diff
@@ -4,6 +4,9 @@
 
 #### _Black_
 
+- `Black` now remove leading non-breaking spaces in comments before adding a space
```
```suggestion
- `Black` now cleans up leading non-breaking spaces in comments
```
I think this is better; we don't need to be too specific about the exact output.

JelleZijlstra (changes requested):
Thanks! The code looks good now, but I think the tests will be cleaner in a separate file.

src/black/__init__.py
```diff
@@ -4614,7 +4621,7 @@ def TErr(err_msg: str) -> Err[CannotTransform]:
 def contains_pragma_comment(comment_list: List[Leaf]) -> bool:
     """
     Returns:
-        True iff one of the comments in @comment_list is a pragma used by one
+        True if one of the comments in @comment_list is a pragma used by one
```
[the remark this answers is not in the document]
Pierre-Sassoulas: My bad 😅

tests/data/comments7.py
```diff
@@ -6,7 +6,7 @@
     Int,
     Path,
     #  String,
-    #  resolve_to_config_type,
+    #  resolve_to_config_type,
```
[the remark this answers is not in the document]
Pierre-Sassoulas: 👍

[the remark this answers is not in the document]
Pierre-Sassoulas: I copy pasted ``comment7.py`` with a new name and added NBSP as the first character of each comments. The reflexion is that there was a lot of different test cases in comment7 but let me know if simple test cases (normal comment, commen with a lot of leading spaces, type ignore comment, type: Optional[Squre] comment ?) is preferable.

tests/data/comments_non_breaking_space.py
```diff
@@ -0,0 +1,271 @@
+from .config import (
+    Any,
```
JelleZijlstra, line 2: Could you remove some extraneous code here? I think most of the imports on this line can go for example, and most of the cases in `class C` below aren't really relevant either. I'd want this test file to have only comments with nbsps and elements that are directly related (e.g., to test behavior around moving comments).

JelleZijlstra (approved):
Thanks!

Pierre-Sassoulas:
I made the test a little too nasty and found a bug:
```python
def function(a:int=42):
    """ This docstring start with a NBSP
        There is one column 5 here
        This is 7 NBSP
    """
```
Error log:
```
--- src
+++ dst
@@ -162,11 +162,11 @@
       body=
         Expr(
           value=
             Constant(
               value=
-                'This docstring start with a NBSP\n\xa0   There is one column 5 here\n\xa0\xa0\xa0\xa0This is 7 NBSP',  # str
+                'This docstring start with a NBSP\nThere is one column 5 here\nThis is 7 NBSP',  # str
             )  # /Constant
         )  # /Expr
         Pass(
         )  # /Pass
       decorator_list=
```
Should I open a new issue or address it here ?

JelleZijlstra:
Oops, too late. Could you just file a new PR?

Issue #538, opened 2018-09-27
Allow no space between operator **


paugier:
This is a follow up of issue https://github.com/ambv/black/issues/148 (see in particular https://github.com/ambv/black/issues/148#issuecomment-384184113 by @njsmith)

Black still adds spaces around the operator `**`.

It is really bad for readability of math formula and it is not consistent with PEP 8 (https://www.python.org/dev/peps/pep-0008/#other-recommendations).

This operator has a so high precedence (https://docs.python.org/3/reference/expressions.html#operator-precedence) so I think it should be allowed not to put spaces around it.

I know that we can use 
```
# fmt: off
t = a**2 + b**3
# fmt: on
```
for math, but many times, the spaces around `**` are the worst problems, so allowing `a**2` would save us from adding too many `# fmt: off ... # fmt: on`.

ambv:
Black does not take existing formatting into account, instead it applies uniform formatting everywhere. Hugging math operators is hard to do uniformly. See this and the next comment:
https://github.com/ambv/black/issues/148#issuecomment-383332150

I am not philosophically opposed to making Black smarter about this but I'm afraid this is less obvious if you include non-trivial operands.

ambv:
Oh, and in case this is not clear: it's out of the question for Black to "preserve user's choice". The entire program is built around the tenet of consistent formatting without looking at preexisting style. The only place where we break this rule is vertical whitespace (people are really attached to their blank lines!) and this was a source of a lot of pain for awhile (still not *fully* solved but *mostly* fine).

We want to solve this but it's rather tricky: we will need to model an approach that Black can execute on its own. Currently Black decides on whitespace for all tokens just by looking at the current one and sometimes also at its immediate predecessor. We are not looking at what comes **after** the current token (we'd have to change the algorithm not to be simply visitor-based anymore) nor make decisions based on whether other tokens received whitespace or not (with the exception of [a simple hack](https://github.com/ambv/black/blob/ed9b31b1394bc4c57bad08a065477323c458e9c3/black.py#L1666) around type-annotated function arguments).

paugier:
OK. I now understand that it is out of the question for Black to "preserve user's choice".

Then a simple solution would be to just remove white spaces around `**`.

I realize that it wouldn't always be consistent with PEP 8 (actually only when `**` is the only operator in the expression) but most of the time it would give a nicer result than with spaces around `**`.

ambv:
[...]

Issue #53, opened 2018-03-21
Create PEP 8-style document with the intended formatting style explicitly described


ambv:
This will help users identify whether what they see is a bug or a feature.

A document like this also helps with giving reasons for specific behavior.

ambv:
Yes, Sphinx and RTFD would be my first choice :-)

ambv:
PEP 8 itself is good but wordy. [Google Style Guide](https://google.github.io/styleguide/pyguide.html) is also good.

willingc:
@ambv @carljm Giving some thought to this. In some cases, *Original* and *After Black* will be helpful to illustrate what will happen. Initially, I'm starting with something similar to: 

- Guide contents: https://test-black.readthedocs.io/en/style-guide/style_guide/style_guide.html
- Rule example: https://test-black.readthedocs.io/en/style-guide/style_guide/guards.html

My current thinking is that Original/After Black examples will cut down on wordiness in PEP8.

Side by side code displays could be cool too.

SanketDG:
@ambv If I understand correctly, this has already been superseded by "The Black Code Style", yes?

mcsitter:
[The Black code style](https://black.readthedocs.io/en/stable/the_black_code_style.html) appears to b quite complete. Are there things missing?

felix-hilden:
Another ping, since this issue is on the "Amazing documentation" project. I do second the earlier commenters. Is there a need for further documentation on this front?

ichard26:
@felix-hilden I think the state of the style documentation is way better as of lately, but there's still a few areas which are missing:

1. the extra string handling Black has (currently locked away under a `--experimental-string-processing` flag); and
2. the docstring formatting Black does (I wouldn't recommend trying to add docs for them right now though since they will probably change since the current handling is IIRC a mistake from some other patch)

felix-hilden:
I see, thanks! Would you accept contributions for the first one?

ichard26:
@felix-hilden indeed! I'll be *ever so annoyed* at you for worsening the eventual merge conflict I'll have to deal with when I submit my documentation reorganization PR (hopefully soon!), but that's more on me for not finishing it quick enough ;-)

felix-hilden:
I'll make sure then to beat you to the punch by the *smallest possible margin* 😄

From the discussions in #1740 I might also try to include some info on the docstring processing.

felix-hilden:
[...]

PR #2095, opened 2021-04-08
Allow no space in doublestar when math operation (#538) (#2095)

dpalmasan:
Simple fix for (#538), changes include:

- Allow no space `doublestar` in math operations
- Add tests for such scenario in which the option is provided.

The approach is simple, but it might be a good solution for the simple case of `**` as operator. When creating the destination contents, it will check on `**` based on `Line` `leaves` and recreate the line changing the `prefix` for the requested tokens (e.g. `a**b` is found, then `**` has a space prefix and `b` has a space prefix, we just re-prefix with empty `""` string). If there is just one `**` it will hug the operands.


tests/data/doublestar_no_spaces.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3.7
+
+
+def function(**kwargs):
+    t = a**2 + b**3
```
JelleZijlstra, line 5: Should also test what happens if there *is* a space in the input.

JelleZijlstra (commented):
The new option is actually a bit different than existing `--skip` options: it means "format things differently", not "leave this area of formatting alone", as the previous options did.

I'd prefer we don't merge this change.

tests/data/doublestar_no_spaces.py
dpalmasan: Maybe is it more a naming convention for the new option, maybe `skip` the chosen name is not the best, and I can change it. Regarding to the spaces question, it will yield the same output (meaning, `DOUBLESTAR` without spaces).

Does your comment mean, update the code and not merge yet, or does it mean: These changes are not convenient, forget about them?

I am sorry if I don't understand correctly it is my first PR on an open-source repo.

ichard26:
To second Jelle's dislike, I would not be happy merging this as long as the new flag is introduced. Black is growing a lot of new options and this makes the situation worse. The lack of configuration is a defining feature of Black and I wouldn't be pleased to see that wither away. Because I'm quite busy ATM, I recommend reading the discussion on this PR: https://github.com/psf/black/pull/1904 where in particular I share the same concerns: https://github.com/psf/black/pull/1904#issuecomment-757101552.

Thanks for the PR and the time investment.

dpalmasan:
Okay, I understand. The only reason I added the flag, was to not alter the default behavior. However, if I remove the flag and always do that on `DOUBLESTAR` operators, would that suffice for a #538 fix?

Thanks for the reviews!

dpalmasan:
Okay, I updated the PR removing the flag. I also added what was discussed on (#538 ) if just one operator is found, not hug the operands with the operators.

tests/data/doublestar_no_spaces.py
felix-hilden: Not sure, but I think Jelle meant you should add test cases which have spaces around the operators, which I think you've done in the code just below! @JelleZijlstra what do you reckon?

I'm not sure about the test writing practices of Black, but I see lots of tests being written a bit more condensed. In your file it's nice to see immediately what's expected from the function name, but another option would be to simply have all the cases in one succession. That's up for debate though!

felix-hilden:
As per #53, I'm asking the maintainers: should the changes here be also added to the Black code style document?

src/black/__init__.py
```diff
@@ -1078,6 +1078,7 @@ def f(
         for line in transform_line(
             current_line, mode=mode, features=split_line_features
         ):
+            line = fix_doublestar_in_op(line)
```
ichard26, line 1081: I'm a little concerned by the performance impact this will have. Doing some rough testing, formatting `src/black/__init__.py` spends ~6500ms in the `format_str` method and ~4% (~300ms) of the total runtime was spent in the `fix_doublestar_in_op`... which is IMO too much for such a simple case like this. 

<details>

<summary>Profiling data (warning: extremely large image)</summary>

out

</details>

Is it possible to make this more efficient? Cloning the leaf seems to be where most of 300ms cost is coming from. Perhaps moving the logic somewhere into one of the transformers called in `transform_line` could help since some lines might not need this fixing? I'm sorry for my unadvice, I usually don't review formatting code.

ichard26 (commented):
While I can't review the actual formatting code since I'm too inexperienced, I do have other comments I'd like to mention. 

Thank you for the time and effort!

CHANGES.md
```diff
@@ -4,6 +4,8 @@
 
 #### _Black_
 
+- `Black` wraps `**` if in math expression with more than 1 op (#2095)
```
felix-hilden, line 7: I like that more verbose wording too! And since it's not updated yet let's grammar-nazi it up just a wee bit: "unless it's *the* only operation". First I thought it read something like "unless it's *just* operation" which confused me a bit. Obvious on closer inspection but worth a correction in my mind 😄

felix-hilden: Damn this should have been a reply to @ichard26, see his suggested changes above.

src/black/__init__.py
dpalmasan: Ohh okay, gonna take a look. Thanks for pointing out!

tests/data/expression.diff
```diff
@@ -19,7 +19,7 @@
  (~int) and (not ((v1 ^ (123 + v2)) | True))
 -+really ** -confusing ** ~operator ** -precedence
 -flags & ~ select.EPOLLIN and waiters.write_task is not None
-++(really ** -(confusing ** ~(operator ** -precedence)))
+++(really**-(confusing**~(operator**-precedence)))
```
ambv, line 22: Wait... what? You really think this is more readable in the new form? I think that once you hit a complex base or exponent, then there is no value in skipping the spaces.

src/black/__init__.py
ambv: Agreed with @ichard26, it would be cleaner if it were a StringTransformer like the other transformation. This would additionally solve the performance issue. Few enough lines in Python have DOUBLESTAR tokens that the simplest way to improve performance is just look whether there even is doublestar on the line in `do_match()`.

tupui:
Hey what is the status here?

As Ralf noted in the linked issue,`**`is not covered by the PEP8 and most (all) codes like NumPy/SciPy never add space around this operator.

In general there it would be great to allow more flexibility around for writing maths. Nobody writes maths like black does... Also remember the second section of PEP8 https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

ichard26:
Congrats @dpalmasan your change has found its way into Black eventually :tada: Thanks for submitting this PR!

Yikun:
Sorry to post comments, just curious, I would like to know have we discussed the case of `x ** 'a string'`, looks like it still not allow in current version, looks a bit inconsistent?

```
# Not allow
3 ** "a string"
# Allow
2**63 - 1
```

ichard26:
@Yikun I mean, that seems like such an edge case since strings can't be a valid operand for power operations with any of the built-in types AFAIK. Do you have a real life example where you'd write this kind of code?

PR #2096, opened 2021-04-08
Fix error from upcoming typeshed change

JelleZijlstra:
See python/typeshed#5190


JelleZijlstra:
Looks like this actually produces a type error now; we'll just have to change it as we upgrade mypy next.

cooperlees (approved):
This makes sense. Lets watch CI.

JelleZijlstra:
This variant passes CI, so I'm merging it

PR #2098, opened 2021-04-11
fix typing issue around lru_cache arguments

JelleZijlstra:
This was found by python/mypy#10308

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -302,7 +302,7 @@ def supports_feature(target_versions: Set[TargetVersion], feature: Feature) -> b
     return all(feature in VERSION_TO_FEATURES[version] for version in target_versions)
 
 
-def find_pyproject_toml(path_search_start: Iterable[str]) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


ichard26 (approved):
SGTM. Thanks!

The usage of `lru_cache()` feels a bit overkill, but it does step around the very real issue of having pass around even more variables across function calls.

P.S. hat offs to mypy_primer, it has found quite a few issues nowadays!

JelleZijlstra:
This one is a real success for mypy-primer: it found that the mypy PR fixed a bug that we didn't even know existed! That's something that would have been very hard to figure out without it.

cooperlees:
> This one is a real success for mypy-primer: it found that the mypy PR fixed a bug that we didn't even know existed! That's something that would have been very hard to figure out without it.

Mypy has a primer too? Or is that a typo (to lazy to go search)

JelleZijlstra:
Yes, it was inspired by black-primer

cooperlees:
Even down to the name! Haha.

I used “primer” as black has the whole Ford car color theme. So priming before you paint is a good thing! Nice.

PR #2099, opened 2021-04-11
Fix crash for non breaking spaces in docstring

### LINKED ISSUE #1762: INTERNAL ERROR when there is an non breaking space in a docstring
1. With the following code:
```
def plot():
    """ adds a variable to the plot according to it's conf

    >   variables : 
    >       foo:
    >           y: variable(bar)
    >       baz:
    >           color: red

    """
```
Notice that there is a non breaking space just after ` variables : `.

2. Run _Black_ on it with no argument
3. The following error happens:
```
INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_xn6q3h3t.log

--- src
+++ dst
@@ -16,11 +16,11 @@
       body=
         Expr(
           value=
             Constant(
               value=
-                "adds a variable to the plot according to it's conf\n\n>   variables :\xa0\n>       foo:\n>           y: variable(bar)\n>       baz:\n>           color: red",  # str
+                "adds a variable to the plot according to it's conf\n\n>   variables :\n>       foo:\n>           y: variable(bar)\n>       baz:\n>           color: red",  # str
             )  # /Constant
         )  # /Expr
       decorator_list=
       name=
         'plot',  # str
```

The error disappear by removing the non breaking space.

**Environment:**

- Version: black-20.8b1
- OS and Python version: Linux/Python 3.8.5
- The bug also happens on master

Pierre-Sassoulas:
This is a follow up of #2092. (Work in progress)

Closes #1762

### DIFF
--- a/tests/data/comments_non_breaking_space.py
+++ b/tests/data/comments_non_breaking_space.py
@@ -2,18 +2,19 @@ from .config import (  ConfigTypeAttributes,    Int,    Path,    # String,
     # DEFAULT_TYPE_ATTRIBUTES,
 )
 
-result = 1  # A simple comment
-result = (    1, ) # Another one
+result = 1  # A simple comment with a NBSP
+result = (    1, ) # Another NBSP
 
 result = 1    # type: ignore
-result = 1# This comment is talking about type: ignore
+result = 1# This comment is talking about type: ignore and start with NBSP
 square = Square(4) # type: Optional[Square]
 
 def function(a:int=42):
-    """ This docstring is already formatted
-       a
-       b
+    """ This docstring start with a NBSP
+        There is one NBSP column 5 here
+       This is 7 NBSP
     """
+
     #    There's a NBSP + 3 spaces before
     #    And 4 spaces on the next line
     pass


ichard26:
I'mma attach a `skip news` label under the impression that this PR finishes or fixes what was set out in PR GH-2092, which hasn't been released yet so a changelog entry isn't necessary.

JelleZijlstra:
I feel like this should actually get a different news entry because it's about nbsps in docstrings, not comments.

ichard26:
Ah, I was under the impression that PR GH-2092 had an unfixed bug which this PR fixes, now looking into it you are right @JelleZijlstra, thanks for letting me know!

ichard26:
I was going through the issue tracker and found this issue: https://github.com/psf/black/issues/1762, it seems to be what this PR aims to fix. If so, please link it so we don't have to remember to go close that issue when this is merged. Thank you in advance!

Pierre-Sassoulas:
Funny, I opened this issue in October. I was thinking it might be the same problem. Thank you for digging it !

ambv:
I added some more tests on this and found it to be fixed: https://github.com/psf/black/commit/eaa337f176b086a9ebd91884c0b9d9a96772aeb3

Let's reopen if needed but it looks solved. Thanks for your contribution.

Pierre-Sassoulas:
I just tested on the test I created, everything is working (no more crash for sure), thank you for fixing this so fast :)

Maybe a little something unrelated to the bug, with this input:
```python
def function(a:int=42):
    """ This docstring start with a NBSP
        There is one NBSP column 5 here
       This is 7 NBSP
    """

    #    There's a NBSP + 3 spaces before
    #    And 4 spaces on the next line
    pass
```
Here's the diff compared to what I was expecting:
```diff
    def function(a: int = 42):
        """This docstring start with a NBSP
-     There is one NBSP column 5 here
+      There is one NBSP column 5 here
      This is 7 NBSP
      """
  
      #    There's a NBSP + 3 spaces before
      #    And 4 spaces on the next line
      pass
```
Probably because there is 7 spaces and not 8 on the last line of the docstring ? I'm going to close #1762 either way.

PR #2100, opened 2021-04-12
Add missing instructions to make test passed

petrus-v:


petrus-v:
Get annoyed while reporting https://github.com/psf/black/issues/2101 !

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -28,7 +28,8 @@ assignees: ""
 2. Or run _Black_ on your machine:
    - create a new virtualenv (make sure it's the same Python version);
    - clone this repository;
-   - run `pip install -e .`;
+   - run `pip install -e .[d,python2]`;
+   - ensure you have install dev dependencies `pip install -r test_requirements.txt`
```
JelleZijlstra, line 32:
```suggestion
   - run `pip install -r test_requirements.txt`
```

CHANGES.md
```diff
@@ -65,6 +65,8 @@
 
 #### _Black_
 
+- Improve github issue template "running on your machine"'s instructions (#2100)
```
JelleZijlstra, line 68: I think we can skip the news entry because this change isn't relevant to people looking to upgrade

JelleZijlstra: Also, this change isn't going to be in 20.8b0

petrus-v: > I think we can skip the news entry because this change isn't relevant to people looking to upgrade

travis state will be red then, but I supposed it's not a matter here, I haven't read contributions guidlines ;)

petrus-v: reverted !

PR #1132, opened 2019-11-02
Improve String Handling

bbugyi200:
This pull request's main intention is to wrap long strings (as requested by #182); however, it also provides better string handling in general and, in doing so, closes the following issues:

Closes #26 
Closes #182
Closes #933 
Closes #1183
Closes #1243

# Examples
* f-strings will be split if they are too long (just like a normal long string would be) and the `f` prefix will be dropped when possible:
``` python
##### INPUT
fstring = f"f-strings definitely make things more {difficult} than they need to be for {{black}}. But boy they sure are handy. The problem is that some lines will need to have the 'f' whereas others do not. This {line}, for example, needs one."

##### OLD OUTPUT
fstring = f"f-strings definitely make things more {difficult} than they need to be for {{black}}. But boy they sure are handy. The problem is that some lines will need to have the 'f' whereas others do not. This {line}, for example, needs one."

##### NEW OUTPUT
fstring = (
    f"f-strings definitely make things more {difficult} than they need to be for"
    " {black}. But boy they sure are handy. The problem is that some lines will need"
    f" to have the 'f' whereas others do not. This {line}, for example, needs one."
)
```

* Manual user splits will be respected when it is possible to do so without violating the line length limit:
``` python
##### INPUT
good_split_func(
    xxx, yyy, zzz,
    long_string_kwarg="But what should happen when code has already "
                      "been formatted but in the wrong way? Like "
                      "with a space at the end instead of the "
                      "beginning. Or what about when it is split too "
                      "soon?",
)

##### OLD OUTPUT
good_split_func(
    xxx,
    yyy,
    zzz,
    long_string_kwarg="But what should happen when code has already "
    "been formatted but in the wrong way? Like "
    "with a space at the end instead of the "
    "beginning. Or what about when it is split too "
    "soon?",
)

##### NEW OUTPUT
good_split_func(
    xxx,
    yyy,
    zzz,
    long_string_kwarg=(
        "But what should happen when code has already "
        "been formatted but in the wrong way? Like "
        "with a space at the end instead of the "
        "beginning. Or what about when it is split too "
        "soon?"
    ),
)
```

* If a manual user split violates the line length limit, however, it will NOT be respected:
``` python
##### INPUT
bad_split_func(
    xxx, yyy, zzz,
    long_string_kwarg="But what should happen when code has already been formatted but in the wrong way? Like with "
                      "a space at the end instead of the beginning. Or what about when it is split too soon?",
)

##### OLD OUTPUT
bad_split_func(
    xxx,
[...]

PR #2106, opened 2021-04-14
Document experimental string processing and docstring indentation

felix-hilden:
From #53: this PR adds short descriptions of #1053 and #1132 to the Black code style document.

Possible points of review/contention:
- Has the processing changed since? Reading the changelog I don't think it has, but I haven't been around.
- Doc placement: I think the "Strings" section is appropriate for both, unless another section for indentation is created. I appended the text to the section by default, and because "experimental" hints that maybe it shouldn't be the first thing to be introduced to the reader. I'm not sure about the docstring paragraph though. It is quite short, so maybe it ought to be merged with some other paragraph.

I don't think any of the checks apply, and `skip news` could also be added. Pre-commit did line up the text though.

Happy to make any changes you suggest!

---

Also, while running the install stuff, there was some sort of process pool termination when formatting the repository. Is that something that is expected to happen on Windows?

### DIFF
--- a/docs/the_black_code_style.md
+++ b/docs/the_black_code_style.md
@@ -289,6 +289,14 @@ If you are adopting _Black_ in a large project with pre-existing string conventi
 you can pass `--skip-string-normalization` on the command line. This is meant as an
 adoption helper, avoid using this for new projects.
 
+As an experimental option, _Black_ splits long strings (using parentheses where
+appropriate) and merges short ones. When split, parts of f-strings that don't need
+formatting are converted to plain strings. User-made splits are respected when they do
+not exceed the line length limit. Line continuation backslashes are converted into
+parenthesized strings. Unnecessary parentheses are stripped. To enable experimental
+string processing, pass `--experimental-string-processing` on the command line. Because
+the functionality is experimental, feedback and issue reports are highly encouraged!
+
 ### Numeric literals
 
 _Black_ standardizes most numeric literals to use lowercase letters for the syntactic
--- a/docs/the_black_code_style.md
+++ b/docs/the_black_code_style.md
@@ -297,6 +297,13 @@ parenthesized strings. Unnecessary parentheses are stripped. To enable experimen
 string processing, pass `--experimental-string-processing` on the command line. Because
 the functionality is experimental, feedback and issue reports are highly encouraged!
 
+_Black_ also processes docstrings. Firstly the indentation of docstrings is corrected
+for both quotations and the text within, although relative indentation in the text is
+preserved. Superfluous trailing whitespace on each line and unnecessary new lines at the
+end of the docstring are removed. All leading tabs are converted to spaces, but tabs
+inside text are preserved. Whitespace leading and trailing one-line docstrings is
+removed. The quotations of an empty docstring are separated with one space.
+
 ### Numeric literals
 
 _Black_ standardizes most numeric literals to use lowercase letters for the syntactic
--- a/tests/data/docstring.py
+++ b/tests/data/docstring.py
@@ -115,6 +115,10 @@ def oneline_empty():
     '''      '''
 
 
+def oneline_nothing():
+    """"""
+
+
 def single_quotes():
     'testing'
 
@@ -264,6 +268,10 @@ def empty():
 
 
 def oneline_empty():
+    """ """
+
+
+def oneline_nothing():
     """ """
 
 


felix-hilden:
Self-review!

So I had a look over the issue tracker, and #1740 implies the existence of further docstring processing (i.e. stripping white space), but I couldn't find the original source. I did find the original Docstring opinion issue, and several by Jelle fixing crashes on docstring modifications like removing trailing whitespace. But I got pretty lost pretty quickly.

Should we just add a mention about stripping white space? Alternatively, #1740 (or a follow-up PR) could add all the information about docstring contents.

I think the long-term solution to closing #53 could be to add a section to `CONTRIBUTING.md` where contributors are instructed to modify the Black code style document when making style changes.

JelleZijlstra:
> So I had a look over the issue tracker, and #1740 implies the existence of further docstring processing (i.e. stripping white space), but I couldn't find the original source. I did find the original Docstring opinion issue, and several by Jelle fixing crashes on docstring modifications like removing trailing whitespace. But I got pretty lost pretty quickly.

Did you try looking at the test cases for docstrings? That's probably the best way to find out all the things we change. I believe it's mostly about reindentation and stripping whitespace.

Also, thanks for contributing this! I'll take a more detailed look later.

ichard26:
Yep, the docstring formatting Black does mostly boils down to reindenting and stripping whitespace, but there is an edge case where Black does move the quotes:

```diff
ichard26@acer-ubuntu:~/programming/oss/black$ cat test.py 
def test8():
    """Or like this.
    """
ichard26@acer-ubuntu:~/programming/oss/black$ black test.py --diff -q
--- test.py	2021-04-14 22:59:18.638257 +0000
+++ test.py	2021-04-14 22:59:33.679733 +0000
@@ -1,3 +1,2 @@
 def test8():
-    """Or like this.
-    """
+    """Or like this."""
```

This behaviour does have a report: https://github.com/psf/black/issues/1635 but it's unclear whether this should exist (and therefore be documented) (@JelleZijlstra I'm personally of the opinion we should get consistent about this by either uniformingly moving quotes or not moving them at all)

>  Alternatively, #1740 (or a follow-up PR) could add all the information about docstring contents.

I am leaning towards letting the proposed PRs add the information on docstring reformatting just in case we reject one (seems unlikely but always a possibility), but I don't feel strongly about this.  

> I think the long-term solution to closing #53 could be to add a section to CONTRIBUTING.md where contributors are instructed to modify the Black code style document when making style changes. It would be nice to enforce it automatically, but I don't know if a CI check about requiring that is even possible. So manual review it is I guess.

Yeah I agree to add such a section in the contributing documentation (I'll probably end up doing it myself). In terms enforcing it automatically... yeah it would difficult to do that. I do want to write PR review documentation though, so I suppose we could mention checking for this there.

> Also, while running the install stuff, there was some sort of process pool termination when formatting the repository. Is that something that is expected to happen on Windows?

One, that certainly shouldn't happen regardless of the platform and sounds like a bug. And two, is it `I/O operation on closed file` errors? This issue has been reported and fixed already but the fix hasn't been released yet (see GH-1664).

docs/the_black_code_style.md
```diff
@@ -289,6 +289,16 @@ If you are adopting _Black_ in a large project with pre-existing string conventi
 you can pass `--skip-string-normalization` on the command line. This is meant as an
 adoption helper, avoid using this for new projects.
 
+As an experimental option, _Black_ splits long strings (using parentheses where
+appropriate) and merges short ones. When split, parts of f-strings that don't need
+formatting are converted to plain strings. User-made splits are respected when they do
+not exceed the line length limit. Line continuation backslashes are converted into
+parenthesized strings. Unnecessary parentheses are stripped. To enable experimental
+string processing, pass `--experimental-string-processing` on the command line.
```
ichard26, line 297: It might be a good idea to explicitly call out that feedback (and bug reports too!) for this feature is welcomed and even encouraged. The experimental nature of it already sends that message, but IMO it can't hurt to make it explicit.

docs/the_black_code_style.md
```diff
@@ -289,6 +289,16 @@ If you are adopting _Black_ in a large project with pre-existing string conventi
...
+
+_Black_ indents docstrings to the appropriate level, but the actual contents are not
+processed further.
```
ichard26, line 300: Even without taking into account any of the many docstring PRs currently proposed, this is false. For multi-line docstrings, starting and ending whitespace is stripped (one of PRs is fixing Black to also do this for single-line docstrings). 

```console
ichard26@acer-ubuntu:~/programming/oss/black$ cat test.py --show-ends 
def test8():$
    """  $
    A$
    Or like this.  $
    """$
ichard26@acer-ubuntu:~/programming/oss/black$ black test.py 
reformatted test.py
All done! ✨ 🍰 ✨
1 file reformatted.
ichard26@acer-ubuntu:~/programming/oss/black$ cat test.py --show-ends 
def test8():$
    """$
    A$
    Or like this.$
    """$
```

ichard26 (changes requested):
Other than the docstring docs being incorrect I can't find anything to fault your documentation on. 

> Has the processing changed since?

Other than _a lot_ of bugfixes, no IIRC the processing hasn't changed.

> RE: Doc placement

I agree with your reasoning and I'm happy with where it is as of right now.

Excellent work and thank you for picking this up! I'll leave this for Jelle to review.

docs/the_black_code_style.md
felix-hilden: Yeah that's a good idea!

docs/the_black_code_style.md
felix-hilden: I think I'll have a look at the tests as Jelle proposed, add in information from them (since that is *certainly* expected behavior), and leave the rest for other issues to deal with.

felix-hilden:
> Did you try looking at the test cases for docstrings? That's probably the best way ...

Oh you're totally right! What an oversight 😅

> One, that certainly shouldn't happen regardless of the platform and sounds like a bug. And two, is it I/O operation on closed file errors?

It occured while Blackening with pre-commit for the first time (but does reproduce with it now too). It's an access denied on `semlock.__enter__()` leading to a bunch of process terminations. Seems like it's worth another bug report once I narrow it down a bit.

felix-hilden:
Sorry about the force noise 😅 I got a bit carried away with small details.

I squashed the changes into the two commits, so you can choose if you'd like to separate the commits in `master` or not, up to you! And the docstring paragraph became a lot longer, so I think it looks nice where it is.

felix-hilden:
Heya once more! With the closing of #1740, I added its style description to this PR, and added a missing test to validate that *completely* empty docstrings get the space inserted too.

Ready if you are!

JelleZijlstra:
Thank you!

PR #2112, opened 2021-04-16
Fix small comment typo

LaurentTreguier:
We probably don't need to fall back on "polling" when setting up an asyncio loop


ichard26 (approved):
Thanks for noticing the typo!

PR #2119, opened 2021-04-19
Simplify GitHub Action entrypoint (#1909)

simaki:
This PR simplifies `entrypoint.sh` for GitHub Actions by removing duplication of `args` and `black_args` (cf. #1909).

The reason why #1909 uses the input id `black_args` is to avoid an overlap with `args`, but this naming seems redundant.
So let me suggest `option` and `src`, which are consistent with CLI. Backward compatibility is guaranteed; Users can still use `black_args` as well.

Any suggestions for improvements are more than welcome. Thanks!

This PR may not require a changelog entry.

### DIFF
--- a/action/entrypoint.sh
+++ b/action/entrypoint.sh
@@ -1,17 +1,8 @@
-#!/bin/bash
-set -e
+#!/bin/bash -e
 
-# If no arguments are given use current working directory
-black_args=(".")
-if [ "$#" -eq 0 ] && [ "${INPUT_BLACK_ARGS}" != "" ]; then
-  black_args+=(${INPUT_BLACK_ARGS})
-elif [ "$#" -ne 0 ] && [ "${INPUT_BLACK_ARGS}" != "" ]; then
-  black_args+=($* ${INPUT_BLACK_ARGS})
-elif [ "$#" -ne 0 ] && [ "${INPUT_BLACK_ARGS}" == "" ]; then
-  black_args+=($*)
-else
-  # Default (if no args provided).
-  black_args+=("--check" "--diff")
-fi
+if [ -n $INPUT_BLACK_ARGS ]; then
+  echo 'WARNING: input `with.black_args` is deprecated. Use `with.options` and `with.src` instead.'
+  black $INPUT_BLACK_ARGS
+  exit $?
 
-sh -c "black . ${black_args[*]}"
+black $INPUT_OPTIONS $INPUT_SRC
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Sorry I missed this. Looks good to me. Lets merge with the deprecated option and remove that in a few months.

My only question, is there a way we can make CI using `black_args` indicate deprecate warnings to the users?

action/entrypoint.sh
```diff
@@ -1,17 +1,8 @@
...
-elif [ "$#" -ne 0 ] && [ "${INPUT_BLACK_ARGS}" != "" ]; then
-  black_args+=($* ${INPUT_BLACK_ARGS})
-elif [ "$#" -ne 0 ] && [ "${INPUT_BLACK_ARGS}" == "" ]; then
-  black_args+=($*)
-else
-  # Default (if no args provided).
-  black_args+=("--check" "--diff")
-fi
+if [ -n $INPUT_BLACK_ARGS ]; then
+  echo 'WARNING: input `with.black_args` is deprecated. Use `with.options` and `with.src` instead.'
```
simaki, line 4:
```suggestion
  echo '::warning::WARNING: input `with.black_args` is deprecated. Use `with.options` and `with.src` instead.'
```

simaki:
Yes, we can indicate deprecate warnings to the users: [Setting a warning message - GitHub Actions](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message).

Please commit the above suggestion to make this change.

ichard26:
If no one objects before then, I'll be merging this in 30ish minutes. Cooperlees already approved and his non-binding comment was also resolved.

ichard26:
Thank you @simaki for your contribution! This should help keep this action maintainable and easy to understand which is always nice since not everyone on the core team is great at this kind of stuff (e.g. me!)

simaki (commented):
@ichard26 Hi, could you kindly merge these fixes? I noticed these mistakes by my local run. Sorry about bothering you... Many thanks! (btw continuous testing of GH Actions would be nice but I've got no idea how to do it. seeking idea.)

ichard26:
> ichard26 Hi, could you kindly merge these fixes? I noticed these mistakes by my local run. Sorry about bothering you... Many thanks! (btw continuous testing of GH Actions would be nice but I've got no idea how to do it. seeking idea.)

yeah don't worry about, I'm honestly more surprised I didn't catch that error :disappointed:, here's the hotfix: https://github.com/psf/black/pull/2202

I got some ideas to test out our Action but they will have to wait till later once I get some time.

Secret158:
- [ ] [Missmia][NASA](Forever)Onlygirl

PR #2126, opened 2021-04-25
Work around stability errors due to optional trailing commas

ambv:
Optional trailing commas put by Black become magic trailing commas on another pass of the tool.  Since they are influencing formatting around optional parentheses, on rare occasions the tool changes its mind in terms of putting parentheses or not.

Ideally this would never be the case but sadly the decision to put optional parentheses or not (which looks at pre-existing "magic" trailing commas) is happening around the same time as the decision to put an optional trailing comma.  Untangling the two proved to be impractically difficult.

This shameful workaround uses the fact that the formatting instability introduced by magic trailing commas is deterministic: if the optional trailing comma becoming a pre-existing "magic" trailing comma changes formatting, the second pass becomes stable since there is no variable factor anymore on pass 3, 4, and so on.

For most files, this will introduce no performance penalty since `--safe` is already re-formatting everything twice to ensure formatting stability.  We're using this result and if all's good, the behavior is equivalent.  If there is a difference, we treat the second result as the binding one, and check its sanity again.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1016,7 +1016,17 @@ def format_file_contents(src_contents: str, *, fast: bool, mode: Mode) -> FileCo
 
     if not fast:
         assert_equivalent(src_contents, dst_contents)
-        assert_stable(src_contents, dst_contents, mode=mode)
+
+        # Forced second pass to work around optional trailing commas (becoming
+        # forced trailing commas on pass 2) interacting differently with optional
+        # parentheses.  Admittedly ugly.
+        dst_contents_pass2 = format_str(dst_contents, mode=mode)
+        if dst_contents != dst_contents_pass2:
+            dst_contents = dst_contents_pass2
+            assert_equivalent(src_contents, dst_contents, pass_num=2)
+            assert_stable(src_contents, dst_contents, mode=mode)
+        # Note: no need to explicitly call `assert_stable` if `dst_contents` was
+        # the same as `dst_contents_pass2`.
     return dst_contents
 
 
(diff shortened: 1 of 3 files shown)


src/black/__init__.py
```diff
@@ -1016,7 +1016,17 @@ def format_file_contents(src_contents: str, *, fast: bool, mode: Mode) -> FileCo
...
+
+        # Forced second pass to work around optional trailing commas (becoming
+        # forced trailing commas on pass 2) interacting differently with optional
+        # parentheses.  Admittedly ugly.
+        dst_contents_pass2 = format_str(dst_contents, mode=mode)
+        if dst_contents != dst_contents_pass2:
+            dst_contents = dst_contents_pass2
+            assert_equivalent(src_contents, dst_contents, pass_num=2)
+            assert_stable(src_contents, dst_contents, mode=mode)
+        # Note: no need to explicitly call `assert_stable` if `dst_contents` was
```
JelleZijlstra, line 1028: We still need to call assert_equivalent though.

JelleZijlstra (changes requested):
Thanks for submitting this! Hopefully we can get it in soon.

src/black/__init__.py
ambv: We did, above in line 1018.

JelleZijlstra (approved):

src/black/__init__.py
JelleZijlstra: So we do. Sorry for the noise!

jayaddison:
@ambv Does this make it possible to remove the `expectedFailure` decorators [here](https://github.com/psf/black/blob/8672af35f052a636545e38110f0419ea92aeca0f/tests/test_black.py#L236-L255)?

ambv:
No. Those are still failing. I added new tests that pass after two calls to `format_str()`.

PR #2131, opened 2021-04-25
fix magic comma and experimental string cache flags

### LINKED ISSUE #2130: Cache does not get wiped when `--skip-magic-trailing-comma` changes
**Describe the bug** The cache is still used after you change a configuration option.

1. Format some code with Black
2. Change some formatting flag (e.g., add `skip-magic-trailing-comma = true` to your pyproject.toml)
3. Run Black on your code again and see that nothing changes
4. Figure out where Black stores the cache and `rm -rf` it
5. Run Black on your code again and see that it changes formatting

**Does this bug also happen on master?** Yes

This is because `FileMode.get_cache_key` doesn't include this flag.

JelleZijlstra:
Fixes #2130


src/black/__init__.py
```diff
@@ -273,9 +273,9 @@ class Mode:
     target_versions: Set[TargetVersion] = field(default_factory=set)
     line_length: int = DEFAULT_LINE_LENGTH
     string_normalization: bool = True
+    is_pyi: bool = False
```
JelleZijlstra, line 276: Made them the same order as they appear in the cache file name for ease of understanding.

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 ## Change Log
 
+### Unreleased
+
+- Reflect the `--skip-magic-trailing-comma` and `--experimental-string-processing` flags
```
cooperlees, line 5: NIT: To maintain the same format ...

```suggestion

#### _Black_

- Reflect the `--skip-magic-trailing-comma` and `--experimental-string-processing` flags
```

cooperlees (approved):
Looks good to me ... Just changes formatting ...

JelleZijlstra:
@cooperlees I moved some code around in the tests to unbreak CI, could you take another look?

It turned out the DEFAULT_MODE in tests wasn't the same as the actual default on the command-line, which seemed wrong. To fix it I explicitly turned on `--experimental-string-processing` for the tests that rely on it.

cooperlees (commented):
These fixes make sense to me, but would love a second look from someone with a better handle on the cache and these tests implementation ...

PR #2132, opened 2021-04-26
Add automatic version tagging to Docker CI Pushes

cooperlees:
- If the git comment has a tag, set that on the docker images pushed
- If we don't have a tag, we just set `latest_non_release`

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - "master"
+  release:
+    types: created
 
 jobs:
   docker:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -24,13 +24,16 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Check + set version tag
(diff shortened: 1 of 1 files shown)


cooperlees:
The tags takes a csv of tags. So was thinking of doing a git command to grab the tag and have a fallback default when we don't.

Does anyone have better ideas / solutions here? 
GOAL: When we cut a new release and tag with a version (e.g. 21.4b1) the docker push gets the version tag too

ichard26 (commented):
Tested git logic locally and it works well. Good work!

I'd appreciate if someone else could review though to make sure I'm not missing anything.

JelleZijlstra (approved):

cooperlees:
> Tested git logic locally and it works well. Good work!
> 
> I'd appreciate if someone else could review though to make sure I'm not missing anything.
> 
> edit: also please shut up prettier for us :)

Cheers, I'm going to apply this to a personal project first to test. Probably https://github.com/cooperlees/speedtest as I'd like arm64 builds too.

friendyou1:
Thank you for these thinkers (:

cooperlees:
This now works - Tested with my personal project. You can see a 1.0.0 here that was done by this CI: https://hub.docker.com/r/cooperlees/speedtest-wrapper/tags?page=1&ordering=last_updated

PR #2152, opened 2021-04-27
Add pyanalyze and typeshed to black-primer

JelleZijlstra:
pyanalyze is one of my projects and it uses `--experimental-string-processing`.

typeshed has a lot of stub files.

### DIFF
--- a/src/black_primer/primer.json
+++ b/src/black_primer/primer.json
@@ -75,6 +75,13 @@
       "long_checkout": false,
       "py_versions": ["all"]
     },
+    "pyanalyze": {
+      "cli_arguments": [],
+      "expect_formatting_changes": false,
+      "git_clone_url": "https://github.com/quora/pyanalyze.git",
+      "long_checkout": false,
+      "py_versions": ["all"]
+    },
     "pyramid": {
       "cli_arguments": [],
       "expect_formatting_changes": true,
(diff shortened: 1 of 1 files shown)


src/black_primer/primer.json
```diff
@@ -75,6 +75,13 @@
       "long_checkout": false,
       "py_versions": ["all"]
     },
+    "pyanalyze": {
+      "cli_arguments": [],
```
cooperlees, line 79: I think we need to add `--experimental-string-processing` here
```suggestion
      "cli_arguments": ["--experimental-string-processing"],
```

src/black_primer/primer.json
```diff
@@ -110,6 +117,13 @@
       "long_checkout": false,
       "py_versions": ["all"]
     },
+    "typeshed": {
```
cooperlees, line 120: Nice! Didn't know type shed had python we would format!

cooperlees (changes requested):

src/black_primer/primer.json
JelleZijlstra: It's in the pyproject.toml, does that not get picked up?

src/black_primer/primer.json
JelleZijlstra: We can format stubs too! That makes typeshed a good way to test our stub formatting.

cooperlees (approved):

Issue #2153, opened 2021-04-27
primer.json is not installed


hroncok:
**Describe the bug** The black-primer tool uses primer.json on runtime, but the file is not installed.

**To Reproduce** Steps to reproduce the behavior:

```
[tmp]$ python -m venv venv
[tmp]$ . venv/bin/activate
(venv) [tmp]$ pip install black
...
Installing collected packages: toml, click, appdirs, mypy-extensions, pathspec, regex, black
Successfully installed appdirs-1.4.4 black-21.4b1 click-7.1.2 mypy-extensions-0.4.3 pathspec-0.8.1 regex-2021.4.4 toml-0.10.2
(venv) [tmp]$ which black-primer
~/tmp/venv/bin/black-primer
(venv) [tmp]$ black-primer
Usage: black-primer [OPTIONS]
Try 'black-primer -h' for help.

Error: Invalid value for '-c' / '--config': Path '/home/churchyard/tmp/venv/lib64/python3.9/site-packages/black_primer/primer.json' does not exist.
```

**Expected behavior** `site-packages/black_primer/primer.json` should be installed

**Environment (please complete the following information):**

- Version: at least 20.8b1 to 21.4b1 but possible any
- OS and Python version: Fedora Linux / Python 3.9.4 (but also happens on any other version)

**Does this bug also happen on master?** Yes

```
(venv) [tmp]$ SETUPTOOLS_SCM_PRETEND_VERSION=21.5 pip install -U https://github.com/psf/black/archive/refs/heads/master.zip
...
Successfully installed black-21.5
(venv) [tmp]$ black-primer
...

Error: Invalid value for '-c' / '--config': Path '/home/churchyard/tmp/venv/lib64/python3.9/site-packages/black_primer/primer.json' does not exist.
```

hroncok:
Working on a fix

ichard26:
A simple package_data mod should fix it:

https://github.com/psf/black/blob/1728bb441e6cc3ba5a5ce52ec62014bbfa435a82/setup.py#L67

Thanks for noticing!

hroncok:
Yes, I have that ready, now going trough https://github.com/psf/black/blob/master/CONTRIBUTING.md

hroncok:
Ready in https://github.com/psf/black/pull/2154

cooperlees:
[...]

PR #2154, opened 2021-04-27
Install primer.json with black

hroncok:
Fixes https://github.com/psf/black/issues/2153

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 ## Change Log
 
+### Unreleased
+
+#### _Packaging_
+
+- Install `primer.json` (used by `black-primer` by default) with black. (#2154)
+
 ### 21.4b1
 
 #### _Black_
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,11 @@ setup(
     ext_modules=ext_modules,
     packages=["blackd", "black", "blib2to3", "blib2to3.pgen2", "black_primer"],
     package_dir={"": "src"},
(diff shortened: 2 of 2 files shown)


JelleZijlstra (approved):

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 ## Change Log
 
+### Unreleased
+
+#### _Packaging_
+
+- Install `primer.json` with black. (#2153)
```
cooperlees, line 7: Bummer ... so close!
```suggestion
- Install `primer.json` with black. (#2154)
```

cooperlees (changes requested):

CHANGES.md
ichard26: Might be worth it for mention this is only for black-primer (not general black users)?

JelleZijlstra: It should be the PR number, makes it easier to check automatically that it's correct.

cooperlees: I updated via web.

ichard26: Our changelog policy to give the contributor the credit by linking the PR number. I was going to link our changelog policy but I just realized we don't mention that it's the PR number we're looking for...

I should fix that, sorry !

hroncok: That was non-obvious for me. ack

hroncok:
Amended.

cooperlees:
> Amended.

Thanks - But the fact you had to force push should have set off alarm bells. I had already done the change.

hroncok:
You've only changed the number ;)

PR #2158, opened 2021-04-27
Ignore inaccessible user config

### LINKED ISSUE #2157: Exception raised if installed under root.
**Describe the bug**
If black is installed under root and run as restricted user following exception is raised:
```
python3 -m black --diff doc gcovr
Running chmod with sudoTraceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.6/dist-packages/black/__main__.py", line 3, in <module>
    patched_main()
  File "/usr/local/lib/python3.6/dist-packages/black/__init__.py", line 7002, in patched_main
    main()
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 829, in __call__
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 781, in main
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 700, in make_context
    self.parse_args(ctx, args)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1048, in parse_args
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1630, in handle_parse_result
    value = invoke_param_callback(self.callback, ctx, self, value)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 123, in invoke_param_callback
    return callback(ctx, param, value)
  File "/usr/local/lib/python3.6/dist-packages/black/__init__.py", line 337, in read_pyproject_toml
    value = find_pyproject_toml(ctx.params.get("src", ()))
  File "/usr/local/lib/python3.6/dist-packages/black/__init__.py", line 315, in find_pyproject_toml
    return str(path_user_pyproject_toml) if path_user_pyproject_toml.is_file() else None
  File "/usr/lib/python3.6/pathlib.py", line 1362, in is_file
    return S_ISREG(self.stat().st_mode)
  File "/usr/lib/python3.6/pathlib.py", line 1158, in stat
    return self._accessor.stat(self)
  File "/usr/lib/python3.6/pathlib.py", line 387, in wrapped
    return strfunc(str(pathobj), *args)
PermissionError: [Errno 13] Permission denied: '/root/.config/black'
```

**To Reproduce**

1. Install black as root
2. Run `python3 -m black --diff .`
3. See error (e.g. https://github.com/Spacetown/gcovr/runs/2450843804?check_suite_focus=true)

**Expected behavior**

The code shall formatted the same way as in `20.8b1, see https://github.com/Spacetown/gcovr/runs/2433190163?check_suite_focus=true.

- Version: 21.4b1
- OS and Python version: Docker image based on ubuntu:18.04 with Python 3.6.7.

This is the current master.

**Additional context**
-

JelleZijlstra:
Fixes #2157

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -313,7 +313,11 @@ def find_pyproject_toml(path_search_start: Tuple[str, ...]) -> Optional[str]:
 
     try:
         path_user_pyproject_toml = find_user_pyproject_toml()
-        return str(path_user_pyproject_toml) if path_user_pyproject_toml.is_file() else None
+        return (
+            str(path_user_pyproject_toml)
+            if path_user_pyproject_toml.is_file()
+            else None
+        )
     except PermissionError:
         # We do not have access to the user-level config directory, so ignore it.
         return None
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -318,8 +318,9 @@ def find_pyproject_toml(path_search_start: Tuple[str, ...]) -> Optional[str]:
             if path_user_pyproject_toml.is_file()
             else None
         )
-    except PermissionError:
+    except PermissionError as e:
         # We do not have access to the user-level config directory, so ignore it.
+        err(f"Ignoring user configuration directory due to {e!r}")
         return None
 
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -311,8 +311,12 @@ def find_pyproject_toml(path_search_start: Tuple[str, ...]) -> Optional[str]:
     if path_pyproject_toml.is_file():
         return str(path_pyproject_toml)
 
-    path_user_pyproject_toml = find_user_pyproject_toml()
-    return str(path_user_pyproject_toml) if path_user_pyproject_toml.is_file() else None
+    try:
+        path_user_pyproject_toml = find_user_pyproject_toml()
+        return str(path_user_pyproject_toml) if path_user_pyproject_toml.is_file() else None
+    except PermissionError:
+        # We do not have access to the user-level config directory, so ignore it.
+        return None
 
 
 def parse_pyproject_toml(path_config: str) -> Dict[str, Any]:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### Unreleased
 
+#### _Black_
+
+- Fix crash if the user configuration directory is inaccessible. (#2158)
+
 #### _Packaging_
 
 - Install `primer.json` (used by `black-primer` by default) with black. (#2154)


Spacetown:
@JelleZijlstra How can I use this in my requirements.txt? I've only usedversions yet.

src/black/__init__.py
```diff
@@ -311,8 +311,16 @@ def find_pyproject_toml(path_search_start: Tuple[str, ...]) -> Optional[str]:
...
-    return str(path_user_pyproject_toml) if path_user_pyproject_toml.is_file() else None
+    try:
+        path_user_pyproject_toml = find_user_pyproject_toml()
+        return (
+            str(path_user_pyproject_toml)
+            if path_user_pyproject_toml.is_file()
+            else None
+        )
+    except PermissionError:
+        # We do not have access to the user-level config directory, so ignore it.
```
cooperlees, line 322: Lets add some logging (atleast @ debug level) so we at least tell the user why their config is not being picked up.
- Could save hours of debugging for someone one day
- Silent swallowing at debug level is evil

cooperlees (changes requested):
Looks good, lets just not be silent about it to save peoples time potentially ...

JelleZijlstra:
@Spacetown for quick testing I'd just manually edit the file (`/usr/local/lib/python3.6/dist-packages/black/__init__.py` in your case, based on the stack trace in the issue). Beyond that, you can install something from git directly (documented at https://pip.pypa.io/en/stable/cli/pip_install/#git), or you can wait for us to release the fix.

It's up to @cooperlees but I think we might want to do another release to get this fix out, since the bug seems to make Black completely unusable for use cases like yours.

cooperlees (approved):

src/black/__init__.py
JelleZijlstra: Good point. I made it just unconditionally print for now. Maybe we'd want to do it only if `-v` is passed, but that's hard because this code gets executed during argument parsing.

cooperlees:
I feel this is a super low risk bug fix and releases are easy now.

Let's wait till tomorrow morning pacific time and I'll evaluate cutting another release.

JelleZijlstra:
Yes, we can also wait for other issues to come in.

Spacetown:
@JelleZijlstra On my machine it's working. The result is also available here https://github.com/Spacetown/gcovr/runs/2451934895?check_suite_focus=true.
```
python3 -m black --diff doc gcovr
Running chmod with sudoIgnoring user configuration directory due to PermissionError(13, 'Permission denied')
```

JelleZijlstra:
Thanks @Spacetown! Looks like a missing newline in the output but I think the first message isn't from Black, so that needs to be fixed elsewhere.

PR #2161, opened 2021-04-28
Remove useless shebangs in non-executable files

hroncok:
Such shebangs are only ever used if the file is executed directly, i.e.:

    $ /usr/lib/python3.9/site-packages/black_primer/cli.py

But that doesn't work:

    $ /usr/lib/python3.9/site-packages/black_primer/cli.py
    bash: /usr/lib/python3.9/site-packages/black_primer/cli.py: Permission denied

The lib file even has: "lib is a library, funnily enough"

----

I consider this **trivial** change, no changelog entry required.

### DIFF
--- a/src/black_primer/cli.py
+++ b/src/black_primer/cli.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 # coding=utf8
 
 import asyncio
--- a/src/black_primer/lib.py
+++ b/src/black_primer/lib.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 import asyncio
 import errno
 import json


src/black_primer/cli.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 # coding=utf8
```
JelleZijlstra, line 1: Maybe remove this comment too? As far as I know utf8 encoding is the default in Python 3.

cooperlees: Not on windows.

cooperlees (approved):
This removal should be fine, but they are not useless. There is tooling out there that uses the shebang to workout if a file is destined to be python2 vs. python3 and do appropriate things. For example, linters, code modding tools, as there was very little other way to know during this migration. Also, if you just `chmod +x cli.pi`, I would expect black primer to work, but we do not test that use case.

The world has moved on to python3 since ambv first wrote black I guess, so these do indeed have less significance.

Just for my curiosity, what were these hurting? Just academic / OCD reasons?

hroncok:
> Just for my curiosity, what were these hurting? Just academic / OCD reasons?

OCD and linters. In this case, rpmlint:

```
black.noarch: E: non-executable-script /usr/lib/python3.9/site-packages/black_primer/cli.py 644 /usr/bin/env python3
black.noarch: E: non-executable-script /usr/lib/python3.9/site-packages/black_primer/lib.py
```

PR #2163, opened 2021-04-28
Extend use of optional parentheses

### LINKED ISSUE #2156: Use optional parentheses more often
**Describe the style change**

For short if statements (just one `and`/`or`) black tries to get rid of optional parentheses whenever possible. One common result of that behavior is that the line is instead spilt on a function call which often reduces the readability compared to optional parentheses. Even if the optional parentheses are inserted manually, black will reformat the code.

**Examples in the current _Black_ style with _desired_ style**

Some examples taken from the pylint code base.

```py
if not is_node_in_type_annotation_context(node) and isinstance(
    node.parent, astroid.Subscript
):
    pass

# better
if (
    not is_node_in_type_annotation_context(node)
    and isinstance(node.parent, astroid.Subscript)
):
    pass
```

```python
def _supports_mapping_protocol(value: astroid.node_classes.NodeNG) -> bool:
    return _supports_protocol_method(
        value, GETITEM_METHOD
    ) and _supports_protocol_method(value, KEYS_METHOD)

# better
def _supports_mapping_protocol(value: astroid.node_classes.NodeNG) -> bool:
    return (
        _supports_protocol_method(value, GETITEM_METHOD)
        and _supports_protocol_method(value, KEYS_METHOD)
    )
```

```python
                    if getattr(
                        func, "name", None
                    ) == "iter" and utils.is_builtin_object(func):

# better
                    if (
                        getattr(func, "name", None) == "iter"
                        and utils.is_builtin_object(func)
                    ):
```

**Additional context**
For any statement with more at least two priority delimiters the above suggested syntax is already the default.
```py
return (
    is_postponed_evaluation_enabled(node)
    and value.qname() in SUBSCRIPTABLE_CLASSES_PEP585
    and is_node_in_type_annotation_context(node)
)
```

One possible implementation could be the modify these lines here and use `> 0` instead of `> 1`.
https://github.com/psf/black/blob/76e1602d60391cca1bb3b21a9ef6ed07a1bca17d/src/black/__init__.py#L6751-L6754

cdce8p:
For statements with just one priority delimiter it's often better / easier to read if optional parentheses are inserted. Some examples (from black itself):

```py
        return subscript_start is not None and any(
            n.type in TEST_DESCENDANTS for n in subscript_start.pre_order()
        )

# new
        return (
            subscript_start is not None
            and any(n.type in TEST_DESCENDANTS for n in subscript_start.pre_order())
        )
```

```py
        if id(line.leaves[string_idx]) in line.comments and contains_pragma_comment(
            line.comments[id(line.leaves[string_idx])]
        )

# new
        if (
            id(line.leaves[string_idx]) in line.comments
            and contains_pragma_comment(line.comments[id(line.leaves[string_idx])])
        ):
```

Additional examples can be seen here: https://github.com/psf/black/commit/74b47c1d8e02e71ea7ba1d1637c73f25aa905589

This is already the default if at least two delimiters with the same max priority are present:
https://github.com/psf/black/blob/b39999da7f451c285befac217f1f9a685774b34d/src/black/__init__.py#L6760-L6763
```py
# example formatting
return (
    is_postponed_evaluation_enabled(node)
    and value.qname() in SUBSCRIPTABLE_CLASSES_PEP585
    and is_node_in_type_annotation_context(node)
)
```

### Notes

The only code change is the first commit: https://github.com/psf/black/commit/4a4a3bc58a395680d53b94307bfa2e283249f4c8

Closes #2156

### Remaining Todo's
- [x] Fix broken tests
- [x] Write changelog entry

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -6776,6 +6776,13 @@ def can_omit_invisible_parens(
         # With more than one delimiter of a kind the optional parentheses read better.
         return False
 
+    if (
+        max_priority >= LOGIC_PRIORITY
+        and bt.delimiter_count_with_priority(max_priority) > 0
+    ):
+        # With at least one delimiter of kind LOGIC or above, optional parentheses are beter
+        return False
+
     if max_priority == DOT_PRIORITY:
         # A single stranded method call doesn't require optional parentheses.
         return True
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -772,12 +772,16 @@ def reformat_one(
                 res_src_s = str(res_src)
                 if res_src_s in cache and cache[res_src_s] == get_cache_info(res_src):
                     changed = Changed.CACHED
-            if changed is not Changed.CACHED and format_file_in_place(
-                src, fast=fast, write_back=write_back, mode=mode
+            if (
+                changed is not Changed.CACHED
+                and format_file_in_place(
+                    src, fast=fast, write_back=write_back, mode=mode
+                )
             ):
                 changed = Changed.YES
-            if (write_back is WriteBack.YES and changed is not Changed.CACHED) or (
-                write_back is WriteBack.CHECK and changed is Changed.NO
+            if (
+                (write_back is WriteBack.YES and changed is not Changed.CACHED)
+                or (write_back is WriteBack.CHECK and changed is Changed.NO)
             ):
                 write_cache(cache, [src], mode)
         report.done(src, changed)
@@ -884,8 +888,9 @@ async def schedule_formatting(
                 changed = Changed.YES if task.result() else Changed.NO
                 # If the file was written back or was successfully checked as
                 # well-formatted, store this information in the cache.
-                if write_back is WriteBack.YES or (
-                    write_back is WriteBack.CHECK and changed is Changed.NO
+                if (
+                    write_back is WriteBack.YES
+                    or (write_back is WriteBack.CHECK and changed is Changed.NO)
                 ):
                     sources_to_cache.append(src)
                 report.done(src, changed)
@@ -1614,9 +1619,10 @@ class Line:
     @property
     def is_stub_class(self) -> bool:
         """Is this line a class definition with a body consisting only of "..."?"""
-        return self.is_class and self.leaves[-3:] == [
-            Leaf(token.DOT, ".") for _ in range(3)
-        ]
+        return (
+            self.is_class
+            and self.leaves[-3:] == [Leaf(token.DOT, ".") for _ in range(3)]
+        )
 
     @property
     def is_def(self) -> bool:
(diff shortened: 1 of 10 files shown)


ichard26:
Hi @cdce8p,

I see you're a new contributor here, welcome! I took a quick look at your changes and I really like the new output. I probably won't be reviewing this (not qualified to do so yet) but I just wanted to provide my support for the formatting changes. Although unfortunately this does seem like the type of change that would be rejected for being too disruptive.

src/black/__init__.py
```diff
@@ -2441,10 +2461,14 @@ def whitespace(leaf: Leaf, *, complex_subscript: bool) -> str:  # noqa: C901
...
-            if prevp.type == token.COLON and prevp_parent.type in {
-                syms.subscript,
-                syms.sliceop,
-            }:
+            if (
+                prevp.type == token.COLON
+                and prevp_parent.type
+                in {
+                    syms.subscript,
+                    syms.sliceop,
```
JelleZijlstra, line 2469: This interacts a bit poorly with the magic trailing comma; we could have put this all in one line.

JelleZijlstra (commented):
I like this change too, but it's unfortunately a pretty big change.

cdce8p:
> I like this change too, but it's unfortunately a pretty big change.

I saw that too. As you might have noticed, the required code change itself isn't that big which is why I though a PR would be a good starting point for a discussion if it should be persuade any further. It also has the added benefit that anyone can try it for him-/herself.

The question now becomes how to best continue with it. The way I see it there are a few options, some better suited then others:
1. Implement it as is. This will almost definitely be quite disruptive, although the output is almost always at least equally readable if not clearly better. I first noticed this issue while working on `pylint`. Most `if` statements there tend to use `isinstance` which are often difficult to read with the current formatting. This change is a clear improvement in that regard. https://github.com/PyCQA/pylint/commit/559d778cc29703c78cafe333846ddac3c94181d4
2. Leave it as is and close PR because it's too disruptive.
3. Add feature flag to disable/enable it. In case of **disable**, there is already some president with `--skip-string-normalization` and `--skip-magic-trailing-comma`. For **enable**, it would allow the use of it with the drawback that most will probably never even know about it.
4. Instead of the current change, add logic to keep user inserted optional parentheses. That way a developer does at least have the option to use them and improve the formatting. However, that also means that they would need to be inserted **manually** each time.

Tbh I'm not sure what to do and in the end it's not my decision to make. Those are just some things I could think of and maybe there is also another way I haven't mentioned. If so, I'm eager to hear about it.

src/black/__init__.py
cdce8p: I would like to address those things once we have decided how to best continue with this PR.

cdce8p:
As a side node, I just started to look at the result of the `Primer` test and, at least IMO, almost all the changes appear to be clear improvements. https://github.com/psf/black/pull/2163/checks?check_run_id=2461939530#step:5:42

cdce8p:
@JelleZijlstra What would be your opinion on how to continue with it?

cooperlees:
I prefer this and think it makes us more consistent. So possibly worth the disruption to codebases.

Wonder if it's worth a community vote going forward with invasive changes?

cdce8p:
> I vote to merge, but I'd like to hear what other maintainers think.

> I prefer this and think it makes us more consistent. So possibly worth the disruption to codebases.

👍🏻 I've gone ahead and added a changelog entry as well as fixed the remaining tests. Should be ready for review then

cdce8p:
@ambv I appreciate the feedback. Although I don't fully agree, I get your point.

Since everyone so far agreed that the style itself is an improvement and consistent with the current one, would it at least be an option to keep user inserted parentheses in that case? That way we could use them when it makes sense without black overwriting the decision.

PR #2170, opened 2021-04-30
Do not use gitignore if explicitly passing excludes

### LINKED ISSUE #2164: gitignore rules are not prominently documented and ignore checked in files
My `.gitignore` is quite simple. It forces me to explicitly add new files or use `-u`; which is good practice and I don't have to worry about adding a bunch of exclusion rules as things pile up.

```
*
```

However black's "convenient" functionality of using the `.gitignore` to decide which files to exclude *without taking into account whether or not the file is checked in* means I cannot simply apply `black` to my source directory. 

```sh
> black src/
> rm .gitignore 
> black src/
All done! ✨ 🍰 ✨
6 files left unchanged.
> black --version
black, version 21.4b2
```

This is not the first large project I've seen where "convenient" features eventually weasel their way in. They might help the common case, but can totally defy other user's expectations and are *rarely* documented (\*cough cough\* `setuptools_scm` \*cough\*). It's mad. Resist the temptation please. Code formatting has *nothing* to do with version control. If someone thinks it's important, let it be a plugin and extra: `black[vcs]`

ktbarrett:
Closes #2164.

Changes behavior of how `.gitignore` is handled. With this change, the rules in `.gitignore` are only used as a *fallback* if no exclusion rule is explicitly passed on the command line or in `pyproject.toml`. Previously they were used regardless if explicit exclusion rules were specified, preventing any overriding of `.gitignore` rules.

Those that depend only on `.gitignore` for their exclusion rules will not be affected. Those that use both `.gitignore` and `exclude` will find that `exclude` will act more like actually specifying `exclude` and not just another `extra-excludes`. If the previous behavior was desired, they should move their rules from `exclude` to `extra-excludes`.


ktbarrett:
If this is deemed "messy" a superior solution might involve making the `exclude` parameter a callable object that defaults to a function that checks the gitignore and default rules, or just returns `match.search` bound method.

ktbarrett:
Fixed the CHANGE.md issue.

cooperlees (changes requested):
LGTM and a good outcome from your issue chat.

Please just make the linter happy.

ichard26 (changes requested):
While I'm still not 100% happy with the level of disruption this might cause, this does seem like the best solution in terms of costs vs benefit. The semantics should make much more sense after this.

> If this is deemed "messy" 

Nah, it's fine, I don't like it too much since `include` used to work the same as `exclude` but now it doesn't, but the complexity of a callable is not worth it IMO. I'd appreciate a comment describing this behaviour in the code, but other than that, I'll approve this once mine and Cooper's comments have been addressed.

**Other than my specific comments, I'd like to see some tests testing this functionality. I don't want this behaviour to change suddenly.** Honestly the fact no test failed from this change makes me feel the existing tests for this logic is not extensive enough. 

Thank you!

ktbarrett:
I find it odd that the primers all passed last night with the same logic. I'll investigate.

ichard26:
@ktbarrett looks like tox reformatted their code to the newest version of Black. In that case, just mark tox as a project with no expected changes:

https://github.com/psf/black/blob/b39999da7f451c285befac217f1f9a685774b34d/src/black_primer/primer.json#L113-L119

ktbarrett:
The failing test decided to take >1 hour and then fail, while the rest completed in <2 minutes. I'm not entirely sure what that test is doing? Is it just testing running black more than once on a source does the same thing each time? I'm just going to try it again...

ichard26:
Sounds like either bad luck or a hypothesmith bug. Anyway, I'll review this sometime soon, probably tomorrow (it's once again night where I am).

ambv:
I'm not a big fan of hiding the default value of `--exclude`. Now it no longer appears in the tool's `--help`. See what it did before and re-add it manually to the help text.

src/black/__init__.py
```diff
@@ -492,13 +492,14 @@ def validate_regex(
...
-    default=DEFAULT_EXCLUDES,
     callback=validate_regex,
     help=(
         "A regular expression that matches files and directories that should be"
         " excluded on recursive searches. An empty value means no paths are excluded."
         " Use forward slashes for directories on all platforms (Windows, too)."
-        " Exclusions are calculated first, inclusions later."
+        " Exclusions are calculated first, inclusions later. [default: /(\\.direnv|"
+        "\\.eggs|\\.git|\\.hg|\\.mypy_cache|\\.nox|\\.tox|\\.venv|venv|\\.svn|_build|"
+        "buck-out|build|dist)/]"
```
ichard26, line 502: Unless I'm missing something, substituting the `DEFAULT_EXCLUDES` variable into the help text at runtime works just fine. If we ever update the `DEFAULT_EXCLUDES` we won't have go and update the help text (that's if we even remember to do so).
```suggestion
        f" Exclusions are calculated first, inclusions later. [default: {DEFAULT_EXCLUDES}]"
```

src/black/__init__.py
```diff
@@ -492,13 +492,14 @@ def validate_regex(
...
     ),
     show_default=True,
```
ichard26, line 504: Useless now.
```suggestion
```

src/black/__init__.py
```diff
@@ -6227,8 +6234,8 @@ def gen_python_files(
         if normalized_path is None:
             continue
 
-        # First ignore files matching .gitignore
-        if gitignore.match_file(normalized_path):
+        # Ignore files matching .gitignore if no ex
```
ichard26, line 6237: ex? You mean "exclude"?

tests/test_black.py
```diff
@@ -1389,6 +1389,32 @@ def test_include_exclude(self) -> None:
         )
         self.assertEqual(sorted(expected), sorted(sources))
 
+    def test_gitingore_used_as_default(self) -> None:
```
ichard26, line 1392: Is it worth it to add another test that's like this one but the complete opposite? So like making sure .gitignore is not used when --exclude is present. OR is that already tested somewhere else (wouldn't surprised if so)?

ichard26 (changes requested):
We're getting closer to this becoming ready to merge! Good work! I got some final suggestions / comments but they're nothing major.

p.s. apologies for the wait for my review

tests/test_black.py
ktbarrett: All of the existing tests test this already since they all pass `exclude`. This is the only one that doesn't.

src/black/__init__.py
ktbarrett: I think I had more there. I guess it never got saved.

ktbarrett, line 6237:
```suggestion
        # Ignore files matching .gitignore or default if 'exclude' is not explicitly passed
```

ktbarrett:
Resolved conflicts.

ktbarrett:
Thanks everyone for the reviews and comments, and for getting this change in.

PR #2173, opened 2021-05-01
Add ability to pass posargs to pytest run in tox.ini

ktbarrett:
I found it useful with #2170 be to able to specify `-k test_im_interested_in`, `--trace`, or `--pdb`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ deps =
 commands =
     pip install -e .[d,python2]
     coverage erase
-    coverage run -m pytest tests
+    coverage run -m pytest tests {posargs}
     coverage report
 
 [testenv:fuzz]


CHANGES.md
```diff
@@ -14,6 +14,8 @@
 
 - Install `primer.json` (used by `black-primer` by default) with black. (#2154)
 
+- Add passing `posargs` to `pytest` in `tox.ini`. (#2173)
```
ichard26, line 17: I don't think this change log entry is necessary since this is only for Black contributors. The change log for user-facing changes. I'll give your PR a `skip news` label instead.

ichard26 (changes requested):
Sounds good to me! Please address my comment and I'll approve. Thank you!

ichard26 (approved):

ichard26:
I can't imagine this being controversial and it's a simple improvement to what we have right now, so I'll merge now.

PR #2175, opened 2021-05-01
Fix Primer: Tox has been formatted with Black 21.4b0

hugovk:
Tox was formatted with Black 21.4b0 in https://github.com/tox-dev/tox/pull/2046 and there are now no expected formatting changes.

I don't think this needs a changelog entry, please add a `skip news` label.


JelleZijlstra:
Thanks!

PR #2182, opened 2021-05-03
Detect `'@' dotted_name '(' ')' NEWLINE` as a simple decorator

### LINKED ISSUE #2181: False-alarm in `RELAXED_DECORATORS` detection breaks compatibility with Python < 3.9
Running Black 21.4b2 on the following code:

```python
@get_decorator()
def f():
    return u"This should be Unicode on Python 2"
```

incorrectly strips the `u`-prefix from the string, because the minimum Python version is incorrectly detected as 3.9 (via a false detection of relaxed decorator syntax).  Removing the `()` in `@get_decorator()` avoids this problem.

I'm pretty sure that the solution is to fix [`black.is_simple_decorator_expression()`](https://github.com/psf/black/blob/35e8d1560d68e8113ff926a9f832582f8f4a694f/src/black/__init__.py#L5775-L5796).

ichard26:
Previously the RELAXED_DECORATOR detection would be falsely True on that
example. The problem was that an argument-less parentheses pair didn't
pass the `is_simple_decorator_trailer` check even it should. OTOH a
parentheses pair containing an argument or more passed as expected.

Fixes #2181

I whipped up this fix really quickly so apologies if this fix is garbage :)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@
 
 - Set `--pyi` mode if `--stdin-filename` ends in `.pyi` (#2169)
 
+- Stop detecting target version as Python 3.9+ with pre-PEP-614 decorators that are
+  being called but with no arguments (#2182)
+
 ### 21.4b2
 
 #### _Black_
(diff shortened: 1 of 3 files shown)


ichard26:
Cancelled the 3.9 fuzz workflow cause it was getting stuck on Setup Python and then also seemed to exhibit the same behavior that leads to it taking forever and then failing. All of the other runs are passing just fine anyway...

JelleZijlstra (approved):

PR #2184, opened 2021-05-03
Enable ` --experimental-string-processing` on most primer projects

cooperlees:
- We want to make this default so need to test it more
- Fixed splat/star bug in extending black args for each project

Helps pave way for #2188

### DIFF
--- a/src/black_primer/lib.py
+++ b/src/black_primer/lib.py
@@ -120,7 +120,7 @@ async def black_run(
     """Run Black and record failures"""
     cmd = [str(which(BLACK_BINARY))]
     if "cli_arguments" in project_config and project_config["cli_arguments"]:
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):

cooperlees:
Hmm - Primer didn't trigger ... no idea why ...

ichard26:
> Hmm - Primer didn't trigger ... no idea why ... 

I thought I had let you know I manually disabed the primer workflow since I was iterating a LOT on that travis CI migration PR. Check the group chat again. I assumed this PR was done, but that was unreasonable. Only the fuzz workflow is disabled right now. Apologies for the disruption!

cooperlees:
I don't know how to retrigger the Primer action now :| ... But it should pass ...

Issue #1971, opened 2021-02-06
Migrate off Travis CI (.org)


ichard26:
> Looks like Travis is just never finishing, do we still need it?

Travis CI for Open Source is shutting down in a few weeks so the queue for jobs is insane due to lower resources. I'm 99.99% sure we don't need it as our Test, Lint, Docs, Upload / Package, Primer, and Fuzz workflows are all on GitHub Actions. So even though we *can* migrate to the .com version with its 1000 free Linux minutes(?), I don't think we need to.

([more info here](https://blog.travis-ci.com/oss-announcement), [here](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing), and [also here](https://docs.travis-ci.com/user/migrate/open-source-repository-migration))

_Originally posted by @ichard26 in https://github.com/psf/black/issues/1953#issuecomment-771213516_

---

Assuming that we are not migrating to the .com version, there's a few things we should do before travis-ci.org shuts down:

- [Remove Travis CI build badge in docs ](https://github.com/psf/black/blob/master/docs/conf.py#L317) and the [readme](https://github.com/psf/black/blob/master/README.md#the-uncompromising-code-formatter)
- [Remove Travis CI config ](https://github.com/psf/black/blob/master/.travis.yml)
- Migrate Coveralls to use GitHub Actions as the coverage collection source instead of Travis CI
  - [We'll probably use this](https://github.com/marketplace/actions/coveralls-github-action) (I have no idea how to configure this correctly)

edit: removed a stupid colon my brain automatically puts after parentheses lol

ichard26:
>   -  - We'll probably use this (I have no idea how to configure this correctly)

Nope! It doesn't support Python unfortunately

LoopThrough-i-j:
Just out of curiosity: Why use coveralls and not codecov?

ichard26:
> Why use coveralls and not codecov?

Can't tell ya, I joined this project somewhat recently (March 2020 I think?) so core decisions like this one were made far before I was ever involved. It's probably @ ambv who knows why (purposefully not @ ing them since they don't need more notifications).

LoopThrough-i-j:
ahh ok, I understand.

PR #2186, opened 2021-05-03
Drop Travis CI and migrate Coveralls

ichard26:
Travis CI for Open Source is shutting down in a few weeks so the queue
for jobs is insane due to lower resources. I'm 99.99% sure we don't need
it as our Test, Lint, Docs, Upload / Package, Primer, and Fuzz workflows
are all on GitHub Actions. So even though we *can* migrate to the .com
version with its 1000 free Linux minutes(?), I don't think we need to.

more information here:
- https://blog.travis-ci.com/oss-announcement
- https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
- https://docs.travis-ci.com/user/migrate/open-source-repository-migration

This commit does the following:
- delete the Travis CI configuration
- add to the GHA test workflows so coverage continues to be recorded
  - tweaked coverage configuration so this wouldn't break
- remove any references to Travis CI in the docs (i.e. readme + sphinx
  docs)

Regarding the Travis CI to GitHub Actions Coveralls transition, the
official action doesn't support the coverage files produced by coverage.py
unfornately. Also no, I don't really know what I am doing so don't @ me
if this breaks :p (well you can, but don't expect me to be THAT useful).

The Coveralls setup has two downfalls AFAIK:
- Only Linux runs are used because AndreMiras/coveralls-python-action
  only supports Linux. Although this isn't a big issue since the Travis
  Coveralls configuration only used Linux data too.
- Pull requests from an internal branch (i.e. one on psf/black) will be
  marked as a push coverage build by Coveralls since our anti-duplicate-
  workflows system runs under the push even for such cases.

Resolves GH-1971.


cooperlees (approved):
LGTM ...

ichard26:
It's still very disfunctional.

ichard26:
Oh dear that took a long time to sort out, the silver lining is that 1) this works! :tada: and 2) I learned a lot about GitHub Actions.

PR #2196, opened 2021-05-04
Speed up test suite via distributed testing

ichard26:
Since we now run the test suite twice, one with Python 2 and another
without, full test runs are getting pretty slow. Let's try to
fix that with parallization.

Also use verbose mode on CI since more logs is usually better since
getting more is quite literally impossible.

The main issue we'll face with this is we'll hit
pytest-dev/pytest-xdist#620 sometimes
(although pretty rarely). I suppose we can test this and see if how bad
this bug is for us, and revert if necessary down the line.

Also let's have some colours :tada:

---

Comparison:

After PR:

Screenshot from 2021-05-04 16-10-44
https://github.com/psf/black/actions/runs/811254652

Before PR:

Screenshot from 2021-05-04 16-12-25
https://github.com/psf/black/actions/runs/811220250

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,9 +9,9 @@ deps =
 commands =
     pip install -e .[d]
     coverage erase
-    pytest tests --run-optional=no_python2 --numprocesses auto --cov {{posargs}
+    pytest tests --run-optional no_python2 --numprocesses auto --cov {posargs}
     pip install -e .[d,python2]
-    pytest tests --run-optional=python2 --numprocesses auto --cov --cov-append {posargs}
+    pytest tests --run-optional python2 --numprocesses auto --cov --cov-append {posargs}
     coverage report
 
 [testenv:fuzz]
--- a/tox.ini
+++ b/tox.ini
@@ -9,9 +9,9 @@ deps =
 commands =
     pip install -e .[d]
     coverage erase
-    coverage run -m pytest tests -m "not python2" {posargs}
+    pytest tests -m "not python2" --numprocesses auto --cov {posargs}
     pip install -e .[d,python2]
-    coverage run -m pytest tests -m "not without_python2" {posargs}
+    pytest tests -m "not without_python2" --numprocesses auto --cov --cov-append {posargs}
     coverage report
 
 [testenv:fuzz]
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ omit =
   src/blib2to3/*
   tests/data/*
   */site-packages/*
+  .tox/*
 
 [run]
 relative_files = True
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Unit tests
         run: |
-          tox -e py
+          tox -e py -- -v --color=yes
 
       - name: Publish coverage to Coveralls
         # If pushed / is a pull request against main repo AND
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -3,5 +3,7 @@ pre-commit
 pytest >= 6.1.1
 pytest-mock >= 3.3.1
 pytest-cases >= 2.3.0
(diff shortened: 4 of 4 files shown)


ichard26:
I love how after the force push the [Test workflow](https://github.com/psf/black/actions/runs/811285467) takes 7 minutes to run, I swear there are improvements and this isn't useless :rofl:

cooperlees:
> I _love_ how after the force push the [Test workflow](https://github.com/psf/black/actions/runs/811285467) takes 7 minutes to run, I swear there are improvements and this isn't useless

Compare it locally on a controlled platform (i.e. you box). That's enough evidence.

tox.ini
```diff
@@ -9,9 +9,9 @@ deps =
 commands =
     pip install -e .[d]
     coverage erase
-    coverage run -m pytest tests -m "not python2" {posargs}
+    pytest tests -m "not python2" --numprocesses auto --cov src/ --cov tests/ {posargs}
```
cooperlees, line 12: pytest does not respect `.coveragerc` ?

cooperlees (changes requested):
GitHub Actions gives us 1 CPU core, so that is why `auto` is not helping here:

From https://pypi.org/project/pytest-xdist/:
```
Pass -n auto to use as many processes as your computer has CPU cores.
```

We might need to just pass `-n 2` here?

The dream is `auto` for developers and something smarter for actions if doable.

tox.ini
ichard26: I'm not actually sure, but good point, I'll check again.

ichard26:
> GitHub Actions gives us 1 CPU core

This is false, but I see where you could get that understanding. The reason for my previous comment was because the GHA Test workflow run before actually saw performance improvements but the one after the forcepush didn't. This change does have improvements, but I was laughing because that workflow run wasn't exactly good at proving this change is beneficial. 

Anyway, in these logs you can tell the tests are running in parallel:

Linux: https://github.com/psf/black/pull/2196/checks?check_run_id=2503919997#step:6:60 
MacOS: https://github.com/psf/black/pull/2196/checks?check_run_id=2503920027#step:5:59
Windows: https://github.com/psf/black/pull/2196/checks?check_run_id=2503920051#step:5:59

Matter of fact, MacOS has three cores! These claims are also backed up the [GitHub Actions documentation](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

But I checked again locally and there are performance improvements:

`4m 33s 362ms` without this change and `3m 53s 547ms ` with this change for a `tox -e py38 -r` run.

tox.ini
ichard26: Okay so `pytest-cov` does respect `.coveragerc`. I just assumed that `--cov` needed an argument so coverage would be collected but I was wrong. `pytest tests -m "not python2" --numprocesses auto --cov {posargs}` works just fine. Great catch! Thank you!

cooperlees (approved):
Sweet - Thanks for double checking. Seems 2 cores for linux/windows and 3 for Mac OS X reading docs.

ichard26:
This \*should\* be ready to go now that I fixed the broken syntax in tox.ini (thanks Łukasz for fixing the conflict!):

Screenshot from 2021-05-07 10-35-39

I'm not merging my own PR though.

Dwalinnewyork:
[AUTHORS.md](https://github.com/psf/black/files/6896889/AUTHORS.md)
[CHANGES.md](https://github.com/psf/black/files/6896890/CHANGES.md)
[CONTRIBUTING.md](https://github.com/psf/black/files/6896892/CONTRIBUTING.md)
[README.md](https://github.com/psf/black/files/6896893/README.md)
[test_requirements.txt](https://github.com/psf/black/files/6896894/test_requirements.txt)

cooperlees:
@Dwalinnewyork - Why did you share this? What's your goal here?

PR #2204, opened 2021-05-07
Mark blackd tests with the `blackd` optional marker

ambv:
This is a follow-up of #2203 that uses a pytest marker instead of a bunch of `skipUnless`.  Similarly to the Python 2 tests, they are running by default and will crash on an unsuspecting contributor with missing dependencies.  This is by design, we WANT contributors to test everything.  Unless we actually don't
and then we can run:

```
pytest --run-optional=no_blackd
```

Relatedly, bump required aiohttp to 3.6.0 at least to get rid of expected failures on Python 3.8 (see 6b5eb7d4651c7333cc3f5df4bf7aa7a1f1ffb45b).

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -20,7 +20,7 @@ wheel = ">=0.31.1"
 black = {editable = true, extras = ["d"], path = "."}
 
 [packages]
-aiohttp = ">=3.3.2"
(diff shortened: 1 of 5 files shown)


zsol (approved):
Looks great!

PR #2206, opened 2021-05-07
Refactor `src/black/__init__.py` into many files

ambv:
It's the same code but in smaller chunks.


src/black/strings.py
```diff
@@ -0,0 +1,319 @@
...
+def sub_twice(regex: Pattern[str], replacement: str, original: str) -> str:
+    """Replace `regex` with `replacement` twice on `original`.
+
+    This is used by string normalization to perform replaces on
+    overlapping matches.
+    """
+    return regex.sub(replacement, regex.sub(replacement, original))
+
+
+def format_hex(text: str) -> str:
```
JelleZijlstra, line 25: This is for numeric literals, not strings. It should go wherever `normalize_numeric_literals` went.

src/black/strings.py
```diff
@@ -0,0 +1,319 @@
...
+    """
+    Returns:
+        True iff @string starts with three quotation characters.
+    """
+    raw_string = string.lstrip(STRING_PREFIX_CHARS)
+    return raw_string[:3] in {'"""', "'''"}
+
+
+@mypyc_attr(patchable=True)
+def dump_to_file(*output: str, ensure_final_newline: bool = True) -> str:
```
JelleZijlstra, line 86: This has little to do with strings, it should go into debug.py or files.py I think.

ichard26:
Cooper wants this refactor to publicized in the CHANGELOG. I don't feel strongly, but I'm putting the idea out there. I think their rational was that this may introduce bugs and crashes so warning users (explicitly or implying) is a good idea.

JelleZijlstra:
> they want this refactor to publicized in the CHANGELOG

I agree. This change will likely break users who `import black` and use some functions. They shouldn't do that since it's undocumented and all, but the change has a reasonable chance of affecting users, so we should publicize it.

JelleZijlstra:
Also, thanks @ambv for getting this done! This will make Black's code a lot easier to maintain and understand and I'm happy we're finally doing it!

CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 #### _Black_
 
+- Refactor `src/black/__init__.py` into many files (#2206)
```
ichard26, line 7: not released yet, although will be soon!

```suggestion
### Unreleased

#### _Black_

- Refactor `src/black/__init__.py` into many files (#2206)

#### _Black_

```

JelleZijlstra: Thanks, fixed

PR #2209, opened 2021-05-08
Stop docker CI from looking for 'master' branch

cooperlees:


ichard26 (approved):
\*insert dramatic intro music\*

> **Starting today, one (important) reference to `master` shall be killed.** 

\*slowly reduces volume to make it more sombre\*

> _ Although, the docs, gallery, and probably other places have more refs to `master`._ We must eliminate them all, before, they kill us all.

\*cuts music because this is dumb\*

Anyway, thank you!

PR #2210, opened 2021-05-09
Replace references to master branch

panos:
- I left out one instance of `master` under [`CHANGES.md`](CHANGES.md).
- I manually checked the URL and adjusted one URL that was invalid either way:
https://github.com/panos/black/blob/b59c54ad75f7f6c11c791a03b5fe84ab2b52f37c/.flake8#L4


JelleZijlstra (approved):

panos:
Is there something specific that I've happened to miss that broke the checks, or was it because of the tree I used?

cooperlees:
Just seem `.github/ISSUE_TEMPLATE/bug_report.md` formatting just need to be fixed by prettier.

panos:
> Just seems `.github/ISSUE_TEMPLATE/bug_report.md` formatting just need to be fixed by prettier.
> 
> If you run `pre-commit` locally it'll fix it.

On it, I'll also add a [`CHANGE.md`](CHANGE.md) message and force-push. Thank you!

ichard26:
Sorry, but there's more references to `master` in `gallery/`, an utility script. Could you please fix them? Thank you!

Also good call on the changelog entry, this next release is going to break a lot of things... and this change will probably make that breakage even worse :laughing:  I'm on the fence about putting this change under documentation since while these changes are in the docs, for users any CI, random scripts, etc. references to `master` they have don't make sense from a documentation point of view. But then again, I can't think of anything better (Black and packaging also aren't IMO good for similar reasons).

panos:
> Sorry, but there's more references to `master` in `gallery/`, an utility script. Could you please fix them? Thank you!

Sure thing, I was actually just very hesitant to fix them. I'll see what I can do and get back to you if there's anything wrong.

I'll note that I'm also going to change this line, which is a reference to `cpython` and irrelevant to black:

https://github.com/psf/black/blob/1bedc176d19f0e71c24149c696c5a2a71f21e941/gallery/gallery.py#L77

> Also good call on the changelog entry, this next release is going to break a lot of things... and this change will probably make that breakage even worse 

I don't think it will, it's mostly just documentation-related changes. Unless if you're referring to the transition from the `master` branch to `main`. GitHub just redirects that sort of stuff, sometimes.

panos:
Additional note: I frankly have no idea why the CI check is freaking out. The test passes on my machine, can't see any tab/space-related sort of difference.

panos:
I added an additional note regarding some source code changes, but I didn't see the changes as substantial enough to move the log away from the `Documentation` section.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -18,13 +18,13 @@ assignees: ""
...
+- Version: \[e.g. main\]
 - OS and Python version: \[e.g. Linux/Python 3.7.4rc1\]
 
-**Does this bug also happen on master?** To answer this, you have two options:
+**Does this bug also happen on main?** To answer this, you have two options:
 
-1. Use the online formatter at <https://black.vercel.app/?version=master>, which will
-   use the latest master branch.
+1. Use the online formatter at <https://black.vercel.app/?version=main>, which will use
+   the latest main branch.
```
ichard26, line 27: Looks like the Black Playground breaks on this version value, I'll see I can submit a PR over there later today so this doesn't stay broken.

.flake8
```diff
@@ -1,7 +1,7 @@
 [flake8]
 extend-ignore = E203, E266, E501
 # line length is intentionally set to 80 here because black uses Bugbear
-# See https://github.com/psf/black/blob/master/docs/the_black_code_style.md#line-length for more details
+# See https://github.com/psf/black/blob/main/docs/the_black_code_style/current_style.md#line-length for more details
```
ichard26, line 4: I'd rather move references to documentation to ReadTheDocs. We're moving away from GitHub as a documentation host to only RTD because it's makes our lives easier creating good docs. I know this link is dead right now, but it won't be once we release a new version with the documentation reorganization changes (which should be soon!).

```suggestion
# See https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length for more details
```

ichard26 (approved):
Thank you! Congrats on your first contribution to _Black_.

I'll commit my suggestion myself so you don't have to spend more time on this. I'll merge once CI passes.

ichard26:
I may or may not have forgotten about this PR... whoops!

Clicking the magical green button now.

PR #2211, opened 2021-05-09
docs(README): Update dead ReadTheDocs links

jidicula:
I found some dead ReadTheDocs links in the README and fixed them. Each link fix is in a separate commit - please let me know if any should be pointing to a different page in RTD altogether and I'll modify my changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ python -m black {source_file_or_directory}
 
 Further information can be found in our docs:
 
-- [Usage and Configuration](https://black.readthedocs.io/en/stable/usage_and_configuration/index.html)
+- [Usage and Configuration](https://black.readthedocs.io/en/stable/installation_and_usage.html#usage)
 
 ### NOTE: This is a beta product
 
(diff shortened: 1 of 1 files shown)


ichard26:
Sorry, but this was intentional, we recently refactored and reorganized *everything*. These links will become functional once these changes have been released which we're planning to do soon. The reason why most links point to stable (even though the reorg changes haven't been pushed yet) is because we don't users to be reading docs that may contain references to features / fixes that haven't actually been released yet.

I'm sorry your work was not useful in this case, but thank you regardless (I was wondering how long it would take for someone to notice the broken links haha).

jidicula:
> Sorry, but this was intentional, we recently refactored and reorganized _everything_. These links will become functional once these changes have been released which we're planning to do soon. The reason why most links point to stable (even though the reorg changes haven't been pushed yet) is because we don't users to be reading docs that may contain references to features / fixes that haven't actually been released yet (that was an issue when ALL of our docs were in the README).
> 
> I'm sorry your work was not useful in this case, but thank you regardless (I was wondering how long it would take for someone to notice the broken links haha).

@ichard26 no worries 🙂

I should have tried submitting an issue first, but I figured since it was such a small change I'd just submit the PR.

PR #2212, opened 2021-05-09
Autogenerate black(d|-primer)? help in usage docs

ichard26:
So these won't go out of date. This does mean the environment has be
setup a bit more carefully so the right version of the tool is used,
but thankfully the build environment is rebuilt on change on RTD anyway.

Also since the HTML docs are known to build fine, let's provide
downloadable HTMLzips of our docs.

This change needs RTD and GH to install Black with the [d] extra so
blackd's help can generated. While editing RTD's config file, let's
migrate the file to a non-deprecated filename.

Also I missed adding AUTHORS.md to the files key in the doc GHA config.

---

You can see built docs with these changes here: https://ichard26-testblackdocs.readthedocs.io/en/autogenerate-help/

cc @cooperlees - this is what I want to use for bandersnatch too


docs/contributing/gauging_changes.md
```diff
@@ -37,25 +37,6 @@ If you're running locally yourself to test black on lots of code try:
 
 ### CLI arguments
 
-```text
-Usage: black-primer [OPTIONS]
+```{program-output} black-primer --help
```
cooperlees, line 40: zOMG - This is amazing!

cooperlees (approved):
This is exactly what I've wanted. Amazing. I'll be stealing this for my other projects.

Issue #2178, opened 2021-05-02
Provide a more useful error when parsing fails during AST safety checks


mrdeveloperdude:
### Summary & why

Currently when Black encounters a SyntaxError while parsing during its AST safety check, it tries older versions of Python to see if it will parse then under its grammar (i.e. the syntax - which is an oversimplification but generally words). In other words, if you have a file that is valid Python 2.7 code and are running on Python 3.8, Black will first try will try Python 3.8 which fails, then 3.7 which also fails, then 3.6, 3.5, and finally 2.7. Since 2.7 is last version it tried parsing with, the error message from that version is the one ultimately emitted to the user.

The behaviour described above logically makes sense but is of limited use and sometimes even annoyance of the end-user since chances are that their file is supposed to be valid Python 3, so a message for Python 2 ain't useful (and can be downright confusing). 

### Solutions & technical details 

We should change Black so when a SyntaxError is encountered while parsing during the AST safety checks it reparses with the latest Python it can try. Then we can take its error message and emit that one to the user, and voilà, a much more useful error message.  

Another option (from @ichard26 personally) is that we could record the error message of the latest Python parse attempt when it fails and then if all of the parse attempts fail, emit the recorded error message instead of freshest error message (probably Python 2). 

One slight issue is that we emit an error telling users if they need Python 2 support they should reinstall Black with the `python2` extra when we try Python 2 parsing (which fails due to missing dependencies). This could be downgraded to a warning so the rest of this enhancement can still do its thing. Or it could just be eliminated which is probably fine since Python 2 is basically dead. 

To potential contributors: 

- [Checkout our contributing docs to get started](https://github.com/psf/black/blob/master/CONTRIBUTING.md). If you need help, feel free to reach out!
- You'll need to definitely touch this part of the codebase: https://github.com/psf/black/blob/14c76e89716b5b53c97ece80bb935ea956b7dd89/src/black/__init__.py#L6422-L6443
- You'll probably also need to add / modify some tests in `tests/` 
- And finally add a changelog entry

\- Richard (@ichard26) (added for anyone looking to contribute)

---

## Original Report

This one had me scratching my head for a while.

**How to reproduce**

Feed the following into black with -t py37:

```
reported_error = f""

e=""
# Actual error
malformed_fstring = f"{e}  ({type(e).__name__], {e.args})"

```
Please note the typo in `malformed_fstring`: I put a square closing bracket "]" in place of a curly closing bracket "}".
[...]

PR #2218, opened 2021-05-10
Provide a more useful error when parsing fails during AST safety checks (#2218)

hramezani:
Fixes https://github.com/psf/black/issues/2178

### DIFF
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -108,18 +108,23 @@ def lib2to3_unparse(node: Node) -> str:
 
 def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
     filename = "<unknown>"
+    error = None
     if sys.version_info >= (3, 8):
         # TODO: support Python 4+ ;)
         for minor_version in range(sys.version_info[1], 4, -1):
             try:
                 return ast.parse(src, filename, feature_version=(3, minor_version))
-            except SyntaxError:
+            except SyntaxError as e:
+                if error is None:
+                    error = str(e)
                 continue
     else:
         for feature_version in (7, 6):
             try:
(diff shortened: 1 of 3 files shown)


src/black/parsing.py
```diff
@@ -108,26 +108,34 @@ def lib2to3_unparse(node: Node) -> str:
...
         raise SyntaxError(
             "The requested source code has invalid Python 3 syntax.\n"
             "If you are trying to format Python 2 files please reinstall Black"
             " with the 'python2' extra: `python3 -m pip install black[python2]`."
         )
-    return ast27.parse(src)
+    try:
+        return ast27.parse(src)
+    except SyntaxError as e:
+        raise SyntaxError(error or str(e))
```
felix-hilden, line 138: Firstly, thanks for the PR!

If I'm not mistaken, this means that 2.7 errors are never passed onwards. To get to this `try` some other `try` above must fail, then `error` always has a value and the `or` short circuits. Indeed this only ever prints the error for Python 3.6, because it's the last to fail before 2.7. So this moves the problem from 2.7 to 3.6, but I'm pretty sure this is not quite the optimal solution.

The linked issue spoke of reparsing with the latest Python version available, and that idea had some agreement. That could be possible using `sys.version_info` and iffing the parser versions again.

felix-hilden:
Reading the issue once more, my previous review was a bit off the mark still. But I took your branch and improved on it in the referenced PR above. Many thanks for the initial work!

ichard26:
Felix's PR (#2304) which is based off this work has been merged now so this will be closed. Thank you for the initial work @hramezani! Your ideas will continue to live in this codebase :)

hramezani:
Thanks @felix-hilden for taking care

PR #2221, opened 2021-05-10
Remove useless flake8 config + test support code

ichard26:
Also remove the primer tests import in tests/test_black.py because it's
annoying when just trying to actually target tests/test_black.py tests.
`pytest -k test_black.py` doesn't do what you expect due to that import.


ichard26:
Low risk of things breaking and the things that could break (the tests) are passing so merging. We've depended on Click 7.x ever since we broke CI systems across the world (oops lol) and flake8-mypy was purged a fair bit back: #1867.

PR #2222, opened 2021-05-11
Fix urls in README.md

dcermak:
The urls in the ## Configuration section were outdated

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,11 +118,11 @@ project.
 
 You can find more details in our documentation:
 
-- [The basics: Configuration via a file](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file)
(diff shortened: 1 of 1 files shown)


ichard26:
Sorry, but this was intentional, we recently refactored and reorganized everything. These links will become functional once these changes have been released which we're planning to do soon. The reason why most links point to stable (even though the reorg changes haven't been pushed yet) is because we don't users to be reading docs that may contain references to features / fixes that haven't actually been released yet (that was an issue when ALL of our docs were in the README).

Thank you for being proactive and opening a PR though!

I'll leave this open since it appears a pinned issue isn't enough to stop people from opening issues / PRs :-)

dcermak:
@ichard26 oops, sorry about that, I was not aware of this. Please excuse the wasted time and proceed with this PR however you wish.

PR #2223, opened 2021-05-11
Change old link in README

kozachynskyi:
Changed old config documentation to respective link

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,11 +118,7 @@ project.
 
 You can find more details in our documentation:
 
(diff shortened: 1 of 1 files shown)


ichard26:
Sorry, but this was intentional, we recently refactored and reorganized everything. These links will become functional once these changes have been released which we're planning to do soon. The reason why most links point to stable (even though the reorg changes haven't been pushed yet) is because we don't users to be reading docs that may contain references to features / fixes that haven't actually been released yet (that was an issue when ALL of our docs were in the README).

Thank you for being proactive and opening a PR though!

I'll leave this open since it appears a pinned issue isn't enough to stop people from opening issues / PRs :-)

PR #2227, opened 2021-05-12
Click 8.0 renamed its "die on LANG=C" function so we need to look for that one too

### LINKED ISSUE #566: Running from the command-line still has issues with Click
I've just run into the same problem with PyCharm and Click that was reported in issue #277:

```
RuntimeError: Click will abort further execution because Python 3 was configured to use ASCII as encoding for the environment.
```
I think there is still a problem and I believe the problem may be that the generated command line script _only_ calls the `main()` function from `black.py` (see the entry point on line 60 in `setup.py`: `black=black:main`). 

So, after running `pip install black`, the generated command-line script looks like this:

```
#!/usr/local/opt/python/bin/python3.6

# -*- coding: utf-8 -*-
import re
import sys

from black import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

If you try to use this script to invoke black (e.g. as the command for an "external tool" in PyCharm), the `patch_click()` function won't be called and you'll get the original error reported in #277. 

To verify, I edited this script by hand, like so:

```
#!/usr/local/opt/python/bin/python3.6

# -*- coding: utf-8 -*-
import re
import sys

from black import main, patch_click

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    patch_click()
    sys.exit(main())
```
This appears to work, but is probably is not the best way to permanently fix the issue. 

One possible fix (although I don't claim that this is the best approach, especially since I'm not sure how it might interact with `blackd`), would be to create a new `main` that can be used as the entry point for the cli script. Maybe something like this at the end of `black.py`:

```
def patched_main():
    patch_click()
    main()

if __name__ == "__main__":
    patched_main()
````
along with this change in `setup.py`:
```
entry_points={"console_scripts": ["black=black:patched_main", "blackd=blackd:main [d]"]},
```
Anyway, as I said, I don't know if this is a good solution. I mostly wanted to post it as a means of clarifying what I think the problem is. 

Thank you for your time and for creating this awesome tool!

### PR DESCRIPTION
**Pull request by ambv**
Fixes #566

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### _Black_
+
+- Restored compatibility with Click 8.0 on Python 3.6 when LANG=C used (#2227)
+
 ## 21.5b1
 
 ### _Black_
--- a/mypy.ini
+++ b/mypy.ini
@@ -33,5 +33,7 @@ cache_dir=/dev/null
 
 [mypy-aiohttp.*]
 follow_imports=skip
-[mypy-_version]
-follow_imports=skip
+[mypy-black]
+# The following is because of `patch_click()`. Remove when
+# we drop Python 3.6 support.
+warn_unused_ignores=False
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1029,7 +1029,7 @@ def nullcontext() -> Iterator[None]:
 
 
 def patch_click() -> None:
-    """Make Click not crash.
+    """Make Click not crash on Python 3.6 with LANG=C.
 
     On certain misconfigured environments, Python 3 selects the ASCII encoding as the
     default which restricts paths that it can access during the lifetime of the
@@ -1047,7 +1047,9 @@ def patch_click() -> None:
 
     for module in (core, _unicodefun):
         if hasattr(module, "_verify_python3_env"):
-            module._verify_python3_env = lambda: None
+            module._verify_python3_env = lambda: None  # type: ignore
+        if hasattr(module, "_verify_python_env"):
+            module._verify_python_env = lambda: None  # type: ignore
 
 
 def patched_main() -> None:

### CODE REVIEW

File: CHANGES.md [hunk 1]
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### _Black_
+
+- Restored compatibility with Click 8.0 on Python 3.6 when LANG=C used (#566)
```
```suggestion
- Restored compatibility with Click 8.0 on Python 3.6 when LANG=C used (#2227)
```
Should be the PR not the issue

JelleZijlstra:
Would be nice to update click in our Pipfile so we test that this works correctly.

ambv:
For the time being I'd rather keep Pipfile.lock as is. Tests with Click 8.0 are passing:
```
❯ pytest --numprocesses=auto
=============================================================================================== test session starts ================================================================================================
platform darwin -- Python 3.6.13, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /Users/ambv/Dropbox/Python/black, configfile: pyproject.toml
plugins: cases-3.5.0, cov-2.11.1, mock-3.6.1, xdist-2.2.1, forked-1.3.0
gw0 [208] / gw1 [208] / gw2 [208] / gw3 [208] / gw4 [208] / gw5 [208] / gw6 [208] / gw7 [208] / gw8 [208] / gw9 [208] / gw10 [208] / gw11 [208]
........................................ss................................x...xx............................................................................................................................ [ 98%]
....                                                                                                                                                                                                         [100%]
================================================================================================= warnings summary =================================================================================================
tests/test_black.py::BlackTestCase::test_docstring_no_string_normalization
tests/test_black.py::BlackTestCase::test_docstring_no_string_normalization
  <unknown>:107: DeprecationWarning: invalid escape sequence \

tests/test_black.py::BlackTestCase::test_docstring_no_string_normalization
tests/test_black.py::BlackTestCase::test_docstring_no_string_normalization
  <unknown>:113: DeprecationWarning: invalid escape sequence \

tests/test_black.py::BlackTestCase::test_docstring_no_string_normalization
tests/test_black.py::BlackTestCase::test_docstring_no_string_normalization
  <unknown>:118: DeprecationWarning: invalid escape sequence \

tests/test_blackd.py: 14 warnings
  /Users/ambv/.virtualenvs/black-3.6/lib/python3.6/site-packages/aiohttp/web_urldispatcher.py:188: DeprecationWarning: Bare functions are deprecated, use async ones
    "Bare functions are deprecated, " "use async ones", DeprecationWarning

tests/test_format.py::TestSimpleFormat::test_simple_format_15_docstring
  <unknown>:163: DeprecationWarning: invalid escape sequence \

tests/test_format.py::TestSimpleFormat::test_simple_format_15_docstring
  <unknown>:169: DeprecationWarning: invalid escape sequence \

tests/test_format.py::TestSimpleFormat::test_simple_format_15_docstring
  <unknown>:174: DeprecationWarning: invalid escape sequence \

tests/test_format.py::TestSimpleFormat::test_simple_format_15_docstring
  <unknown>:160: DeprecationWarning: invalid escape sequence \

tests/test_format.py::TestSimpleFormat::test_simple_format_15_docstring
  <unknown>:166: DeprecationWarning: invalid escape sequence \

tests/test_format.py::TestSimpleFormat::test_simple_format_15_docstring
  <unknown>:171: DeprecationWarning: invalid escape sequence \

-- Docs: https://docs.pytest.org/en/stable/warnings.html
============================================================================== 203 passed, 2 skipped, 3 xfailed, 26 warnings in 5.43s ==============================================================================

❯ python -c "import click; print(click.__version__)"
8.0.0

```

PR #2228, opened 2021-05-13
Use codespell to find typos

cclauss:
Fix typos discovered by https://github.com/codespell-project/codespell
```
codespell --ignore-words-list="ans,hart,ot,poit,reallllly,tring" .
```

### DIFF
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -576,7 +576,7 @@ class StringParenStripper(StringTransformer):
             - The target string is NOT the only argument to a function call.
             - The target string is NOT a "pointless" string.
             - If the target string contains a PERCENT, the brackets are not
-              preceeded or followed by an operator with higher precedence than
+              preceded or followed by an operator with higher precedence than
               PERCENT.
 
     Transformations:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### _Black_
 
+- Fix typos discovered by codespell (#2228)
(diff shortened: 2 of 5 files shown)


src/black/trans.py
```diff
@@ -576,7 +576,7 @@ class StringParenStripper(StringTransformer):
             - The target string is NOT the only argument to a function call.
             - The target string is NOT a "pointless" string.
             - If the target string contains a PERCENT, the brackets are not
-              preceeded or followed by an operator with higher precedence than
+              proceeded or followed by an operator with higher precedence than
```
JelleZijlstra, line 579: This should be "preceded"

.pre-commit-config.yaml
```diff
@@ -29,3 +29,9 @@ repos:
     hooks:
       - id: prettier
         args: [--prose-wrap=always, --print-width=88]
+
+  - repo: https://github.com/codespell-project/codespell
```
JelleZijlstra, line 33: I'm not sure adding this to pre-commit is worth it; it may just slow things down. Open to other opinions though.

cclauss:
```
% time codespell --ignore-words-list="ans,hart,ot,poit,reallllly,tring" .

0.24s user 0.02s system 91% cpu 0.284 total
```
Could also run it in a GitHub Action.

cclauss: Dropped the pre-commit.

JelleZijlstra (approved):

PR #2235, opened 2021-05-15
Update vim plugin manual installation instructions.

### LINKED ISSUE #2234: vim plugin needs new installation instructions
Installing the current vim plugin (v1.2) now needs both `plugin/black.vim` and `autoload/black.vim`  This is not reflected in the official vim plugin installation instructions:
https://black.readthedocs.io/en/stable/integrations/editors.html#vim

At least not for the manual file installation instructions--I don't use vim-plug or Vundle.

Previously it was only necessary to install `black.vim` in ~/.vim/plugins, so the new behavior could use some documentation.

itsayellow:
Closes #2234 

Since the vim.plugin was updated to v1.2 in #1157 , the installation procedure has changed.

This PR updates the manual installation instructions for the new vim plugin to include BOTH `plugin/black.vim` and `autoload/black.vim`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 - Fix typos discovered by codespell (#2228)
 - Restored compatibility with Click 8.0 on Python 3.6 when LANG=C used (#2227)
+- Fix vim plugin installation instructions. (#2235)
 
 ### _Blackd_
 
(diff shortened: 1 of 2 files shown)


itsayellow:
I should say that I can only vouch for the manual vim plugin instructions.  I have no idea if the Vundle or vim-plug instructions on the vim plugin installation page still work (I don't use those tools.)

ambv:
Thanks! ✨ 🍰 ✨

Issue #1741, opened 2020-10-05
Line endings


spapanik:
**Is your feature request related to a problem? Please describe.**

It is frustrating to see a git diff the all it does is changing the line endings from LF to CRLF. In a language with significant whitespace, I don't want to do diffs with the hide whitespace option. 

**Describe the solution you'd like**

I would like either black to enforce LF, or have LF as default and be able to choose in `pyproject.toml`, in something along the lines:
```toml
[tool.black]
line-ending = "LF" or "CR" or "CRLF"
```

**Describe alternatives you've considered**
Viewing diffs with "hide whitespace changed" on.

**Additional context**
N/A

hugovk:
As another alternative, there are a number of pre-commit hooks to do this: https://pre-commit.com/hooks.html

* `mixed-line-ending` - Replaces or checks mixed line ending
* `forbid-crlf` - Forbid files containing CRLF end-lines to be committed
* `remove-crlf` - Replace CRLF end-lines by LF ones before committing

zsol:
IIRC we have considered this, but ultimately decided to be extra careful and preserve the original CRLF line endings, as we felt that's more in the domain of the version control software being used. The only opinion Black has on this matter is that line endings should be consistent (which is a fairly basic requirement :) )

The alternatives @hugovk offers above are also excellent.

hugovk:
Oh, and Git itself can be configured to handle line endings:

* https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings

felix-hilden:
Since reasonable alternatives have been proposed, there has been no discussion for a while and no additional points have been raised, should this issue be closed? I agree that line endings should not be handled by Black.

ichard26:
I also agree line ending shouldn't be handled by Black (it would probably be one of the more risky things Black would do if accepted) so closing!

Thanks for the suggestion though.

Issue #2238, opened 2021-05-17
Improving the contributing experience


ichard26:
### If you came here because you were asked to provide feedback:

Hi there! First of all I'd like to personally thank you for your contribution(s) to Black!   This project is only possible by your contributions :black_heart:. Two, one of our top priorities is making it a good experience to contribute here. We would love to hear from you about your feedback and suggestions (context why can be found further below). 

**Please comment with your feedback and suggestions.** **Any sort of constructive feedback is welcomed.** It can be as general as "the contributing documentation is lacking" or as specific as "on X environment, the test suite breaks", **although in general the more specific and actionable the better.** Bonus points if it's about the contributing documentation since I personally (as in me @ichard26) have been working on improving them lately.

---

### Context & why

_Black_ is a popular project and probably gets a mountain worth of potential contributors. _Black_ is also a more mature and older project with a lot of bugs to squash so we need a fair of constant development to keep the project running smooth. One of my top concerns is that we are wasting / losing a lot of potential contributions because a poor contributing experience (fits in the general maintainability goal I've been striving for). It's not like the maintainers of this project can handle everything. 

A few reasons why we might be losing potential contributions:

- Lack of good contributing documentation: yes the basics are there, but nothing else, you're left on your own for stuff like "how do I write a test?", "why are the docs structured this way?", etc. 
- Difficulty with testing, linting, or other commands used during development (our requirements / environment system is currently IMO a bit of a mess)
- Poor responsiveness to issues and pull requests (this one is hard to fix but through delegation, automation, and other ideas it can be improved): it's discouraging to see your issue / pull request sit there without activity for long periods of time
- Lack of personalized support for experienced contributors: mentorship is a great way of turning a causal but somewhat experience contributor into a serious longterm contributor, but we don't provide that (and I don't think we can with our current resources) but at least providing a discussion area for maintainers and contributors to chat and ask / answer questions would be a good middle ground (the solution to this would probably be advertising the #blackformatter IRC channel a bit more)

**But the thing is that I'm a maintainer so a lot of things that used to annoy me don't anymore since I got used to them. Asking contributors (especially first-time contributors!) for feedback and suggestions is better.** Sooo... hence this issue.

_fyi to fellow maintainers: I plan to use this issue as a place to collect feedback and suggestions as I ping contributors during the PR process_

felix-hilden:
Nice! I agree for the most part. Some thoughts:

> "How do I write a test?"

[...]

PR #2244, opened 2021-05-17
Make Prettier preserve line ending type

felix-hilden:
Hi! This PR, should you choose to accept it, makes Prettier preserve the line ending types of files it touches.

Why? The default in Prettier 2.0 was [changed](https://prettier.io/docs/en/options.html#end-of-line) from `auto` to `LF`. This makes development on Windows awkward, because every file is marked with changes both by Prettier and then by Git regardless of repository line ending settings, making committing harder than it should be.

As discussed in #1741, line ending settings can be configured, and there are other tools for enforcing line ending consistency. I don't know whether they have the same effect of converting all files regardless of settings though.

Suggested: `skip-news`. If this is not to your liking, feel free to close. This PR will self-destruct in five years.

---

Aside from that: I noticed that runnin pre-commit manually seems to add line endings to symlink files, but they disappear when actually committing. Don't know if that's a known.. quirk..(?) or not.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,4 +28,3 @@ repos:
     rev: v2.2.1
     hooks:
       - id: prettier
-        args: [--prose-wrap=always, --print-width=88, --end-of-line=auto]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,4 +28,4 @@ repos:
     rev: v2.2.1
     hooks:
       - id: prettier
-        args: [--prose-wrap=always, --print-width=88]
+        args: [--prose-wrap=always, --print-width=88, --end-of-line=auto]
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,0 +1,3 @@
+proseWrap: always
+printWidth: 88
+endOfLine: auto


JelleZijlstra (approved):
Makes sense to me.

Unrelatedly it'd be nice to have our prettier config in a `.prettierrc` file instead of just in the precommit configuration.

ichard26:
side note:

> This PR will self-destruct in five years.

uhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh :laughing:

felix-hilden:
Thanks! It seems that some Mac build is failing, but it's the same as on `main`.

I'll add the `.prettierrc`. Actually that sounds like the reason for the symlink line endings changing.

felix-hilden:
Hmm, the symlinks still toggle, but there you go with the config file at least! I'm happy for now. It's not like running pre-commit manually is absolutely necessary or anything.

JelleZijlstra (approved):

ichard26:
Thanks much for your contribution! This project is only possible by contributions like these :black_heart:.  You're awesome, @felix-hilden. To be honest, I haven't touched a Windows machine in a long time so I've never seen this as an issue *nervous laughter*. Also, I've started a new project to improve the contributing experience for this project, so I'd love to hear your feedback and suggestions. More details can be found here: GH-2238. 

<sub>*don't mind me testing out this new manner of saying thanks haha :D*</sub>

felix-hilden:
You're most welcome! Yeah I think that's a very nice thank you for any new contributor, albeit a little over the top for my modest Finnish self :D but I know that won't read so sarcastic for other potential contributors 😅 Also, that black heart is a nice touch.

Hmm, I'll have a look at the issue!

PR #2245, opened 2021-05-17
Fix test requirements file name in documentation

felix-hilden:
No mountain too high to climb, no fix too small to shoot a PR and write a description that's a hundred times longer than the diff. Right?

A no-brainer, really. Another `skip-news`, if you'd be so kind!

### DIFF
--- a/docs/contributing/the_basics.md
+++ b/docs/contributing/the_basics.md
@@ -22,7 +22,7 @@ from the cloned _Black_ repo. It will do the correct thing.
 Non pipenv install works too:
 
 ```console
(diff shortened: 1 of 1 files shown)


cooperlees (approved):

ichard26 (approved):
Thank you again! No idea how we managed to miss this mistake. Honestly I'm scared there's another mistake like this one somewhere.

felix-hilden:
Yeah well this is the kind of thing you just run, look at the file and then just run again and forget about it :D I know I did. So no biggie, but still worth fixing!

PR #2246, opened 2021-05-18
Link isort profile to Black code style mention

felix-hilden:
Hi, the isort configuration currently in the [Black code style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines) document is duplicated in [Using Black with other tools](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort). I think it would be better to consolidate information and simply link to the tool guide, mentioning the easy profile in the original document.

I think `skip news`, what do you think about the change?

---

I changed the link to isort PyPI page because for users it could be better to see the Black docs on why that configuration is necessary and what isort *is* from Black's perspective.

### DIFF
--- a/docs/the_black_code_style/current_style.md
+++ b/docs/the_black_code_style/current_style.md
@@ -128,29 +128,9 @@ indentation level (like the arguments list and the docstring in the example abov
 If a data structure literal (tuple, list, set, dict) or a line of "from" imports cannot
 fit in the allotted length, it's always split into one element per line. This minimizes
 diffs as well as enables readers of code to find which commit introduced a particular
-entry. This also makes _Black_ compatible with [isort](https://pypi.org/p/isort/) with
-the following configuration.
-
-<details>
-<summary>A compatible `.isort.cfg`</summary>
-
-```cfg
-[settings]
-multi_line_output = 3
-include_trailing_comma = True
-force_grid_wrap = 0
-use_parentheses = True
-ensure_newline_before_comments = True
-line_length = 88
-```
-
-The equivalent command line is:
-
-```
-$ isort --multi-line=3 --trailing-comma --force-grid-wrap=0 --use-parentheses --line-width=88 [ file.py ]
-```
-
-</details>
+entry. This also makes _Black_ compatible with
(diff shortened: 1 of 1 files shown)


ichard26:
Thanks for the heads up! I'll review this soon, just a little busy today. Although, what is CI doing!? Why is everything cancelled? I guess it's because they're all very stuck. 

Nevermind... there's an [active incident with GitHub Actions](https://www.githubstatus.com/incidents/m16jzl31gnqt). Time to wait then huh.

felix-hilden:
It seems that the issue is resolved and Actions is recovering!

ichard26:
I reran the jobs and they still failed. I don't think GHA has fully recovered yet :(

ichard26 (approved):
Thank you for your contribution! LGTM, I like reducing duplication in the docs :)

WhyNotHugo:
Is that profile considered part of black's style or isort?

I'd like to bring up for discussion the idea of changing that profile to use `force_single_line=true`, which reduces merge conflicts by a lot (I feel this aligns with black's approach too).

However, I'm not sure if I should bring it up here, or it `isort`'s tracker.

ichard26:
It's probably better if you bring it up on the isort issue tracker since we don't know isort well. I had to search up with `force_single_line` is. FWIW, it doesn't seem like something Black would learn to enforce.

MarcoGorelli:
> which reduces merge conflicts by a lot (I feel this aligns with black's approach too)

yeah, imports were a major source of merge conflicts in pandas, setting `force_grid_wrap` helped resolve this. It works fine with `profile=black`

https://github.com/pandas-dev/pandas/blob/daec2e73b48fe34086dcdeeb0858d29536b6ca6a/pyproject.toml#L115

> However, I'm not sure if I should bring it up here, or it isort's tracker.

pretty sure this would be isort, as that's where the profile's defined

WhyNotHugo:
`force_grid_wrap` _can_ have weird merge conflicts though, whereas `force_single_line` is a bit more dead simple to deal with.

I've opened [an issue](https://github.com/PyCQA/isort/issues/1791) to discuss the topic on the isort side.

PR #2249, opened 2021-05-22
Removed adding a space into empty docstrings.

### LINKED ISSUE #2168: Unable to include empty docstring
Empty docstrings are used by sphinx to indicate methods that should not be included in the documentation. However, as of 21.4b0 black reformats empty docstrings to contain a space.

**To Reproduce**:
Source:
```
def foo():
    """"""
    ...
```

gets reformatted to

```
def foo():
    """ """
    ...
```

**Expected behavior** I'm generally supportive of the style changes introduced in #1740. However, in this case black is changing the code in a meaningful way. I suggest that empty docstrings be ignored as a special case. I am agnostic on whether the empty docstring should be spelled `""` or `""""""`.

**Environment:**

- Version: 21.4b2
- OS and Python version: debian 10/Python 3.7.10

**Does this bug also happen on master?** Yes

MarkCBell:
Resolves #2168 by disabling the insertion of a " " when the docstring is entirely empty.

Note that this PR is focussed only on the case of empty docstrings. In particular this does not make any changes to the behaviour that a " " is inserted if a non-empty docstring begins with the quoting character. That is, black still prefers:

    """ "something" """

to:

    """"something" """

and that:

    """"Something""""

is not a legal docstring.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### _Black_
 
+- A space is no longer inserted into empty docstrings (#2249)
 - Respect `.gitignore` files in all levels, not only `root/.gitignore` file (apply
   `.gitignore` rules like `git` does) (#2225)
 - Restored compatibility with Click 8.0 on Python 3.6 when LANG=C used (#2227)
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -255,9 +255,6 @@ class LineGenerator(Visitor[Line]):
                         # Odd number of tailing backslashes, add some padding to
                         # avoid escaping the closing string quote.
                         docstring += " "
-            else:
-                # Add some padding if the docstring is empty.
-                docstring = " "
 
             # We could enforce triple quotes at this point.
             quote = quote_char * quote_len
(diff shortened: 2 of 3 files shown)


JelleZijlstra:
Thanks, I agree we should do this. Can we still make it output a single space if the original docspace is nonempty?

JelleZijlstra (approved):
Looks good to me. With the change to preserve docstrings that already contain a space, this shouldn't cause thrash for people who applied the formatting from the current release, while preserving completely empty docstrings for those who need them.

ichard26 (approved):
I don't see anything wrong with this and this is a good pragmatic change. Thanks!

johnthagen:
> Can we still make it output a single space if the original docspace is nonempty?

Does this mean if I reformatted the code with the previous `black` I need to hunt through and find the docstrings `""" """` and fix them manually again?

ichard26:
> Does this mean if I reformatted the code with the previous black I need to hunt through and find the docstrings """ """ and fix them manually again?

Yes. It's not great for docstrings that were erroneously (like for sphinx docs) changed to `""" """` but the main reason why is reduce churn by not effectively flip-flopping the original rule. Why the original rule? IIRC it was to indicate to users about the (probably) wrong empty or whitespace only docstring (it turns out an empty docstring isn't always wrong, unfortunately we didn't know this until after we got a report).

PR #2258, opened 2021-05-25
Add optional uvloop import

### LINKED ISSUE #2257: Add optional extra uvloop install
Lets offer support for using [uvloop](https://github.com/MagicStack/uvloop) optionally via a extra install and `try/except` import + loop policy set.

- If we can import it, use and set it as the loop
  - Naturally this will only work on POSIX OS's (e.g. there is now windows uvloop)

cooperlees:
- If we find `uvloop` in the env for black, blackd or black-primer lets try and use it
- Add a uvloop extra install

Fixes #2257

Test:
- Add ci job to install black[uvloop] and run a primer run with uvloop
  - Only with latest python (3.9)
  - Will be handy to compare runtimes as a very unoffical benchmark

### DIFF
--- a/.github/workflows/uvloop_test.yml
+++ b/.github/workflows/uvloop_test.yml
@@ -1,0 +1,46 @@
+name: test uvloop
+
+on:
+  push:
+    paths-ignore:
+      - "docs/**"
+      - "*.md"
+
+  pull_request:
+    paths-ignore:
+      - "docs/**"
+      - "*.md"
+
+jobs:
+  build:
+    # We want to run on external PRs, but not on our own internal PRs as they'll be run
+    # by the push to the branch. Without this if check, checks are duplicated since
+    # internal PRs match both the push and pull_request events.
+    if:
+      github.event_name == 'push' || github.event.pull_request.head.repo.full_name !=
+      github.repository
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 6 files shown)


cooperlees:
Why are tests and lints skipped? Cause last commit only changed the `CHANGES.md` file?

ichard26:
Yep. I made that change because of all of the documentation changes lately which don't need the (relatively) expensive primer and test workflows. In hidesight this isn't great UX even if it's technically speaking totally fine safety wise. We can revert it by simply removing the `path-ignore` keys in relevant workflow files.

.github/workflows/uvloop_test.yml
```diff
@@ -0,0 +1,45 @@
...
+      - "*.md"
+
+  pull_request:
+    paths-ignore:
+      - "docs/**"
+      - "*.md"
+
+jobs:
+  build:
+    # We want to run on external PRs, but not on our own internal PRs as they'll be run
```
zsol, line 16: what does internal vs external mean in this context?

zsol (approved):
lgtm

.github/workflows/uvloop_test.yml
cooperlees: This is copy pasta to maintain other CI behavior - this stops running the same CI twice for maintainers like me who make a branch and push to the black repo itself. @ichard26 did this and I like that it saves some resources. GitHub should really default to this to save CPU. One of us should go work there and claim the reward points.

PR #2259, opened 2021-05-26
Code Flag Options

### LINKED ISSUE #1801: --code doesn't work with --check or --diff
**Describe the bug**

- `--code` is supposed to take the code as an input string
- `--check` is supposed to change the exit code based on whether edits were made while formatting
- `--diff` is supposed to only output a diff, instead of the entire formatted source

What I found was the the `--check` and `--diff` flags did not work when passing input using `--code`; it behaved as if the extra flags hadn't been passed at all.

1. Run this command (notice the double space after `=`):

```sh
black --check --code 'a =  5'
```

2. The command exits with a status `0`, despite making edits (removing the double space).

**Expected behavior** The command should have exited with a status of `1`, since edits were made.

- Version: `20.8b1`
- OS and Python version: Arch Linux/Python 3.8.6

### LINKED ISSUE #2104: Safety checks are disabled when passing in code using `-c` / `--code`
I've been writing some docs during my documentation reorganization effort, and I've found myself writing this warning:

> --check, --diff, and --safe / --fast have no effect when using -c / --code. Safety checks normally turned on by default that verify Black's output are disabled as well.

~Is this by design? IMO it seems like an oversight / bug that should fixed.~

edit: yep this is in fact a bug, [thanks Jelle](https://github.com/psf/black/issues/2104#issuecomment-818302531)!

cc @JelleZijlstra @cooperlees

HassanAbouelela:
Closes #2104, closes #1801.

Fixes a bug in the `--code` flag which would ignore options such as `--diff`, `--check`, and `--fast`.

This PR implements the fix described in the issue. It also adds tests for this fix.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -29,7 +29,6 @@ from typing import (
 from dataclasses import replace
 import click
 
-from black import output
 from black.const import DEFAULT_LINE_LENGTH, DEFAULT_INCLUDES, DEFAULT_EXCLUDES
 from black.const import STDIN_PLACEHOLDER
 from black.nodes import STARS, syms, is_simple_decorator_expression
@@ -385,75 +384,56 @@ def main(
     )
     if config and verbose:
         out(f"Using configuration from {config}.", bold=False, fg="blue")
-    if code is not None:
-        then = datetime.utcnow()
-
-        try:
-            formatted = format_file_contents(code, fast=fast, mode=mode)
-        except NothingChanged:
-            formatted = code
-
-        stream = io.TextIOWrapper(sys.stdout.buffer)
-        try:
-            if write_back is WriteBack.CHECK:
-                ctx.exit(0 if code == formatted else 1)
-            elif write_back in (WriteBack.DIFF, WriteBack.COLOR_DIFF):
-                now = datetime.utcnow()
-                src_name = f"STDIN\t{then} +0000"
-                dst_name = f"STDOUT\t{now} +0000"
-
-                diff_contents = output.diff(code, formatted, src_name, dst_name)
-                if write_back == WriteBack.COLOR_DIFF:
-                    diff_contents = color_diff(diff_contents)
-                    stream = wrap_stream_for_windows(stream)
-
-                stream.write(diff_contents)
-            else:
-                stream.write(formatted)
-        finally:
-            # Ensure the stream is cleaned up properly
-            stream.detach()
-
-        ctx.exit(0)
 
     report = Report(check=check, diff=diff, quiet=quiet, verbose=verbose)
-    sources = get_sources(
-        ctx=ctx,
-        src=src,
-        quiet=quiet,
-        verbose=verbose,
-        include=include,
-        exclude=exclude,
-        extend_exclude=extend_exclude,
-        force_exclude=force_exclude,
-        report=report,
-        stdin_filename=stdin_filename,
-    )
 
-    path_empty(
-        sources,
-        "No Python files are present to be formatted. Nothing to do 😴",
-        quiet,
-        verbose,
-        ctx,
-    )
-
-    if len(sources) == 1:
-        reformat_one(
-            src=sources.pop(),
-            fast=fast,
-            write_back=write_back,
-            mode=mode,
-            report=report,
+    if code is not None:
+        reformat_code(
+            content=code, fast=fast, write_back=write_back, mode=mode, report=report
         )
     else:
-        reformat_many(
-            sources=sources, fast=fast, write_back=write_back, mode=mode, report=report
+        sources = get_sources(
+            ctx=ctx,
+            src=src,
+            quiet=quiet,
+            verbose=verbose,
+            include=include,
+            exclude=exclude,
+            extend_exclude=extend_exclude,
+            force_exclude=force_exclude,
+            report=report,
+            stdin_filename=stdin_filename,
         )
+
+        path_empty(
+            sources,
+            "No Python files are present to be formatted. Nothing to do 😴",
+            quiet,
+            verbose,
+            ctx,
+        )
+
+        if len(sources) == 1:
+            reformat_one(
+                src=sources.pop(),
+                fast=fast,
+                write_back=write_back,
+                mode=mode,
+                report=report,
+            )
+        else:
+            reformat_many(
+                sources=sources,
+                fast=fast,
+                write_back=write_back,
+                mode=mode,
+                report=report,
+            )
 
     if verbose or not quiet:
         out("Oh no! 💥 💔 💥" if report.return_code else "All done! ✨ 🍰 ✨")
-        click.secho(str(report), err=True)
+        if code is None:
+            click.secho(str(report), err=True)
     ctx.exit(report.return_code)
 
 
@@ -538,6 +518,29 @@ def path_empty(
     if not src and (verbose or not quiet):
         out(msg)
         ctx.exit(0)
+
+
+def reformat_code(
+    content: str, fast: bool, write_back: WriteBack, mode: Mode, report: Report
+) -> None:
+    """
+    Reformat and print out `content` without spawning child processes.
+    Similar to `reformat_one`, but for string content.
+
+    `fast`, `write_back`, and `mode` options are passed to
+    :func:`format_file_in_place` or :func:`format_stdin_to_stdout`.
+    """
+    try:
+        changed = Changed.NO
+        if format_stdin_to_stdout(
+            content=content, fast=fast, write_back=write_back, mode=mode
+        ):
+            changed = Changed.YES
+        report.done(content, changed)
+    except Exception as exc:
+        if report.verbose:
+            traceback.print_exc()
+        report.failed(content, str(exc))
 
 
 def reformat_one(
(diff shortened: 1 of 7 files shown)


HassanAbouelela:
I've fixed the linting issues now, could someone rerun the workflows.

cooperlees (approved):
Wow, great tests here! Thanks for your contribution.

src/black/__init__.py
```diff
@@ -385,8 +386,29 @@ def main(
...
+            ctx.exit(0 if code == formatted else 1)
+        elif write_back in (WriteBack.DIFF, WriteBack.COLOR_DIFF):
+            now = datetime.utcnow()
+            src_name = f"STDIN\t{then} +0000"
+            dst_name = f"STDOUT\t{now} +0000"
+
+            diff_contents = output.diff(code, formatted, src_name, dst_name)
+            if write_back == WriteBack.COLOR_DIFF:
+                diff_contents = color_diff(diff_contents)
+            print(diff_contents)
```
ichard26, line 406: I don't know whether `echo` is better than `print` here (and we use both inconsistently in the codebase) but `echo` seems to be safer?

ichard26 (commented):
Thanks for working on this fix! I got some feedback but nothing truly blocking (except for possibly a Windows bug).

Great tests btw!

src/black/__init__.py
```diff
@@ -29,6 +29,7 @@
 from dataclasses import replace
 import click
 
+from black import output
```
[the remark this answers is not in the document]
HassanAbouelela: You'd be right for the most part, but `diff` is a parameter of the main function, and overrides the global diff function. I figured doing it this way would avoid changing the interface of the function.

src/black/__init__.py
HassanAbouelela: That may be the case, but if we're going to be using a textiowrapper on stdout, we could just use the write method, as is done in the linked example.

ichard26: *facepalm* should've caught that, my bad

src/black/__init__.py
```diff
@@ -385,8 +386,29 @@ def main(
...
+            now = datetime.utcnow()
+            src_name = f"STDIN\t{then} +0000"
+            dst_name = f"STDOUT\t{now} +0000"
+
+            diff_contents = output.diff(code, formatted, src_name, dst_name)
+            if write_back == WriteBack.COLOR_DIFF:
+                diff_contents = color_diff(diff_contents)
+            print(diff_contents)
+        else:
+            print(formatted)
```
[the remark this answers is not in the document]
HassanAbouelela: I completely missed that, thanks.

Resolved in 6221b880.

src/black/__init__.py
HassanAbouelela: No longer relevant. 6221b88.

tests/test_black.py
```diff
@@ -2069,6 +2073,102 @@ def test_docstring_reformat_for_py27(self) -> None:
...
 
+    def test_code_option(self) -> None:
+        """Test the code option with no changes."""
+        code = 'print("Hello world")\n'
+        args = ["--code", code]
+        result = CliRunner().invoke(black.main, args)
+
+        # The CLI prints an extra linebreak when exiting.
+        self.assertEqual(result.output, code + "\n")
+        self.assertEqual(result.exit_code, 0)
```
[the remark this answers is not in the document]
HassanAbouelela: Thanks, I've updated the tests. 081a8e32.

tests/test_black.py
```diff
@@ -2069,6 +2073,99 @@ def test_docstring_reformat_for_py27(self) -> None:
...
 
+    @staticmethod
+    def compare_results(
+        result: click.testing.Result, expected_value: str, expected_exit_code: int
+    ) -> None:
+        """Helper method to test the value and exit code of a click Result."""
+        assert (
+            result.output == expected_value
+        ), "The output did not match the expected value."
+        assert result.exit_code == expected_exit_code, "The exit code is incorrect."
```
ichard26, line 2084: Nice clean up!

ichard26 (changes requested):
This time I did a hands-on review / test and this still isn't quite right. The error message when formatting fails isn't right. For normal cases, this is handled by the `Report` class. Check out the difference between these cases:

`--code`:

Screenshot from 2021-05-27 18-51-43

STDIN:

Screenshot from 2021-05-27 18-50-41

It honestly might be better to generalize `reformat_one` to handle the `--code` case too. 

Thanks for being patient with me!

src/black/__init__.py
```diff
@@ -385,8 +386,36 @@ def main(
...
+            formatted = code
+
+        stream = io.TextIOWrapper(sys.stdout.buffer)
+        try:
+            if write_back is WriteBack.CHECK:
+                ctx.exit(0 if code == formatted else 1)
+            elif write_back in (WriteBack.DIFF, WriteBack.COLOR_DIFF):
+                now = datetime.utcnow()
+                src_name = f"STDIN\t{then} +0000"
+                dst_name = f"STDOUT\t{now} +0000"
```
ichard26, line 403: Also technically misleading but I can't think of anything better.

HassanAbouelela:
Thanks for the feedback, and I totally agree. Originally I wanted to maintain the whole “thrown together, mainly for testing” behavior, but as the PR has progressed, the feature has become quite bloated, and repeated a lot of the same code from the other functions. I’ll see what I can do to formalize this flag a bit more.

HassanAbouelela:
Updated the code to use the STDIN formatter.

As a note for maintainers: we should probably squash/rebase this PR before merging. Most of the commits prior to switching to the STDIN formatter are irrelevant.

JelleZijlstra:
We always squash merge, fyi. Also, could you fix the merge conflict in CHANGES.md?

HassanAbouelela:
Thanks, good to know. The conflicts have been fixed.

src/black/__init__.py
```diff
@@ -511,6 +520,29 @@ def path_empty(
...
+    `fast`, `write_back`, and `mode` options are passed to
+    :func:`format_file_in_place` or :func:`format_stdin_to_stdout`.
+    """
+    try:
+        changed = Changed.NO
+        if format_stdin_to_stdout(
+            content=content, fast=fast, write_back=write_back, mode=mode
+        ):
+            changed = Changed.YES
+        report.done(content, changed)
```
JelleZijlstra, line 539: This will make it print something like "<all of your code> already well formatted". That's going to be pretty ugly if you input multiple lines. I'd suggest passing something like `Path("<string>")`, matching what Python itself does:
```
% python -c 'def f(x): 1/0
quote> f(1)
quote> '''
  File "<string>", line 2, in <module>
  File "<string>", line 1, in f
ZeroDivisionError: division by zero
```

src/black/report.py
```diff
@@ -4,6 +4,7 @@
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
+import typing
```
JelleZijlstra, line 7: Prefer `from typing import Union` (but with my suggested change above you should be able to undo all changes to this file)

JelleZijlstra (changes requested):
Thanks, this looks significantly cleaner

src/black/__init__.py
JelleZijlstra: Just applied this change myself so @cooperlees can do a release if he wants

JelleZijlstra:
Just tried this out locally and decided that it's better to make `-c` apply `--quiet` by default because the extra "All done" output isn't really useful.

There might also be a discrepancy regarding whether `pyproject.toml` is used; looking into that more now.

JelleZijlstra:
OK, there's indeed a problem with pyproject.toml:
```
(black) jelle@mbpt-root tmp % black -c "a = 'b'"  
a = "b"
(black) jelle@mbpt-root tmp % echo "a = 'b'" | black -
a = 'b'
All done! ✨ 🍰 ✨
1 file left unchanged.
(black) jelle@mbpt-root tmp % cat pyproject.toml 
[tool.black]
```
I think the correct behavior is to pick up the pyproject.toml. That way, users can run `black -c` in their project directory to see how black would format code in that project.

The reason for the discrepancy is that in `black.read_pyproject_toml`, we pass `()` to `black.files.find_pyproject_toml`, and that function returns `Path("/").resolve()` if no src is provided. I think it should instead default to searching for a toml in the cwd. The current behavior would make it pick up `/pyproject.toml` at the file system root, which doesn't make much sense.

src/black/__init__.py
HassanAbouelela: Resolved in 39c3ae1.

src/black/report.py
HassanAbouelela: Resolved in 5f85ff5.

src/black/files.py
```diff
@@ -38,7 +38,7 @@ def find_project_root(srcs: Sequence[str]) -> Path:
     project root, the root of the file system is returned.
     """
     if not srcs:
-        return Path("/").resolve()
+        return Path(".").resolve()
```
JelleZijlstra, line 41: I think this should be `srcs = [Path.cwd()]` to make this work like it does for `-`. In practice, this will make a difference if your pyproject.toml is in a parent directory.

HassanAbouelela: That makes sense, updated along with the tests.

ichard26: Unfortunately now tests are failing on Python 3.6 and 3.7. It's not clear what the issue is from an initial lookover. If you look into it that would be great but I wouldn't mind looking into it myself.

HassanAbouelela: I managed to replicate and found out it's due to the mock.call interface changing between [3.7](https://docs.python.org/3.7/library/unittest.mock.html#call) and [3.8](https://docs.python.org/3.8/library/unittest.mock.html#call). 3.8+ is backward compatible with the <3.8 format, so it was an easy fix.

This solution feels a bit cryptic, especially if you aren't familiar with the call structure, but it gets the job done. If anyone has better solutions or a different way to test this change, I'd be happy to implement it.

JelleZijlstra (approved):

ichard26 (approved):
I am happy with the final product for the most part. The only thing I'd dock against this patch is the general awkwardness of using `format_stdin_to_stdout` in `reformat_code`, but this is already quite complex so I'd rather refactor this (and other parts of the "main called to formatting will start" pathway) in a separate PR. 

Thank you so much for your contribution! This project is only possible by contributions like these :black_heart:. You're awesome, @HassanAbouelela! Congratulations on your first PR for psf/black. It's great to be able to get rid of that documentation warning for once and all. Also thanks for playing a role in the very nice welcome to Python Discord.

P.S. since you seem to be \*the\* person who manages to break all of our development workflows, it would be even more awesome if you could provide some feedback on your experience contributing here, we got a lot to learn (especially in the Windows department) so yeah. More details in #2238.

cooperlees:
This needs the `CHNAGES.md` fixed ... Wonder how we can automate checking for that ...

ichard26:
Oh fun! Probably automatically although that will make it harder to cleanly separate the logic into own repo when I *eventually* do that.

cooperlees:
https://github.com/psf/black/commit/5de8c5f2f798d97133de3a1b391c04813c314c48

Issue #2252, opened 2021-05-24
Move Binary building CI to latest python (3.9)


cooperlees:
Today we hard set python version to 3.7 for building our static binaries we push to GitHub. Lets move this to 3.9

https://github.com/psf/black/blob/main/.github/workflows/upload_binary.yml

- Bonus Points: Just don't set the version and we automatically move as new version of Python come out.

felix-hilden:
I think this can be closed now, my bad for not automating it in the PR itself.

ichard26:
And I totally forget to close this issue when I originally noticed this was still opened :laughing:

PR #2260, opened 2021-05-26
Use latest Python in uploading binaries

felix-hilden:
From #2252, here to claim my bonus points: I found [a way](https://docs.github.com/en/actions/guides/building-and-testing-python#using-a-specific-python-version) to always use the latest Python minor in GitHub Actions. It seemed easy enough to just drop in, although I'm not really sure how it could be verified without actually uploading.

I do trust the documentation, but if you're a bit scared (like I am too), we could just have the version be set to 3.9 for a guaranteed success. What do ya reckon?

---

`skip news`? I'm still not really sure what changes should end up in the changelog but I'm slowly getting there 😅

### DIFF
--- a/.github/workflows/upload_binary.yml
+++ b/.github/workflows/upload_binary.yml
@@ -10,7 +10,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7]
         os: [windows-2019, ubuntu-20.04, macos-latest]
         include:
           - os: windows-2019
@@ -29,10 +28,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up latest Python
         uses: actions/setup-python@v2
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.x"
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/upload_binary.yml
+++ b/.github/workflows/upload_binary.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Set up latest Python
         uses: actions/setup-python@v2
         with:
-          python-version: "3.x"
+          python-version: "*"
 
       - name: Install dependencies
         run: |
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@
 ### _Packaging_
 
 - Release self-contained macOS binaries as part of the GitHub release pipeline (#2198)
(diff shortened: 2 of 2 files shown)


.github/workflows/upload_binary.yml
```diff
@@ -10,7 +10,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7]
```
cooperlees, line 13: I was happy to just delete this line

.github/workflows/upload_binary.yml
```diff
@@ -29,10 +28,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up latest Python
         uses: actions/setup-python@v2
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.x"
```
cooperlees, line 34: My idea is to just move to latest, and only oin if we have to.

mom a big believer in move forward and rather than add tech debt by pinning by default, only do it when problems arise.

cooperlees (changes requested):
I feel no pinning at all is needed here. I also feel this minor version pinning won’t do much as there is no idea when a 4.0 will happen …

Let’s see what other maintainers say here.

.github/workflows/upload_binary.yml
felix-hilden: Yeah that sounds reasonable to me, but I'll wait for other comments as you requested before changing it!

JelleZijlstra: Yes, let's just remove the pinning.

felix-hilden:
For some reason I prefer `*` to `x` when it's the only one there. I think it's a more familiar wildcard. I also added a changelog entry because I felt like it. Maybe it's better to err on the side of informing users more 😄

Good to go?

JelleZijlstra (approved):

.github/workflows/upload_binary.yml
```diff
@@ -29,10 +28,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up latest Python
         uses: actions/setup-python@v2
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "*"
```
cooperlees, line 34: I leave this line out in other CI jobs but I guess this is more explicit wildcard …

cooperlees (approved):

PR #2262, opened 2021-05-26
Fix and test docs on Windows

ichard26:
There's some weird interaction between Click and
sphinxcontrib-programoutput on Windows that leads to an encoding error
during the printing of black-primer's help text.

Also symlinks aren't well supported on Windows so let's just use
includes which actually work because we now use MyST :D

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -11,14 +11,17 @@ jobs:
       github.event_name == 'push' || github.event.pull_request.head.repo.full_name !=
       github.repository
 
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+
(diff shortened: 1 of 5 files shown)


HassanAbouelela (approved):
Tested locally, and everything works as expected.

Adding windows to the CI is an appreciated touch.

.github/workflows/doc.yml
```diff
@@ -11,14 +11,17 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Python 3.9
+      - name: Set up latest Python
```
cooperlees, line 23: 🥰

cooperlees (approved):
LGTM - Thanks doc guru.

Issue #2254, opened 2021-05-24
fmt:skip fails with internal error


memeplex:
**Describe the bug** A clear and concise description of what the bug is.

`# fmt:skip` fails with internal error message.

**To Reproduce** Steps to reproduce the behavior:

1. Open the link to vercel.app below
2. See the log at the right hand side

**Expected behavior** A clear and concise description of what you expected to happen.

Not sure, that it doesn't abort with an error at the bare minimum, I was just giving a try to the new `# fmt: skip` directive, my intention was to avoid a blank line before a local class definition, maybe it's not supposed to work like that, it's not clear to me, but that's one thing and another is:

```
error: cannot format /Users/carlos/Repositories/Jampp/predex-docs/dox.py: INTERNAL ERROR: Black produced different code on the second pass of the formatter.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /var/folders/st/d_bcbs_j6033r_g_dfztgmxc0000gn/T/blk_7soxut5i.log
Oh no! 💥 💔 💥
1 file failed to reformat.
```

**Environment (please complete the following information):**

I'm not sure this is relevant because I'm reproducing the issue in vercel.app, but here it is:

- Version: 21.5b0
- OS and Python version: 3.8.5

**Does this bug also happen on main?** To answer this, you have two options:

I hope this resource is persistent enough, because of the large URL I wasn't allowed to post directly from vercel.app:

[...]

PR #2281, opened 2021-05-30
Possible fix for issue with indentation and fmt: skip

enzet:
Possible fix for #2254 

Not sure the fix is right.  Here is what I found: the issue is connected with the line

```python    
first.prefix = prefix[comment.consumed :]
```
    
in `comments.py`.  `first.prefix` is a prefix of the line, that ends with `# fmt: skip`, but `comment.consumed` is the length of the `"  # fmt: skip"` string.  If prefix length is greater than 14, `first.prefix` will grow every time we apply formatting.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### _Black_
 
+- Fix failure caused by `fmt: skip` and indentation (#2281)
 - Correct max string length calculation when there are string operators (#2292)
 
 ## 21.5b2
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -159,7 +159,10 @@ def convert_one_fmt_off_pair(node: Node) -> bool:
             first = ignored_nodes[0]  # Can be a container node with the `leaf`.
             parent = first.parent
             prefix = first.prefix
-            first.prefix = prefix[comment.consumed :]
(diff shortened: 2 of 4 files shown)


ichard26:
Hey,

I just want to let you know that we should be able to get started on the review process for your changeset soon. Apologies for the wait. Your fix looks promising, but I'm bad when it comes to AST mangling code so take my opinion with a grain of salt :)

Looking forward to reviewing your work, thanks for working on this!

JelleZijlstra (approved):
This looks right and the tests look good, but I don't know this part of the code well.

ichard26 (approved):
I'm in roughly the same boat as Jelle, but I did spent a few hours somewhat recently in this specific part of the codebase trying to fix a bug (I wasn't successful) and this looks good to me as well. 

Thank you for the PR! It's awesome to see one more bug fixed. This project is only possible by contributions like these :black_heart:. You're awesome, @enzet! Perhaps we should have done some more testing before rolling out this new feature :sweat_smile:. Anyway, it would be even more awesome if you could provide some feedback on your experience contributing to psf/black. No pressure to do so, but it would be helpful. More details here: #2238. Thanks again.

enzet:
Thank you, @ichard26! Hope the fix is helpful.

PR #2282, opened 2021-05-30
Fix --experiemental-string-processing crash when matching parens not found

### LINKED ISSUE #2271: `--experimental-string-processing` crash on string + % {
```
% cat legacy.py                                   

def _legacy_listen_examples():
    text += (
        "    \"listen for the '%(event_name)s' event\"\n"
        "\n    # ... (event logic logic logic) ...\n"
        % {
            "since": since,
        }
    )
% black --experimental-string-processing legacy.py
error: cannot format legacy.py: Apparently, old parentheses do NOT exist?!
Oh no! 💥 💔 💥
1 file failed to reformat.
```
Minified from a file in sqlalchemy that we originally found this crash on.

The crash is in the `StringParenWrapper` class in `trans.py`, which apparently doesn't know how to handle the parentheses enclosing the string + `% {`. I spent some time looking at it but I'm not sure what the best solution is.

bbugyi200:
Fixes #2271.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -4466,9 +4466,10 @@ class StringParenWrapper(CustomSplitMapMixin, BaseStringSplitter):
                 right_leaves.pop()
 
             if old_parens_exist:
-                assert (
-                    right_leaves and right_leaves[-1].type == token.RPAR
-                ), "Apparently, old parentheses do NOT exist?!"
+                assert right_leaves and right_leaves[-1].type == token.RPAR, (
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Thanks! There's a pretty severe merge conflict, probably because you based this PR on a checkout from before we split up `__init__.py`.

bbugyi200:
@JelleZijlstra Yeah, I didn't realize the `master` branch was renamed to `main`. I'll close and resubmit.

PR #2283, opened 2021-05-30
Fix --experiemental-string-processing crash when matching parens not found

### LINKED ISSUE #2271: `--experimental-string-processing` crash on string + % {

bbugyi200:
Fixes #2271

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
   `.gitignore` rules like `git` does) (#2225)
 - Restored compatibility with Click 8.0 on Python 3.6 when LANG=C used (#2227)
 - Add extra uvloop install + import support if in python env (#2258)
+- Fix --experimental-string-processing crash when matching parens are not found (#2283)
 
 ### _Blackd_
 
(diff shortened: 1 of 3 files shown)


JelleZijlstra (approved):

ichard26:
Thank you so much for your contribution! This project is only possible by contributions like these :black_heart:. You're awesome, @bbugyi200! You've contributed **a lot** over the years to our string handling code. And while it may not be perfect, it's definitely an improvement. If you wouldn't mind, we would appreciate if you could give us some feedback on your experience contributing to psf/black. I bet you have a ton :wink: For more details, see #2238 

Afterwards, go have a break (whatever that means to you), you've done a lot. Thank you.

PR #2285, opened 2021-05-30
Regression fix: leave R prefixes capitalization alone

ichard26:
`black.strings.get_string_prefix` used to lowercase the extracted
prefix before returning it. This is wrong because 1) it ignores the
fact we should leave R prefixes alone because of MagicPython, and 2)
there is dedicated prefix casing handling code that fixes issue 1.
`.lower` is too naive.

This was originally fixed in 20.8b0, but was reintroduced since 21.4b0.

(Re)fixes GH-1244.

### DIFF
--- a/src/black/strings.py
+++ b/src/black/strings.py
@@ -87,7 +87,7 @@ def get_string_prefix(string: str) -> str:
     prefix = ""
     prefix_idx = 0
     while string[prefix_idx] in STRING_PREFIX_CHARS:
-        prefix += string[prefix_idx].lower()
+        prefix += string[prefix_idx]
         prefix_idx += 1
 
     return prefix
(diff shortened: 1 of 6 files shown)


JelleZijlstra:
I think there's code in `trans.py` that relies on the result of this function being lowercased. See trans.py line 438 for example, where we check for `"f" in prefix`.

felix-hilden:
Yeah after #2297 I was wondering why there are essentially two functions for getting docstring prefixes. Looking at their usage it seems that `get_prefix` is used more widely for just information about the prefix and `normalize` in one place to actually transform the lines. Maybe they should be combined in some way, or separated entirely into strictly getter and transformer. The dependencies to lowercase could very easily be implemented in the calling code.

src/black/strings.py
```diff
@@ -87,7 +87,7 @@ def get_string_prefix(string: str) -> str:
     prefix = ""
     prefix_idx = 0
     while string[prefix_idx] in STRING_PREFIX_CHARS:
-        prefix += string[prefix_idx].lower()
+        prefix += string[prefix_idx]
```
felix-hilden, line 90: I wonder if this could not repeatedly append to the string and just calculate the index, and then slice the whole prefix at once 🤔 Not really a comment to the PR but came to my mind anyway.

felix-hilden:
Apart from the comment above, the changes seem good to me at least. Leaving all processing to calling code seems appropriate in a function called `get`.

src/black/strings.py
felix-hilden: Sadly `min(s.find("'"), s.find('"'))` doesn't quite work because no matches is returned as -1 😄

ichard26: I'm taking this as that no you have no nits or objections for this PR :)

Thanks for the review though! You're doing great adjusting to your new role!

tests/data/long_strings__regression.py
```diff
@@ -1100,3 +1120,25 @@ async def foo(self):
...
+    f"<<{author.display_name}>>\n"
+)
+
+fstring = f"We have to remember to escape {braces}. Like {{these}}. But not {this}."
+
+welcome_to_programming = R"hello," R" world!"
+
+fstring = (
+    f"f-strings definitely make things more {difficult} than they need to be for"
+    " {black}. But boy they sure are handy. The problem is that some lines will need"
```
JelleZijlstra, line 1137: I like this style of test

JelleZijlstra (approved):

PR #2288, opened 2021-05-30
[primer] Enable everything

cooperlees:
- See if we pass all our repos with experimental string processing enabled.

Django probably needed:
- Ignores
- >= 3.8 only
  - We could support PEP440 version specifiers, but that would introduce the `packaging` module as a dependency that I'd like to avoid ... Or I could implement a poor persons version or vendor

### DIFF
--- a/src/black_primer/lib.py
+++ b/src/black_primer/lib.py
@@ -39,7 +39,7 @@ class Results(NamedTuple):
 
 async def _gen_check_output(
     cmd: Sequence[str],
-    timeout: float = 300,
+    timeout: float = 600,
     env: Optional[Dict[str, str]] = None,
     cwd: Optional[Path] = None,
 ) -> Tuple[bytes, bytes]:
--- a/src/black_primer/primer.json
+++ b/src/black_primer/primer.json
@@ -30,8 +30,6 @@
       "py_versions": ["all"]
     },
     "django": {
-      "disabled_reason": "black --check --diff returned 123 on tests_syntax_error.py",
-      "disabled": true,
       "cli_arguments": ["--experimental-string-processing"],
       "expect_formatting_changes": true,
       "git_clone_url": "https://github.com/django/django.git",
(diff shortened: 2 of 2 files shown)


src/black_primer/lib.py
```diff
@@ -39,7 +39,7 @@ class Results(NamedTuple):
 
 async def _gen_check_output(
     cmd: Sequence[str],
-    timeout: float = 300,
+    timeout: float = 600,
```
cooperlees, line 42: Maybe we could also move this to be per project ... So we don't wait for ever for small projects ...

src/black_primer/primer.json
```diff
@@ -30,13 +30,16 @@
...
+        "--experimental-string-processing",
+        "--skip-string-normalization",
+        "--extend-exclude",
+        "/((docs|scripts)/|django/forms/models.py|tests/gis_tests/test_spatialrefsys.py|tests/test_runner_apps/tagged/tests_syntax_error.py)"
+      ],
       "expect_formatting_changes": true,
       "git_clone_url": "https://github.com/django/django.git",
       "long_checkout": false,
-      "py_versions": ["all"]
+      "py_versions": ["3.8", "3.9"]
```
cooperlees, line 42: Here is where PEP440 version specifiers could be nice:
```suggestion
      "py_versions": [">=3.8"]
```

ichard26: The development environment already mandates `packaging` as required due to Sphinx, tox, and pytest. We could lock away primer under an extra to avoid the extra dependency for production installs.

src/black_primer/primer.json
```diff
@@ -53,8 +56,6 @@
       "py_versions": ["all"]
     },
     "pandas": {
-      "disabled_reason": "black-primer runs failing on Pandas - #2193",
-      "disabled": true,
       "cli_arguments": ["--experimental-string-processing"],
```
ichard26, line 59: Hmm, my local copy of Black from main failed to format a single file for pandas. Looks like either my copy of mypyc is subtly broken or my source files are out of date.

ichard26 (approved):
I want this for my mypyc work and everything looks OK.

Issue #2251, opened 2021-05-23
[Python 2 has been deprecated] - Formalise Python support policy


felix-hilden:
### Status update

_November 1st, 2021 @ 10:36AM EST_ by @ichard26 

**We've deprecated Python 2 support starting with version 21.10b0**. Assuming things go as planned the support will be **dropped sometime in January of 2022** along side the first stable release of Black.

---

From the discussions in #2247. The new FAQ (once merged) contains information on the supported Python versions. @ichard26 mentioned that it would be nice to have such documentation elsewhere in the docs too.

As discussed, the support plan seems to essentially be: "support whatever CPython versions are currently not EOL'd". This is for running Black. Then another question is formatting files, which is currently also done for 2.7.

- Should the Python version support of Black be considered identical to CPython official support?
- Does file formatting follow that, or do we intend to support older versions too?

cooperlees:
I'd like to propose dropping Python 2 formatting support in the near future. It's still causing formatting decisions that are not beneficial to modern python3 users - e.g. `fr""` vs `rf""` strings.

felix-hilden:
To go a bit further, would we also like for supported formatting versions follow the supported runtime versions? Currently `black/mode/TargetVersion` has versions from 3.3 onwards too.

cooperlees:
> To go a bit further, would we also like for supported formatting versions follow the supported runtime versions? Currently `black/mode/TargetVersion` has versions from 3.3 onwards too.

I believe this is the goal. e.g. as soon as Python 3.6 is EOL'd we'll remove that target version and CI from black + no longer maintain comparability etc.

septatrix:
I strongly believe that teams which still use Python 2 can simply fix their black version to a version still supporting py2. They have to do that for other tools anyway (like pip which also does not support py2 any longer).

pradyunsg:
Honestly, I don't see any reason to maintain support for a single day beyond EoL.

The main beneficiaries of extended post-EoL support for Python versions are organisations/institutions. These have taken on risk by staying on EoL software. One of the risk factors is lack of support in tooling that they don't pay for. The onus is on them to figure out how to keep their lights on. However they do so, it must cost them $$$ -- either by paying maintainers enough to justify their effort spent maintaining that support, by having the developers they have hired maintain the support for them internally or by paying a redistributor to maintain that support for them.

From https://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html:

[...]

PR #2297, opened 2021-06-01
Normalise string prefix order

### LINKED ISSUE #2171: Normalize raw f-strings prefix

Black should normalize the prefix used for raw f-strings (https://www.python.org/dev/peps/pep-0498/#raw-f-strings). Both `rf` and `fr` work, but one should be preferred.

Currently, the following code is not changed:

```
    f = rf"hello {world}"
    f = fr"hello {world}"
```

Pick one and normalize. I don't really care which. `rf` matches the name "raw f-string", but `fr` matches the example in PEP-498 and Jupyterlab syntax highlighting seems to prefer it.

Doing nothing leaves open the opportunity for inconsistency and disagreement.

**Additional context**

None

felix-hilden:
Closes #2171: this PR makes _Black_ normalise string prefix order. Since no other maintainer but Jelle responded to the last poll, I went ahead and put <strike>"r" first since it was also the winner of the first poll. But that's easy to change if you'd like.</strike> (<strike>"r" last now, see issue and below</strike> ("r" first now again lol))

- [x] Tests and changed existing tests and source to comply
- [x] Changelog and style documentation
- [x] 100% money back guarantee

---
Also, don't worry: I wrote "normalize" to the source for consistency. And for freedom! 🦅

### DIFF
--- a/docs/the_black_code_style/current_style.md
+++ b/docs/the_black_code_style/current_style.md
@@ -235,10 +235,10 @@ escapes than before.
 
 _Black_ also standardizes string prefixes. Prefix characters are made lowercase with the
 exception of [capital "R" prefixes](#rstrings-and-rstrings), and in the case of multiple
-characters "r" is put first as in spoken language: "raw f-string". On top of that, if
-your code is already Python 3.6+ only or it's using the `unicode_literals` future
-import, _Black_ will remove `u` from the string prefix as it is meaningless in those
-scenarios.
+characters "r" is put last as the common prefix and as required in Python 2. On top of
+that, if your code is already Python 3.6+ only or it's using the `unicode_literals`
+future import, _Black_ will remove `u` from the string prefix as it is meaningless in
+those scenarios.
 
 The main reason to standardize on a single form of quotes is aesthetics. Having one kind
 of quotes everywhere reduces reader distraction. It will also enable a future version of
--- a/docs/the_black_code_style/current_style.md
+++ b/docs/the_black_code_style/current_style.md
@@ -233,8 +233,10 @@ _Black_ prefers double quotes (`"` and `"""`) over single quotes (`'` and `'''`)
 will replace the latter with the former as long as it does not result in more backslash
 escapes than before.
 
-_Black_ also standardizes string prefixes, making them always lowercase. On top of that,
-if your code is already Python 3.6+ only or it's using the `unicode_literals` future
+_Black_ also standardizes string prefixes. Prefix characters are made lowercase with the
+exception of [capital "R" prefixes](#rstrings-and-rstrings), and in the case of multiple
+characters "r" is put first as in spoken language: "raw f-string". On top of that, if
+your code is already Python 3.6+ only or it's using the `unicode_literals` future
 import, _Black_ will remove `u` from the string prefix as it is meaningless in those
 scenarios.
 
--- a/docs/the_black_code_style/current_style.md
+++ b/docs/the_black_code_style/current_style.md
@@ -235,10 +235,10 @@ escapes than before.
 
 _Black_ also standardizes string prefixes. Prefix characters are made lowercase with the
 exception of [capital "R" prefixes](#rstrings-and-rstrings), and in the case of multiple
-characters "r" is put last as the common prefix and as required in Python 2. On top of
-that, if your code is already Python 3.6+ only or it's using the `unicode_literals`
-future import, _Black_ will remove `u` from the string prefix as it is meaningless in
-those scenarios.
+characters "r" is put first as in spoken language: "raw f-string". On top of that, if
+your code is already Python 3.6+ only or it's using the `unicode_literals` future
+import, _Black_ will remove `u` from the string prefix as it is meaningless in those
+scenarios.
 
 The main reason to standardize on a single form of quotes is aesthetics. Having one kind
 of quotes everywhere reduces reader distraction. It will also enable a future version of
--- a/docs/the_black_code_style/current_style.md
+++ b/docs/the_black_code_style/current_style.md
@@ -234,11 +234,9 @@ will replace the latter with the former as long as it does not result in more ba
 escapes than before.
 
 _Black_ also standardizes string prefixes. Prefix characters are made lowercase with the
-exception of [capital "R" prefixes](#rstrings-and-rstrings), and in the case of multiple
-characters "r" is put first as in spoken language: "raw f-string". On top of that, if
-your code is already Python 3.6+ only or it's using the `unicode_literals` future
-import, _Black_ will remove `u` from the string prefix as it is meaningless in those
-scenarios.
+exception of [capital "R" prefixes](#rstrings-and-rstrings), unicode literal markers
+(`u`) are removed because they are meaningless in Python 3, and in the case of multiple
+characters "r" is put first as in spoken language: "raw f-string".
 
 The main reason to standardize on a single form of quotes is aesthetics. Having one kind
 of quotes everywhere reduces reader distraction. It will also enable a future version of
--- a/src/black/strings.py
+++ b/src/black/strings.py
@@ -151,7 +151,7 @@ def normalize_string_prefix(s: str) -> str:
     )
 
     # Python syntax guarantees max 2 prefixes and that one of them is "r"
-    if len(new_prefix) == 2 and "r" != new_prefix[0].lower():
+    if len(new_prefix) == 2 and "r" == new_prefix[0].lower():
         new_prefix = new_prefix[::-1]
     return f"{new_prefix}{match.group(2)}"
 
@@ -188,9 +188,9 @@ def normalize_string_quotes(s: str) -> str:
         return s  # There's an internal error
 
     prefix = s[:first_quote_pos]
-    unescaped_new_quote = _cached_compile(rf"(([^\\]|^)(\\\\)*){new_quote}")
-    escaped_new_quote = _cached_compile(rf"([^\\]|^)\\((?:\\\\)*){new_quote}")
-    escaped_orig_quote = _cached_compile(rf"([^\\]|^)\\((?:\\\\)*){orig_quote}")
+    unescaped_new_quote = _cached_compile(fr"(([^\\]|^)(\\\\)*){new_quote}")
+    escaped_new_quote = _cached_compile(fr"([^\\]|^)\\((?:\\\\)*){new_quote}")
+    escaped_orig_quote = _cached_compile(fr"([^\\]|^)\\((?:\\\\)*){orig_quote}")
     body = s[first_quote_pos + len(orig_quote) : -len(orig_quote)]
     if "r" in prefix.casefold():
         if unescaped_new_quote.search(body):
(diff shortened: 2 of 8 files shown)


tests/data/string_prefixes.py
```diff
@@ -1,18 +1,24 @@
...
+(u"", U"")
+(r"", R"")
+
+(rf"", fr"", Rf"", fR"", rF"", Fr"", RF"", FR"")
+(rb"", br"", Rb"", bR"", rB"", Br"", RB"", BR"")
 
 # output
 
 
 #!/usr/bin/env python3.6
```
cooperlees, line 15: NIT: I feel we should just go to python3 here
```suggestion
#!/usr/bin/env python3
```

src/black/strings.py
```diff
@@ -136,7 +136,10 @@ def normalize_string_prefix(s: str, remove_u_prefix: bool = False) -> str:
     match = re.match(r"^([" + STRING_PREFIX_CHARS + r"]*)(.*)$", s, re.DOTALL)
     assert match is not None, f"failed to match string {s!r}"
     orig_prefix = match.group(1)
+
     new_prefix = orig_prefix.replace("F", "f").replace("B", "b").replace("U", "u")
+    if len(new_prefix) == 2 and "r" != new_prefix[0].lower():
```
cooperlees, line 141: Just double checking - There is never a chance of > 2 here right?

- Also, lets set a const for 2 to explain it's value

Feel free to pick a better name etc.
```suggestion
    TWO_CHAR_PREFIX_LEN = 2
    if len(new_prefix) == TWO_CHAR_PREFIX_LEN and "r" != new_prefix[0].lower():
```

cooperlees (approved):
LGTM. Just some unimportant suggestions + NITs.

tests/data/string_prefixes.py
felix-hilden: Consider it done!

src/black/strings.py
felix-hilden: Yeah about that, at least with Python 3 it's max 2. Never used Python 2 tho, but trying on an online interpreter, "rb" is invalid and "f" doesn't exist. "ru" and "ub" are also invalid. So we should be ok.

I'll fight you a bit with adding the constant, if I may. Having "two char prefix len" be 2 is both a bit too obvious and not explaining enough in my opinion. Would `len(pf) > 1` or maybe a comment saying max 2 is guaranteed be better? I won't die on this hill though 😄 so feel free to just put your foot down.

cooperlees: I also don't care enough to fight this. I've just always found it easier to come back to code with consts explaining a naked integer years / even weeks later. Happy with this if other maintainers are.

felix-hilden: Added a comment there explaining it a bit more, that is definitely appropriate and enables a more verbose explanation. Will add the const if others also agree!

Numerlor: 2 Seems to be the max according to the lexical analysis page on both versions.
I think a comment would be best here, as a var name won't imo explain why it's two without being overly verbose.

While not directly related, is there any reason for the yoda condition? `... new_prefix[0].lower() != "r"` reads better to me

felix-hilden: I guess "r is not the first prefix char" seemed more intuitive for me when writing it, but it could be "the first char is not r" just as well!

tests/data/python2.py
```diff
@@ -5,10 +5,11 @@
 print >> sys.stderr , "Warning:" ,
 print >> sys.stderr , "this is a blast from the past."
 print >> sys.stderr , "Look, a repr:", `sys`
+print >> sys.stderr , U"Even these prefixes are normalized!"
 
 
 def function((_globals, _locals)):
-    exec ur"print 'hi from exec!'" in _globals, _locals
+    exec "print 'hi from exec!'" in _globals, _locals
```
JelleZijlstra, line 12: Why are the `ur` removed here?

felix-hilden: Oh wow. "ru" is invalid but "ur" isn't? 😅 Good catch. Let me go through the others once more.

felix-hilden: "br" is valid and "rb" invalid. Well. This is a bummer. They go exactly against the rule. Should we go with "r last" after all?

JelleZijlstra: Oh you're right. And in Python 3 they're both invalid.

felix-hilden: Yeah "u" with anything else seems to be invalid, but "br" naturally works.

JelleZijlstra: Sorry that comment was in reference to `ru` not `rb`.

In summary:
- Python 3 allows `rb`, `br`, but not `ur` or `ru`
- Python 2 allows `br` and `ur` but not `rb` or `ru`

felix-hilden: I commented on the original issue and marked this PR as a draft. We ought to discuss it through before merging anything.

src/black/strings.py
jamesmyatt: What about some logic along the lines of "if there's an r in the prefix, then move it to the front" which will also work for future multi-character prefixes? You could also put the rest of the prefix in alphabetical order at the same time for future consistency. Might be gold-plating though.

felix-hilden: Yeah, reversing is essentially a shorthand in that case. I don't expect many more prefixes to be added so I just left it at that. Are there any on the horizon? If not, I'd say it's just making it too fancy, although strictly correct 😄

felix-hilden:
In the absense of comments from maintainers on the linked issue I went ahead and changed the rule to "r" last to take Python 2 into account more easily. As mentioned in the issue, since there were no strong opinions either way I think this is a reasonable switch.

src/black/strings.py
```diff
@@ -137,6 +137,10 @@ def normalize_string_prefix(s: str, remove_u_prefix: bool = False) -> str:
     assert match is not None, f"failed to match string {s!r}"
     orig_prefix = match.group(1)
     new_prefix = orig_prefix.replace("F", "f").replace("B", "b").replace("U", "u")
+
+    # Python syntax guarantees max 2 prefixes and that one of them is "r"
+    if len(new_prefix) == 2 and "r" == new_prefix[0].lower():
+        new_prefix = new_prefix[::-1]
```
felix-hilden, line 143: Sadly `sorted` isn't quite enough here because of `ur` :/

src/black/strings.py
```diff
@@ -137,6 +137,10 @@ def normalize_string_prefix(s: str, remove_u_prefix: bool = False) -> str:
     assert match is not None, f"failed to match string {s!r}"
     orig_prefix = match.group(1)
     new_prefix = orig_prefix.replace("F", "f").replace("B", "b").replace("U", "u")
+
+    # Python syntax guarantees max 2 prefixes and that one of them is "r"
+    if len(new_prefix) == 2 and "r" == new_prefix[0].lower():
```
felix-hilden, line 142: And technically the first condition is redundant, but I like it because it clarifies the meaning.

felix-hilden:
A friendly bump! My reviews above are only remarks, so I think this should be ready for review.

tests/data/string_prefixes.py
```diff
@@ -1,18 +1,24 @@
-#!/usr/bin/env python3.6
+#!/usr/bin/env python3
 
-name = R"Łukasz"
```
JelleZijlstra, line 3: Any reason to change the input section here?

felix-hilden: Just for reducing a redundant test, since all the prefixes are tested below! I thought about removing it altogether, but seems fine to test f-string substitution there at least. Would you like to keep it?

src/black/strings.py
Numerlor: Could use a key like this with `sorted`, I think it's nicer than the reverse but I'm not sure if it reads better overall.
```py
new_prefix = "".join(sorted(new_prefix, key="fubrR".index))
```
`STRING_PREFIX_CHARS` could also be reordered to be used as the index string

felix-hilden: That could actually be quite fun 😄 but with #2251 we are still second guessing the decisions, so this is a bit blocked.

jamesmyatt: https://github.com/psf/black/pull/2740 has been merged now

src/black/strings.py
felix-hilden: Also it short circuits the (likely) more complex subscript and `.lower`.

jamesmyatt:
FWIW, I liked the `new_prefix = "".join(sorted(new_prefix, key="fbrR".index))` version, especially since "fbrR" could be replaced with a module-level variable, e.g. `PREFIX_ORDER`, which can then be modified externally if someone wants to hack it to change the order. OTOH, having options for changing the behaviour is not in black's MO.

felix-hilden:
@jamesmyatt Fair enough, but I don't much care for how it reads. It's easy enough to read the current version as "if "r" isn't first, switch the characters around". Using a join and sorting by an index without a condition is quite a bit more complex and less efficient. And yeah, we won't introduce a hackable value 😄 although I don't know what'll stop a dedicated monkey patcher.

docs/the_black_code_style/current_style.md
```diff
@@ -233,8 +233,10 @@ _Black_ prefers double quotes (`"` and `"""`) over single quotes (`'` and `'''`)
 will replace the latter with the former as long as it does not result in more backslash
 escapes than before.
 
-_Black_ also standardizes string prefixes, making them always lowercase. On top of that,
-if your code is already Python 3.6+ only or it's using the `unicode_literals` future
+_Black_ also standardizes string prefixes. Prefix characters are made lowercase with the
+exception of [capital "R" prefixes](#rstrings-and-rstrings), and in the case of multiple
+characters "r" is put first as in spoken language: "raw f-string". On top of that, if
+your code is already Python 3.6+ only or it's using the `unicode_literals` future
```
JelleZijlstra, line 239: This should be changed (somewhat unrelated to this PR): we only support formatting 3.6+, so we currently always remove the `u`.

JelleZijlstra (commented):
I like r-first, since we say "raw f-string" not "f-raw string".

docs/the_black_code_style/current_style.md
felix-hilden: Good catch 👍 We still support 3.3-3.5 right? Not that it changes the message but anyways.

JelleZijlstra: Oh right, we do. But I'm not sure what the "3.6+" is about in the first place in that case, since u prefixes are redundant in all of Python 3. (And in Python 3.0-3.2 they did nothing.)

felix-hilden:
The primer fail seems unrelated (thanks Jelle! ❤️) So this could now finally be ready for the last comments and a merge!

JelleZijlstra (approved):

felix-hilden:
Last force to fix the primer after #2769

docs/the_black_code_style/current_style.md
```diff
@@ -233,10 +233,10 @@ _Black_ prefers double quotes (`"` and `"""`) over single quotes (`'` and `'''`)
...
 escapes than before.
 
-_Black_ also standardizes string prefixes, making them always lowercase. On top of that,
-if your code is already Python 3.6+ only or it's using the `unicode_literals` future
-import, _Black_ will remove `u` from the string prefix as it is meaningless in those
-scenarios.
+_Black_ also standardizes string prefixes. Prefix characters are made lowercase with the
+exception of [capital "R" prefixes](#rstrings-and-rstrings), unicode literal markers
+(`u`) are removed because they are meaningless in Python 3, and in the case of multiple
+characters "r" is put first as in spoken language: "raw f-string".
```
cooperlees, line 239: Love this and agree with all!

cooperlees (approved):
All looks good to me. I'm looking forward to this normalization. Thanks!

ichard26:
Yay we did it, it only took half a year haha :)

cooperlees:
> Yay we did it, it only took half a year haha :)

We can blame Python 2 ... Which we will never have to every again :D

jamesmyatt:
Thanks so much @felix-hilden and @JelleZijlstra

PR #2300, opened 2021-06-02
Add option to require a specific version to be running

### LINKED ISSUE #1246: Feature Request - add required black version in pyproject.toml
Many projects that I contribute to have adopted black. (Congrats and Thank You!)

An issue which has come up in multiple projects is automatic tests failing because of differences in the installed `black` versions between environments.

For example: a project may indicate a particular version of black in a pre-commit configuration in `.pre-commit-config.yaml` (https://black.readthedocs.io/en/stable/version_control_integration.html) which will re-format the code to that version... but that will not-necessarily be the same version which is invoked by the local developer or the build/deploy service - as black will consult the `pyproject.toml` file instead (https://black.readthedocs.io/en/stable/pyproject_toml.html).

This can create issues on automated tests, because the server environment may be configured to install `19.3b0` while the hook or developer's environment is formatting to `19.10b0`.  This can suggest a developer didn't run `black` or made an error, when the build/deploy system is what requires adjustment.

A potential solution is to extend the `pyproject.toml` configuration to have an argument which states the required version:

    [tool.black]
    line-length = 88
    rev = 19.10b0

If specified, black would then raise a fatal exception if invoked with a version that does not match. Perhaps a commandline flag is needed to avoid this for certain causes.  Some projects specify a specific version of black for a particular test environment in `tox.ini`, but nothing is global.

An alternative might also include migrating the version declaration from the `.pre-commit-config.yaml` into the `pyproject.toml` file, where it seems to be more applicable.  However, I assume the revision was declared in this file for important reasons I do not understand yet.

To summarize:

* pre-config integration is great, but if a contributor doesn't have pre-commit installed, they'll invoke black directly
* if a contributor invokes black directly, they may format the code with a different version than the project uses
* automated systems can pick up a new version of black; by may not pin the version in the tox config for the test. or they use travis.

having a centralized version pin, instead of requiring a version pin for each context, would eliminate many problems.

Thank you.

see also: #420

felix-hilden:
Closes #1246: This PR adds a new option (and automatically a toml entry, hooray for existing configuration management 🎉) to require a specific version of Black to be running.

For example: `black --required-version 20.8b -c "format = 'this'"`

Execution fails straight away if it doesn't match `__version__`. Some points of contention:
- Option placement: higher or lower on the CLI and help
- Option name: <strike>revision was suggested in the issue, I agree that it's the most appropriate, but it isn't perfect. Other possibilities: `runtime-version`, `black-version`..?</strike> `required-version` was chosen
- Option shorthand: I don't think it's needed
- Error message: <strike>particularly printing the configuration location</strike>
- Documentation: new section name, placement, anything else

### DIFF
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -176,18 +176,18 @@ $ black --version
 black, version 21.5b0
 ```
 
-An option to requiring a specific version to be running is also provided.
+An option to require a specific version to be running is also provided.
 
 ```console
-$ black --revision 21.5b2 -c "format = 'this'"
+$ black --required-version 21.5b2 -c "format = 'this'"
 format = "this"
-$ black --revision 31.5b2 -c "still = 'beta?!'"
-Oh no! 💥 💔 💥 The required revision does not match the running version!
+$ black --required-version 31.5b2 -c "still = 'beta?!'"
+Oh no! 💥 💔 💥 The required version does not match the running version!
 ```
 
 This is useful for example when running _Black_ in multiple environments that haven't
-necessarily installed the correct version. Setting this option can be done in a
-configuration file for consistent results across environments.
+necessarily installed the correct version. This option can be set in a configuration
+file for consistent results across environments.
 
 ## Configuration via a file
 
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -167,7 +167,7 @@ $ black src/ -q
 error: cannot format src/black_primer/cli.py: Cannot parse: 5:6: mport asyncio
 ```
 
-### Getting the version
+### Versions
 
 You can check the version of _Black_ you have installed using the `--version` flag.
 
@@ -175,6 +175,19 @@ You can check the version of _Black_ you have installed using the `--version` fl
 $ black --version
 black, version 21.5b0
 ```
+
+An option to requiring a specific version to be running is also provided.
+
+```console
+$ black --revision 21.5b2 -c "format = 'this'"
+format = "this"
+$ black --revision 31.5b2 -c "still = 'beta?!'"
+Oh no! 💥 💔 💥 The required revision does not match the running version!
+```
+
+This is useful for example when running _Black_ in multiple environments that haven't
+necessarily installed the correct version. Setting this option can be done in a
+configuration file for consistent results across environments.
 
 ## Configuration via a file
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -250,6 +250,14 @@ def validate_regex(
     help="If --fast given, skip temporary sanity checks. [default: --safe]",
 )
 @click.option(
+    "--revision",
+    type=str,
+    help=(
+        "Require a specific version of Black to be running (useful for unifying results"
+        " across many environments e.g. with a pyproject.toml file)."
+    ),
+)
+@click.option(
     "--include",
     type=str,
     default=DEFAULT_INCLUDES,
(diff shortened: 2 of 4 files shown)


src/black/__init__.py
```diff
@@ -436,9 +456,9 @@ def main(
             )
 
     if verbose or not quiet:
-        out("Oh no! 💥 💔 💥" if report.return_code else "All done! ✨ 🍰 ✨")
+        out(error_msg if report.return_code else "All done! ✨ 🍰 ✨")
         if code is None:
-            click.secho(str(report), err=True)
+            click.echo(str(report), err=True)
```
felix-hilden, line 461: This is a bit unrelated but noticed that styling isn't used here, so I changed it.

docs/usage_and_configuration/the_basics.md
```diff
@@ -176,6 +176,19 @@ $ black --version
 black, version 21.5b0
 ```
 
+An option to requiring a specific version to be running is also provided.
```
JelleZijlstra, line 179:
```suggestion
An option to require a specific version to be running is also provided.
```

docs/usage_and_configuration/the_basics.md
```diff
@@ -176,6 +176,19 @@ $ black --version
...
+
+```console
+$ black --revision 21.5b2 -c "format = 'this'"
+format = "this"
+$ black --revision 31.5b2 -c "still = 'beta?!'"
+Oh no! 💥 💔 💥 The required revision does not match the running version!
+```
+
+This is useful for example when running _Black_ in multiple environments that haven't
+necessarily installed the correct version. Setting this option can be done in a
```
JelleZijlstra, line 189:
```suggestion
necessarily installed the correct version. This option can be set in a
```
Fewer words is better

src/black/__init__.py
```diff
@@ -368,6 +377,17 @@ def main(
     config: Optional[str],
 ) -> None:
     """The uncompromising code formatter."""
+    error_msg = "Oh no! 💥 💔 💥"
+    if revision and revision != __version__:
+        msg = (
+            f"{error_msg} The required revision `{revision}` does not match"
+            f" the running version `{__version__}`!"
```
JelleZijlstra, line 384: Half of this message says "revision" and the other half says "version" for the same concept, which might be an indication that the option name has to change :)

I don't really like the term "revision" because it makes me think of a Git commit instead of a release.

JelleZijlstra (commented):
A couple of wording comments.

Also, what about `--release` for the version name? After all, it only works with released versions.

Now going to check if other tools (Prettier, gofmt?) also have options like this that we could steal the name of.

src/black/__init__.py
felix-hilden: Yep this was very awkward :D "revision" does remind me of smaller changes too, "release" sounds much better.

JelleZijlstra:
Rustfmt uses `--required-version`: https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=#required_version

Prettier, gofmt, and clang-format don't appear to have a similar option.

felix-hilden:
That or "release", both sound good to me! I think I'll go with "required version" in the absense of other comments, because it's more explicit. Thanks a lot for the wording improvements!

src/black/__init__.py
```diff
@@ -368,6 +377,17 @@ def main(
     config: Optional[str],
 ) -> None:
     """The uncompromising code formatter."""
+    if config and verbose:
+        out(f"Using configuration from {config}.", bold=False, fg="blue")
```
felix-hilden, line 381: I deleted the previous printing of config file location, I think it's better to leave it to --verbose.

felix-hilden:
There ya go, keep the comments coming if something else is off!

JelleZijlstra:
Sounds good! The test failure indeed looks random.

It occurred to me that this option name could be confused with `--target-version`, which is a different kind of version. That's probably fine though.

JelleZijlstra (approved):

ichard26 (approved):

jvanasco:
Thank you so much!

PR #2301, opened 2021-06-02
Switch `toml` TOML library for `tomli`

### LINKED ISSUE #2280: Valid toml syntax causes black to crash
**Describe the bug**

I have a pyproject.toml for configuring various other tools (not black), which includes a list with mixed type contents, as explicitly support by the v1.0.0 of the toml spec.

When I run black it crashes with a cryptic traceback from inside the toml library.

**To Reproduce**

```bash
mkdir new_project
cat <<EOT > ./new_project/pyproject.toml
# this example is from the toml spec
contributors = [
  "Foo Bar <foo@example.com>",
  { name = "Baz Qux", email = "bazqux@example.com", url = "https://example.com/bazqux" }
]
EOT
cd new_project
black .
```

**Expected behavior**

It shouldn't crash.

- Version: black, version 21.5b1
- OS and Python version: MacOs Python 3.8.8

---

Since the presently used "toml" library explicitly only supports up to version 0.5.0 of the spec (from about two years ago), I'd suggest switching to [tomlkit](https://github.com/sdispater/tomlkit), which supports the current toml spec (officially up to 1.0.0rc1 but nothing has materially changed since then).

I'd be happy to put together a PR if this approach is deemed acceptable in principle.

ichard26:
`toml` unfortunately has a lack of maintainership issue right now. It's
evident by the fact `toml` only supports TOML v0.5.0. TOML v1.0.0 has
been recently released and right now Black crashes hard on its usage.

`tomli` is a brand new parse only TOML library. It supports TOML
v1.0.0. Although TBH we're switching to this one mostly because
pip is doing the same.

--- 

Blocked on https://github.com/pypa/pip/pull/10035, fixes #2280.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ setup(
     install_requires=[
         "click>=7.1.2",
         "appdirs",
-        "toml>=0.10.1",
+        "tomli>=0.2.6,<2.0.0",
         "typed-ast>=1.4.2; python_version < '3.8'",
         "regex>=2020.1.8",
         "pathspec>=0.8.1, <1",
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 ### _Black_
 
+- Configuration files using TOML features higher than spec v0.5.0 are now supported
+  (#2301)
 - Add primer support and test for code piped into black via STDIN (#2315)
 - Fix internal error when `FORCE_OPTIONAL_PARENTHESES` feature is enabled (#2332)
 - Accept empty stdin (#2346)
--- a/Pipfile
+++ b/Pipfile
@@ -32,7 +32,7 @@ click = ">=8.0.0"
 mypy_extensions = ">=0.4.3"
 pathspec = ">=0.8.1"
 regex = ">=2020.1.8"
-toml = ">=0.10.1"
+tomli = ">=0.2.6, <2.0.0"
 typed-ast = "==1.4.2"
 typing_extensions = {"python_version <" = "3.8","version >=" = "3.7.4"}
 black = {editable = true,extras = ["d"],path = "."}
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "0c126694d1a1cd1c4dc875a99f330f6b9a5994e51a1f870a98dbacd0185dae53"
+            "sha256": "e6b7c19c10b5fba5ba20296b68049fab625b4c6d61db97b79d83a6dfa36d9f6d"
         },
         "pipfile-spec": 6,
         "requires": {},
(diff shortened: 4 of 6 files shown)


src/black/__init__.py
```diff
@@ -112,9 +112,7 @@ def read_pyproject_toml(
     try:
         config = parse_pyproject_toml(value)
     except (OSError, ValueError) as e:
-        raise click.FileError(
-            filename=value, hint=f"Error reading configuration file: {e}"
-        )
+        raise click.UsageError(f"unable to read configuration file ({value}): {e}")
```
JelleZijlstra, line 115: What is the reason for this change?

cooperlees:
Sweet - I am all for this. I've been watching the pip issue and was planning to do this once accepted. It's been verbally so I am happy to go this route too.

src/black/__init__.py
JelleZijlstra: Sorry, it's in the PR description.

ichard26: During manual testing I didn't really like the existing error message when an error happens during reading or parsing:

main:

```console
❯ black --config=tests/invalid.toml
Error: Could not open file 'tests/invalid.toml': Error reading configuration file: Expected "=" after a key in a key-to-value mapping
```

this PR:

```console
❯ black --config=tests/invalid.toml
Usage: black [OPTIONS] [SRC]...
Try 'black -h' for help.

Error: unable to read configuration file (tests/invalid.toml): Expected "=" after a key in a key-to-value mapping
```

JelleZijlstra: I actually prefer the message from main. The usage information is noise here, since the problem isn't with the command line flag syntax, but with the contents of the config file.

ichard26: Fair enough, I'll revert.

pyproject.toml
```diff
@@ -31,4 +31,4 @@ build-backend = "setuptools.build_meta"
 optional-tests = [
   "no_python2: run when `python2` extra NOT installed",
   "no_blackd: run when `d` extra NOT installed",
-]
\ No newline at end of file
+]
```
cooperlees, line 34: I take it tailing \n is part of standard?

cooperlees (approved):
Nice. LGTM especially with the test.toml being loaded with new syntax.

pyproject.toml
ichard26: honestly I don't recall how this change slipped in here (whether it was manual or by some tool), but I can't imagine it being harmful so I'll leaving it alone :p

setup.py
```diff
@@ -74,7 +74,7 @@ def get_long_description() -> str:
     install_requires=[
         "click>=7.1.2",
         "appdirs",
-        "toml>=0.10.1",
+        "tomli>=0.2.6",
```
hukkin, line 77:
```suggestion
        "tomli>=0.2.6,<2.0.0",
```
I would much appreciate this upper bound for the version. There's many open source projects afraid to make any changes, ever, due to the fear of downstream projects breaking. Luckily we have a solution: semantic versions. Let's use them :wink:

cooperlees: I get the goals here, but I feel this kind of pinning is counterintuitive and causes unneeded stagnation. A recent example here is "myst-parser" blocking a sphinx upgrade on another project I help maintain: https://github.com/pypa/bandersnatch/pull/919
- I went and looked why and it's some formatting / mathematical syntax that's blocking it - i.e. it would work fine for 99.9999% of people ...

I feel packages should give people the two options version freedom and also offer pinned version. Thus I feel black should not do the '<2.0.0' here. Users of libraries should run their own CI to prove the dependencies works for them. Nice packages should also offer pinned (package==X/X) requirement files ([requirements.txt](https://github.com/pypa/bandersnatch/blob/master/requirements.txt)) or container images with tested versions for people who want to go that route. With Black, we do this via docker releases that give you the whole dependency tree pinned.

hukkin: > Users of libraries should run their own CI to prove the dependencies works

I agree 100%. Generally the rule should be that libraries allow unbounded dep versions (unless something is known not to work) and end-user applications should pin. My understanding is that Black is an end-user application, not a library (although can be used as one with no guarantees). Most users are installing via `pip install`, not using Docker, and will probably not be very happy when an upstream project makes a perfectly valid semantic version major version bump, and suddenly the `pip install black` is broken.

As to why myst-parser does the pinning, the reason is exactly the same: it considers itself an application first and foremost, not a library.

Also note that `tomli`'s public API is literally two functions and one error. If anything there changes, it is a thousand times more likely for something to break downstream than in the myst-parser example.

If you still disagree, I can totally respect that, but like said would appreciate the upper bound. I find it unlikely I'll ever tag a tomli v2.0.0 but if that does happen, these pins are incredibly nice to have.

pyproject.toml
WhyNotHugo: It'll add noise to the diff next time someone adds lines at the end of this file.

It might also be annoying because a lot of tools out there (e.g.: vim) automatically add a trailing newline, so I can imagine people having issues with that (e.g.: since `neovim` automatically adds this, I have to use another tool to actually delete that).

jack1142: Now that #2330 is merged, this should no longer be an issue (or at least, it won't be in the future :)). Looks like there are some conflicts that need to be resolved now due to EOF normalization though.

setup.py
ichard26: My opinion here is that while I also disagree semantic versioning does a good job at avoiding unnecessary breakage (https://hynek.me/articles/semver-will-not-save-you/) I don't think adding an upper bound for tomli is out of the question. While it would be the first upper-bound Black specified, I think giving tomli some room to change is fine (we hope that it doesn't but who knows what will happen) given the points hukkin has layed out. Another point is that since tomli is so focused, it's not *that* bad to be unable to upgrade. The only major exception is that brand new TOML syntax will be blocked once their support has been implemented in 2+, but assuming a version major bump does mean a broken API (of which Black uses 2/3 of) that's kinda moot :)

hukkin: > https://hynek.me/articles/semver-will-not-save-you/

The individual teachings in this write up make sense, but the main point seems to be "if you do things poorly, then the result is bad", :smile: from which I'm not sure how much there is to learn.

**cooperlees:**
> (pinning) causes unneeded stagnation

The way I see it, some stagnation happens regardless (unless nobody cares or pins in which case there is no stagnation but a lot of breakage). If application maintainers don't make the effort of testing and bounding to major versions then libraries stagnate in the fear of breaking user facing apps. I think the game is more fair if everyone takes care of their own stagnation and doesn't inflict it on others.

**ichard26:**
> I think giving tomli some room to change is fine

Related to what I wrote above, I don't think applications should have the power to _give_ libraries freedom to change. That freedom (with semantic version constraints) should be the default.

**TL;DR**
Just me expressing my views. I don't think this is that important, but arguing is fun :smile:

ambv:
While I agree with Hynek's article, if the maintainer of a library we want to use is explicitly asking us for an upper bound then I think we should respect that. Thanks for your work on tomli, @hukkin.

setup.py
ichard26: > Related to what I wrote above, I don't think applications should have the power to give libraries freedom to change. That freedom (with semantic version constraints) should be the default.

Well okay, clearly that phrase of mine ended up being more of a hot take than I intended :slightly_smiling_face: The main point is that I don't think semantic versioning works as well as claimed. I do agree libraries should be free to change regardless of what versioning scheme they choose, and while I still don't really think semantic versioning is ideal, it is what tomli chose, and this is how their freedom is maintained so OK, you've convinced a little bit :smiley:

JelleZijlstra (approved):

cooperlees:
Damn, I can't re-approve this.

JelleZijlstra:
Now you can!

cooperlees (approved):
One small step for Black, one giant leap for Python kind.

ichard26:
I totally forgot about this PR after updating it, merging!

PR #2303, opened 2021-06-03
don't uvloop.install on import

### LINKED ISSUE #2302: Black as a library import should check for running event loop before installing
https://github.com/psf/black/blob/f2a3fee15c3bc0f4068b2b2fb64f255b417cb4ef/src/black/__init__.py#L57-L63

See comments I left here https://github.com/psf/black/commit/754eecf69eed0bdc75fe224c19a702a4f0676807#r51686621

# what is the issue?

We are using the black library in a system that already had an eventloop set. Importing black the library now clears the event loop settings we had. Impact to us is that we cannot use this latest version of black. 

# Suggestion

Black the library should not assume that it can create its own event loop -- instead it should check to see if a running event loop exists before calling `uvloop.install()`.  

Actually, I'd go further to ask why the library should be thinking about controlling an eventloop in the first place? Much like configuring loggers should not be set in libraries, event loops should not be set in libraries either?

JelleZijlstra:
Fixes #2302

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@
 
 - Correct max string length calculation when there are string operators (#2292)
 - Fixed option usage when using the `--code` flag (#2259)
-- Do not call `uvloop.install()` when *Black* is used as a library (#2303)
+- Do not call `uvloop.install()` when _Black_ is used as a library (#2303)
 
 ## 21.5b2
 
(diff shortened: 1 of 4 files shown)


skrawcz (approved):
thanks for the quick turn around. Looks good!

PR #2304, opened 2021-06-03
Improve AST safety parsing error message

### LINKED ISSUE #2178: Provide a more useful error when parsing fails during AST safety checks

felix-hilden:
Closes #2178, alternative to #2218: This PR continues on the work of Hasan Ramezani in the alternative PR to improve the error messages when parsing formatted AST. His original commit is till at the beginning of the branch, which I then modified (Thank you for the initial push!)

A number of non-obvious details and questions:
- As suggested by Richard in the linked issue, I removed the error messages (and a test) for Python 2 parsing
- Previously versions 3.6 and 3.7 were parsed twice, first by the builtin `ast` and then by `typed_ast`. On [Discord](https://discord.com/channels/267624335836053506/846434317021741086/850083982459535400) we determined this was an accident so I removed it.
- The parsing was refactored a bit for clearer flow. It's now two functions: one for parsing a single version, and the wrapping function for constructing a list of versions to parse and then doing it. Probably easier to view the source than the diff on that one.
- Mypy linting fails, even though none of the type signatures were changed. Should we ignore it or is there some trickery that could be done to make Mypy recognise the types?

Everything's in separate commits for an easier review. I'll leave this as a draft at least until the Mypy issue has been resolved.


felix-hilden:
Oops. Didn't realise tests wouldn't run in a draft!

JelleZijlstra:
Not sure what's going on with the mypy error. Can look in more detail in a few hours.

felix-hilden:
Also, target versions include Pythons all the way from 3.3. Should they be here too?

src/black/parsing.py
```diff
@@ -106,28 +106,37 @@ def lib2to3_unparse(node: Node) -> str:
...
 
 
-def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
+def parse_single_version(
+    src: str, version: Tuple[int, int]
+) -> Union[ast.AST, ast3.AST, ast27.AST]:
     filename = "<unknown>"
+    if (3, 8) <= version < (4, 0):
+        return ast.parse(src, filename, feature_version=version)
+    elif version >= (3, 6):
```
felix-hilden, line 115: Also, target versions include Pythons all the way from 3.3. Should they be here (and in the version list below) too?

ichard26:
I said this on PyDis but I'll say it here for the record and also just in case someone misses it:

> @[felix] I just looked into it and the mypy error seems to be about how ast.parse only started to accept feature_version in 3.8 and higher

> ast.parse(source, filename='<unknown>', mode='exec', *, type_comments=False, feature_version=None)
> [...]
> Changed in version 3.8: Added type_comments, mode='func_type' and feature_version.

> And mypy isn't smart enough to know that ast will only be used on versions where feature_version is supported :slight_smile:
 
The easiest solution would be just adding an assert that makes sure we're running on 3.8 in that branch. Other than that, I can't think of a solution that wouldn't require redoing and restructuring the patch which is kinda defeating the point :/

felix-hilden:
Thanks for digging around! I see now why the error happens only now. So a couple of things we could do:

- `# type: ignore`, since we know that the calling code has made sure we are running on 3.8+ (tested and works)
- Assert or redundantly test in the if `sys.version_info` once more (tested and works)
- Revert the refactoring, if you're not too keen on having it (I know it's only a slight improvement)

I'll leave it up to you!

src/black/parsing.py
ambv: Looks like the previous behavior was mistakenly parsing 3.3-3.5 with ast27. And if it was somehow deliberate the logic fails to explain it so let's remove that special case and try to parse 3.3-3.5 like you're suggesting. Good find!

felix-hilden: I had to go and verify myself how the versions were parsed again. Here's a summary and new versions we'd like to have:
```
typed | sysver | parsed                              | new parsed
-----------------------------------------------------------------------
 yes  | >= 3.8 | ast: sysver down to 3.5, typed: 2.7 | ast: sysver down to 3.3, typed: 2.7
 yes  |  < 3.8 | typed: 3.7, 3.6, 2.7                | typed: 3.7 down to 3.3, 2.7
  no  | >= 3.8 | ast: sysver down to 3.5             | ast: sysver down to 3.3
  no  |  < 3.8 | error on import
```
So what I've understood is that the builtin ast is good for all versions except 2.7 if we're above 3.8, but below it we need `typed_ast`. I think I understood it differently when I first refactored, so I'll change it if I get a nod of approval that this time around it's solid 😄 Thanks for the patience!

felix-hilden: Also: would we like to tie the parsed version to target versions? I.e. when we add or remove something from the enumeration, we won't have to change this part of the code.

ichard26: IIRC we need typed_ast for <3.8 because the built-in `ast` module has limited feature_version support. In other words, using `ast` <3.8 would force Black to be able to only parse using the syntax of the runtime version. I'm not entirely sure though, I would strongly encourage either putting this information in a comment or the dev documentation because it's been like a year of me being here and I still don't 100% understand this area.

ichard26: Also, apparently `ast` and `typed_ast` have feature_version minimums of (3, 4): 

> The lowest supported version is (3, 4); the highest is sys.version_info[0:2].

https://docs.python.org/3/library/ast.html#ast.parse

> Currently only fully supported for Python 3.5+ with partial support for Python 3.4. So, feature_version=3 or less are all equivalent to feature_version=4.

https://github.com/python/typed_ast/blob/8eed936014f81a55a3e17310629c40c0203327c3/typed_ast/ast3.py#L50-L54

The good news is that it really doesn't matter since 3.3 and 3.4 seem to be practically the same syntax / grammar wise:
https://github.com/psf/black/blob/aa31a117b16ed849710cf13cea5c064839beb29e/src/black/mode.py#L47-L48

felix-hilden: That makes sense, thank you! I'll shoot changes later today and hopefully we can get this forward!

felix-hilden: Seems like the builting `ast` also accepts feature versions below 3.4 as long as they are above 3. Also, I didn't link the code to `TargetVersion`, because assuming the correct enumeration values for minors or parsing the names felt hacky 😅

felix-hilden:
I corrected the parsing logic now! I think this could be ready for another round of review (hopefully the last as well) 👌

src/black/parsing.py
```diff
@@ -106,28 +106,36 @@ def lib2to3_unparse(node: Node) -> str:
...
+        return ast.parse(src, filename, feature_version=version)
+    elif version >= (3,):
+        return ast3.parse(src, filename, feature_version=version[1])
+    elif version == (2, 7):
+        return ast27.parse(src)
+    raise AssertionError("INTERNAL ERROR: Tried parsing unsupported Python version!")
+
+
+def parse_ast(src: str) -> Union[ast.AST, ast3.AST, ast27.AST]:
+    # TODO: support Python 4+ ;)
```
ichard26, line 124: Technically a bigger concern is the fact 3.10 syntax is not supported but oh well ;)

ichard26 (approved):
I don't see any issue with this and it's a nice readability improvement while implementing a nice enhancement. Thank you!

src/black/parsing.py
```diff
@@ -106,28 +106,36 @@ def lib2to3_unparse(node: Node) -> str:
...
-            except SyntaxError:
-                continue
-    if ast27.__name__ == "ast":
-        raise SyntaxError(
-            "The requested source code has invalid Python 3 syntax.\n"
-            "If you are trying to format Python 2 files please reinstall Black"
-            " with the 'python2' extra: `python3 -m pip install black[python2]`."
-        )
-    return ast27.parse(src)
+    # typed_ast is needed because of feature version limitations in the builtin ast
```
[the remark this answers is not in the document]
felix-hilden: Yeah, I think I like this a bit more so let's stick with it unless someone really wants more detail. Once again we should definitely continue the discussion on #2165 or even make a dedicated issue diving into exactly why `typed_ast` is needed. But I think this will do for the current PR.

src/black/parsing.py
felix-hilden: Yeah 😅

JelleZijlstra (approved):

PR #2312, opened 2021-06-05
Account for += assignment when deciding whether to split string

### LINKED ISSUE #2294: String processing produces code exceeding length limit
**Describe the bug**

String processing produces code exceeding the length limit when it was already formatted properly before.

**To Reproduce**

For example:
1. Take this file:
```py
class X:
    async def foo(self):
        msg = ""
        for candidate in CANDIDATES:
            msg += (
                "**{candidate.object_type} {candidate.rev}**"
                " - {candidate.description}\n"
            )
```
2. Run _Black_ on it with these arguments `--experimental-string-processing`
3. See the reformatted code that exceeds line length limit:
```py
class X:
    async def foo(self):
        msg = ""
        for candidate in CANDIDATES:
            msg += "**{candidate.object_type} {candidate.rev}** - {candidate.description}\n"
```

**Expected behavior**

I expected the code to remain unchanged.

- Version: main (`cf75673` at the time of testing)
- OS and Python version: Windows 10/Python 3.8.9

Yes.

**Additional context**

I was unable to find another issue that would seem to be the same as this though it's possible I missed something. Just to be sure it's not about the same thing #2284 is, I also tried out the changes from #2292 (since they apparently fix some edge cases that weren't fixed by #2286.
Also, the original code actually contained f-string but to rule out that it has anything to do with #2293, I changed it to use regular strings which as you can see from the issue description didn't change the end result.

bbugyi200:
Fixes #2294

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### _Black_
 
+- Account for += assignment when deciding whether to split string (#2312)
 - Correct max string length calculation when there are string operators (#2292)
 - Fixed option usage when using the `--code` flag (#2259)
 - Do not call `uvloop.install()` when _Black_ is used as a library (#2303)
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -867,7 +867,7 @@ class BaseStringSplitter(StringTransformer):
                 # WMA4 a space, a comma, and a closing bracket [e.g. `), STRING`].
                 offset += 3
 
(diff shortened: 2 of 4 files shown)


JelleZijlstra (approved):
Should this also cover other augmented assignment operators? I don't think any appear with strings in practice much though, maybe `%=`.

Thanks for the PR!

PR #2315, opened 2021-06-08
Add STDIN test to primer

### LINKED ISSUE #2310: Add STDIN primer test
Test feeding a file in vía STDIN to primer

Expected usage:
```
cat awesome.py | black -
```

primer will do with with asyncio.subprocess

cooperlees:
- Check that out STDIN black support stays working
- Add asyncio.subprocess STDIN pip via communicate
- We just check we format python code from primer's `lib.py`

Fixes #2310

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 - Added `--required-version` option to require a specific version to be running (#2300)
 - Fix incorrect custom breakpoint indices when string group contains fake f-strings
   (#2311)
+- Add a test for code piped into black via STDIN (#2315)
 
 ## 21.5b2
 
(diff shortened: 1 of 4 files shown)


cooperlees:
I think we need to revise the `.md` only file push causing no unit tests running ...

JelleZijlstra:
Why do we need to test this with primer specifically? There are already various unit tests for stdin in test_black.py.

cooperlees:
This ensures we don’t break the ‘-‘ cli and the whole pipeline from stdin to return value using this read from STDIN option. I've seen Unit tests miss things before and this was simple to add and fast to execute so I see no downside to this.

It also keeps our integrations testing testing as much as it can. Next I’d like to get some different CLI usages too.

Issue #2314, opened 2021-06-07
Performance regression with experimental string processing


JelleZijlstra:
```
% time black --diff ~/some/file.py
All done! ✨ 🍰 ✨
1 file would be left unchanged.
black --diff ~/some/file.py  17.78s user 0.18s system 99% cpu 18.067 total
% time black --experimental-string-processing --diff ~/some/file.py
< many changes snipped >
would reformat /Users/jelle/some/file.py
All done! ✨ 🍰 ✨
1 file would be reformatted.
black --experimental-string-processing --diff   389.93s user 3.25s system 99% cpu 6:35.03 total
```
I can't share the file since it's private code, but it's 14000 lines of a single list display.

JelleZijlstra:
Assigning to myself to do profiling to figure out the cause

JelleZijlstra:
Here's some profiling output on the latest release:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000 1161.996 1161.996 /home/jelle/ans/venv/lib/python3.6/site-packages/black/__init__.py:720(format_file_in_place)

        1    0.034    0.034 1161.996 1161.996 /home/jelle/ans/venv/lib/python3.6/site-packages/black/__init__.py:820(format_file_contents)

        2    0.095    0.047 1150.835  575.418 /home/jelle/ans/venv/lib/python3.6/site-packages/black/__init__.py:850(format_str)

1577774/28514    4.211    0.000 1092.089    0.038 /home/jelle/ans/venv/lib/python3.6/site-packages/black/linegen.py:296(transform_line)

105815/46    0.483    0.000 1091.390   23.726 /home/jelle/ans/venv/lib/python3.6/site-packages/black/linegen.py:947(run_transformer)

37028/123    0.998    0.000 1070.323    8.702 {method 'extend' of 'list' objects}

    88012    0.519    0.000  958.530    0.011 /home/jelle/ans/venv/lib/python3.6/site-packages/black/trans.py:126(__call__)

  2761738   48.296    0.000  942.670    0.000 /home/jelle/ans/venv/lib/python3.6/site-packages/black/lines.py:535(append_leaves)

 10635526   92.206    0.000  737.378    0.000 /home/jelle/ans/venv/lib/python3.6/site-packages/black/lines.py:48(append)

      244    0.142    0.001  624.820    2.561 /home/jelle/ans/venv/lib/python3.6/site-packages/black/trans.py:683(do_transform)

[...]

PR #2319, opened 2021-06-09
Support named escapes (`\N{...}`) in string processing

### LINKED ISSUE #1468: Black produces code with invalid syntax by splitting `\n` in half
**Describe the bug**:
Black produces code with invalid syntax by splitting `\n` in half when reformatting string.

**To Reproduce**:
1. Put this code in a file:
```py
message = (
    f"1. Go to Google Developers Console and log in with your Google account.\n"
    "(https://console.developers.google.com/)\n"
    "2. You should be prompted to create a new project (name does not matter).\n"
    "3. Click on Enable APIs and Services at the top.\n"
    "4. In the list of APIs choose or search for YouTube Data API v3 and "
    "click on it. Choose Enable.\n"
    "5. Click on Credentials on the left navigation bar.\n"
    "6. Click on Create Credential at the top.\n"
    '7. At the top click the link for "API key".\n'
    "8. No application restrictions are needed. Click Create at the bottom.\n"
    "9. You now have a key to add to `[p]set api youtube api_key <your_api_key_here>`"
)
```
2. Run _Black_ on it with no arguments.
3. See error:
```
error: cannot format test.py: INTERNAL ERROR: Black produced invalid code: invalid syntax (<unknown>, line 2). Please report a bug on https://github.com/psf/black/issues.  This invalid output might be helpful: C:\Users\Jakub\AppData\Local\Temp\blk_6n5icmuc.log
```
Contents of the log file:
```diff
  File "c:\users\jakub\onedrive\dokumenty\red-discordbot\.venv\lib\site-packages\black\__init__.py", line 6037, in assert_equivalent
    dst_ast = parse_ast(dst)
  File "c:\users\jakub\onedrive\dokumenty\red-discordbot\.venv\lib\site-packages\black\__init__.py", line 5948, in parse_ast
    return ast27.parse(src)
  File "c:\users\jakub\onedrive\dokumenty\red-discordbot\.venv\lib\site-packages\typed_ast\ast27.py", line 50, in parse
    return _ast27.parse(source, filename, mode)
message = (
    f"1. Go to Google Developers Console and log in with your Google account.\n"
    f"(https://console.developers.google.com/)\"
    f"n2. You should be prompted to create a new project (name does not matter)."
    f"\n3. Click on Enable APIs and Services at the top"
    f".\n4. In the list of APIs choose or search for YouTube Data API v3 "
    f"and click on it. Choose Enab"
    f"le.\n5. Click on Credentials on the left navigation "
    f"bar.\n6. Click on Create Credential at the"
    f' top.\n7. At the top click the link for "API'
    f' key".\n8. No application restrictions are needed. Click Create at the '
    f"bottom.\n9. You now have a key to add to `[p]set api youtube api_key"
    f" <your_api_key_here>`"
)
```

**Expected behavior**
_Black_ shouldn't produce invalid syntax.

**Environment:**
- Version: master (commit 3461338)
- OS and Python version: Windows/Python 3.8.3

**Does this bug also happen on master?**:
It *only* happens on master, it doesn't happen on current release as there's no string reformatting in it yet. Online formatter doesn't error, but it does produce this output which is obviously invalid syntax:
```py
message = (
    f"1. Go to Google Developers Console and log in with your Google account.\n"
    f"(https://console.developers.google.com/)\"
    f"n2. You should be prompted to create a new project (name does not matter)."
    f"\n3. Click on Enable APIs and Services at the top"
    f".\n4. In the list of APIs choose or search for YouTube Data API v3 "
    f"and click on it. Choose Enab"
    f"le.\n5. Click on Credentials on the left navigation "
    f"bar.\n6. Click on Create Credential at the"
    f' top.\n7. At the top click the link for "API'
    f' key".\n8. No application restrictions are needed. Click Create at the '
    f"bottom.\n9. You now have a key to add to `[p]set api youtube api_key"
    f" <your_api_key_here>`"
)
```

**Additional context**
None

I'm sorry for the amount of issues but I tried to run Black's master on Cog-Creators/Red-DiscordBot's code base which resulted in quite a few of errors and also some undesired (by me) changes which I wanted to at least report even if they're not gonna be changed.

jack1142:
Fixes #1468 (or at least the issue that's in the comment there since they're not as closely related as I initially thought)

Not an awfully big amount of tests added here but I'm not sure if there's a need for more.


src/black/trans.py
```diff
@@ -1243,6 +1243,41 @@ def more_splits_should_be_made() -> bool:
...
+            if not backslash_count:
+                if c == "\\":
+                    backslash_count += 1
+                continue
+            if c == "\\":
+                backslash_count += 1
+                continue
+            backslash_count = 0
+            if c != "N":
+                continue
```
felix-hilden, line 1266: If I'm not mistaken, with escaped backslashes this would function incorrectly: `\\N{something}` would result in `count = 2` and still finding a slice. So how about something like:

```python
previous_was_backslash = False
for idx, c in it:
    if c == "\\":
        if not previous_was_backslash:
            previous_was_backslash = True
        else:
            previous_was_backslash = False    
        continue

    if not previous_was_backslash or c != "N":
        continue

    # Do the slice
```
Other potential names: `unescaped`, `previous_was_unescaped`, `previous_backslash_was_unescaped` (lol maybe not)

tests/data/long_strings.py
```diff
@@ -207,6 +207,18 @@ def foo():
...
+)
+
+x = (
+    "........................................................................... \N{LAO KO LA}"
+)
+
+x = (
+    "............................................................................ \N{LAO KO LA}"
+)
+
```
felix-hilden, line 221: Maybe it would be appropriate to test the escaping too, that would help in testing the backslash algorithm.

felix-hilden (commented):
Hi, thanks for submitting! I'm not familiar with this part of the code base so take it with a grain of salt, but I left some thoughts below. I'd like for the more experienced maintainers to take a look too.

src/black/trans.py
```diff
@@ -1307,6 +1342,20 @@ def breaks_fstring_expression(i: Index) -> bool:
...
 
+        nameescape_slices = self._get_nameescape_slices(string)
+
+        def breaks_nameescape_expression(i: Index) -> bool:
+            """
+            Returns:
+                True iff returning @i would result in the splitting of an
+                \\N{...} expression (which is NOT allowed).
+            """
+            for (start, end) in nameescape_slices:
```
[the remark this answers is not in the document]
jack1142: I mostly used this because I got suggested by the very similar function for f-strings above, but this one's definitely simpler so thanks! 😄

src/black/trans.py
jack1142: `backslash_count` was supposed to be used for escapes but looks I completely forgot to actually put such handling in...
I'll look into fixing this and adding more tests to cover it later today.

felix-hilden: Yeah I think testing for an odd number of backslashes works just as well if you were going for that!

felix-hilden: Actually now I see that my version wouldn't reset if some other character was after a backslash, so just keep yours I think 😅

jack1142: Escapes should work now, I decided to go for a tri-bool.

src/black/trans.py
```diff
@@ -1243,6 +1243,45 @@ def more_splits_should_be_made() -> bool:
...
+        # None - looking for first backslash
+        # True - the previous backslash was unescaped
+        # False - the previous backslash was escaped
+        backslash_tracking_state: Optional[bool] = None
+        it = iter(enumerate(string))
+        for idx, c in it:
+            if backslash_tracking_state is None:
+                if c == "\\":
+                    backslash_tracking_state = True
+                continue
```
felix-hilden, line 1263: Is this redundant? Because `not None is True`, so this could be combined with below. I think. Then it would start flipping between `True` and `False`.

jack1142: It isn't because we want to do an early-continue *if* it's not a backslash.

felix-hilden: I get that you're signifying "no backslashes" with `None`, but `False` seems to do the same thing, since we don't care where the escaping happens. But at this point it's splitting hairs 😅 so feel free to ignore me.

jack1142: Although I guess I could change the order and simplify this a little bit:
```py
if c == "\\":
    backslash_tracking_state = not backslash_tracking_state
    continue
if backslash_tracking_state is None:
    continue
```

jack1142: Okay, I see what you mean now 😄 Pushed a commit that should simplify this a fair bit.

felix-hilden (approved):
In my opinion this looks much better now, thanks a ton for being so speedy to modify!

JelleZijlstra:
I wonder if this could be generalized into other string pieces that can't be split, so we don't need special-case logic for the different kinds. Instead, we could just generate a single list of unsplittable slices.

felix-hilden:
Yep it should definitely be done!

jack1142:
> I wonder if this could be generalized into other string pieces that can't be split, so we don't need special-case logic for the different kinds. Instead, we could just generate a single list of unsplittable slices.

I can look into this but before I do I would like to know if we care that some spans might overlap? Because I could either just make a list of slices from all functions that generate the slices (so `_get_nameescape_slices` and `fexpr_slices` at the current time) and just use that *or* I could go further - sort it and make a new list of non-overlapping spans.

JelleZijlstra:
Overlap seems fine with how we're using these slices. Actually it may be more efficient to generate a set of illegal indices, since the current code looks quadratic to me. Which reminds me I should run some profiling for #2314.

jack1142:
I went with the set idea as that indeed seems like a more performant way and isn't really hard to do either.

Sadly this causes the diff to be more complicated, if that's a problem I can split this into a separate PR that can be reviewed separately *after* this one is merged.

src/black/trans.py
```diff
@@ -1273,9 +1273,32 @@ def _get_nameescape_slices(self, string: str) -> List[Tuple[int, int]]:
...
+    def _get_illegal_split_indices(self, string: str) -> Set[Index]:
+        illegal_indices: Set[Index] = set()
+        iterators = [
+            self._iter_fexpr_slices(string),
+            self._iter_nameescape_slices(string),
+        ]
+        for it in iterators:
+            for begin, end in it:
+                for idx in range(begin, end + 1):
+                    illegal_indices.add(idx)
```
JelleZijlstra, line 1300:
```suggestion
                illegal_indices.update(range(begin, end + 1))
```

felix-hilden:
If we wanted, I bet we could construct a single function that loops the string through once, cycling through N escape and f-string modes and generating the index ranges. But if this works, then it could be enough for this PR!

ichard26:
Thank you so much for your contribution! This project is only possible by contributions like these :black_heart:.  You're awesome, @jack1142. Many thanks for pretty much beta testing the experimental string processing handling extremely early. I'm sure you're the reason why this feature will be a lot less buggy. Which is great since we want to push a release with it enabled by default soon.

PR #2323, opened 2021-06-09
Add coverage files to gitignore

jack1142:
I've been getting a bunch of these when running `tox -e py`
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .venv
 .coverage
+.coverage.*
 _build
 .DS_Store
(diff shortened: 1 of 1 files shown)


felix-hilden (approved):
Now that I run that exact command it happens to me as well, so seems like a no-brainer. Thanks!

---

Actually, the files are deleted afterwards. They are not deleted for you? Maybe they should be aggregated to the single `.coverage` file.

JelleZijlstra (approved):

PR #2330, opened 2021-06-13
Add EOF and trailing whitespace fixer to pre-commit config

jack1142:
Unix tools often rely upon the files ending with a newline character to show the output properly (e.g. when the file has no newline at the end, `cat file` will cause the prompt to show right after its output without a newline in between). This may help minimize the diffs and ensures consistency across all files.

I've also added auto-removal of trailing whitespace since I don't think that outside of test data (which is excluded from pre-commit) anyone would actually want to have trailing whitespace in their files. Let me know if that's not something that you want though.

I am doing this PR because I've seen that people were basically wasting their time on this problem during review [here](https://github.com/psf/black/pull/2301/files#r645797416) and this could prevent that in the future.


ichard26 (approved):
I'm all for less pointless discussion regarding trailing whitespace :D

Thank you again!

PR #2331, opened 2021-06-13
Update pre-commit config

sadikkuzu:
Update pre-commit config

via `pre-commit autoupdate`

```
Updating https://gitlab.com/pycqa/flake8
... updating 3.9.0 -> 3.9.2.
Updating https://github.com/pre-commit/mirrors-mypy
... updating v0.812 -> v0.902.
Updating https://github.com/pre-commit/mirrors-prettier
... updating v2.2.1 -> v2.3.1.
```


Pipfile
```diff
@@ -10,6 +10,9 @@ docutils = "==0.15"  # not a direct dependency, see https://github.com/pypa/pipe
 flake8 = "*"
 flake8-bugbear = "*"
 mypy = ">=0.812"
+types-dataclasses = ">=0.1.3"
```
JelleZijlstra, line 13: This should only be needed <3.7, can we express that here?

ichard26: Good call, will do that since IIRC it's possible.

ichard26:
Sorry for not updating until now. Your PR was mostly A-OK to land except for the fact that mypy 0.9XX introduces modular typeshed which means we need to pull in a few `types-*` packages. That is what I've been working on but I'm out of time today.

Thanks for the PR!

sadikkuzu:
Maybe this can help: https://twitter.com/pre_commit_ci/status/1404525673100705794

Pipfile
ichard26: For the record, this is actually not possible due to `python_version=3.6` in `mypy.ini`.

JelleZijlstra (approved):

sadikkuzu:
🚀

PR #2343, opened 2021-06-20
Fix bad formatting of error messages about EOF in multi-line statements

### LINKED ISSUE #2317: Bad formatting of error messages about EOF in multi-line statements
**Describe the bug**

"EOF in multi-line statement" error is shown to the user as the raw tuple it was returned as, rather than in a nicely formatted message.

**To Reproduce**

For example:
1. Take this (invalid) file:
```py
print(

```
2. Run _Black_ on it with no arguments
3. See a message with the EOF error being shown as a raw tuple representation:
```
error: cannot format test_black.py: ('EOF in multi-line statement', (2, 0))
Oh no! 💥 💔 💥
1 file failed to reformat.
```

**Expected behavior**

A bit more nicely formatted error message.

- Version: 
- OS and Python version: 

Yes.

**Additional context**

For comparison, here's how I'm assuming the error should look like:
```
error: cannot format test_black.py: Cannot parse: 1:7: print([)
Oh no! 💥 💔 💥
1 file failed to reformat.
```
which is what you get in output when you try to format:
```py
print([)
```

tanvimoharir:
Closes #2317


src/black/parsing.py
```diff
@@ -92,6 +93,16 @@ def lib2to3_parse(src_txt: str, target_versions: Iterable[TargetVersion] = ()) -
...
+
+        except TokenError as te:
+            lineno, column = te.args[1]
+            lines = src_txt.splitlines()
+            try:
+                faulty_line = lines[lineno - 1]
+            except IndexError:
+                faulty_line = "<line number missing in source>"
+            exc = InvalidInput(f"Cannot parse: {lineno}:{column}: {faulty_line}")
+
```
tanvimoharir, line 105: I have tried to keep the format consistent with ParseError above but still need to correct the values for lineno,column and faulty_line (which are currently wrong)

ichard26: I think we can do without the duplication of code by conditionalizing (made that up right now :p) the lineno. and col. information extraction code. So in the end it would look something like this (warning: untested):

```python
        try:
            result = drv.parse_string(src_txt, True)
            break

        except (ParseError, TokenError) as err:
            if isinstance(err, ParseError):
                lineno, column = err.context[1]
            else:
                lineno, column = err.args[1]

            lines = src_txt.splitlines()
            try:
                faulty_line = lines[lineno - 1]
            except IndexError:
                faulty_line = "<line number missing in source>"
            exc = InvalidInput(f"Cannot parse: {lineno}:{column}: {faulty_line}")
```

ichard26 (changes requested):
Hi there,

Thanks for the PR! It's great to see this getting closer to being fixed. I looked over your work and I have a few suggestions and requests to make, so let's get started:

It would be extra fantastic to catch and print out a prettier error message for IndentationError as well. The main difficulty would probably be getting the line no. and col. information out of the exception object since this exception is a built-in exception and has a different argument setup than TokenError (or ParseError) which are custom. Although I wouldn't mind ignoring IndentationError because the current error message isn't that bad actually:

```console
~ via Python v3.8.5 took 1m10s232ms 
123❯ cat test.py
def t():
    hello = "world"

  a = 123

~ via Python v3.8.5 
1❯ black test.py --check
error: cannot format test.py: unindent does not match any outer indentation level (<tokenize>, line 4)
Oh no! 💥 💔 💥
```

Here's the only place where IndentationError is raised in blib2to3: https://github.com/psf/black/blob/1bedc176d19f0e71c24149c696c5a2a71f21e941/src/blib2to3/pgen2/tokenize.py#L522

Second thing, it would be awesome if a test ensuring TokenError doesn't hit the top-level exception handler would be added, although I wouldn't block on this. 

Finally, you'll need a changelog entry in `CHANGES.md` that describes what your patch does. Don't forget to add your PR number to the end. While you're doing that, you can also add yourself to the `AUTHORS.md` file!

Oh, by the way, you can use special keywords in the PR description so GitHub will automatically close an issue upon merge. You can read more here: https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword. This makes sure any issue that should be closed are in fact closed (and I won't have to go proactively hunt for such issues!), keeping our very chaotic issue tracker ever so slightly less chaotic and scary :) **Although, if this PR doesn't end up covering IndentationError, closing the issue might not be the right call.**

Thanks for your work! We appreciate it!

src/black/parsing.py
tanvimoharir: Thanks I will try this. I had initially thought of making changes here https://github.com/psf/black/blob/main/src/blib2to3/pgen2/tokenize.py#L177

tanvimoharir:
> Hi there,
> 
> Thanks for the PR! It's great to see this getting closer to being fixed. I looked over your work and I have a few suggestions and requests to make, so let's get started:
> 
> It would be extra fantastic to catch and print out a prettier error message for IndentationError as well. The main difficulty would probably be getting the line no. and col. information out of the exception object since this exception is a built-in exception and has a different argument setup than TokenError (or ParseError) which are custom. Although I wouldn't mind deferring IndentationError because the current error message isn't that bad actually:
> 
> ```
> ~ via Python v3.8.5 took 1m10s232ms 
> 123❯ cat test.py
> def t():
>     hello = "world"
> 
>   a = 123
> 
> ~ via Python v3.8.5 
> 1❯ black test.py --check
> error: cannot format test.py: unindent does not match any outer indentation level (<tokenize>, line 4)
> Oh no! 💥 💔 💥
> 1 file would fail to reformat.
> ```
> 
> Here's the only place where IndentationError is raised in blib2to3:
> 
> https://github.com/psf/black/blob/1bedc176d19f0e71c24149c696c5a2a71f21e941/src/blib2to3/pgen2/tokenize.py#L522
> 
> Second thing, it would be awesome if a test ensuring TokenError doesn't hit the top-level exception handler would be added, although I wouldn't block on this.
> 
> Finally, you'll need a changelog entry in `CHANGES.md` that describes what your patch does. Don't forget to add your PR number to the end. While you're doing that, you can also add yourself to the `AUTHORS.md` file!
> 
> Oh, by the way, you can use special keywords in the PR description so GitHub will automatically close an issue upon merge. You can read more here: https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword. This makes sure any issue that should be closed are in fact closed (and I won't have to go proactively hunt for such issues!), keeping our very chaotic issue tracker ever so slightly less chaotic and scary :) **Although, if this PR doesn't end up covering IndentationError, closing the issue might not be the right call.**
> 
> Thanks for your work! We appreciate it!

Thanks @ichard26 I will check the changes which might be needed for IndentationError and I will also be adding tests.
Also to your point of updating CHANGES.md and AUTHORS.md,I thought I could do that once I'm done with all the other changes:)

src/black/parsing.py
```diff
@@ -84,14 +85,18 @@ def lib2to3_parse(src_txt: str, target_versions: Iterable[TargetVersion] = ()) -
             result = drv.parse_string(src_txt, True)
             break
 
-        except ParseError as pe:
-            lineno, column = pe.context[1]
+        except (TokenError, ParseError) as err:
+            if isinstance(err, ParseError):
+                lineno, column = err.context[1]
+            else:
+                lineno, column = err.args[1]
```
tanvimoharir, line 92: I'm trying to understand why i'm getting linenumber value as 2 for token error:
```
(black) bash-3.2$ python -m src.black --check --code "print("
error: cannot format <string>: Cannot parse: 2:0: <line number missing in source>
(black) bash-3.2$ python -m src.black --check --code "print([)" 
error: cannot format <string>: Cannot parse: 1:7: print([)
```

ichard26: Yeahhh that seems to be a bug with the blib2to3 library. We can just ignore that for this PR, it's better if we can just get this landed so while the error message won't be perfect, it should be better than `error: cannot format <string>: ('EOF in multi-line statement', (2, 0))` :)

Although a problem is that the source isn't printed at all which is unfortunate since it would be quite useful. Perhaps it would be good to also include that it was an "unexpected EOF" error in the message?

Good catch tho!

tanvimoharir:
@ichard26 
Changed the error message for TokenError,
```
black) bash-3.2$ python -m src.black --check --code "print("
error: cannot format <string>: Cannot parse: 2:0: Unexpected EOF
(black) bash-3.2$ python -m src.black --check --code "print([)"
error: cannot format <string>: Cannot parse: 1:7: print([)
```

src/black/parsing.py
```diff
@@ -92,6 +93,12 @@ def lib2to3_parse(src_txt: str, target_versions: Iterable[TargetVersion] = ()) -
             except IndexError:
                 faulty_line = "<line number missing in source>"
             exc = InvalidInput(f"Cannot parse: {lineno}:{column}: {faulty_line}")
+
+        except TokenError as te:
+            lineno, column = te.args[1]
+            lines = src_txt.splitlines()
+            exc = InvalidInput(f"Cannot parse: {lineno}:{column}: Unexpected EOF")
+
```
tanvimoharir, line 101: I included it as a separate block since that seemed more readable to me instead of adding multiple if (isintance,TokenError) in the above block where were handling ParseError

tanvimoharir:
@ichard26  Per your comment above about adding a test `to ensure TokenError doesn't hit the top-level exception handler `
Could you please explain this a little more?

ichard26:
> @ichard26 Per your comment above about adding a test to ensure TokenError doesn't hit the top-level exception handler

Sorry I took so long to get back to you on this. Basically I want a test which throws this input at Black and asserts that `TokenError` isn't the error thrown. Now depending on what level you do this at, running black at the command line (via BlackRunner) or calling black.format_str directly, how to check for that changes. At the command line level, checking the error output is your only option, but with black.format_str, using something fancier like pytest.raises(black.parsing.InvalidInput). There's some good examples of both options in `tests/test_black.py`. If you need more help or have questions, lemme know!

src/black/parsing.py
```diff
@@ -92,6 +93,12 @@ def lib2to3_parse(src_txt: str, target_versions: Iterable[TargetVersion] = ()) -
             except IndexError:
                 faulty_line = "<line number missing in source>"
             exc = InvalidInput(f"Cannot parse: {lineno}:{column}: {faulty_line}")
+
+        except TokenError as te:
+            lineno, column = te.args[1]
+            lines = src_txt.splitlines()
+            exc = InvalidInput(f"Cannot parse: {lineno}:{column}: Unexpected EOF")
```
ichard26, line 100: The `lines = src_txt.splitlines()` line is unnecessary now.

src/black/parsing.py
ichard26: Fair enough, I agree the duplication isn't a problem here.

src/black/parsing.py
tanvimoharir: Removed

tanvimoharir:
> > @ichard26 Per your comment above about adding a test to ensure TokenError doesn't hit the top-level exception handler
> 
> Sorry I took so long to get back to you on this. Basically I want a test which throws this input at Black and asserts that `TokenError` isn't the error thrown. Now depending on what level you do this at, running black at the command line (via BlackRunner) or calling black.format_str directly, how to check for that changes. At the command line level, checking the error output is your only option, but with black.format_str, using something fancier like pytest.raises(black.parsing.InvalidInput). There's some good examples of both options in `tests/test_black.py`. If you need more help or have questions, lemme know!

Checking tests/test_black.py.

tanvimoharir:
@ichard26  Added two tests at 
https://github.com/psf/black/blob/813c8ddded61e668efda2dd32a41909f3ea0ed83/tests/test_black.py#L2201-L2219

Please let me know if these are okay or if you were expecting something different.
Thanks!

ichard26 (approved):
@tanvimoharir hey sorry for such the long wait. The tests you added were good but the second one wasn't necessary as we already have tests that cover that area. I made the first (still relevant) test a bit more of an unit test so changes to --code don't affect it down the line. 

Finally I tweaked the error message opting to reuse the message already stored in the TokenError, and fixed the merge conflict. Thank you so much and my apologies once again!

PR #2344, opened 2021-06-22
Get `click` types from main repo

hukkin:
Click types have been moved to click repo itself. See https://github.com/pallets/click/pull/1856

I've had some issues with typeshed types being outdated in another project so might be good to avoid that here.


.pre-commit-config.yaml
```diff
@@ -27,7 +27,7 @@ repos:
           - types-dataclasses >= 0.1.3
           - types-toml >= 0.1.1
           - types-typed-ast >= 1.4.1
-          - types-click >= 7.1.2
+          - click >= 7.1.2
```
felix-hilden, line 30: The linked Click issue was milestoned to version 8. Should this dependency be upgraded then?

felix-hilden (commented):
Kiitos! I think @JelleZijlstra and others might know more about this, but I left a small comment below.

---

Also, the lint check fails with mypy errors, so perhaps the type hints are slightly different. If this is accepted, should that be addressed in our own code?

hukkin:
> Kiitos!

Olepas hyvä!

> The linked Click issue was milestoned to version 8. Should this dependency be upgraded then?

You're right, the annotations are not present in pre v8 click distributions. I the requirement to v8 for type annotations.

> Also, the lint check fails with mypy errors

Fixed now. This was just a drive-by PR made in web browser. I had fingers crossed everything will work with no other changes added, haha.

JelleZijlstra (approved):

JelleZijlstra:
We still require `click>=7.1.2` in setup.py. I'm hesitant to change it there since that will require all users to also upgrade click in their environments.

ichard26 (approved):
ship it, thanks for the pr!

PR #2346, opened 2021-06-22
Accept empty stdin (close #2337)

### LINKED ISSUE #2337: Zero length stdin input causes black to fail (regression)
**Describe the bug**

Using latest version 21.6b0 piping an empty file into black as below causes "string index out of range".  This is a regression from 21.5b2.  

NOTE: referencing an empty file works as expected this is only when piping into - as example below.
i.e. this works:
```
touch __init__.py
python -m black __init__.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

**To Reproduce**

```
touch __init__.py
cat __init__.py | python -m black -
error: cannot format -: string index out of range
Oh no! 💥 💔 💥
1 file failed to reformat.
```

**Expected behavior**

Using stdin with empty file works as previous version without error.

- Version: 
- OS and Python version: 

Yes

**Additional context**

A bit of a tails case but we're using this approach within a CI script.

simaki:
fix #2337: Now Black does not fail for empty stdin file 👍 

- Accept empty stdin (#2337)

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -6,6 +6,7 @@ from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
 from dataclasses import replace
 import inspect
+import io
 from io import BytesIO
 import os
 from pathlib import Path
@@ -1680,16 +1681,18 @@ class BlackTestCase(BlackBaseTestCase):
             report.done.assert_called_with(expected, black.Changed.YES)
 
     def test_reformat_one_with_stdin_empty(self) -> None:
-        with patch("sys.stdout", open(os.devnull, "w")):
-            # Patch `sys.stdout` to null device lest
-            # `format_stdin_to_stdout` detaches stdout at the end and causes
-            # "ValueError: I/O operation on closed file" in subsequent tests
-            black.format_stdin_to_stdout(
-                fast=True,
-                content="",
-                write_back=black.WriteBack.YES,
-                mode=DEFAULT_MODE,
-            )
+        output = io.StringIO()
+        with patch("io.TextIOWrapper", lambda *args, **kwargs: output):
+            try:
+                black.format_stdin_to_stdout(
+                    fast=True,
+                    content="",
+                    write_back=black.WriteBack.YES,
+                    mode=DEFAULT_MODE,
+                )
+                assert output.getvalue() == ""
+            except io.UnsupportedOperation:
+                pass  # StringIO does not support detach
 
     def test_gitignore_exclude(self) -> None:
         path = THIS_DIR / "data" / "include_exclude_tests"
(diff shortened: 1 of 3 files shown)


tests/test_black.py
```diff
@@ -1679,6 +1679,19 @@ def test_reformat_one_with_stdin_and_existing_path(self) -> None:
             # __BLACK_STDIN_FILENAME__ should have been stripped
             report.done.assert_called_with(expected, black.Changed.YES)
 
+    def test_reformat_one_with_stdin_empty(self) -> None:
+        """Check that black doesn't fail for empty stdin (issue/#2337)"""
```
felix-hilden, line 1683: The docstring may be a bit redundant, because the "doesn't fail" could be communicated in the test name, which also is displayed when the tests are run. But take it or leave it.

felix-hilden (approved):
Thanks a bunch! The change itself looks good, and the command line result is completely fine without the newline. I left a small nitpick below, but feel free to ignore it!

felix-hilden:
Could you also write a change log entry? I think this is worth notifying users.

tests/test_black.py
simaki: Exactly. Thank you!

```suggestion
```

felix-hilden:
Thanks for the fast turnaround, LGTM!

tests/test_black.py
```diff
@@ -1679,6 +1679,18 @@ def test_reformat_one_with_stdin_and_existing_path(self) -> None:
             # __BLACK_STDIN_FILENAME__ should have been stripped
             report.done.assert_called_with(expected, black.Changed.YES)
 
+    def test_reformat_one_with_stdin_empty(self) -> None:
+        with patch("sys.stdout", open(os.devnull, "w")):
+            # Patch `sys.stdout` to null device lest
+            # `format_stdin_to_stdout` detaches stdout at the end and causes
+            # "ValueError: I/O operation on closed file" in subsequent tests
+            black.format_stdin_to_stdout(
```
JelleZijlstra, line 1687: Can you assert that it reformats to the empty string? Maybe by setting stdout to a StringIO object.

simaki: Sure! Here it is.

tests/test_black.py
```diff
@@ -1680,16 +1681,18 @@ def test_reformat_one_with_stdin_and_existing_path(self) -> None:
...
+        output = io.StringIO()
+        with patch("io.TextIOWrapper", lambda *args, **kwargs: output):
+            try:
+                black.format_stdin_to_stdout(
+                    fast=True,
+                    content="",
+                    write_back=black.WriteBack.YES,
+                    mode=DEFAULT_MODE,
+                )
+                assert output.getvalue() == ""
```
JelleZijlstra, line 1693: This line should be outside the try-except to ensure it's executed.

simaki: Right, thanks. It's fixed!

ichard26 (approved):
LGTM! Thanks a lot for the quick patch, it's well appreciated :black_heart:

JelleZijlstra (approved):

Issue #704, opened 2019-02-08
The CLI "SRC" argument is listed as optional in --help display, but black does nothing if no paths are supplied.


bskinn:
**Operating system:** Windows 10 (presume not platform specific)
**Python version:** 3.6.6
***Black* version:** 18.9b0
**Does also happen on master:** Yes

**My question/problem:**

The CLI help looks like this:

```
>black --help
Usage: black [OPTIONS] [SRC]...

  The uncompromising code formatter.

Options:
...
```

To my mind, this implies that the `[SRC]...` argument is fully optional.  However, this happens:

```
>black
No paths given. Nothing to do 😴
```

For sure, the extra two keystrokes of `>black .` are not very burdensome. That said, to my mind, the natural default for `SRC` would be `pwd`. 

Would this be smart? If so, doable? I can definitely see a footgun here, though, where someone runs `black` thinking they'll get a help screen and instead end up reformatting their entire directory tree.

Thanks for a nifty tool!

zsol:
I actually think the default should be reading from stdin, so the same as `black -`.

ambv:
Thanks for your feedback! [As I explained](https://github.com/python/black/pull/740#issuecomment-490095950) on #740, at this point I want to keep the current behavior. It shields us from running `black` unexpectedly on a potentially confusing directory if a workflow integration ends up with an empty string.

bskinn:
<nod>, I definitely understand the rationale for the current behavior. Lots of footgun potential otherwise.

This is nitpicky, but is there any way in `click` to change the help message to remove the `[]` around `SRC`, to read like the following?

```
>black --help
Usage: black [OPTIONS] SRC ...

  The uncompromising code formatter.

Options:
...
```

Having `Usage: black [OPTIONS] [SRC] ...` makes it look like a zero-argument invocation should be legal.

ichard26:
@bskinn great suggestion! I know this is so late, but I've submitted a PR that avoids the brackets :) Thanks for the suggestion!

PR #2356, opened 2021-07-03
Avoid src being marked as optional in help

ichard26:
https://github.com/psf/black/issues/704#issuecomment-490251352

I *hope* this is trivial enough where it doesn't need a changelog entry (mostly because I forgot about it and am too lazy to add one).


JelleZijlstra:
I think this is better for clarity even if it's not strictly accurate.

PR #2362, opened 2021-07-06
Don't include profiling/ to cut down sdist by ~2x

ichard26:
They seem to be used as test cases for a specific region of formatting
that was slow. Now performance testing is probably something end users
won't be needing to do, so this is an easy way of reducing the sdist
size sigificantly.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+prune profiling


cooperlees (commented):
Why do we want to do this? I've had experiences where I've wanted to use projects Docker files, run CI + linting locally before, so I feel we'd want most of these files ... Will go with that others think, but I like having all the config Giles. the `.github` and action etc. make sense to prune.

I guess we can say to do this (and I guess the way I usually do it) is you clone the repo. dist is the minimum to just "build" the python project ...

ichard26:
Looking back at our messenger logs, it was brought up that the sdist was significantly larger than the wheel and that we wanted to bring bring it down. We agreed on getting rid of `profiling` and then everything else was debatable :) I personally ended up going with a policy of keeping only what I'd assume a redistributor would need (tests, docs, source code, etc.). I'll remove all of the `exclude` rules since they don't really save many bytes and I guess they could be useful.

I'm also happy to just get rid of `profiling` and then call it a day since excluding it is responsible for like 92% of the space savings. I'm also down to just close this PR if it's just better to leave the sdist basically a bookmark of the git repo (this would've more useful when Black only has a sdist but our automated CD should prevent such cases).

felix-hilden (approved):
Yay! This seems pretty reasonable 👌 

Maybe also:
- `global-exclude *.py[cod] __pycache__ *.so`
- `prune docs/_build`

although I'm not sure if they are needed because we release with a workflow. Perhaps the first should be added just to be sure.

Are all other files really included by default? I was under the impression that each data file needed to be included explicitly. That might have something to do with the setup configuration though.

hukkin:
> Are all other files really included by default? I was under the impression that each data file needed to be included explicitly.

I also regularly forget the answer to these questions :smile: . [check-manifest](https://github.com/mgedmin/check-manifest) provides a pre-commit hook to address this, can recommend.

felix-hilden:
Cheers mate! That looks quite useful 👌

cooperlees (approved):
LGTM - Lets merge it!

PR #2363, opened 2021-07-06
Use setuptools.find_packages in setup

felix-hilden:
Hi! Just a quick PR changing our setup a bit for the better. If we were to introduce new modules or refactor the source further down into submodules, maintaining the list in setup may become a pain to deal with. No more.

`find_packages` only includes proper packages, so I had to add an init file to primer. Shouldn't affect anything though, since it's already used as a package.

Feel free to shoot this down if it's not what we want! I didn't feel like this was big enough to go through an issue first.

---

I also thought about changing the data file specs to use the newly introduced (#2362) MANIFEST.in and `include_package_data=True`, but excluding the license and readme of blib2to3 may be a bit complicated with too little gain, when the original spec is 3 lines and very readable.

### DIFF
--- a/tests/test_primer.py
+++ b/tests/test_primer.py
@@ -189,7 +189,7 @@ class PrimerLibTests(unittest.TestCase):
         with patch("black_primer.lib.git_checkout_or_rebase", return_false):
             with TemporaryDirectory() as td:
                 return_val = loop.run_until_complete(
-                    lib.process_queue(str(config_path), td, 2)
+                    lib.process_queue(str(config_path), Path(td), 2)
                 )
                 self.assertEqual(0, return_val)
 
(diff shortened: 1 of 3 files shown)


felix-hilden:
Hmm, the lint job is failing. It *is* primer things, but I'm not sure why it's not caught in main. Also running locally on this branch (and excluding test data folders for some reason to avoid duplicate modules) mypy is fine. I'm even on the same Python minor. I'll get back to this.

MarcoGorelli:
>  I'm not sure why it's not caught in main

I presume because, without the `__init__.py` file you've added, `mypy` wouldn't recognise `black_primer` as a package? https://mypy.readthedocs.io/en/latest/running_mypy.html#specifying-code-to-be-checked

tests/test_primer.py
```diff
@@ -210,7 +210,7 @@ def test_async_main(self) -> None:
             "no_diff": False,
         }
         with patch("black_primer.cli.lib.process_queue", return_zero):
-            return_val = loop.run_until_complete(cli.async_main(**args))
+            return_val = loop.run_until_complete(cli.async_main(**args))  # type: ignore
```
felix-hilden, line 213: Decided to ignore here, because of the dict's convenience

PR #2368, opened 2021-07-11
Don't add trailing comma in one argument func defs

### LINKED ISSUE #2355: No trailing comma for one argument function definitions
**Describe the style change**

Black should **not** insert a trailing comma for one argument function definitions. This would be consistent with the style for two and more arguments.

```python
# trailing comma is added after 'some_argument'
def onnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnne(
    some_argument,
):
    pass


# for two+ args -> no comma is added
def twwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwo(
    some_argument, another_argument
):
    pass
```

**Desired style**
No trailing comma should be added, unless the user adds it.

```python
def onnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnne(
    some_argument
):
    pass
```

**Additional context**
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines

The issue seems to be caused by this part of the code:
https://github.com/psf/black/blob/017aafea992ca1c6d7af45d3013af7ddb7fda12a/src/black/linegen.py#L571-L587

cdce8p:
Closes #2355

No added trailing comma in one argument function definitions
```py
def no_trailing_comma_one_argument_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx(
    some_argument
):
    pass
```

This is consistent with the existing style for two and more arguments
```py
def no_trailing_comma_multiple_arguments_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx(
    first_arg, second_arg
):
    pass
```

### DIFF
--- a/tests/data/function_trailing_comma.py
+++ b/tests/data/function_trailing_comma.py
@@ -21,6 +21,15 @@ def xxxxxxxxxxxxxxxxxxxxxxxxxxxx() -> Set[
 ]:
     json = {"k": {"k2": {"k3": [1,]}}}
 
+def no_trailing_comma_one_argument_short(arg): pass
+def no_trailing_comma_multiple_arguments_short(a, b, c): pass
+def no_trailing_comma_one_argument_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx(
+    some_argument
+): pass
+def no_trailing_comma_multiple_arguments_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx(
+    first_arg, second_arg
+): pass
+
 # output
 
 def f(
@@ -86,3 +95,23 @@ def xxxxxxxxxxxxxxxxxxxxxxxxxxxx() -> Set[
             }
         }
     }
+
+
+def no_trailing_comma_one_argument_short(arg):
+    pass
+
+
+def no_trailing_comma_multiple_arguments_short(a, b, c):
+    pass
+
+
+def no_trailing_comma_one_argument_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx(
+    some_argument
+):
+    pass
+
+
+def no_trailing_comma_multiple_arguments_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx(
+    first_arg, second_arg
+):
+    pass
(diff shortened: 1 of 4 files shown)


tests/data/function_trailing_comma.py
```diff
@@ -21,6 +21,15 @@ def xxxxxxxxxxxxxxxxxxxxxxxxxxxx() -> Set[
 ]:
     json = {"k": {"k2": {"k3": [1,]}}}
 
+def no_trailing_comma_one_argument_short(arg): pass
+def no_trailing_comma_multiple_arguments_short(a, b, c): pass
+def no_trailing_comma_one_argument_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx(
```
JelleZijlstra, line 26: Needs more test cases:
- One like this but currently on one line
- One like this but with a trailing comma already present

JelleZijlstra (commented):
For what it's worth I'm not convinced we should change anything here.

Thanks for your contribution though!

tests/data/function_trailing_comma.py
cdce8p: Done. Just for reference here's the link to the failed primer run (for `pyanalyze`):
https://github.com/cdce8p/black/runs/3039749494?check_suite_focus=true#step:5:40

As best as I can tell, the change only causes a reformat if black is run with `skip-magic-trailing-comma = true`.

JelleZijlstra: Thanks! I don't like this change from the pyanalyze run:
```
         def forward_ref_bad_attribute(
-            x: "collections.defalutdict",  # E: undefined_name
+            x: "collections.defalutdict"  # E: undefined_name
         ):
```
Now if I want to add an argument, I'll also have to insert a comma in the middle instead of just adding a new line

cdce8p: That does make sense. Surprisingly easy to add after reverting some of the previous changes. See the last commit.

This here is the latest primer output for `pyanalyze`:
https://github.com/cdce8p/black/runs/3041092246?check_suite_focus=true#step:5:69

cdce8p:
For reference, the argument why this change would be useful: https://github.com/psf/black/issues/2355#issuecomment-877841081

JelleZijlstra:
This has been open for a long time and there isn't agreement on this change. I am going to close this and keep the current behavior. Thanks for your contribution!

PR #2369, opened 2021-07-11
Second run of tox -e py results in a test error for test marked with no_python2

### LINKED ISSUE #2367: Second run of `tox -e py` results in a test error for test marked with `no_python2`
**Describe the bug**

`no_python2` test errors on `main` while following Black's docs for contributing.

**To Reproduce**

Per [docs](https://black.readthedocs.io/en/stable/contributing/the_basics.html):
```
pipenv install --dev
pipenv shell
pre-commit install
pre-commit run -a
tox -e py
```
The above works properly, however *another* run of `tox -e py` results in an error due to tox's usage of existing `py` environment causing `no_python2` mark to be fulfilled:
```
___________________________ BlackTestCase.test_python2_should_fail_without_optional_install ___________________________
[gw2] win32 -- Python 3.8.9 c:\users\jakub\desktop\black\.tox\py\scripts\python.exe

self = <tests.test_black.BlackTestCase testMethod=test_python2_should_fail_without_optional_install>

    @pytest.mark.no_python2
    def test_python2_should_fail_without_optional_install(self) -> None:
        if sys.version_info < (3, 8):
            self.skipTest(
                "Python 3.6 and 3.7 will install typed-ast to work and as such will be"
                " able to parse Python 2 syntax without explicitly specifying the"
                " python2 extra"
            )

        source = "x = 1234l"
        tmp_file = Path(black.dump_to_file(source))
        try:
            runner = BlackRunner()
            result = runner.invoke(black.main, [str(tmp_file)])
>           self.assertEqual(result.exit_code, 123)
E           AssertionError: 0 != 123

C:\Users\Jakub\Desktop\black\tests\test_black.py:468: AssertionError
```

**Expected behavior**

No test errors.

- Version: main
- OS and Python version: Windows/Python 3.9.5

Yes

**Additional context**

Originally posted in #2238 (https://github.com/psf/black/issues/2238#issuecomment-857569590)

MarcoGorelli:
closes #2367 

I ran
```
rm -rf .tox
tox -e py
tox -e py
```
and it worked fine

cc @jack1142

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,6 +4,7 @@ envlist = {,ci-}py{36,37,38,39},fuzz
 [testenv]
 setenv = PYTHONPATH = {toxinidir}/src
 skip_install = True
+recreate= True
 deps =
     -r{toxinidir}/test_requirements.txt
 ; parallelization is disabled on CI because pytest-dev/pytest-xdist#620 occurs too frequently
@@ -18,7 +19,6 @@ commands =
     pytest tests --run-optional python2 \
         !ci: --numprocesses auto \
         --cov --cov-append {posargs}
-    pip uninstall typed_ast -y
     coverage report
 
 [testenv:fuzz]
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ envlist = {,ci-}py{36,37,38,39},fuzz
 [testenv]
 setenv = PYTHONPATH = {toxinidir}/src
 skip_install = True
-recreate= True
+recreate = True
 deps =
     -r{toxinidir}/test_requirements.txt
 ; parallelization is disabled on CI because pytest-dev/pytest-xdist#620 occurs too frequently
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -4,6 +4,7 @@ envlist = {,ci-}py{36,37,38,39},fuzz
 [testenv]
 setenv = PYTHONPATH = {toxinidir}/src
 skip_install = True
+recreate= True
```
JelleZijlstra:
```suggestion
recreate = True
```

tox.ini
```diff
@@ -4,6 +4,7 @@ envlist = {,ci-}py{36,37,38,39},fuzz
 [testenv]
 setenv = PYTHONPATH = {toxinidir}/src
 skip_install = True
+recreate = True
```
JelleZijlstra, line 7: I think this could use a comment explaining why we do this and linking back to the issue.

MarcoGorelli: that's a good point, thanks!

ichard26 (approved):
Thanks mate!

PR #2374, opened 2021-07-16
Create Docker tag 'latest_release'

### LINKED ISSUE #2373: Docker release - tags used for created images
**Describe the bug**

When you are creating Docker image, you always tag it with `latest` tag and either:
- Git tag
or:
- `latest_non_release`

This is done in line # 41 of [docker.yml](https://github.com/psf/black/blob/main/.github/workflows/docker.yml#L41).
Earlier `GIT_TAG` is prepared in line # 32 of [docker.yml](https://github.com/psf/black/blob/main/.github/workflows/docker.yml#L32).

**To Reproduce**

Pull both images:
```
$ docker pull pyfound/black:latest
$ docker pull pyfound/black:latest_non_release
```

Check version of Black for both images:
```
$ docker run --rm -it pyfound/black:latest black --version
black, version 0.1.dev1+g91773b8
$ docker run --rm -it pyfound/black:latest_non_release black --version
black, version 0.1.dev1+g91773b8
```

Of course, there can be a point in time when `latest` is same as image tagged with Git tag, but every push to `main` branch will produce new Docker image.

**Expected behavior**

`latest` tag should correspond to latest released version of Black. So in my automation I can always pull Docker image related to the newest released version of Black.

N/A

N/A

**Additional context**

N/A

pszlazak:
Docker images created during release process will have extra tag 'latest_release'.

This closes #2373.

With this change image with following tags was created when **release was created**:

docker-release-images

Steps in workflow:

docker-release-steps

And image with following tags was created for **regular commit** (you can still see previously created tags `99.9b0` and `latest_release` for image with digest ids `f9846dab1900` for `linux/amd64` and `7e76799a752c` for `linux/arm64`):

docker-regular-commit-images

Steps in workflow:

docker-regular-commit-steps

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -40,5 +40,14 @@ jobs:
           push: true
           tags: pyfound/black:latest,pyfound/black:${{ env.GIT_TAG }}
 
+      - name: Build and push latest_release tag
+        if: ${{ github.event_name == 'release' && github.event.action == 'created' }}
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: pyfound/black:latest_release
+
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}


.github/workflows/docker.yml
```diff
@@ -40,5 +40,14 @@ jobs:
           push: true
           tags: pyfound/black:latest,pyfound/black:${{ env.GIT_TAG }}
 
+      - name: Build and push latest_release tag
+        if: ${{ github.event_name == 'release' && github.event.action == 'created' }}
```
cooperlees, line 44: Awesome. Thanks for testing + finding these if conditions!

cooperlees (approved):
LGTM - For bonus points or I can do later - We should describe the tags and their intended usages somewhere in our documentation.

If you have time that would be awesome, otherwise I'll add one day.

cooperlees:
- Also, let's add a CHANGES.md entry about this since it might be useful for a lot of people / docker users.

PR #2357, opened 2021-07-03
Jupyter notebook support

MarcoGorelli:
closes #2345 

To summarise, based on what was discussed in that issue:
- due to not being able to parse automagics (e.g. `pip install black`) without a running IPython kernel, cells with syntax which is parseable by neither `ast.parse` nor `IPython` will be skipped
- cells with multiline magics will be skipped
- trailing semicolons will be preserved, as they are often put there intentionally in Jupyter Notebooks to suppress unnecessary output

----

some TODOs:
- [x] add black-primer test (e.g. pymc-examples)
- [x] support `--diff`
- [x] handle when explicitly trying to run on Jupyter notebook but extra dependencies aren't found

PR #2377, opened 2021-07-17
add context manager to temporarily change the cwd

JelleZijlstra:
@ichard26 noticed that the tests from #2357 fail locally due to order dependence. This change makes it so tests don't pollute the global state by making sure any chdir() is reversed.

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -44,6 +44,7 @@ from pathspec import PathSpec
 # Import other test classes
 from tests.util import (
     THIS_DIR,
+    change_directory,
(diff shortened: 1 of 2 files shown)


ichard26 (approved):
We still gotta fix that issue where the AST equivalence checks aren't always run in situations they should be in the test suite but this is an improvement :tada:

JelleZijlstra:
Ideally I feel like we shouldn't rely on the cwd at all in tests, just invoke subprocesses with cwd=. But that may be hard to get right in all cases.

PR #2379, opened 2021-07-18
Additional refactors based on the PR #2206

Anselmoo:
Additional refactors based on the PR #2206

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -833,6 +833,7 @@ def maybe_make_parens_invisible_in_atom(node: LN, parent: LN) -> bool:
 
     return True
 
+
 def _extracted_from_maybe_make_parens_invisible_in_atom_28(node, first, last, parent):
     middle = node.children[1]
     # make parentheses invisible
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -423,10 +423,9 @@ def left_hand_split(line: Line, _features: Collection[Feature] = ()) -> Iterator
         ):
             current_leaves = tail_leaves if body_leaves else head_leaves
         current_leaves.append(leaf)
-        if current_leaves is head_leaves:
-            if leaf.type in OPENING_BRACKETS:
-                matching_bracket = leaf
-                current_leaves = body_leaves
+        if current_leaves is head_leaves and leaf.type in OPENING_BRACKETS:
+            matching_bracket = leaf
+            current_leaves = body_leaves
     if not matching_bracket:
         raise CannotSplit("No brackets found")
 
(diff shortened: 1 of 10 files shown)


src/black/linegen.py
```diff
@@ -812,35 +813,40 @@ def maybe_make_parens_invisible_in_atom(node: LN, parent: LN) -> bool:
...
-        if is_atom_with_invisible_parens(middle):
-            # Strip the invisible parens from `middle` by replacing
-            # it with the child in-between the invisible parens
-            middle.replace(middle.children[1])
+    return True
 
-        return False
 
-    return True
+def _extracted_from_maybe_make_parens_invisible_in_atom_28(node, first, last, parent):
```
JelleZijlstra, line 837: What does 28 mean here?

src/blib2to3/pgen2/conv.py
```diff
@@ -67,12 +67,10 @@ def parse_graminit_h(self, filename):
...
         self.symbol2number = {}
         self.number2symbol = {}
-        lineno = 0
-        for line in f:
-            lineno += 1
+        for lineno, line in enumerate(f, start=1):
             mo = re.match(r"^#define\s+(\w+)\s+(\d+)$", line)
             if not mo and line.strip():
-                print("%s(%s): can't parse %s" % (filename, lineno, line.strip()))
+                print("%s(%s): can't parse %s" % (filename, lineno+1, line.strip()))
```
JelleZijlstra, line 73: You need either this or the start=1 but not both.

src/black/parsing.py
```diff
@@ -102,8 +102,7 @@ def lib2to3_parse(src_txt: str, target_versions: Iterable[TargetVersion] = ()) -
 
 def lib2to3_unparse(node: Node) -> str:
     """Given a lib2to3 node, return its string representation."""
-    code = str(node)
-    return code
+    return str(node)
```
MarcoGorelli, line 105: just a suggestion - perhaps make separate PRs, one for anything totally non-controversial (like this), and another for things like factoring out `_extracted_from_maybe_make_parens_invisible_in_atom_28`?

Anselmoo: Sorry for the in `atom_28`, this was a passed issue!
I didn't know that you prefer separated PRs.  Go it

src/black/linegen.py
Anselmoo: Sorry, the number `28` is just a filler for the help-functions, but actually not needed

src/blib2to3/pgen2/conv.py
Anselmoo: Yes, you are right!

src/black/parsing.py
MarcoGorelli: I'm not a maintainer so it's not my call, but in general small and focused PRs are easier to review

PR #2380, opened 2021-07-18
Update editors.md

Andrew-Chen-Wang:
Make PyCharm integration docs less confusing


HassanAbouelela (commented):
I'm not quite sure if this change is due to some perceived enhancement in behavior, or if it's a purely cosmetic one. Removing the quotes had a negative effect, at least in my case, as it meant the file path was treated as multiple separate arguments when there was a space in the path. Having the quotes worked as expected.

Andrew-Chen-Wang:
Interesting; yes it was just cosmetics. Closing since it looks like it doesn't fully work across all version!

PR #2384, opened 2021-07-19
Stop changing return type annotations to tuples

### LINKED ISSUE #2381: Internal Error when formatting bracketed return annotations
**Describe the bug**

When formatting a `*.py` file with bracketed returns annotations the error below is returned:
```
error: cannot format test.py: INTERNAL ERROR: Black produced code that is not equivalent to the source on pass 1.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /var/folders/9m/cps802s92k77tpy3z1w56hxr0000gn/T/blk_fu7q8w3k.log
```

**To Reproduce**

For example:
`test.py`
```
import typing

returning_a_deeply_nested_import_of_a_type = str

def some_method_with_a_really_long_name() -> (
    returning_a_deeply_nested_import_of_a_type
):
    pass

```

run `black -l 79` on the above file

**Expected behavior**

The file is formatted and the bracketed form is maintained.

- Version: main
- OS and Python version: macOS v11.4/Python 3.9.4

<!-- To answer this, you have two options:

Yes with the error:
```
INTERNAL ERROR: Black produced code that is not equivalent to the source on pass 1.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_aa17x_pj.log

--- src
+++ dst
@@ -47,17 +47,23 @@
         )  # /Pass
       decorator_list=
       name=
         'some_method_with_a_really_long_name',  # str
       returns=
-        Name(
+        Tuple(
           ctx=
             Load(
             )  # /Load
-          id=
-            'returning_a_deeply_nested_import_of_a_type',  # str
-        )  # /Name
+          elts=
+            Name(
+              ctx=
+                Load(
+                )  # /Load
+              id=
+                'returning_a_deeply_nested_import_of_a_type',  # str
+            )  # /Name
+        )  # /Tuple
       type_comment=
         None,  # NoneType
     )  # /FunctionDef
   type_ignores=
 )  # /Module
\ No newline at end of file
```

ichard26:
Fixes #2381.

---

This fixes a bug where a trailing comma would be added to a
parenthesized return type annotation changing its type to a tuple.
Here's one case where this bug shows up:

```python
def spam() -> (
    this_is_a_long_type_annotation_which_should_NOT_get_a_trailing_comma
):
    pass
```

The root problem was that the type annotation was treated as if it was
a collection (is_body=True to linegen::bracket_split_build_line) where
a trailing comma is usually fine. Now there's another check in the
aforementioned function to make sure the body it's operating on isn't
a parenthesized return type annotation before truly adding a trailing
comma.


JelleZijlstra:
Great find! I wonder if there are other syntactic contexts where this can happen, maybe a with statement on a really long variable name?

ichard26:
Ping! I'd appreciate some attention on this ^^

PR #2386, opened 2021-07-20
Replace for loop with list comprehension and set() with {}

monosans:


ichard26:
Thanks for the pull request, but I will be rejecting this as the readability improvement is basically zero IMO. The set() vs {...} form is 50-50 and honestly the for-loop -> list comprehension is worse (since the for-loop body is rather complicated). This doesn't outweigh the blame cost this has. And I don't believe these changes will have a negligible performance impact either.

PR #2398, opened 2021-07-24
Clarify contributing docs

JelleZijlstra:
"as configurable as gofmt" means little to people who haven't used gofmt.


ichard26 (approved):
Much better! Thank you! Hopefully this sets the stage of what Black is all about (even as we're trying to figure out the consistency bit right now) a bit better.

PR #2399, opened 2021-07-24
Fix issue templates + add docs template + PyDis

ichard26:
The template weren't applying the default labels ever since I renamed
the labels.

There has been enough issues about documentation opened recently so it's
probably worth a template for it.

And oh yeah PyDis exists as a communication method, let's advertise that :)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,8 +1,8 @@
 ---
 name: Bug report
-about: Create a report to help us improve
+about: Create a report to help us improve Black's quality
 title: ""
-labels: bug
+labels: "T: bug"
 assignees: ""
 ---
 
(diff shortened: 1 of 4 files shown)


JelleZijlstra (approved):

JelleZijlstra:
Feel free to merge this whenever you're happy with it.

ichard26:
Yea I assumed that was what implied, I'm in no rush so I'll leave it for a day for any extra feedback (although I doubt there will be any - we'll see).

felix-hilden (approved):
LGTM, disabling empty issues when we have good templates and redirect to a general discussion board would be better in my opinion. But you've probably seen more issues and have more experience than I 😄 and it's an easy change afterwards if we change our mind!

felix-hilden:
I'm sorry, my comment was too passive. I meant to ask: why do you value empty issues still? Have there been some valid issues that don't have a clear category, or would you not like to e.g. force redirect users to Discord right away?

ichard26:
I like empty issues because they give flexibility to me when I want to open an issue that either 1) doesn't quite fit any of the templates (and is too niche to get one), or 2) I am time limited but still need to get an issue posted so I don't forget. I think the fact it's at the bottom and in smaller text should discourage normal users from creating empty issues unless they really know what they want / are doing :)

ichard26:
It's been a couple of days and there's been no strong feedback so I'mma merge this. We can play with the new fancy issue forms later :)

Issue #2242, opened 2021-05-17
Support Python 3.10 pattern matching syntax


tirkarthi:
**Is your feature request related to a problem? Please describe.**

Python 3.10 has added support for pattern matching. It will be good to support it. https://www.python.org/dev/peps/pep-0622/

**Describe the solution you'd like** .

**Describe alternatives you've considered** 

**Reproducer**

```python
from dataclasses import dataclass

@dataclass
class Person:
    name: str

person = Person("Kate")

match person:
    case Person(name):
        print(f"Found {name}")
    case _:
        print("Not found")
```

```
 black /tmp/dataclass_match.py
error: cannot format /tmp/dataclass_match.py: Cannot parse: 10:6: match person:
Oh no! 💥 💔 💥
1 file failed to reformat.
```

**Additional context** Add any other context or screenshots about the feature request
here.

brandtbucher:
Note that the syntax has changed slightly since PEP 622 was superseded. [PEP 634](https://www.python.org/dev/peps/pep-0634/) (the accepted PEP) has the updated grammar.

I'm happy to help review or answer any questions on this once the work begins. I'm definitely looking forward to this support!

JelleZijlstra:
This will probably require #2318.

LordOfPolls:
May I ask if there's any news in regards to this?

graingert:
and here https://github.com/Instagram/LibCST/issues/285#issuecomment-894699168

LordOfPolls:
So, re the above. Does this mean black will be getting support for the new syntax shortly? Or are we to switch to using one of the forks?

JelleZijlstra:
What I'd like to see is that we land support based on @isidentical's patch, but somebody needs to step up to actually make a PR and get it landed. I might work on that if I have time, but I can't give any guarantees.

isidentical:
I would love to contribute @JelleZijlstra! If the lack of full support and other issues I have mentioned is considered and agreed upon I can convert my PR to a patch. And iterate over it for better support.

JelleZijlstra:
That would be great! I'd be OK with landing partial support first and iterating on the rest later.

isidentical:
Cool, will send a patch within this week!

JelleZijlstra:
I am going to close this because thanks to @isidentical's awesome work we shipped 21.11b1 with basic match support. There may be some cases that don't work, but we should open new issues for those as we run into them.

AlozoR:
[...]

PR #2406, opened 2021-07-28
Test on Python 3.10-dev

hugovk:
Python 3.10 is currently in its last beta with a release candidate due next week (2021-08-02).

* https://www.python.org/dev/peps/pep-0619/#release-schedule

This PR tests `3.10-dev` on GitHub Actions, currently Python 3.10.0-beta.4.

---

I put the versions in quotes in the yaml, because when 3.10 final is released and `-dev` is removed from `3.10-dev`, quoteless `3.10` would evaluate to 3.1!

* https://dev.to/hugovk/the-python-3-1-problem-85g

### DIFF
--- a/.github/workflows/primer.yml
+++ b/.github/workflows/primer.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10-dev"]
         os: [ubuntu-latest, windows-latest]
 
     steps:
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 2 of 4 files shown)


JelleZijlstra (approved):
Thanks!

.github/workflows/primer.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10-dev"]
```
cooperlees, line 27: The float rounding indicates what a fail 3.10 is :P

cooperlees (approved):

cooperlees:
Stoked we pass! Do we just avoid changing Pattern matching syntax today or we somehow understand it?

- I feel we need some code with Pattern matching in our test suite added - https://www.python.org/dev/peps/pep-0636/
- Also Union types - https://www.python.org/dev/peps/pep-0604/

https://docs.python.org/3.10/whatsnew/3.10.html

hugovk:
> Do we just avoid changing Pattern matching syntax today or we somehow understand it?

It's not yet supported, see https://github.com/psf/black/issues/2242.

PR #2408, opened 2021-07-30
Disable universal newlines when reading TOML

hukkin:
Reading a file with universal newlines is not TOML compatible.

See e.g. https://github.com/toml-lang/toml/issues/837 for context.


JelleZijlstra:
Thanks! Do you think this requires a changelog entry? I'm leaning yes since it does change user-visible behavior.

hukkin:
You're probably right. I added what I think best describes the change from user perspective

JelleZijlstra (approved):

ichard26 (commented):
Currently it appears this is under discussion and the changes contained here [may end up contested](https://github.com/pypa/pip/pull/10238#discussion_r680121388). Given we've already copied pypa/pip when it comes to our toml story, we are going to continue following them and wait for how they resolve this first.

JelleZijlstra:
Yes, I'll monitor this and merge this and the corresponding mypy PR if and when the pip and toml discussions are resolved.

wbolster:
given that this is gonna pop up more often b/c #2703, maybe it's time to merge this?

JelleZijlstra (approved):

PR #2411, opened 2021-08-06
Fix type dependencies of mypy invocation

hukkin:
Removes `types-toml` from pre-commit mypy invocation's addiditional dependencies. Adds `tomli` and `platformdirs`.


.pre-commit-config.yaml
```diff
@@ -25,9 +25,10 @@ repos:
         exclude: ^docs/conf.py
         additional_dependencies:
           - types-dataclasses >= 0.1.3
-          - types-toml >= 0.1.1
+          - tomli >= 0.2.6
```
ichard26, line 28: Do you want an upper bound here too?

Pipfile
```diff
@@ -11,7 +11,6 @@ flake8 = "*"
 flake8-bugbear = "*"
 mypy = ">=0.812"
 types-dataclasses = ">=0.1.3"
-types-toml = ">=0.1.1"
```
ichard26, line 14: Good catch!

.pre-commit-config.yaml
hukkin: :+1: changed it for consistency now that you mentioned. Though it's just for mypy so shouldn't matter too much :smile_cat:

ichard26 (approved):
Thank you!

PR #2412, opened 2021-08-06
Docker image usage description

pszlazak:
As [mentioned](https://github.com/psf/black/pull/2374#pullrequestreview-708433510) by @cooperlees I described Docker image tags and provided two basic examples of Docker image usage.

Please note that I'm not native English speaker, so be forgiving 😀

### DIFF
--- a/docs/usage_and_configuration/black_docker_image.md
+++ b/docs/usage_and_configuration/black_docker_image.md
@@ -5,13 +5,13 @@ https://hub.docker.com/r/pyfound/black
 
 The following tags are used for _Black_ images:
 
-- `latest` - tag used for either
-  [the latest commit of `main` branch](https://github.com/psf/black/commits/main) or
-  [the latest release](https://github.com/psf/black/releases/latest)
-- `latest_release` - tag used for
-  [the latest release](https://github.com/psf/black/releases/latest)
-- `latest_non_release` - tag used for
-  [the latest commit of `main` branch](https://github.com/psf/black/commits/main)
+- `latest` - tag used for the newest image of _Black_. It can point to same image as tag
+  `latest_release` or `latest_non_release`
+- `latest_release` - tag used for image created when new version of _Black_ is released.
+  It maps to [the latest release](https://github.com/psf/black/releases/latest) of
+  _Black_
+- `latest_non_release` - tag used for image created for every single
+  [commit of `main` branch](https://github.com/psf/black/commits/main)
 - release numbers, e.g. `21.5b2`, `21.6b0`, `21.7b0` etc.
 
 ## Usage
--- a/docs/usage_and_configuration/black_docker_image.md
+++ b/docs/usage_and_configuration/black_docker_image.md
@@ -3,16 +3,21 @@
 Official _Black_ Docker images are available on Docker Hub:
 https://hub.docker.com/r/pyfound/black
 
-The following tags are used for _Black_ images:
+From end-user perspective the following tags are used for _Black_ images:
 
-- `latest` - tag used for the newest image of _Black_. It can point to same image as tag
-  `latest_release` or `latest_non_release`
-- `latest_release` - tag used for image created when new version of _Black_ is released.
-  It maps to [the latest release](https://github.com/psf/black/releases/latest) of
-  _Black_
-- `latest_non_release` - tag used for image created for every single
-  [commit of `main` branch](https://github.com/psf/black/commits/main)
-- release numbers, e.g. `21.5b2`, `21.6b0`, `21.7b0` etc.
+- `latest` - tag used for the newest image of _Black_.\
+  ℹ Recommended for users who always want to use the latest version of _Black_.
+- `latest_release` - tag used for image created when new version of _Black_ is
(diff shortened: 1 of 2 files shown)


docs/usage_and_configuration/black_docker_image.md
```diff
@@ -0,0 +1,40 @@
+# Black Docker image
+
+Official _Black_ Docker images are available on Docker Hub:
+https://hub.docker.com/r/pyfound/black
+
+Following tags are used for _Black_ images:
```
JelleZijlstra, line 6:
```suggestion
The following tags are used for _Black_ images:
```

docs/usage_and_configuration/black_docker_image.md
```diff
@@ -0,0 +1,40 @@
...
+
+- `latest` - tag used for either
```
JelleZijlstra, line 8: I don't understand what this would mean to a user. When is it one versus the other?

docs/usage_and_configuration/black_docker_image.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Check code
+
+```console
+$ docker run --rm --volume $(pwd):/src --workdir /src pyfound/black:latest_release black --check .
+```
+
+_Remark_: besides [regular _Black_ exit codes](./the_basics.md) returned by `--check`
+option, [Docker exit codes](https://docs.docker.com/engine/reference/run/#exit-status)
+should be also considered.
```
JelleZijlstra, line 40:
```suggestion
should also be considered.
```

docs/usage_and_configuration/black_docker_image.md
pszlazak: It's quite hard to express the difference without going too much into technical details.

I tried to address it via bd03bf8. Human-readable version to be found here:  
https://github.com/pszlazak/black/blob/bd03bf86ff76f930b0f2bc14f7ee01891e267acf/docs/usage_and_configuration/black_docker_image.md

JelleZijlstra: Thanks! Can't we just say it's from the latest commit to main? Releases of black are always made from main.

pszlazak: You mean `latest`?
- `latest` is created when there is new commit or new tag (and tags are created during release):  
https://github.com/psf/black/blob/b1d060101626aa1c332f52e4bdf0ae5e4cc07990/.github/workflows/docker.yml#L41
- `latest_non_release` is created when commit does not map to tag:
https://github.com/psf/black/blob/b1d060101626aa1c332f52e4bdf0ae5e4cc07990/.github/workflows/docker.yml#L32
https://github.com/psf/black/blob/b1d060101626aa1c332f52e4bdf0ae5e4cc07990/.github/workflows/docker.yml#L33
- `latest_release` is created during realease:
    - condition:
https://github.com/psf/black/blob/b1d060101626aa1c332f52e4bdf0ae5e4cc07990/.github/workflows/docker.yml#L44
    - creation:
https://github.com/psf/black/blob/b1d060101626aa1c332f52e4bdf0ae5e4cc07990/.github/workflows/docker.yml#L50

pszlazak: I guess from end user perspective only two tags are important:
- `latest` - always the newest version of _Black_ (bleeding edge)
- `latest_release` - latest "stable" version

Third option - if anyone wants to use specific version of _Black_, image tags corresponding to Git tags can be used.

pszlazak: @JelleZijlstra - please take a look at e25cafe. Human-readable version to be found here:  
https://github.com/pszlazak/black/blob/e25cafe/docs/usage_and_configuration/black_docker_image.md

I hope I finally found sweet-spot between being to technical and end-user friendly.

docs/usage_and_configuration/black_docker_image.md
```diff
@@ -0,0 +1,45 @@
...
+enough to run _Black_ commands for chosen image denoted as `:tag`. In examples below
+`latest_release` tag is used. If `:tag` is omitted, `latest` tag will be used.
+
+More about _Black_ usage can be found in
+[Usage and Configuration: The basics](./the_basics.md).
+
+### Check Black version
+
+```console
+$ docker run --rm pyfound/black:latest_release black --version
```
cooperlees, line 34: We could state --rm is optional, but not needed I don't think.
```suggestion
$ docker run [--rm] pyfound/black:latest_release black --version
```

docs/usage_and_configuration/black_docker_image.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+### Check code
+
+```console
+$ docker run --rm --volume $(pwd):/src --workdir /src pyfound/black:latest_release black --check .
```
cooperlees, line 40: Cool - `--workdir` is super handy here.

cooperlees (approved):
All sounds good to me. Many thanks for this! Once @JelleZijlstra is happy lets merge.

PR #2414, opened 2021-08-07
Present a more user-friendly error if .gitignore is invalid

### LINKED ISSUE #2359: Present a more user-friendly error if  .gitgignore

Yes. When `.gitignore` is invalid, Black crashes with a confusing `IndexError` exception and it's not obvious that the `.gitignore` file is the issue. This is worsen by the fact, it's non-obvious Black uses `.gitignore` at all (maintainer note: it's been documented so it's fine, just describing how this is even worse of a problem than at first glance).

This is also a problem if say reading fails for some reason.

If an exception is raised within the `get_gitignore` function, capture it and present a more user-friendly error without an unnecessary traceback and with a quick mention of `.gitignore`.

- Continue operation even if `.gitignore` can't be parsed, just emit a warning and go on (fine, but I'm hesitant)
- Stop using `.gitignore` automatically (probably too much disruption potential)

**Additional context**

I thought of this when triaging https://github.com/psf/black/issues/2351.

nipunn1313:
Fixes #2359


tests/test_black.py
```diff
@@ -1727,6 +1727,16 @@ def test_nested_gitignore(self) -> None:
...
 
+    def test_invalid_gitignore(self) -> None:
+        path = THIS_DIR / "data" / "invalid_gitignore_tests"
+        empty_config = THIS_DIR / "data" / "invalid_gitignore_tests" / "pyproject.toml"
+        result = BlackRunner().invoke(
+            black.main, ["--verbose", "--config", str(empty_config), str(path)]
+        )
+        assert result.exit_code == 1
+        assert result.stderr_bytes is not None
+        assert b"Invalid git pattern: '!\\n'" in result.stderr_bytes
```
nipunn1313, line 1738: This is the error emitted by pathspec 0.9.0 thanks to https://github.com/cpburnz/python-path-specification/pull/50

JelleZijlstra: It seems fragile to test the exact error message produced by pathspec. Maybe we should check only for the "Could not parse" bit from black's own code?

JelleZijlstra (commented):
Thank you!

tests/test_black.py
nipunn1313: Makes sense to me! Especially if we'd be upgrading the dependency quietly.

tests/test_black.py
```diff
@@ -1727,6 +1727,18 @@ def test_nested_gitignore(self) -> None:
...
+        path = THIS_DIR / "data" / "invalid_gitignore_tests"
+        empty_config = THIS_DIR / "data" / "invalid_gitignore_tests" / "pyproject.toml"
+        result = BlackRunner().invoke(
+            black.main, ["--verbose", "--config", str(empty_config), str(path)]
+        )
+        assert result.exit_code == 1
+        assert result.stderr_bytes is not None
+
+        gitignore = path / ".gitignore"
+        assert f"Could not parse {gitignore}" in result.stderr_bytes.decode()
```
nipunn1313, line 1740: Here - it only checks for the presence of black's own error message

ichard26 (approved):
This seems fine to me, but I will note that code wasn't immediately clear and I was quite scared about "if gitignore is None: sys.exit(1)" bit since I thought it would also exit on cases where a gitignore file didn't exist. 

Nevertheless, thank you!

nipunn1313:
> This seems fine to me, but I will note that code wasn't immediately clear and I was quite scared about "if gitignore is None: sys.exit(1)" bit since I thought it would also exit on cases where a gitignore file didn't exist.
> 
> Nevertheless, thank you!

Good point. One idea could be to lift the `except GitWildMatchPatternError` to the callsite - instead of converting it to a `None` within the function. That might be more obvious at the callsite. Thoughts?

Another option might be to move the `exit(1)` downward into the helper method - but that seems a little too magical for my taste (would make it hard to unit test for example).

ichard26 (changes requested):
Actually this doesn't go far enough, it didn't occur to me before but we actually support nested gitignores these days, and uhm, this PR doesn't fully fix the ungraceful crash when a nested gitignore is invalid. 
```console
❯ tree -a test
test
├── a
│   ├── a.py
│   └── .gitignore
└── a.py

1 directory, 3 files

~/programming/oss/black on gitignore [$?] via v3.8.5 (bm-venv) 
❯ dropblackcache -y ; black test
no caching was found!
Could not parse test/a/.gitignore: Invalid git pattern: '!\n'
  File "/home/ichard26/programming/oss/black/bm-venv/bin/black", line 33, in <module>
    sys.exit(load_entry_point('black', 'console_scripts', 'black')())
  File "/home/ichard26/programming/oss/black/src/black/__init__.py", line 1281, in patched_main
    main()
  File "/home/ichard26/programming/oss/black/bm-venv/lib/python3.8/site-packages/click/core.py", line 1137, in __call__
  File "/home/ichard26/programming/oss/black/bm-venv/lib/python3.8/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/home/ichard26/programming/oss/black/bm-venv/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
  File "/home/ichard26/programming/oss/black/bm-venv/lib/python3.8/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/home/ichard26/programming/oss/black/bm-venv/lib/python3.8/site-packages/click/decorators.py", line 26, in new_func
  File "/home/ichard26/programming/oss/black/src/black/__init__.py", line 431, in main
    sources = get_sources(
  File "/home/ichard26/programming/oss/black/src/black/__init__.py", line 536, in get_sources
    sources.update(
  File "/home/ichard26/programming/oss/black/src/black/files.py", line 227, in gen_python_files
    gitignore + get_gitignore(child) if gitignore is not None else None,
TypeError: unsupported operand type(s) for +: 'PathSpec' and 'NoneType'
```
Sorry for not realizing this earlier!

nipunn1313:
ooh thanks for catching! Will cover and add test

nipunn1313:
Updated! Great catch. I went ahead and caught the exception one method higher so that it covers both callsites. Added a test for the nested case as well (it failed before, now it passes).

ichard26 (approved):
Thanks for the quick update!

JelleZijlstra (approved):

ichard26:
Thank you so much for your contribution! This project is only possible by contributions like these :black_heart:.  You're awesome, @nipunn1313. It's great to see the project get easier / more friendly to use :)

Congrats on your first PR here! If you have any feedback on experience contributing please let me know via #2238.

PR #2416, opened 2021-08-07
document jupyter hook

### LINKED ISSUE #420: pre-commit sample configuration should not suggest `rev: stable`
👋 hello, pre-commit maintainer here :)

The [sample configuration currently suggests `rev: stable`](https://github.com/ambv/black#version-control-integration) for `black`'s repository setup.  A mutable ref [poses some problems](https://pre-commit.com/#using-the-latest-version-for-a-repository).  Notably it gives the illusion that the latest version is being used but in reality it uses the version _at install time_.

This setup isn't suitable for those looking for a repeatable and reproducible linting experience. 

The suggestion that we use for our official repositories is to either [list the version explicitly in the docs](https://github.com/pre-commit/pre-commit-hooks#using-pre-commit-hooks-with-pre-commit) (though this adds tedium to the release cycle) or to use [some suitable non-value](https://github.com/pre-commit/mirrors-mypy#using-mypy-with-pre-commit) with instructions on how to substitute an appropriate value.

Yet another option would be to suggest `pre-commit autoupdate` after writing such a configuration as this will _usually_ convert the mutable rev to an immutable tag (or sha if `--bleeding-edge` is passed).

### LINKED ISSUE #2413: Follow-up items for the Jupyter PR
PR #2357 was recently merged :tada: For the sake of getting it merged somewhat quickly and to not tire everyone, we left a few TO-DO items for later. This issue attempts to list 'em so we don't forget :)

- Add documentation about the newly exported `black-jupyter` pre-commit hook 
- Look into whether we should add the Jupyter deps to `Pipfile` or just document the quirk (tox is unaffected)

I probably already forget a few tho, so feel free to edit / comment with more.

MarcoGorelli:
closes #420
~~addresses part of~~ closes #2413

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -21,6 +21,7 @@ flake8-bugbear = "*"
 mypy = ">=0.910"
 types-dataclasses = ">=0.1.3"
 types-typed-ast = ">=1.4.1"
+types-PyYAML = ">=5.4.1"
 
 # Documentation related requirements.
 Sphinx = ">=4.1.2"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "ac07cc9a5cb19ea72381baf4ba0db1689f475538d37e4be3119fc958a722b062"
+            "sha256": "ebf216584cfb2c962a1792d0682f3c08b44c7ae27305a03a54eacd6f42df27db"
         },
         "pipfile-spec": 6,
         "requires": {},
@@ -901,6 +901,14 @@
             "markers": "python_version >= '3.6'",
             "version": "==2.0.1"
         },
+        "matplotlib-inline": {
+            "hashes": [
+                "sha256:5cf1176f554abb4fa98cb362aa2b55c500147e4bdbb07e3fda359143e1da0811",
+                "sha256:f41d5ff73c9f5385775d5c0bc13b424535c8402fe70ea8210f93e11f3683993e"
+            ],
+            "markers": "python_version >= '3.5'",
+            "version": "==0.1.2"
+        },
         "mccabe": {
             "hashes": [
                 "sha256:ab8a6258860da4b6677da4bd2fe5dc2c659cff31b3ee4f7f5d64e79735b80d42",
@@ -1149,7 +1157,7 @@
                 "sha256:c203ec8783bf771a155b207279b9bccb8dea02d8f0c9e5f8ead507bc3246ecc1",
                 "sha256:ef9d7589ef3c200abe66653d3f1ab1033c3c419ae9b9bdb1240a85b024efc88b"
             ],
-            "markers": "python_version >= '2.6' and python_version not in '3.0, 3.1, 3.2'",
+            "markers": "python_version >= '2.6' and python_version not in '3.0, 3.1, 3.2, 3.3'",
             "version": "==2.4.7"
         },
         "pytest": {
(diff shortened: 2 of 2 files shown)


docs/integrations/source_version_control.md
```diff
@@ -7,8 +7,12 @@ Use [pre-commit](https://pre-commit.com/). Once you
 ```yaml
 repos:
-    rev: stable # Replace by any tag/version: https://github.com/psf/black/tags
+    rev: 21.7b0 # Replace by any tag/version: https://github.com/psf/black/tags
```
JelleZijlstra, line 10: We'll have to put something in the release process to update this (or better yet, make it automatic).

MarcoGorelli: sure, I'll add a pre-commit check for this

MarcoGorelli: I've added a check to ensure that the latest version from `CHANGES.md` matches the rev in the example from `source_Version_control.md`

demo:
```console
(venv) marco@marco-Predator-PH315-52:~/black-dev$ git diff
diff --git a/docs/integrations/source_version_control.md b/docs/integrations/source_version_control.md
index b63a7c4..64cc8b1 100644
--- a/docs/integrations/source_version_control.md
+++ b/docs/integrations/source_version_control.md
@@ -7,7 +7,7 @@ Use [pre-commit](https://pre-commit.com/). Once you
 ```yaml
 repos:
-    rev: 21.7b0 # Replace by any tag/version: https://github.com/psf/black/tags
+    rev: 21.6b0 # Replace by any tag/version: https://github.com/psf/black/tags
     hooks:
       - id: black
         language_version: python3 # Should be a command that runs python3.6+
(venv) marco@marco-Predator-PH315-52:~/black-dev$ git commit -a -m 'foo'
black................................................(no files to check)Skipped
Check pre-commit rev in example is correct...............................Failed
- hook id: check-pre-commit-rev-in-example
- exit code: 1

  File "/home/marco/miniconda3/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/marco/miniconda3/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/marco/black-dev/scripts/check_pre_commit_rev_in_example.py", line 32, in <module>
    main(changes, source_version_control)
  File "/home/marco/black-dev/scripts/check_pre_commit_rev_in_example.py", line 19, in main
    raise ValueError(
ValueError: Please set the rev in ``source_version_control.md`` to be the latest one

flake8...............................................(no files to check)Skipped
mypy.................................................(no files to check)Skipped
prettier.................................................................Passed
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
```

ichard26 (approved):
Awesome to see more automation come alive -- love it! Thank you!

MarcoGorelli:
Thank you @ichard26 for adding an extra commit to fix it up, the corrections look good to me!

ichard26:
Hmm, another idea I just thought of was to inject the value during the documentation build. This would further make it less likely we would miss this. OTOH it would be subject to bit-rot and could break (and we probably wouldn't notice until someone pointed it out to us). I don't really care either way since I pretty much always have pre-commit setup locally, but not everyone does.

MarcoGorelli:
`pre-commit` runs in CI so I think this should be the fastest way to be averted if the `rev` hasn't been updated, and if you prep releases in PRs then all should be fine

happy to change if you prefer though

ichard26:
I personally could go either way. Eventually this repo will probably get a custom local Sphinx extension where such a feature could be easily be added, but until then I don't think it's worth the complexity for this. I just brought that up as a potential suggestion someone else could've brought up for the sake of completeness.

PR #2419, opened 2021-08-10
Add jupyter deps to Pipfile.lock

MarcoGorelli:
xref #2413 

I added `jupyter` to `Pipfile` and then, in a new virtual environment, ran `pipenv install --dev`, and committed. Is this what's required for that issue?


ichard26 (approved):
Yup! Thank you!

P.S. the reason why I wanted the jupyter deps in here is to reduce friction when contributing. You shouldn't have to reinstall black with the jupyter extra just to test your ipynb changes ^^

PR #2425, opened 2021-08-12
Update language server links

TomFryers:
Development moved to https://github.com/python-lsp/python-lsp-server.


ichard26 (approved):
Thank you very much!

Issue #2407, opened 2021-07-28
Create/Find a repo that uses every python syntax + Add to Primer


cooperlees:
It would be nice to have a repo cpython developers add in their new syntax sugar every time the latest betas add the new support. That way we can have black run over that code super early and let us know what we fail on with more time before it becomes main stream.

I wonder if we could try format just the cpython test suite .py files for this.

JelleZijlstra:
I think the cpython test suite is our best bet for this. I think there's a few files in there that have syntax errors on purpose so we need to be a little careful, but we can just exclude those files.

cooperlees:
So I spent a few hours getting *most* of  cpython passing black. We have 16 files that fail with HEAD in main today. I also wanted to make sure it was not *WAY* to slow for CI. It seems reasonable, but lets see on a lot less CPU GitHub action container how long this takes.

## Test Checkout

- See how Long python takes with `--depth 1` on my cable (GitHub actions should be faster)
```
time git clone --depth 1 https://github.com/python/cpython.git
Cloning into 'cpython'...

real	0m12.558s
user	0m1.547s
sys	0m0.829s
```
- This is fine to add to CI ...

## Test Runtimes

The following times were captured via testing out on my 8 Core Apple M1 16gb of ram laptop.

Run with esp:
```
time /tmp/tb/bin/black --experimental-string-processing --check . 2>&1 | tee /tmp/black_cpython_esp
...
Oh no! 💥 💔 💥
1859 files would be reformatted, 148 files would be left unchanged, 16 files would fail to reformat.

real	4m8.563s
user	16m21.735s
sys	0m6.000s
```
Run without string processing:
```
time /tmp/tb/bin/black --check . 2>&1 | tee /tmp/black_cpython
Oh no! 💥 💔 💥
1861 files would be reformatted, 149 files would be left unchanged, 13 files would fail to reformat.

real	4m11.308s
user	11m49.160s
sys	0m4.762s
```
- See no real time difference here with and without `--experimental-string-processing`


Total Error Files: 16
- All files paste: https://pastebin.com/AHPQG848
- Without `--experimental-string-processing` we have only 13 files error
`--experimental-string-processing` causes error on these 3 files:
- Lib/test/test_base64.py
- Lib/test/test_tokenize.py
- Lib/test/test_xml_etree.py

PR coming with all failing files excluded for now. We will need to dig into why each file fails ...

PR #2429, opened 2021-08-15
Add cpython Lib/ repository config into primer config - Disabled

cooperlees:
- cpython tests is probably the best repo for black to test on as the stdlib's unittests should use all syntax
  - Limit to running in recent versions of the python runtime - e.g. today >= 3.9
    - This allows us to parse more syntax
- Exclude all failing files for now
  - Definately have bugs to explore there - Refer to #2407 for more details there
  - Some test files on purpose have syntax errors, so we will never be able to parse them
- Add new black command arguments logging in debug mode; very handy for seeing how CLI arguments are formatted

cython now succeeds ignoring 16 files:
```
Oh no! 💥 💔 💥
1859 files would be reformatted, 148 files would be left unchanged.
```

Testing
- Ran locally with and without string processing - Very little runtime difference BUT 3 more failed files
```
time /tmp/tb/bin/black --experimental-string-processing --check . 2>&1 | tee /tmp/black_cpython_esp
...
Oh no! 💥 💔 💥
1859 files would be reformatted, 148 files would be left unchanged, 16 files would fail to reformat.

real	4m8.563s
user	16m21.735s
sys	0m6.000s
```
- Add unittest for new covienence config file flattening that allows long arguments to be broke up into an array/list of strings

Addresses #2407

### DIFF
--- a/src/black_primer/lib.py
+++ b/src/black_primer/lib.py
@@ -133,10 +133,8 @@ def _flatten_cli_args(cli_args: List[Union[Sequence[str], str]]) -> List[str]:
             flat_args.append(arg)
             continue
 
-        new_args_str = ""
-        for arg_str in arg:
-            new_args_str += arg_str
-        flat_args.append(new_args_str)
+        args_as_str = "".join(arg)
+        flat_args.append(args_as_str)
 
     return flat_args
 
--- a/src/black_primer/lib.py
+++ b/src/black_primer/lib.py
@@ -12,7 +12,17 @@ from shutil import rmtree, which
 from subprocess import CalledProcessError
 from sys import version_info
 from tempfile import TemporaryDirectory
-from typing import Any, Callable, Dict, NamedTuple, Optional, Sequence, Tuple
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    NamedTuple,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+)
 from urllib.parse import urlparse
 
 import click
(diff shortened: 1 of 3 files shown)


cooperlees:
Not going to work. But since I did all the work diagnosing all the files we break on, I'd love to leave the config so interested people could use this to fix the bugs it exposes ... But can give up if people don't want the noise there.

cooperlees:
So, should I close this or do we want to merge for people to use in the future when people have cycles for potentially looking into the bugs this finds?

ichard26:
> So, should I close this or do we want to merge for people to use in the future when people have cycles for potentially looking into the bugs this finds?

I don't know how likely it is people will look into the discovered issues, but this can't possibly hurt and would be useful so I'm merging.

src/black_primer/lib.py
```diff
@@ -113,6 +124,23 @@ def analyze_results(project_count: int, results: Results) -> int:
...
+    flat_args = []
+    for arg in cli_args:
+        if isinstance(arg, str):
+            flat_args.append(arg)
+            continue
+
+        new_args_str = ""
+        for arg_str in arg:
+            new_args_str += arg_str
+        flat_args.append(new_args_str)
```
ichard26, line 139:
```suggestion
        args_as_str = "".join(arg)
        flat_args.append(args_as_str)
```

ichard26 (approved):
Hopefully my edit via the web interface doesn't break the tests. Will merge afterwards.

src/black_primer/lib.py
```diff
@@ -133,10 +133,8 @@ def _flatten_cli_args(cli_args: List[Union[Sequence[str], str]]) -> List[str]:
             flat_args.append(arg)
             continue
 
-        new_args_str = ""
-        for arg_str in arg:
-            new_args_str += arg_str
-        flat_args.append(new_args_str)
+        args_as_str = "".join(arg)
+        flat_args.append(args_as_str)
```
cooperlees, line 137: Nice - Much more pythonic

Issue #366, opened 2018-06-21
[Feature Request] Consider using compiled code for common functions


frsfnrrg:
This is similar to #349,  and #109. I am unfortunate enough to deal with 100+ KB python files, on
a system where black runs at only 60KB/s. For the same reason as in #349, black is not a viable save hook, except that here the problem is single-file throughput, not startup time. 

[Low throughput is common to the other python formatters that I've tried (yapf, autopep8). Profiling immediately reveals that, while much of the time is spent by other libraries (esp in `lib2to3_parse`), but at least 10-20% falls within black itself. There are specific solutions to the save hook performance problem (e.g., by detecting which parts of the file have changed and only reformatting those), but they would all make the program logic much more complicated.]

Would it be feasible to stop using a single-file approach for black and progressively move stable, basic types (like `Line` or `BracketTracker`) to a compiled library which is reachable by the main code through Python's C API? This *does* complicate installation, but would finally provide a path to speed comparable with e.g. `gofmt` and `clang-format`.

ambv:
Raw Python C API is not a silver bullet. We'd have to rewrite parts of the program in straight C and since most of it operates on blib2to3.pytree, we'd need that in C, too.

My not-so-secret plan is to cythonize blib2to3 and go from there. But that's something I'll get to after Black becomes more stable and most issues currently open are solved.

In the mean time, Black does safety checks that other formatters don't. If you haven't encountered any problems, use --fast which is over 2X faster than --safe.

boxed:
I ran black on a big file at work and was confused as to why it's so slow when the file needs many changes and so fast when there are no changes? Naively one would think you load a full syntax AST, run some transformers, dump it and if the new string is different from the old you write it. If this was the case then it should be roughly the same for many changes and no changes... So where is my naive thinking incorrect?

JelleZijlstra:
There is a cache. If the file hasn't changed since the last time we ran Black, we don't do any processing.

ambv:
It would be amazing to compile Black with @mypyc/mypyc. If I understand correctly, @msullivan is interested in making it happen. FTR, if this will require changes to the Black codebase, I'm open to that as long as they are not sweeping or overly disruptive.

msullivan:
I can start sending up PRs to support mypyc in the next week building on work done by my intern from last summer @SanjitKal

[...]

PR #2431, opened 2021-08-17
Implementing mypyc support pt. 2 

ichard26:
It's finally happening :tada:, the initial work was done several years ago in GH-1009 by @ msullivan is now being followed through by yours truly. If this work doesn't find its way to PyPI eventually, I'll be disappointed :slightly_smiling_face: 

### Performance

I wrote an entire report which can be read here: https://gist.github.com/ichard26/b996ccf410422b44fcd80fb158e05b0d, but the TL;DR is:

- Formatting _with_ safety checks: **1.93x faster**
- Formatting _without_ safety checks: **2.11x faster**
- Blib2to3 parsing: 1.81x faster
- Import (import black): **1.16x faster**

Windows numbers are to come later. Unfortunately I probably won't be able to get MacOS numbers (I own no MacOS running machine), but I'll see what I can do.

### Stability

I've been running the test suite during the wheel build and as of latest of this branch it's all green. Though it should be noted that the arm64 and the arm portion of the universial2 wheels are completely untested since we don't have access to any M1 CI resources :(

Locally I've ran [diff-shades](https://github.com/ichard26/black-mypyc-wheels/blob/main/diff_shades.py) (Jelle, it's basically mypy-primer but for black) against the CPython 3.8 manylinux wheel and no difference between interpreted and compiled Black was found (across the same project suite used by black-primer + blackbench :wink:). I plan to also test Windows with diff-shades in the following days. I'll also get Cooper to test Intel MacOS + M1 MacOS for me, but I need to come up with better instructions before that.

Anyhow, please don't expect this to be fully tested. I plan to ask the wider community to test the experimental compiled wheels (and push any additional fixes) after this PR is merged. 

### Review notes

This PR contains a lot of things including QOL improvements, optimizations, compatibility changes, hacks, etc. While the diff isn't that painful, following the commit history should give you some context on why certain changes were made.

It should be noted that I plan to handle the merge of this PR myself since this project involves a lot of moving parts and I'd like to maintain control for the time being. Thanks!

---

Related issue: GH-366  
Related repo: https://github.com/ichard26/black-mypyc-wheels

And if you're some passerby who really wants to try the experimental wheels, checkout the workflows here: https://github.com/ichard26/black-mypyc-wheels/actions.


cooperlees:
Nice work. FWIW - I'm totally down for `black-primer` to be refactored to do what it does today as well as what diff-shades does to cut down on the number of tools we all use everywhere. I guess diff-shades like behavior could be more easily added to our CI too if we incorporated it into the black repo.

mypy.ini
```diff
@@ -33,7 +36,23 @@ cache_dir=/dev/null
...
+# TOML people like bare CR as a newline sequence I suppose). Anyway,
+# some versions of tomli still specify str as the type so mypy complains
+# that the type: ignore is unnecessary. Gosh I wish I wasn't telling
+# this story.
+#
+# See also: https://github.com/psf/black/pull/2408
+#           https://github.com/pypa/pip/pull/10238 (because black is
+#             following whatever they do)
+#           https://github.com/toml-lang/toml/issues/837
+warn_unused_ignores=False
```
hukkin, line 58: Now this is an exciting PR, good job!

I sense some frustration towards this particular ignore :smile: 
If you want you can remove all of this by going to `black.files.py` and replacing
```python
with open(path_config, encoding="utf8") as f:
    pyproject_toml = tomli.load(f)  # type: ignore  # due to deprecated API usage
```
with
```python
with open(path_config, encoding="utf8") as f:
    pyproject_toml = tomli.loads(f.read())
```

The behavior 100% the same, no need to worry about that.

src/black/brackets.py
```diff
@@ -49,7 +49,9 @@
 DOT_PRIORITY: Final = 1
 
 
-class BracketMatchError(KeyError):
+# Ideally this would be a subclass of KeyError, but mypyc doesn't like that.
```
JelleZijlstra, line 52: I don't think it really needs to be a subclass of KeyError. I say we just remove this comment and leave it a direct subclass of Exception.

src/black/linegen.py
```diff
@@ -362,6 +365,11 @@ def rhs(line: Line, features: Collection[Feature]) -> Iterator[Line]:
                 line, line_length=mode.line_length, features=features
             )
 
+        # HACK: functions (like rhs) compiled by mypyc don't retain their __name__
```
JelleZijlstra, line 368: I guess it would be a bigger refactor but I'd like it better if we just stopped using `__name__`.

src/black/parsing.py
```diff
@@ -79,7 +91,7 @@ def lib2to3_parse(src_txt: str, target_versions: Iterable[TargetVersion] = ()) -
         src_txt += "\n"
 
     for grammar in get_grammars(set(target_versions)):
-        drv = driver.Driver(grammar, pytree.convert)
+        drv = driver.Driver(grammar)
```
JelleZijlstra, line 94: What is the motivation behind this change?

src/black/strings.py
```diff
@@ -142,6 +152,11 @@ def normalize_string_prefix(s: str, remove_u_prefix: bool = False) -> str:
     return f"{new_prefix}{match.group(2)}"
 
 
+@lru_cache(maxsize=256)
```
JelleZijlstra, line 155: Does this really help? `re.compile` already caches things internally I believe.

JelleZijlstra (approved):
Thanks for all your work on this! I'm really excited to see this PR.

src/black/parsing.py
JelleZijlstra: Never mind it's explained in blib2to3 code.

src/black/brackets.py
ichard26: Sounds good to me!

src/black/linegen.py
ichard26: Yeah, I'd prefer doing this later - tbh I'm a bit tired after all of work! I'll open an issue about this.

ichard26: Filed: https://github.com/psf/black/issues/2432.

src/black/strings.py
ichard26: Looking at the source and testing locally, I can confirm that both re and regex cache compiled patterns. Somehow I still observe a 8-9% speed up on that strings list literal file. Running this test script:
```python
import time
from functools import lru_cache

import re
import regex

_cached_regex_compile = lru_cache(regex.compile)

# This is *roughly* how many times regex.compile was called formatting strings-list.py
loops = 410 
pattern = r"\s*\t+\s*(\S)"

re.compile(pattern)
regex.compile(pattern)

t0 = time.perf_counter()
for i in range(1, loops + 1):
    re.compile(pattern)
elapsed = time.perf_counter() - t0
print(f"{loops} cached compiles with re: {elapsed * 1000:.1f}ms")

t0 = time.perf_counter()
for i in range(1, loops + 1):
    regex.compile(pattern)
elapsed = time.perf_counter() - t0
print(f"{loops} cached compiles with regex: {elapsed * 1000:.1f}ms")

t0 = time.perf_counter()
for i in range(1, loops + 1):
    _cached_regex_compile(pattern)
elapsed = time.perf_counter() - t0
print(f"{loops} cached compiles with regex + lru_cache: {elapsed * 1000:.1f}ms")
```
I get:
- 410 cached compiles with re: 0.4ms
- 410 cached compiles with regex: 1.6ms
- 410 cached compiles with regex + lru_cache: 0.1ms

So it seems like this is the source of the slight speedup, but oddly enough this only accounts for ~half of the 8-9% speedup I saw for fmt-strings-list. Maybe the fact this script only repeated compiles one regex is why the delta isn't 1:1. For real life cases this speedup is probably minimal so I wouldn't mind reverting this to be less confusing :)

JelleZijlstra: Interesting, thanks for the extra information. I looked at `re.compile` and it calls a helper function (`_compile`) and does an `isinstance` call before looking at the cache, and the cache key is a tuple `type(pattern), pattern, flags`. I suppose that adds up to more overhead than whatever lru_cache does internally.

src/black/parsing.py
ichard26: Ah yea the optimizations I made can be confusing! Profiling found that a lot of time parsing was spent in the Parser.push / Parser.shift methods so optimizing them was critical. Lookups at runtime are expensive and so are calls via the generic Python calling convention so I wanted to avoid them.

mypy.ini
ichard26: > The behavior will be 100% the same, no need to worry about that.

If this changes I'll be *very* annoyed, but I still trust you so alright.

JelleZijlstra (approved):

ichard26:
Whoops this isn't PyPy compatible ATM, will fix.

PR #2433, opened 2021-08-19
Throwing a more descriptive warning when reading a bad .gitignore

john-science:
This is in response to the ticket: https://github.com/psf/black/issues/2359

A particularly unhelpful error was thrown before, I feel this will be a lot more helpful to users trying to understand what went wrong.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 - Add support for formatting Jupyter Notebook files (#2357)
 - Move from `appdirs` dependency to `platformdirs` (#2375)
+- Fix opaque error when parsing invalid .gitignore (#2433)
 
 ### Integrations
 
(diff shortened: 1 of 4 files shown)


john-science:
I tried different `.gitignore` errors and found a few possible things that can go wrong, so I grabbed all three error types.

I also converted the error to `SyntaxError`. This is the only questionable choice I think I made in this tiny PR. I think it makes it more clear to the user; it _is_ a gitignore syntax problem, after all.

JelleZijlstra:
#2414 is already open trying to improve the same area.

ichard26:
Sorry @john-science but we've decided that #2414 is a better fit for resolving this issue mostly since it hides away all of the scary traceback and exception messaging and replaces it with a clear error message. Thanks for the contribution and effort tho!

john-science:
> Sorry @john-science but we've decided that #2414 is a better fit...

No worries!  Have a great weekend!

Issue #1688, opened 2020-09-08
Blackd Packaging


mlucool:
**Describe the bug**

The [blackd documenation states](https://black.readthedocs.io/en/stable/blackd.html):
> blackd is not packaged alongside Black by default because it has additional dependencies. You will need to execute pip install black[d] to install it.

But `blackd` is included in [packages](https://github.com/psf/black/blob/master/setup.py#L65) for black.

**To Reproduce** Steps to reproduce the behavior:

```bash
$ python3 -m venv venv; source venv/bin/activate
$ pip install black
$ which blackd
/path/to/venv/bin/blackd
```

**Expected behavior**

No binary named `blackd` is installed because its webserver dependencies are not installed.

**Environment (please complete the following information):**

- Version: master
- OS and Python version: Linux

**Does this bug also happen on master?**
Yes, see above

cooperlees:
Does anyone have suggestions here? I’m trying to read docs on how to do this and not finding many answers. I’ll keep dogging. We do have ‘[d]’ in the console_scripts entry ...

zsol:
I'm fairly sure this used to work as described in the docs. Has something changed in or around `pip`?

zsol:
From [the python packaging guide](https://python-packaging-user-guide.readthedocs.io/specifications/entry-points/?highlight=extras#file-format):

> Using extras for an entry point is no longer recommended. Consumers should support parsing them from existing distributions, but may then ignore them. New publishing tools need not support specifying extras. The functionality of handling extras was tied to setuptools’ model of managing ‘egg’ packages, but newer tools such as pip and virtualenv use a different model.

It's unclear to me what this different model is, but we should probably use that.

cooperlees:
I did see this too, but I can't work out what it is. I'll call in favors from friends to try work this out.

hugovk:
That was added to the guide in https://github.com/pypa/packaging.python.org/pull/390.

I've not read through it, but it also adds an "entry points specification", perhaps there's something useful in there? Could also ask someone from that PR?

PR #2440, opened 2021-08-23
Change sys.exit to Raise.

erykoff:
The fix for #1688 in #1761 breaks `help("modules")` introspection and also leads to unhappy results when inadvertently importing `blackd` from python.  E.g.:

```
$ python
Python 3.8.10 | packaged by conda-forge | (default, May 11 2021, 07:01:05) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> help('modules')

Please wait a moment while I gather a list of all available modules...

aiohttp dependency is not installed: No module named 'aiohttp'. Please re-install black with the '[d]' extra install to obtain aiohttp_cors: `pip install black[d]`
$
```
(Notice that python has been exited).

This PR changes the `sys.exit(-1)` in the `import` code to a proper `Raise`.

### DIFF
--- a/src/blackd/__init__.py
+++ b/src/blackd/__init__.py
@@ -1,6 +1,5 @@
 import asyncio
 import logging
-import sys
 from concurrent.futures import Executor, ProcessPoolExecutor
 from datetime import datetime
 from functools import partial
@@ -11,13 +10,11 @@ try:
     from aiohttp import web
     import aiohttp_cors
 except ImportError as ie:
-    print(
+    raise ImportError(
         f"aiohttp dependency is not installed: {ie}. "
         + "Please re-install black with the '[d]' extra install "
-        + "to obtain aiohttp_cors: `pip install black[d]`",
-        file=sys.stderr,
+        + "to obtain aiohttp_cors: `pip install black[d]`"
     )
-    sys.exit(-1)
 
 import black
 from black.concurrency import maybe_install_uvloop
--- a/src/blackd/__init__.py
+++ b/src/blackd/__init__.py
@@ -14,7 +14,7 @@ except ImportError as ie:
         f"aiohttp dependency is not installed: {ie}. "
         + "Please re-install black with the '[d]' extra install "
         + "to obtain aiohttp_cors: `pip install black[d]`"
-    )
+    ) from None
 
 import black
 from black.concurrency import maybe_install_uvloop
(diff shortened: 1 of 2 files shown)


src/blackd/__init__.py
```diff
@@ -11,13 +11,11 @@
     from aiohttp import web
     import aiohttp_cors
 except ImportError as ie:
-    print(
+    raise RuntimeError(
```
ktlim, line 14: Should this maybe continue to be an `ImportError` instead of changing it to a `RuntimeError`?

ichard26 (approved):
Ah yeah this makes sense. Thanks for the PR!

src/blackd/__init__.py
```diff
@@ -11,13 +10,11 @@
...
     import aiohttp_cors
 except ImportError as ie:
-    print(
+    raise ImportError(
         f"aiohttp dependency is not installed: {ie}. "
         + "Please re-install black with the '[d]' extra install "
-        + "to obtain aiohttp_cors: `pip install black[d]`",
-        file=sys.stderr,
+        + "to obtain aiohttp_cors: `pip install black[d]`"
     )
```
zsol, line 17: Now we could raise this `from ie`

zsol (approved):
Lgtm, just one minor comment

src/blackd/__init__.py
erykoff: Thanks for pointing this out.  I changed to `from None` as suggested by my colleague @timj as that leads to a more friendly output, e.g.:
```
>>> import blackd
  File "<stdin>", line 1, in <module>
  File "/u2/ki/erykoff/src/black_junk/black/src/blackd/__init__.py", line 13, in <module>
    raise ImportError(
ImportError: aiohttp dependency is not installed: No module named 'aiohttp'. Please re-install black with the '[d]' extra install to obtain aiohttp_cors: `pip install black[d]`
>>> 
```

ichard26: I agree with this. While raising from the original error would be more descriptive, it's unlikely to be useful yet is likely to confuse or scare the user. If some power user really wants to dig into the error, they can edit the source code.

ichard26:
Not waiting for all of primer since the chance of them uncovering an issue here is extremely low.

ichard26:
Hey thanks for your contribution! This project is only possible by contributions like these :black_heart:. You're awesome, @erykoff, never really thought about how a sys.exit would be problematic on import (not sure why, but eh who cares :p). If you have any suggestions about your contributing experience here, please let us via #2238, thank you once again!

Issue #2449, opened 2021-08-27
setuptools-scm is causing issues (especially since 20.8b1 doesn't have wheels) 


RafalSkolasinski:
Hi everyone,

We are aware of the recent installation issues especically with 20.8b1. This is due to a problem with the recently published versions of setuptools-scm, one of black's build time dependencies, https://github.com/pypa/setuptools_scm/issues/608. The good news is that all of the offending releases have been yanked on PyPI so this issue should be now gone. 

If you're wondering why 20.8b1 is failing while the others are fine, it's because 20.8b1's wheel was deleted due to a mistake done during the release artifact building process (the build environment wasn't cleaned up). This is important because building only needs to happen with sdists ... and it's the wheel building process that was borked by the setuptools-scm fires. We've moved our artifact building to GHA so these sort of release mistakes should be less common -- hopefully. Sorry for the disruption!  

~ @ichard26 (August 27th, 12:55 PM EST)

---

Currently 
```
conda create --name test python==3.7.10 (just to have fresh environment of python 3.7 for showcase (not a conda issue)
conda activate test
pip install black==20.8b1
```
fails with 
```
Collecting black==20.8b1
Collecting black==20.8b1
  Downloading black-20.8b1.tar.gz (1.1 MB)
     |████████████████████████████████| 1.1 MB 4.3 MB/s 
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /opt/conda/bin/python /opt/conda/lib/python3.7/site-packages/pip/_vendor/pep517/in_process/_in_process.py prepare_metadata_for_build_wheel /tmp/tmp6m1djknn
         cwd: /tmp/pip-install-y3c4u6lp/black_d9e87aa87d32486194850d6807c77340
    Complete output (30 lines):
    Traceback (most recent call last):
      File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 349, in <module>
        main()
      File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 331, in main
        json_out['return_val'] = hook(**hook_input['kwargs'])
      File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 151, in prepare_metadata_for_build_wheel
        return hook(metadata_directory, config_settings)
[...]

PR #2450, opened 2021-08-27
 Pin setuptools-scm dependency version

FabianNiehaus:
### Description

The `setuptools-scm` dependency in `setup.cfg` did not have a version specified, leading to the issues described in https://github.com/psf/black/issues/2449 after a faulty release of `setuptools-scm` was published.
To avoid this issue in the future, the last version before that faulty update is now pinned.

### Checklist - did you ...

- [x] Add / update tests if necessary? -> Not needed
- [x] Add new / update outdated documentation? -> Not needed

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
   trailing comma (#2384)
 - Parsing support has been added for unparenthesized walruses in set literals, set
   comprehensions, and indices (#2447).
+- Pin `setuptools-scm` dependency version (#2449)
 
 ### _Blackd_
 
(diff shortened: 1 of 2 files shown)


ichard26:
I honestly don't know why we specify setuptools-scm in both place, but given pyproject.toml is the newer standard (also isn't setuptools specific) it's probably best to pin there (although setup.cfg is by no means deprecated or anything like that). FYI looks like you got some trailing whitespace in setup.cfg. A comment explaining why we're pinning would be fantastic too but totally not neccesary.

ichard26:
Sorry @FabianNiehaus but I ended up incorporating your PR into a new one with some tweaks applied on top since I need to get a release out today. I hope the credit via the co-author line is good enough! Thanks for your contributing! Feedback welcome to #2238 :3

PR #2460, opened 2021-08-29
Exclude broken typing-extensions version + fix import

ichard26:
### Description

re. import, the ipynb code was assuming that typing-extensions would always be available, but that's not the case! There's an environment marker on the requirement meaning it won't get installed on 3.10 or higher. The test suite didn't catch this issue since aiohttp pulls in typing-extensions unconditionally.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@
 - Parsing support has been added for unparenthesized walruses in set literals, set
   comprehensions, and indices (#2447).
 - Pin `setuptools-scm` build-time dependency version (#2457)
+- Exclude typing-extensions version 3.10.0.1 due to it being broken on Python 3.10
+  (#2460)
 
 ### _Blackd_
 
--- a/setup.py
+++ b/setup.py
@@ -79,7 +79,10 @@ setup(
         "regex>=2020.1.8",
         "pathspec>=0.9.0, <1",
         "dataclasses>=0.6; python_version < '3.7'",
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -14,6 +14,7 @@
 - Parsing support has been added for unparenthesized walruses in set literals, set
   comprehensions, and indices (#2447).
 - Pin `setuptools-scm` build-time dependency version (#2457)
+- Exclude typing-extensions version 3.10.0.1 due to it being broken on 3.10 (#2460)
```
onerandomusername, line 17: Since these versions are nearly the same, would it make sense to specify `on CPython 3.10`?

setup.py
```diff
@@ -79,7 +79,10 @@ def get_long_description() -> str:
         "regex>=2020.1.8",
         "pathspec>=0.9.0, <1",
         "dataclasses>=0.6; python_version < '3.7'",
-        "typing_extensions>=3.10.0.0; python_version < '3.10'",
+        "typing_extensions>=3.10.0.0",
+        # 3.10.0.1 is broken on at least 3.10,
```
onerandomusername, line 83: Same here?

onerandomusername (changes requested):

onerandomusername (approved):
LGTM, lets hope that the Fuzz passes this time!  😄

JelleZijlstra (approved):

PR #2461, opened 2021-08-29
fix: run pypi upload from published draft releases

onerandomusername:
### Description
Fixes pypi_upload.yml not running when @ichard26 published a draft release to release 21.8b0.

Documentation is here, https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release, and since upload_binary ran, with this: https://github.com/psf/black/blob/a8b4665e7d6eb945c47820adb1a3f8b006adce0c/.github/workflows/upload_binary.yml#L3-L5
changing the trigger for pypi_upload.yml to same as the above should fix the problem.

### Checklist - did you ...

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,9 +18,9 @@
     Tests are required for bugfixes and new features. Documentation changes
     are necessary for formatting and most enhancement changes. -->
 
-- [] Add a CHANGELOG entry if necessary?
-- [] Add / update tests if necessary?
-- [] Add new / update outdated documentation?
+- [ ] Add a CHANGELOG entry if necessary?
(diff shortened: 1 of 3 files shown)


ichard26 (approved):
Part of me wants to consolidate the OS native binaries and PyPI build+ upload workflows into one file, but this fixes a possible pitfall with the release system. This is already a win. Thanks @onerandomusername! You're awesome!

onerandomusername:
Just noticed, the docker workflow also uses created, and not released.

onerandomusername:
I think its fixed now, also checked logs and it didn't run either.
https://github.com/psf/black/actions/workflows/docker.yml

@ichard26 as you had to push to pypi manually, there is also not a published docker image for this release. https://hub.docker.com/r/pyfound/black/tags?page=1&ordering=last_updated

PR #2375, opened 2021-07-16
Use platformdirs over appdirs

gaborbernat:
As the appdirs repositories, maintainership has dried up (https://github.com/ActiveState/appdirs/issues/79) we've forked the repository under platformdirs (https://github.com/platformdirs/platformdirs) name.

We've merged some existing open PRs and adding Android support (https://github.com/platformdirs/platformdirs/pull/15) is currently in the works. We encourage black to switch to it.

PR #2466, opened 2021-08-31
Add hidden import to PyInstaller build

### LINKED ISSUE #2464: Error when using black 21.8b0 black binary on linux
**Describe the bug**

When running the latest black binary on linux, the following error is thrown:
```
  File "black/__main__.py", line 1, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "black/__init__.py", line 38, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "black/nodes.py", line 28, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "black/cache.py", line 9, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 546, in exec_module
  File "platformdirs/__init__.py", line 31, in <module>
  File "platformdirs/__init__.py", line 27, in _set_platform_dir_class
  File "importlib/__init__.py", line 127, in import_module
ModuleNotFoundError: No module named 'platformdirs.unix'
[2725] Failed to execute script '__main__' due to unhandled exception!
```

This seems to an issue with PyInstaller missing a module during packaging.

**To Reproduce**

1. Run a linux container: `docker run --rm -it debian:stable-slim bash`
2. Download the black binary: `wget --tries=5 -q -O /usr/local/bin/black https://github.com/psf/black/releases/download/21.8b0/black_linux && chmod +x /usr/local/bin/black`
3. Run black: `black -v`

**Expected behavior**

The version of `black` should be printed.

- Version: 21.8b0
- OS and Python version: Linux

N/A?

**Additional context**

This is likely due to #2375.

This error doesn't occur when using the 21.7b0 binaries. `platformdirs.unix` may need to be added explicitly using PyInstaller's [hidden imports](https://pyinstaller.readthedocs.io/en/stable/usage.html#cmdoption-hidden-import) functionality.

jalaziz:
### Description

Add new platformdirs dependencies as hidden imports when creating
PyInstaller-based binaries.

platformdirs imports the module for each platform dynamically, which
PyInstaller is unable to correctly detect for packing. By adding the
modules as hidden imports, we are telling PyInstaller to include the
modules in the packaged binary.

This issue seems to have been introduce when switching to platformdirs
in #2375.

fixes #2464

### Checklist - did you ...

### DIFF
--- a/.github/workflows/upload_binary.yml
+++ b/.github/workflows/upload_binary.yml
@@ -40,8 +40,11 @@ jobs:
           python -m pip install pyinstaller
 
       - name: Build binary
-        run: |
-          python -m PyInstaller -F --name ${{ matrix.asset_name }} --add-data 'src/blib2to3${{ matrix.pathsep }}blib2to3' src/black/__main__.py
+        run: >
+          python -m PyInstaller -F --name ${{ matrix.asset_name }} --add-data
+          'src/blib2to3${{ matrix.pathsep }}blib2to3' --hidden-import platformdirs.unix
+          --hidden-import platformdirs.macos --hidden-import platformdirs.windows
+          src/black/__main__.py
 
       - name: Upload binary as release asset
         uses: actions/upload-release-asset@v1
--- a/.github/workflows/upload_binary.yml
+++ b/.github/workflows/upload_binary.yml
@@ -16,14 +16,17 @@ jobs:
             pathsep: ";"
             asset_name: black_windows.exe
             executable_mime: "application/vnd.microsoft.portable-executable"
+            platform: windows
           - os: ubuntu-20.04
             pathsep: ":"
             asset_name: black_linux
             executable_mime: "application/x-executable"
+            platform: unix
           - os: macos-latest
             pathsep: ":"
             asset_name: black_macos
             executable_mime: "application/x-mach-binary"
+            platform: macos
 
     steps:
       - uses: actions/checkout@v2
@@ -42,9 +45,8 @@ jobs:
       - name: Build binary
         run: >
           python -m PyInstaller -F --name ${{ matrix.asset_name }} --add-data
-          'src/blib2to3${{ matrix.pathsep }}blib2to3' --hidden-import platformdirs.unix
-          --hidden-import platformdirs.macos --hidden-import platformdirs.windows
-          src/black/__main__.py
+          'src/blib2to3${{ matrix.pathsep }}blib2to3' --hidden-import platformdirs.${{
+          matrix.platform }} src/black/__main__.py
 
       - name: Upload binary as release asset
         uses: actions/upload-release-asset@v1
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### Packaging
+
+- Fix missing modules in self-contained binaries (#2466)
+
 ## 21.8b0
 
 ### _Black_


felix-hilden:
Thanks for the PR! I'm not familiar with the system in place for building binaries, so I can't comment on the solution. But I think this could warrant a changelog entry, since it is clearly fixing a problem!

jalaziz:
@felix-hilden added a changelog entry. Please let me know if it needs any adjustment.

.github/workflows/upload_binary.yml
```diff
@@ -40,8 +40,11 @@ jobs:
           python -m pip install pyinstaller
 
       - name: Build binary
-        run: |
-          python -m PyInstaller -F --name ${{ matrix.asset_name }} --add-data 'src/blib2to3${{ matrix.pathsep }}blib2to3' src/black/__main__.py
+        run: >
+          python -m PyInstaller -F --name ${{ matrix.asset_name }} --add-data
+          'src/blib2to3${{ matrix.pathsep }}blib2to3' --hidden-import platformdirs.unix
+          --hidden-import platformdirs.macos --hidden-import platformdirs.windows
+          src/black/__main__.py
```
cooperlees, line 47: Just a question - Is there a fix we can do to platformdirs to not require this? I guess a follow on question is what is the difference with what platformdirs does vs. appdirs? Does it divide up each platforms into sub modules?

cooperlees (commented):
Just asking a question to see if we should maybe inform (via an issue) platformdirs about this? This is a change of behavior from appdirs.

.github/workflows/upload_binary.yml
jalaziz: It looks like `appdirs` [doesn't use dynamically loaded modules per platform](https://github.com/ActiveState/appdirs/blob/master/appdirs.py), while `platformdirs` [does](https://github.com/platformdirs/platformdirs/blob/main/src/platformdirs/__init__.py#L18-L28).

jalaziz:
I went ahead and updated this PR with a small enhancement that only includes the `platformdirs` module for the platform being built against. I've briefly tested the linux and macos binaries. Test binaries can be found at https://github.com/jalaziz/black/releases/tag/21.8b1.

jalaziz:
> Just asking a question to see if we should maybe inform (via an issue) platformdirs about this? This is a change of behavior from appdirs.

Technically, the change in behavior should be fine. It just break things when packaging via `PyInstaller`. It could be argued that is is better as it allow a slightly smaller binary by excluding modules that aren't needed for a particular OS.

JelleZijlstra (approved):

.github/workflows/upload_binary.yml
ichard26: The fact platformdirs breaks PyInstaller doesn't surprise me knowing how it works, but what does is that we have conditional imports in our code and yet they're not broken? The main culprit I have in mind is `typing-extensions`... it's not like the base install of Black has dependencies that then unconditionally depend on it. I guess sys.version_info guarded imports are treated correctly? 

Or I could be misunderstanding all of this which is totally possible as it's 11 PM over here ^^

jalaziz: According to PyInstaller:

> Hidden imports can occur when the code is using __import__(), importlib.import_module() or perhaps exec() or eval(). Hidden imports can also occur when an extension module uses the Python/C API to do an import. When this occurs, Analysis can detect nothing. There will be no warnings, only an ImportError at run-time.

That seems to suggest that conditional imports are fine, but dynamic imports aren't. I haven't looked into exactly how PyInstaller detects imports, but I suspect they some version of searching for import statements.

ichard26: Oh right, conditional import != hidden import. G'ah I should really get some sleep :sweat_smile:

ichard26:
I can test the Windows binary tomorrow if no one beats me to it. I doubt there's going to be an issue with it, but I'd rather play it safe. Thanks for the awesome work and for reporting the issue in the first place!

ichard26 (approved):
Tested the generated binary on Windows and it works totally fine. Beautiful, thank you!

ichard26:
> Just asking a question to see if we should maybe inform (via an issue) platformdirs about this? This is a change of behavior from appdirs.

They seem to be aware of this issue: https://github.com/platformdirs/platformdirs/issues/40 but given that the only maintainer response has been the addition of the "enhancement" and in particular "help wanted" labels in a ~week... it might take a while for that to be resolved. I'd rather prefer to unbreak this for our users first, and then get rid of it if those folks end up working around or fixing the issue. 

---

Thank you so much for your contribution! This project is only possible by contributions like these :black_heart: @jalaziz. Congrats on your commit to psf/black! If you have any feedback re. contributing, please tell us know via #2238.

jalaziz:
Thank you for the quick turnaround! Always happy to contribute. Turns out I ran into this a couple weeks back and thought it was something that the binary build already took care of. Didn't realize the `platformdirs` change wasn't released yet at the time.

With regards to platformdirs/platformdirs#40, I was actually looking into the hook system that is being recommended there but thought that it would be a bit awkward for a project like `black`. Didn't cross my mind that `platformdirs` could host the hook file.

That being said, any chance of this fix being published soon given that the 21.8b0 binaries are broken?

jalaziz:
@ichard26 Looks like PyInstaller has a contrib repo for hooks and support was added for `platformdirs` recently: https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-platformdirs.py.

Once a new version of the contrib repo is released, we should be able to remove the manual imports (the PyInstaller docs suggest that the contrib package is installed automatically).

PR #2468, opened 2021-08-31
Change docker workflow for latest_release tag

pszlazak:
### Description

Follow-up for #2461.

Triggering conditions changed in one more place of `docker.yml`.  
This is triggering condition for creating `pyfound/black:latest_release` Docker image.

@onerandomusername and @ichard26 to take a look.

### Checklist - did you ...

- [x] Add a CHANGELOG entry if necessary? -> **N/A**
- [x] Add / update tests if necessary? -> **N/A**
- [x] Add new / update outdated documentation? -> **N/A**

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -41,7 +41,7 @@ jobs:
           tags: pyfound/black:latest,pyfound/black:${{ env.GIT_TAG }}
 
       - name: Build and push latest_release tag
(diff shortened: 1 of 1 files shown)


Akarys42 (approved):
For what is worth, this is the right trigger to use

cooperlees (approved):
Thanks for this

PR #2477, opened 2021-09-05
Remove `blackcellmagic` reference

MarcoGorelli:
This package seems to be unmaintained (last commit is from > 2 years ago), and `black` now runs on Jupyter Notebooks directly

### Checklist - did you ...


cooperlees (approved):
Seems legit to me to no longer recommend this …

Issue #2383, opened 2021-07-19
Question: Does the code format differ between versions of python 3 (3.7, 3.8, 3.9) ?


freddyaboulton:
Hello,

Not sure where the best place to ask this question is so I decided to file an issue. In my organization, we've been running black for a couple of months under python versions 3.7 and 3.8 in our CI and I've never noticed that the output is different between the two.

Is there any scenario where the code style between the python 3 versions will be different? I'm wondering if I should suggest that we only run black with python 3.8.

Thank you! I love using your package.

HassanAbouelela:
Off the top of my head, you will get some differences when it comes to version-specific features, but only in very limited cases. For a minimal reproducible example, have a look at the following script:
```py
if x := False:
    print("what?")
```

Walrus was introduced 3.8 I believe, which means a 37 vs 38 black will have slightly different behavior (only in certain cases).
The following commands will work (file.py is the code above):
> (38) black file.py
> (38) black -t py37 file.py
> (38) black -t py37 -c "if x := False: print("wot")"

But not the following:
> (37) black file.py
> (37) black -t py38 file.py
> (37) black -t py38 -c "if x := False: print("wot")"

In short: yes, there is some difference in behavior (mostly in AST parsing in this case), but you may not notice it depending on your code. If your codebase is not intended to support 3.7, then it's probably safe to ignore this. If 3.7 is supported though, it may make sense to keep it.

freddyaboulton:
Thank you @HassanAbouelela ! This is a very helpful answer.

cooperlees:
I think the tl;dr here is just run black under the newest cpython runtime possible (or in use at your company) and you should have less issues.

PR #2481, opened 2021-09-06
DOC: cleanup pre-commit instructions following #2430

neutrinoceros:
### Description

I believe this was forgotten in #2430

### Checklist - did you ...

- [N/A] Add a CHANGELOG entry if necessary?

### DIFF
--- a/docs/integrations/source_version_control.md
+++ b/docs/integrations/source_version_control.md
@@ -10,7 +10,6 @@ repos:
     rev: 21.8b0
     hooks:
       - id: black
-        language_version: python3 # Should be a command that runs python3.6+
 ```
 
 Feel free to switch out the `rev` value to something else, like another
--- a/docs/integrations/source_version_control.md
+++ b/docs/integrations/source_version_control.md
@@ -10,6 +10,11 @@ repos:
     rev: 21.8b0
     hooks:
       - id: black
+        # It is recommended to specify the latest version of Python
+        # supported by your project here, or alternatively use
+        # pre-commit's default_language_version, see
+        # https://pre-commit.com/#top_level-default_language_version
+        language_version: python3.9
 ```
 
 Feel free to switch out the `rev` value to something else, like another


neutrinoceros:
I don't think this change warrants a mention in the changelog. I can add one if required but I'd be content with a "skip news" label.

zsol:
Correct me if I'm wrong but if you don't specify that `language_version` anywhere then precommit will use whatever is on your system. If that's the case, I'd rather have it there as a suggestion as it's a reasonable default.

neutrinoceros:
@zsol I believe this was previously discussed in #2430. Here's [the key comment therein](https://github.com/psf/black/pull/2430#issuecomment-900434142)
Quote:
> black's hooks already require a minimum_pre_commit_version of 2.9.2, which is after pre-commit itself dropped Python 2 support. So, pre-commit will already be running under Python 3, and in the absense of a langauge_version defined here, will reuse the same Python it is running under for the black hook.

MarcoGorelli:
If I can give my 2 cents:

It's true that specifying Python3 is no longer necessary, as there's no longer any risk of `black` being run via Python2. However, `black` differs when run under Python3.7 or Python3.8.

Example:
```console
$ cat t.py 
if a := 3 > 0: print(a)
$ vim .pre-commit-config.yaml 
$ cat .pre-commit-config.yaml 
repos:
  rev: 21.9b0
  hooks:
  - id: black
$ pre-commit run black --all-files
- hook id: black
- exit code: 123

error: cannot format t.py: cannot use --safe with this file; failed to parse source file.  AST error message: invalid syntax (<unknown>, line 1)
Oh no! 💥 💔 💥
1 file failed to reformat.

$ vim .pre-commit-config.yaml 
$ cat .pre-commit-config.yaml 
repos:
  rev: 21.9b0
  hooks:
  - id: black
    language_version: python3.8
$ pre-commit run black --all-files
- hook id: black
- files were modified by this hook

reformatted t.py
All done! ✨ 🍰 ✨
1 file reformatted.
```

So arguably it's still good advice to specify a language version.

Perhaps the docs could just be modified to show:
```
        language_version: python3.8 # Optional, to enforce a specific Python version
```
?

neutrinoceros:
Wouldn't it be redundant with the target-version parameter ?

MarcoGorelli:
Just setting `-t py38` didn't work for me:
```console
$ cat .pre-commit-config.yaml 
repos:
  rev: 21.9b0
  hooks:
  - id: black
    args: ['-t', 'py38']

$ pre-commit run black -a
- hook id: black
- exit code: 123

error: cannot format t.py: cannot use --safe with this file; failed to parse source file.  AST error message: invalid syntax (<unknown>, line 1)
Oh no! 💥 💔 💥
1 file failed to reformat.
```

neutrinoceros:
Interesting. In retrospect, I think I never realized this because the only large project I've used black via pre-commit with has 3.6 as a target, and black never supported an older version I think.
So the correct recommendation would actually be to merge align `language-version` with black's `target-version` ? Given the later can be an array, should we align with the oldest or with the most recent target ?

MarcoGorelli:
> So the correct recommendation would actually be to merge align language-version with black's target-version ? Given the later can be an array, should we align with the oldest or with the most recent target ?

I've not looked much into `target_version`, but from https://github.com/psf/black/issues/2383#issuecomment-882729863 it looks like it should be the newest target

neutrinoceros:
Alright I updated the PR to reflect this discussion.

zsol (approved):

zsol:
Thanks!

Issue #2382, opened 2021-07-19
All-whitespace files do not get their newlines removed


JelleZijlstra:
From @akaihola in #779:

In akaihola/darker#164 ([comment](https://github.com/akaihola/darker/issues/164#issuecomment-882014824)) we noticed that `format_str()` gives different results for some files than running `black` from the command line or `black.main(..., standalone_mode=False)`. This should probably be fixed before making `format_str()` an officially public Python API.

Consider this test script `compare_black.py`:
```python
import io
import sys
from contextlib import redirect_stdout
from subprocess import PIPE, check_output

from black import FileMode, diff, format_str, main

path = sys.argv[1]

print("============================= format_str =============================")
with open(path) as f:
    content = f.read()
format_str_result = format_str(content, mode=FileMode())
print(diff(content, format_str_result, path, path))

print("================================ main ================================")
with redirect_stdout(io.TextIOWrapper(io.BytesIO())) as main_stdout:
    main(["--quiet", "--diff", path], standalone_mode=False)
main_stdout.seek(0)
print(main_stdout.read())

print("============================= subprocess =============================")
subprocess_result = check_output(["black", "--quiet", "--diff", path], encoding="UTF-8")
print(subprocess_result)
```

If you create a valid Python source code file with just a few newlines:
```bash
echo "



" >newlines.py
```
and reformat it using the above script:
```bash
python compare_black.py newlines.py                   
============================= format_str =============================
--- /tmp/myfile.py
+++ /tmp/myfile.py
@@ -1,5 +0,0 @@
-
-
-
-
-

================================ main ================================

============================= subprocess =============================

```
You see that `format_str()` removes those blank lines while `main()` and `black` leave the file untouched.

I haven't tried to found other cases where these outputs don't match. I wonder if it only happens to all-whitespace files. **Edit:** Yes, only with all-whitespace files.

_Originally posted by @akaihola in https://github.com/psf/black/issues/779#issuecomment-882233332_

akaihola:
>All-whitespace files do not get their newlines removed

So @JelleZijlstra the expected behavior is for all-whitespace files to be truncated instead of being left unmodified? Makes sense.

So it's actually `format_str()` which does the right thing, while the `black` command line, `black.main(..., standalone_mode=False)` and `black.format_file_contents()` are wrong since they don't reformat all-whitespace files.

JelleZijlstra:
[...]

PR #2484, opened 2021-09-07
Empty and all-whitespace files become a single empty line and a linefeed

akaihola:
### Description

Currently Python source files with only whitespace (space, tab, linefeed) as well as zero-length files are not reformatted but kept intact. In #2382 the consensus was that all of these should become a file with a single empty line, i.e. just a linefeed character.

Also, currently `black.format_str()` turns all-whitespace strings into zero-length strings, which is inconsistent compared to the command line and the `black.format_file_contents()` function. Wtih this patch, both functions reformat the source into a single newline.

This PR fixes issue #2382.

**Note:** See comments for the discussion on whether **zero-length** files should be kept intact, turned into a single LF, or turned into a single platform-dependent line terminator.

### Checklist - did you ...


CHANGES.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### _Black_
+
+- Reformat zero-length and all-whitespace files as a single empty line, i.e. just a
+  linefeed character (#2382)
```
```suggestion
  linefeed character (#2484)
```
This should be the PR number, not the issue number.

Also, thanks for submitting this PR! I'll take a deeper look soon.

akaihola: Hi @JelleZijlstra,

I amended and force-pushed this correction as soon as the linter warned me about it – I didn't expect any human to notice it this soon :)

I hope the force-push didn't mess anything up.

src/black/__init__.py
```diff
@@ -1061,6 +1060,8 @@ def f(
             current_line, mode=mode, features=split_line_features
         ):
             dst_contents.append(str(line))
+    if not dst_contents:
+        return "\n"
```
cooperlees, line 1064: What do other formatters do here when running on windows vs. POSIX?

Should we (if found) leave Windows like terminators alone? (to leave windows test files alone? i.e. `"\r\n"`)

cooperlees (commented):
I love making this a common black practice.

I only have a question about if we come across windows empty text files - Should we maintain them or just say "black "\n's" all empty .py files etc. - I'm fine with the latter - Just want to see if others are ...

src/black/__init__.py
felix-hilden: That'd be good for consistency with nonempty files 😅

akaihola: >Should we (if found) leave Windows line terminators alone?

I believe this patch doesn't change Black's behavior with regard to existing line terminators. 

Replacing `"\n"` with `os.linesep` here and in corresponding tests should make this behave in a more native way on Windows, right?
```suggestion
    if not dst_contents:
        return os.linesep
```

akaihola: >I believe this patch doesn't change Black's behavior with regard to existing line terminators.

Ah, but of course it does. A whitespace-only file with Windows line terminator first becomes empty, and is then changed into a single Unix linefeed. Hm.

cooperlees: Yeah - This could also induce changes all the time if you have developers on Windows vs. POSIX ... i.e. file gets reformatted on Windows, then committed, then a POSIX dev will reformat and commit (we could have a constant battle)

We will just have to pick the one - which I think `\n` is more accepted in Python source files. Editors have been taught to leave this alone by default usually I think.

felix-hilden: Surely git will take care of changing the endings though!

felix-hilden: Some discussion about the line endings in #1741

akaihola: I think I agree with @zsol's [comment](https://github.com/psf/black/issues/1741#issuecomment-703775430):
>IIRC we have considered this, but ultimately decided to be extra careful and preserve the original line endings, as we felt that's more in the domain of the version control software being used. The only opinion Black has on this matter is that line endings should be consistent (which is a fairly basic requirement :) )

So for zero-length files which Black turns into a single line terminator, I see three viable options:
1. always `\n`
2. `os.linesep`, so `\n` on Unix and `\r\n` on Windows
3. somehow sniff or remember the line terminator in neighboring files and use that

I never develop on Windows, so I don't know if it's a more common case to see CR or CRLF line terminators in Python projects when using most common tooling on that platform.

felix-hilden: I think it should be the system line ending, because git can be used for conversion. I don't really know if it is more common to still use `\n`, but at least I've not encountered it. Might have missed it though as well, if editors are transparent about that.

septatrix: > What do other formatters do here when running on windows vs. POSIX?
>
> Should we (if found) leave Windows line terminators alone? (i.e. "\r\n")

I do not have a windows machine at my disposal but I tested a few formatters from different languages:

* Yapf: Leaves a completely empty file as is,  a whitespace only file without newline gets converted to `\n`, a file with a single newline (`\n` or `\r\n`) is preserved the way it was before and a file with multiple newlines is truncated to a `\n` (even if the newlines were `\r\n`).
* Prettier (for JS): Defaults to `\n` for each platform.
* Rustfmt (for Rust): Seems to convert an empty file to `\n`, otherwise identical to yapf.

Personally I think it would be the best compromise is to leave a completely empty file as is and a file with non-newline whitespace should be converted to a completely empty file. For all other files we can infer the line ending and preserve it.

akaihola: So the "votes" so far for what zero-length files should become on Windows go like:
- **0.5** for keep empty: @JelleZijlstra (in issue 2382)
- **2** for LF: @JelleZijlstra, @septatrix (in issue 2382), @cooperlees (in this issue)
- **1.5** for CRLF: @cooperlees (?), @felix-hilden (in this issue)

I'm counting half a vote for those who are ok with two different alternatives. @JelleZijlstra & @cooperlees, do you want to choose just one preference based on the discussion above? Is "voting" even a good way to reach a design decision in Black, or do we have a BDFL to dictate the choice?

akaihola: Found one linting issue with empty files:
```bash
$ echo >/tmp/empty.py
$ flake8 /tmp/empty.py
/tmp/empty.py:1:1: W391 blank line at end of file
$ pylint /tmp/empty.py      
************* Module empty
/tmp/empty.py:1:0: C0305: Trailing newlines (trailing-newlines)
```

So [Flake8](https://flake8.pycqa.org/) and [Pylint](https://pylint.org/) seem to think empty source files should be zero-length instead of containing any line terminators! I'll add these as "voters" above.

zsol: I agree with Flake8 and Pylint FWIW

akaihola: I now changed the implementation to emit a
- zero-length file for empty or all-whitespace input but now line terminators
- single line terminator for all-whitespace input with at least one line terminator, preserving LF/CRLF line terminators

akaihola: I couldn't figure out why there were linting errors in the GitHub Actions workflow for files this branch didn't even touch.

akaihola:
I couldn't figure out why there were linting errors in the GitHub Actions workflow for files this branch didn't even touch.

ichard26:
if it's flake8 related, it's a known issue, we unfortunately did not pin a plugin we use and now there's new warnings.

akaihola:
@ichard26 yes, I'm seeing Flake8 failures – interestingly different ones for different builds, some for files I didn't modify.

In addition, in the latest build, coverage can't be submitted:
```
Failed submitting coverage with service_name: github
  File "/usr/local/lib/python3.8/site-packages/coveralls/api.py", line 290, in submit_report
    response.raise_for_status()
  File "/usr/local/lib/python3.8/site-packages/requests/models.py", line 953, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 405 Client Error: Not Allowed for url: https://coveralls.io/api/v1/jobs

  File "/src/entrypoint.py", line 62, in run_coveralls
    result = coveralls.wear()
  File "/usr/local/lib/python3.8/site-packages/coveralls/api.py", line 257, in wear
    return self.submit_report(json_string)
  File "/usr/local/lib/python3.8/site-packages/coveralls/api.py", line 293, in submit_report
    raise CoverallsException(
coveralls.exception.CoverallsException: Could not submit coverage: 405 Client Error: Not Allowed for url: https://coveralls.io/api/v1/jobs
Trying submitting coverage with service_name: github-actions...
Patching os.environ with: {'COVERALLS_REPO_TOKEN': '***', 'COVERALLS_PARALLEL': 'true', 'COVERALLS_FLAG_NAME': 'py3.9-ubuntu-latest'}
```

The same happens twice, the second time `with service_name: github-actions` instead of `github`.

PR #2489, opened 2021-09-10
Fix patching of `io.TextWrapper` in tests

akaihola:
### Description

`patch("io.TextIOWrapper")` wouldn't actually cause `io.TextIOWrapper()` to return the `StringIO` object created by the test. This didn't cause any problems though until I added a test in #2484 in which `content=` is non-empty:

```python
>               f.write(dst)
E               ValueError: I/O operation on closed file

src/black/__init__.py:849: ValueError
```

In the debugger we can see that `io.TextIOWrapper` used on `src/black/__init__.py:842–844`:
```python
            f = io.TextIOWrapper(
                sys.stdout.buffer, encoding=encoding, newline=newline, write_through=True
            )
```
is not the patched lambda:
```python
(Pdb) io.TextIOWrapper
<class '_io.TextIOWrapper'>
```

With the changes in the PR, it is properly patched:
```python
> src/black/__init__.py(850)format_stdin_to_stdout()
-> f.write(dst)
(Pdb) io.TextIOWrapper
<Mock name='mock.TextIOWrapper' id='139749471475408'>
```

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@
 - Fix missing modules in self-contained binaries (#2466)
 - Fix missing toml extra used during installation (#2475)
 
+### Tests
+
+- Fix patching of `io.TextWrapper` (#2489)
+
 ## 21.8b0
 
 ### _Black_
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@
 
 ### Tests
 
-- Fix patching of `io.TextWrapper` (#2489)
+- Fix patching of `io.TextIOWrapper` (#2489)
 
 ## 21.8b0
 
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -25,7 +25,7 @@ from typing import (
 )
 import pytest
 import unittest
-from unittest.mock import patch, MagicMock
+from unittest.mock import patch, MagicMock, Mock
 from parameterized import parameterized
 
 import click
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -7,6 +7,10 @@
 - Fix missing modules in self-contained binaries (#2466)
 - Fix missing toml extra used during installation (#2475)
 
+### Tests
+
+- Fix patching of `io.TextWrapper` (#2489)
```
JelleZijlstra, line 12:
```suggestion
- Fix patching of `io.TextIOWrapper` (#2489)
```

JelleZijlstra (commented):
Can you clarify why the previous way didn't work? It looks to me like it should.

akaihola:
>Can you clarify why the previous way didn't work? It looks to me like it should.

@JelleZijlstra, a very good question, which was a tough one to sort out!

Indeed, in the `main` branch, it works just fine. But in #2484 I had to introduce the extra call to `decodebytes()` in `format_str()` (see [diff](https://github.com/psf/black/pull/2484/files#diff-d23a6917d7262ca9815093bffb7148299e516575251959ed9b405b648e76aa29R1064)). And since `decodebytes()` instantiates an `io.TextIOWrapper()` and closes it, the `io.TextIOWrapper()` invocation we intended to patch in `format_stdin_to_stdout()` wouldn't work anymore.

In other words, this is a case of a common problem when patching and asserting library classes/functions: You can't be sure that the patched item won't in the future get extra invocations in addition to the one you're originally testing for.

For some reason, patching only the `io` module imported in `black/__init__.py` fixes this issue. I can't really figure out now why that is the case, since `decodebytes()` is in that same module.

As a conclusion from looking into this deeper, the modification to patching in the test function really belongs in #2484 since it's only the changes there which break the patching. So closing this PR.

JelleZijlstra:
Thanks for the thorough explanation!

PR #2494, opened 2021-09-14
Update CHANGES.md for 21.9b0 release

JelleZijlstra:


hugovk:
RTD stable doesn't show the new release:

https://black.readthedocs.io/en/stable/change_log.html

RTD latest does:

https://black.readthedocs.io/en/latest/change_log.html

Does something need updating?

cooperlees:
We need to move the stable tag. We never do that right at release. We wait a few days. So it's "stable".

PR #2500, opened 2021-09-18
re-implement simple CORS middleware for blackd

zsol:
### Description

This PR removes Black's dependency on `aiohttp-cors` because it seems to be unmaintained and spewing `DeprecationWarnings` (due to the use of `@coroutine` decorators).

### TODO

- [x] Add a CHANGELOG entry 
- [x] Add tests
- [x] Actually remove `aiohttp-cors` from `setup.py`
- [x] Actually remove `aiohttp-cors` from `Pipfile`

### DIFF
--- a/src/blackd/middlewares.py
+++ b/src/blackd/middlewares.py
@@ -1,12 +1,14 @@
 from typing import Iterable, Awaitable, Callable
 from aiohttp.web_response import StreamResponse
 from aiohttp.web_request import Request
+from aiohttp.web_middlewares import middleware
 
 Handler = Callable[[Request], Awaitable[StreamResponse]]
 Middleware = Callable[[Request, Handler], Awaitable[StreamResponse]]
 
 
 def cors(allow_headers: Iterable[str]) -> Middleware:
+    @middleware
     async def impl(request: Request, handler: Handler) -> StreamResponse:
         is_options = request.method == "OPTIONS"
         is_preflight = is_options and "Access-Control-Request-Method" in request.headers
@@ -29,4 +31,4 @@ def cors(allow_headers: Iterable[str]) -> Middleware:
 
         return resp
 
-    return impl
+    return impl  # type: ignore
--- a/src/blackd/middlewares.py
+++ b/src/blackd/middlewares.py
@@ -1,5 +1,4 @@
 from typing import Iterable, Awaitable, Callable
-from aiohttp.web import middleware
 from aiohttp.web_response import StreamResponse
 from aiohttp.web_request import Request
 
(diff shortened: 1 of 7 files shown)


src/blackd/middlewares.py
```diff
@@ -0,0 +1,34 @@
...
+        resp.headers["Access-Control-Expose-Headers"] = "*"
+        if is_options:
+            resp.headers["Access-Control-Allow-Headers"] = ", ".join(allow_headers)
+            resp.headers["Access-Control-Allow-Methods"] = ", ".join(
+                ("OPTIONS", "POST")
+            )
+
+        return resp
+
+    return impl  # type: ignore
```
zsol, line 34: mypy can't figure out the type of `impl` here unfortunately - it complains about returning an `Any` value instead of `Middleware`. @JelleZijlstra any ideas?

ichard26:
I'm very concerned about the changes within Pipfile.lock. I suspect pipenv removed too much, eg. dataclasses. I have some experience (un)fortunately in this regard and I'll try to reduce the lockfile changes to avoid breakage.

zsol:
You're right, `dataclasses` got downgraded to 0.6 accidentally, let me see if I can fix that

zsol:
I gave up on trying to get pipenv to not revert the dataclasses version :)

src/blackd/middlewares.py
JelleZijlstra: What version of aiohttp are we using? On master (https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_middlewares.py#L34) it looks like the decorator does nothing.

zsol: Looks like 3.6 and above: https://github.com/psf/black/blob/d08f95aa87d30c6997b706af22eed945d2b45afb/setup.py#L89-L89

We could update this, but I'd prefer doing it in a separate PR

JelleZijlstra: For what it's worth the decorator does do something on aiohttp 3.6: https://github.com/aio-libs/aiohttp/blob/3.6/aiohttp/web_middlewares.py#L34. I'm not sure why mypy turns the function into Any here though; the type annotations seem like they should work.

zsol: That's true but afaict nothing checks that attribute. Anyway, the tests pass 😎

ichard26:
I wasn't able to get the cleanest Pipfile.lock diff ever but this should do for the time being (ignoring the dataclasses fiasco).

ichard26 (approved):
This looks fine to me but the only knowledge of CORS I have comes from a quick read of the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS). I hope someone else with more web dev experience can pitch in.

PR #2510, opened 2021-09-25
MNT: remove unnecessary test deps + some refactoring

ichard26:
### Description

The main goals of this PR include:
- improving consistency on how strict the test suite is -- Jelle has seen cases where a test did not fail to an incomplete test setup even though it should've
- simplifying tests for both ease of creation and reading via parametrization and helpers
- reorganizing the test suite by grouping more tests
- dropping test suite dependencies that aren't strictly necessary   

The test suite could definitely do with more refactoring, but this is a good first pass. Anyway it would've gotten too big to review effectively if I did continue on this PR.

### Checklist - did you ...

- [x] Add new / update outdated documentation? -> n/a -- I'll add test suite docs later

### Review notes

I strongly suggest reviewing commit-by-commit as the total diff is rather big due to the various moves I did. The first commit will probably be still annoying to review (as I didn't split the move and the refactor into two commits) but everything else should be well scoped -- I hope.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,11 +7,8 @@ verify_ssl = true
 # Testing related requirements.
 coverage = ">= 5.3"
 pytest = " >= 6.1.1"
-pytest-mock = ">= 3.3.1"
-pytest-cases = ">= 2.3.0"
 pytest-xdist = ">= 2.2.1"
 pytest-cov = ">= 2.11.1"
-parameterized = ">= 0.7.4"
 tox = "*"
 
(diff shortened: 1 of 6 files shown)


felix-hilden (approved):
Can't say I went through each case line by line, but the changes and refactorings do make sense 👌 LGTM

JelleZijlstra (approved):
Looks great but there's a merge conflict on Pipfile.lock :(

PR #2523, opened 2021-10-06
Deprecate Python 2 formatting support

ichard26:
### Description

It's 2021. See also #2251.

### Checklist - did you ...

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -74,12 +74,16 @@ disabled-by-default counterpart W504. E203 should be disabled while changes are 
 
 ## Does Black support Python 2?
 
+```{warning}
+Python 2 support has been deprecated since 21.10b0.
+
+This support will be dropped in the first stable release, expected for January 2022.
+See [The Black Code Style](the_black_code_style/index.rst) for details.
+```
+
 For formatting, yes! [Install](getting_started.md#installation) with the `python2` extra
 to format Python 2 files too! In terms of running _Black_ though, Python 3.6 or newer is
 required.
-
-Note that this support will be dropped in the first stable release, expected for
-January 2022. See [The Black Code Style](the_black_code_style/index.rst) for details.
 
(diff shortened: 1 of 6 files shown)


ichard26:
Just tagging this as a "do not merge" as all of this is very much up for debate.

ichard26:
The current structure of the commits require that this is merged via a rebase so the split ends up being useful :)

> **Prepare for Python 2 depreciation** 
>
> - Use BlackRunner and .stdout in command line test
>
> So the next commit won't break this test. This is in its own commit so we can just revert the depreciation commit when dropping Python 2 support completely.

cooperlees (approved):
LGTM - Lets merge?

graingert (changes requested):
The comments aren't consistent

graingert (approved):

ichard26:
One potential problem I do want to bring up is that right now each call to format_str with Python 2 warns ... that may be too noisy.

docs/faq.md
```diff
@@ -74,13 +74,17 @@ disabled-by-default counterpart W504. E203 should be disabled while changes are
 
 ## Does Black support Python 2?
 
+```{warning}
+Python 2 support has been deprecated since 21.10b0.
+
+this support will be dropped in the first stable release, expected for January 2022.
```
JelleZijlstra, line 80:
```suggestion
This support will be dropped in the first stable release, expected for January 2022.
```

JelleZijlstra (commented):
Noticed one typo.

I'm also still thinking about the feature detection stuff here and brought it up on Discord.

JelleZijlstra:
Printing for each file might be noisy, but properly doing it only once might require communication across asyncio workers, which I don't really want to have to deal with.

JelleZijlstra (approved):
Let's merge it.

aarav2you (approved):

PR #2525, opened 2021-10-07
feat(config): Support `--config` in pyproject.toml

### LINKED ISSUE #1826: Support for `--config` in pyproject.toml
I am often dealing with projects that produce multiple libraries from the same repository, which looks a bit like:
```
src/lib1/pyproject.toml
src/lib1/setup.py
src/lib2/pyproject.toml
src/lib2/setup.py
```
All libs need to follow the same formatting rules, which means that the black configuration has to be duplicated in each `pyproject.toml` file.

A possible way to avoid the duplication could be to teach black to handle the already existing `--config` option when it is provided inside the *normal* `pyproject.toml`. Basically something like:
```
src/lib1/pyproject.toml
src/lib1/setup.py
src/lib2/pyproject.toml
src/lib2/setup.py
src/black.toml
```
with each `pyproject.toml` containing:
```
[tool.black]
config = '../black.toml'
```
A relative path would be interpreted relative to the folder containing the `pyproject.toml` file.

Thoughts? Would this be a welcome contribution?

Shivansh-007:
### Description

You can now specify `config` key in pyproject.toml which does the same job as the CLI option `--config`. The path of config should be relative to the CWD it is running from. If the file is not found or is invalid i.e. it is a folder rather a TOML file it would raise `click.exceptions.FileError`, and if the file can't be parsed it would raise `tomli.TOMLDecodeError`. All black config keys in the custom config should be registered under `[black]` and they would overwrite the config specified in pyproject but not the one specified by `--config` flag while running black with the CLI.

I also did a small update to `.gitignore` by grouping the "test" and adding `htmlcov` to it. Which is generated by `coverage html` and `.pytest_cache/` produced when you run `pytest .`.

Closes #1826

### Checklist - did you ...


ichard26:
I haven't taken a deep look at the code nor played with the feature locally but I do quickly want to mention we should probably somehow indicate we using a configuration file via indirection. Honestly I'm not sure what's up with the weird merging / overriding behaviour here, but I'd suggest to keep it simple for now and just treat a config value in config file (regardless if was explicitly given or implicitly discovered) as a HTTP 302 redirect.

Shivansh-007:
Question: Should the config be overwritten if it is passed through the CLI option `--config`? The `config = "black.toml` would be found in that file (which is passed through the CLI).

ichard26:
I'm confused by what you mean, can you further explain what scenario we're talking about here?

Shivansh-007:
There are two ways configs are loaded, (1) they are passed by the user through `--config` flag and (2) the `pyproject.toml` is found in the `src`. So currently when we find the `config` key we check whether if it is found by the config found in `pyproject.toml` or it is found in the one passed by the user via `--config` option. I am not sure why I did this originally, so I was asking should I keep this check or just overwrite it in any case. I would go with the latter.

Shivansh-007:
Yeah, I am on discord also (`Jason Terror#XXXX`) but currently, off from it until my exams ended, I have a break between the next ones till the 17th so opened these PRs - didn't want to keep them hanging, school ones just finished.

felix-hilden (changes requested):
Richard also raised a good point about chaining this value in the [original issue](https://github.com/psf/black/issues/1826#issuecomment-876022270). I tend towards chaining being supported, which would require a loop of some sort to follow the TOML files.

Also, I think it would be more consistent to just require all TOMLs to have the same structure, i.e. a `tool.black` section instead of some other magical section. That way we could probably simplify the implementation and also solve chain cycles by looping through the files with an iteration threshold.

felix-hilden (commented):
Looking better already, here's some comments:

CHANGES.md
```diff
@@ -9,13 +9,15 @@
...
 - Add support for parenthesized with (#2586)
+- Allow specifying `config` in config files (#2525)
 
 ## 21.10b0
 
 ### _Black_
 
 - Document stability policy, that will apply for non-beta releases (#2529)
 - Add new `--workers` parameter (#2514)
+- Bumped typed-ast version minimum to 1.4.3 for 3.10 compatibility (#2519)
```
JelleZijlstra, line 20:
```suggestion
```
I messed up the merge

tests/test_black.py
```diff
@@ -1296,6 +1296,28 @@ def test_read_pyproject_toml(self) -> None:
         self.assertEqual(config["exclude"], r"\.pyi?$")
         self.assertEqual(config["include"], r"\.py?$")
 
+    def test_black_replace_config(self) -> None:
+        test_toml_file = THIS_DIR / "test_replace.toml"
+        fake_ctx = FakeContext()
+        black.read_pyproject_toml(fake_ctx, FakeParameter(), str(test_toml_file))
+        config = fake_ctx.default_map
+        # `0` in `test_replace.toml` and `1` in `_black_config.toml`. Should be `1` as
```
JelleZijlstra, line 1304: What is the reason for this behavior? The original use case was to make it possible to have a base config that other config files in a family of projects can import. For that use case, it makes more sense to me to have `test_replace.toml` win.

Either way, this behavior should be documented.

Shivansh-007: Isn't this the required behaviour? The linked config would over-write the root config.

JelleZijlstra: Why? It doesn't make sense to me to allow other config in the toml that does `config = ` if we throw it away.

felix-hilden: A simple override could be better to not have weird issues and inconsistencies. And we could even warn or *error* on other values beyond `config`. But anyways, merging wasn't required at least in the original use case of multiple libraries sharing the same configuration in #1826. I'm still leaning on not merging. Not sure about it for CLI flags though, intuitively they should be prioritised.

Shivansh-007: We discussed this in the discord channel and the commited solution was decided.

docs/usage_and_configuration/the_basics.md
```diff
@@ -200,6 +200,37 @@ project.
...
+lineno-start: 1
+emphasize-lines: 2
+---
+[tools.black]
+color = false
+line-length = 88
+target-version = ['py36', 'py37', 'py38']
+```
+
+The resulting value for `color` would be `true` as specified in the parent config.
```
felix-hilden, line 231: Surely the "parent" configuration is the linked configuration here, which the child inherits and overrides. But now the behavior is starting to make sense 👍

tests/test_black.py
```diff
@@ -1303,6 +1303,28 @@ def test_read_pyproject_toml(self) -> None:
...
+        self.assertEqual(config["target_version"], ["py36", "py37", "py38"])
+        self.assertEqual(config["exclude"], r"\.pyi?$")
+        self.assertEqual(config["include"], r"\.py?$")
+
+    def test_invalid_black_config(self) -> None:
+        test_toml_file = THIS_DIR / "invalid_test.toml"
+        fake_ctx = FakeContext()
+        with self.assertRaises(click.exceptions.FileError):
+            black.read_pyproject_toml(fake_ctx, FakeParameter(), str(test_toml_file))
+
```
JelleZijlstra, line 1327: Could you add a unit test where two toml files mutually include each other? Unless I'm misreading that would currently cause infinite recursion; in any case it should be tested.

docs/usage_and_configuration/the_basics.md
```diff
@@ -200,6 +200,37 @@ project.
 "No". _Black_ is all about sensible defaults. Applying those defaults will have your
 code in compliance with many other _Black_ formatted projects.
 
+````{note}
+If you are using the linked configuration feature i.e. specifying `config` in the black config,
```
JelleZijlstra, line 204: You should first define the "linked configuration feature". Suggested wording:

```
You can use the `config = ` field to include configuration options from another configuration file.
Values set in the original configuration file have precedence over those in the included file.

Suppose `pyproject.toml` contains:
... (example below)
And `linked_config.toml` contains:
... (example below)
```

Also, no need for the `{note}`.

docs/usage_and_configuration/the_basics.md
JelleZijlstra: I think we should avoid talking about the "parent" config because different people understand it to be either the first or the second config.

JelleZijlstra (changes requested):

JelleZijlstra:
This has a merge conflict that I can't resolve in the UI.

JelleZijlstra (approved):

src/black/__init__.py
```diff
@@ -127,28 +127,44 @@ def read_pyproject_toml(
 
     if not config:
         return None
+
+    black_config = config.get("config") or (
+        ctx.default_map.get("config") if ctx.default_map else None
+    )
+    if black_config:
+        black_config_path = Path(Path(value).parent, black_config).resolve()
```
felix-hilden, line 135:
```suggestion
        black_config_path = (Path(value).parent / black_config).resolve()
```

src/black/__init__.py
```diff
@@ -127,28 +127,44 @@ def read_pyproject_toml(
 
     if not config:
         return None
+
+    black_config = config.get("config") or (
```
felix-hilden, line 131:
```suggestion
    inner_config = config.get("config") or (
```
The name could be improved. At least to me it sounds like "Black's section in the config file". So `inner`, `recursive`, ...?

src/black/__init__.py
Shivansh-007: Any reason you prefer that over the current method?

src/black/__init__.py
Shivansh-007: I am not sure if I understand what you mean by 

> So `inner`, `recursive`, ...?

Otherwise, I am okay with using `inner_config` 👍🏻

src/black/__init__.py
felix-hilden: Just that I had to have another look to immediately understand what was happening. It's a bit faster (at least for me) to read from left to right: "value's parent / config and resolve that path" rather than "value's parent and config in some call, hold on what was it, oh new Path". But no biggie!

src/black/__init__.py
felix-hilden: Name suggestions: `inner_config`, `recursive_config` etc. Inner is fine by me as well!

felix-hilden (approved):

docs/usage_and_configuration/the_basics.md
```diff
@@ -200,6 +200,40 @@ project.
 "No". _Black_ is all about sensible defaults. Applying those defaults will have your
 code in compliance with many other _Black_ formatted projects.
 
+### Linked Configuration feature
```
ichard26, line 203:
```suggestion
### Linked Configuration
```

docs/usage_and_configuration/the_basics.md
```diff
@@ -200,6 +200,40 @@ project.
...
+those in the included file.
+
+Suppose `pyproject.toml` contains:
+
+```{code-block} toml
+---
+lineno-start: 1
+emphasize-lines: 2,3
+---
+[tools.black]
```
ichard26, line 216:
```suggestion
[tool.black]
```
It's `[tool.black]` not `[tools.black]` https://www.python.org/dev/peps/pep-0518/#tool-table

docs/usage_and_configuration/the_basics.md
```diff
@@ -200,6 +200,40 @@ project.
...
+```
+
+And `linked_config.toml` contains:
+
+```{code-block} toml
+---
+lineno-start: 1
+emphasize-lines: 2
+---
+[tools.black]
```
ichard26, line 228:
```suggestion
[tool.black]
```
It's `[tool.black]` not `[tools.black]` https://www.python.org/dev/peps/pep-0518/#tool-table

docs/usage_and_configuration/the_basics.md
```diff
@@ -200,6 +200,40 @@ project.
 "No". _Black_ is all about sensible defaults. Applying those defaults will have your
 code in compliance with many other _Black_ formatted projects.
 
+### Linked Configuration feature
+
+You can use the `config = ` field to include configuration options from another
+configuration file. Values set in the original configuration file have precedence over
+those in the included file.
```
ichard26, line 207: Please explicitly note the path is relative to the linking configuration file.

CHANGES.md
```diff
@@ -58,6 +58,7 @@ and the first release covered by our new stability policy.
 - Enable Python 3.10+ by default, without any extra need to specify
   `--target-version=py310`. (#2758)
 - Make passing `SRC` or `--code` mandatory and mutually exclusive (#2804)
+- Allow specifying `config` in config files (#2525)
```
ichard26, line 61:
```suggestion
- Allow specifying `config` in config files to include configuration options (#2525)
```

ichard26 (changes requested):
I did not do a hands-on test as I'm far too annoyed right now to do so but here's some initial feedback. We also need to check the pypy coverage drop.

JelleZijlstra:
This has a merge conflict without an obvious fix, could you take a look?

JelleZijlstra:
@ichard26 are you happy with this PR now?

danielmitchell:
This PR appears to have stalled? It would be great if it could progress since there are a couple of related issues (https://github.com/psf/black/issues/1826, https://github.com/psf/black/issues/2863) it will resolve and make black much easier to configure in a monorepo.

ichard26:
Hey all, I'll try to review this soon (especially since IIRC I had some outstanding comments I never shared the last time I looked at this), but I can't make promises.

jaklan:
@ichard26 any chance to push it forward?

AbdealiLoKo:
I was looking for the usecase described in https://github.com/psf/black/issues/1826 and came across this PR
Would be great if this can be released!

If I can help with something to push this forward, I'm happy to help

codybak4 (approved):

alfawal:
@ichard26 Just a reminder to take a look please.

hauntsaninja:
See also https://github.com/psf/black/issues/1826#issuecomment-1922891622

cobaltt7:

PR #2526, opened 2021-10-07
Enhance `--verbose`

### LINKED ISSUE #1880: `--verbose` mode should print location of project root

Black 20.8b1 does not respect skip-string-normalization = true in the pyproject.toml

1. pyproject.toml
```
[tool.black]
target-version = ['py36']
line-length = 120
exclude = '(.*/migrations/.*\.py)'
include = '.*/server/.*\.py$'
```

2. create a python file with single strings

```
example = 'This is an example'
```

3. run `black <filename> --check --diff` from same directory as pyproject.toml (which is root repo dir)

No changes would be applied by black given our pyproject.toml. Note that if you pass `--config` to that same command it works. 

- Version: 20.8b1
- OS and Python version: macOS 10.15.7, Python 3.7.2 (but target version is 3.6)

Yes with:

> 2. Or run _Black_ on your machine:
>    - create a new virtualenv (make sure it's the same Python version);
>    - clone this repository;
>    - run `pip install -e .`;
>    - make sure it's sane by running `python -m unittest`; and
>    - run `black` like you did last time.
> 

We haven't normalized our strings yet because it's a ~30k line change (we plan to in the coming months). Noting that it seems to be a general "not respecting pyproject.toml" issue but string normalization was the most obvious thing.

Shivansh-007:
### Description

Black would now echo the location that it determined as the root path for the project if `--verbose` is enabled by the user, according to which it chooses the SRC paths, i.e. the absolute path of the project: `{root}/{src}`.

Screenshot from 2021-10-07 13-40-59

Closes #1880

### Checklist - did you ...


ichard26:
OK yeah, not knowing what project root makes my suggestion a confusing in its own way :disappointed: .

> [...] the sources to check are being taken from here

Technically true but I don't think it makes it much clearer especially since it can be confusing when the SRCs are partially distanced from the root. Let's take this example that would happen if no root was found where we fallback to the filesystem root.

```console
~/programming/oss/black/docs on branch feat/project-root-verbose [$?] via Python v3.8.5 (black) 
$ black _build/html/ --check  --verbose
Checking for `_build/html/` in /
```

This isn't clear it's talking about the project root, and it's unclear what "checking for `_build/html` in ..." means here. The user could easily misunderstand this as checking the whole directory for directory structures that match `_build/html` which isn't right. To be fair this is also a mark against my suggestion since `Identified / as project root` is also invalid (usually as chances are we couldn't identify one).

Shivansh-007:
> To be fair, I kinda like the colours. Hurts the focus of this PR but might not be that big of a deal smile But we need to discuss more (or just get a go-ahead from other maintainers) about the content of the messages. And btw, it seems odd to first declare that config from root is used and then that the root was found.

Should I use the path of the configuration then?

src/black/__init__.py
```diff
@@ -501,10 +507,23 @@ def get_sources(
     force_exclude: Optional[Pattern[str]],
     report: "Report",
     stdin_filename: Optional[str],
-) -> Set[Path]:
+) -> Tuple[Set[Path], Path]:
     """Compute the set of files to be formatted."""
```
JelleZijlstra, line 511: Let's document here what the second return value means.

src/black/files.py
```diff
@@ -30,7 +30,7 @@
 
 
 @lru_cache()
-def find_project_root(srcs: Sequence[str]) -> Path:
+def find_project_root(srcs: Sequence[str]) -> Tuple[Path, Optional[str]]:
     """Return a directory containing .git, .hg, or pyproject.toml.
```
JelleZijlstra, line 34: Same here, docstring should say what the rest of the return value is.

tests/test_black.py
```diff
@@ -1764,7 +1776,7 @@ def test_gitignore_used_as_default(self) -> None:
         src = [base / "b/"]
         assert_collected_sources(src, expected, extend_exclude=r"/exclude/")
 
-    @patch("black.find_project_root", lambda *args: THIS_DIR.resolve())
+    @patch("black.find_project_root", lambda *args: (THIS_DIR.resolve(), "None"))
```
JelleZijlstra, line 1779:
```suggestion
    @patch("black.find_project_root", lambda *args: (THIS_DIR.resolve(), None))
```
This repeats a bunch of times.

src/black/__init__.py
```diff
@@ -501,10 +507,23 @@ def get_sources(
...
+    root, method = find_project_root(src)
+
+    if verbose:
+        if method:
+            out(
+                f"Identified `{root}` as project root containing a '{method}'.",
```
JelleZijlstra, line 517:
```suggestion
                f"Identified `{root}` as project root containing a {method}.",
```
Looks better than the current output:
```
Identified `/Users/jelle/py/black` as project root containing a '.git directory'.
```

JelleZijlstra (changes requested):
A couple of points. We also need to resolve the other issues brought up above.

src/black/report.py
```diff
@@ -10,6 +13,12 @@
 from black.output import out, err
 
 
+def root_relative(path: Path, root: Optional[Path]) -> Path:
+    if not root:
+        return path
+    return Path(os.path.relpath(path, start=Path(f"{root}/../")))
```
felix-hilden, line 19: `pathlib.Path` does have `p.relative_to`. But this works fine too 👍

felix-hilden (approved):
Overall LGTM, sucks that we have to pass `root` around so much, but I guess there's no cleaner way either because having relative paths in the first place is problematic as well.

src/black/report.py
felix-hilden:
```suggestion
    return path.relative_to(root.parent)
```
would be the form I guess. So why the parent?

felix-hilden: Oh I guess we want the root folder to be included?

Shivansh-007: Yeah

felix-hilden:
As discussed on Discord, Jason thought we could use the click context to pass root around more easily, and include it in the report object so that it doesn't have to be in every method call. I think they could be nice simplifications. Thoughts?

src/black/report.py
felix-hilden: The pathlib version is much cleaner to read as well, so maybe we should use it instead.

ichard26 (commented):
[Feedback given on Discord](https://discord.com/channels/267624335836053506/846434317021741086/920484521714016256).

src/black/__init__.py
```diff
@@ -515,8 +520,21 @@ def get_sources(
...
+            out(
+                f"Identified `{root}` as project root containing a {method}.",
+                fg="blue",
+            )
+        else:
+            out(f"Identified `{root}` as project root.", fg="blue")
+        paths = '", "'.join(
+            str(Path(source).resolve().relative_to(root)) for source in src
+        )
+        out(f'Sources to be formatted: "{paths}"', fg="blue")
```
ichard26, line 536:
```suggestion
        normalized_paths = [Path(source).resolve().relative_to(root) for source in src]
        srcs_string = '", "'.join(normalized_paths)
        out(f'Sources to be formatted: "{srcs_string}"', fg="blue")
```
This also makes it harder for future developers to mistake `paths` for `src`.

src/black/files.py
```diff
@@ -31,14 +31,19 @@
...
     That directory will be a common parent of all files and directories
     passed in `srcs`.
 
     If no directory in the tree contains a marker that would specify it's the
     project root, the root of the file system is returned.
+
+    Returns a tuple with the first element as the project root path and the
+    second element as a string which describes the method by which black
+    found out the project root (None, if `None` of the registered method
+    was used).
```
ichard26, line 46:
```suggestion
    Returns a two-tuple with the first element as the project root path and
    the second element as a string describing the method by which the
    project root was discovered (it will be None the fallback root was used).
```

CHANGES.md
```diff
@@ -7,6 +7,7 @@
 - Improve error message for invalid regular expression (#2678)
 - Fix mapping cases that contain as-expressions, like `case {"key": 1 | 2 as password}`
   (#2686)
+- `--verbose` enhancements (project root, how is config found) (#2526)
```
ichard26, line 10: Perhaps instead:

> Verbose mode also now describes how a project root was discovered and which paths will be formatted. (#2526) 

This feels a bit more concrete than what we have right now but then again this might be too technical?

src/black/__init__.py
```diff
@@ -515,8 +520,21 @@ def get_sources(
...
+        if method:
+            out(
+                f"Identified `{root}` as project root containing a {method}.",
+                fg="blue",
+            )
+        else:
+            out(f"Identified `{root}` as project root.", fg="blue")
+        paths = '", "'.join(
+            str(Path(source).resolve().relative_to(root)) for source in src
+        )
```
ichard26, line 535: Would it be worth it to guard this with a try... except? When we normalize path before applying the file discovery regexes (using `normalize_path_maybe_ignore`) we have be able to handle permission or out of tree exceptions gracefully and this code does none of that. Not sure how this would look but I'd prefer a defensive approach to our filepath handing code.

ichard26 (commented):
I like it! This is getting closer to merge!

image

This does look a little bit weird, now I see why you originally chose to log the SRCs and the discovered project root in the same line as the project root would provide the necessary context. I don't know if this is worth fixing but I wanted to make sure you're aware.

Finally, I don't know how I didn't realize this but pleeeeasee add some tests for this functionality as there's a surprising amount of ways handling filepaths can blow up. Make sure to include cases with symlinks (you're gonna have to skip these tests on windows) and out of project root paths (to exercise the edge cases protected against in `normalize_path_maybe_ignore`).

Thank you!

JelleZijlstra (approved):

felix-hilden (changes requested):
Some comments still! Did you have a look at the test Richard requested? Or maybe they are the old tests you modified and I'm just reading things wrong. Anyways, thanks again for the changes and let's give it a push once more 😄 <sub>I can't believe this conversation is already like 30 comments long, seems like no change is small in a project this big :D thanks for holding up</sub>

felix-hilden (approved):
If nothing else comes up and the stars align... 🎉

JelleZijlstra (approved):

PR #2527, opened 2021-10-07
feat(config): Add support for path specific config

### LINKED ISSUE #1370: Override line-length for subdirectories
Currently, black allows changing the line length globally via `pyproject.toml` or with the `--line-length` argument. This setting is global for the whole directory tree.   

In my project, I have a `docs/examples` directory, which contains python files, that are shown in the documentation by using the `literalinclude` directive. While for most of the project the global line length setting (100) is fine, for the examples it's too wide and requires the horizontal scroll bar to display the code samples.

I've tried to add another `pyproject.toml` file in the `docs/examples` directory, but it is respected only when I'll execute black from that directory. It's hard to use it from pre-commit hooks.  

Is there any way to override the line length for subdirectories?

Shivansh-007:
Lol, I am sorry for spamming 3 PRs. Had completed them before vacation and didn't get time to open them until now :)

### Description

Woohoo! Module specific configuration! and a bit of restructuring of code in the `black/__init__.py` file. Mainly shifting all the "reformatting" logic to a separate function to increase the readability of code in the click CLI  function. You can go through the code in the commit, code should be self explanatory, though I still need to update the documentation of the code and this feature.

I haven't aimed at bringing new tests in this commit, if required, I could do that and this rather fixes the tests to work according to the change I have done in this commit.

Summarising the changes:
- Now, you can support file/module specific configuration in the pyproject.toml config under the section of `[tools.black-module]`. I have added a separate "key" for the modules as according to the current code when the module name could clash with a config argument. For example `[tools.black.src]` would clash with `SRC`.

- Module specific configuration can only some arguments which are listed under a global constant in `config.py` and should agree with the type hint present on them in the class variables, the logic for this functionality can be seen in `cast_typehint` function of `config.py`. At some places I have ignored MYPY errors so as I couldn't change the configuration since they were being used for typing.

- Added a new class for storing all the parameters, the order of the arguments is as follows:
  * Read global pyproject.toml/specified config config.
  * Read module specific config, which would take the none specified arguments from the global configuration.
  * If value specified through click options/arguments it would be overwritten, this won't happen if the default value of the option is taken.
  * Tada!

Closes #1370

### Checklist - did you ...

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -98,12 +98,11 @@
         },
         "dataclasses": {
             "hashes": [
-                "sha256:0201d89fa866f68c8ebd9d08ee6ff50c0b255f8ec63a71c16fda7af82bb887bf",
-                "sha256:8479067f342acf957dc82ec415d355ab5edb7e7646b90dc6e2fd1d96ad084c97"
+                "sha256:454a69d788c7fda44efd71e259be79577822f5e3f53f029a22d08004e951dc9f",
+                "sha256:6988bd2b895eef432d562370bb707d540f32f7360ab13da45340101bc2307d84"
             ],
-            "index": "pypi",
             "markers": "python_version < '3.7'",
-            "version": "==0.8"
+            "version": "==0.6"
         },
         "idna": {
             "hashes": [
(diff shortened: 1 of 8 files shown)


ichard26:
FWIW, it would've been a good idea to get sign off on the original idea before implementing this as this has some ethos design considerations which could easily block it. And for the next submission I'd encourage putting refactor changes in their own preceding commits so I don't have to spend too much time discerning what changes are for what.

Also, what's up with the Pipfile.lock changes? They don't very correct to me as many dependencies needed on 3.6/7 seem to have been dropped (you can blame pipenv's strong dependence on the runtime version during locking for that) but I also don't see a reason why Pipfile.lock needs to be touched here. If it's a routine upgrade you're trying to do, I'd suggest moving them to a new PR (although honestly I'd suggest not trying to touch pipenv's lockfiles as it's a shockingly annoying thing to use).

Thanks for the PR though!

Shivansh-007:
Right, I was on an earlier commit while working on this and the `pipenv install` was failing, showing  `Pipfile.lock` is out of date and some deps were failing to be installed, don't remember the error. So I had to relock the Pipfile. It has been fixed in one of the newer commits, so I will remove that section of the PR.

I do keep my commits atomic for most of my PRs, it's just I saw squashed commits being merged to main. On a few PRs originally people were force pushing their changes to a single commit after addressing the reviews, so I thought it was some sort of a policy to keep the changes squashed.

Shivansh-007:
Alright, will remember that!

felix-hilden:
Yeah, I guess in hindsight it would have been better to discuss further on the issue since it's so old 😅

JelleZijlstra:
This has a bunch of merge conflicts now, and there's still no consensus yet on whether we should even do this (but that should be discussed on the issue). Are you still interested in moving this forward?

Shivansh-007:
Yeah, I have been waiting for the design approval, whether we should go with what Felix proposed or the current solution.

making787:
You

JelleZijlstra:
Given the discussion on the issue, I'm going to close this for now. Thanks for your contribution!

PR #2529, opened 2021-10-07
Define a stability policy

### LINKED ISSUE #517: Non-prerelease black release

### LINKED ISSUE #2394: Define a stability policy
We're often hesitant to change formatting now, even though technically we're still in beta (#517). Some questions we should document the answer to:

- When does Black accept formatting changes? How is the decision to change formatting made?
- What are the guarantees Black wants to continue making? When will we make exceptions to the AST safety guarantee?
- As a user, what should I expect when upgrading to a new version of Black?
- How will this change when we have a "stable" release? Will anything change?

The answers to these questions should be in the documentation. Feel free to respond to this issue with other similar questions.

JelleZijlstra:
Fixes #2394. Eventually fixes #517.

This is essentially @pradyunsg's suggestion from #2394. I suggest that at the
same time we start the formal stability policy, we take a few other disruptive steps
and drop Python 2 and the "b" marker.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -31,9 +31,9 @@ pragmatism. However, _Black_ is still in beta so style changes are both planned 
 still proposed on the issue tracker. See
 [The Black Code Style](the_black_code_style/index.rst) for more details.
 
-Starting in 2022, the formatting output will be stable for the releases made in the same year
-(other than unintentional bugs). It is possible to opt-in to the latest formatting styles,
-using the `--future` flag.
+Starting in 2022, the formatting output will be stable for the releases made in the same
+year (other than unintentional bugs). It is possible to opt-in to the latest formatting
+styles, using the `--future` flag.
 
 ## Why is my file not formatted?
 


maxtrem271991:
69375bab5add9c945ba7c7800813d5e9d1c703bf

ichard26:
> re. Python 3.10 rewrite

We're in a slightly better state as the great work by @isidentical means Black supports 3.10 pretty well at this point. We still have the fact the parser is still LL(1) and now has a somewhat permissive grammar meaning we'd have to rewrite the project eventually if Python gets even more new syntax in future releases. This gives us more time and less pressure but then again the whole problem is the complete lack of resources as per usual in FOSS :/

It's probably worth fleshing out a basic transition plan after we finally get Black stable, but to be honest I have no idea where we'd start.

PR #2532, opened 2021-10-12
Fix feature detection for positional-only arguments in lambdas

Zac-HD:
### Description

While working on `shed`, I discovered that feature detection for positional-only arguments doesn't work in lambdas, though the docstring of `get_features_used()` suggests that it should.  I've therefore added `syms.varargslist` to the check, and a test case which demonstrates that it works.

### Checklist

- [x] Added a CHANGELOG entry
- [x] Added test cases
- [x] This change does not need documentation

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 ### _Black_
 
 - Add new `--workers` parameter (#2514)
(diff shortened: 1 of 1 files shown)


BotchJet-Pony:
Shed? It's not black anymore?

PR #2557, opened 2021-10-22
Remove some unneeded exceptions from mypy.ini

nipunn1313:
### Description

I removed these and found that mypy still passed (particularly since it's hard coded to target python 3.6 currently)

### Checklist - did you ...
- [x] Add a CHANGELOG entry if necessary? (not needed)
- [x] Add / update tests if necessary? (not needed)
- [x] Add new / update outdated documentation? (not needed)

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -32,14 +32,6 @@ check_untyped_defs=True
 # No incremental mode
 cache_dir=/dev/null
 
-[mypy-aiohttp.*]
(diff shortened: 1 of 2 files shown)


ichard26:
Huh, I wonder what's different this time as Łukasz had a hell of a time trying to silence mypy for _unicodefun when we first added support for Click 8+.

Issue #727, opened 2019-02-28
Support PyPy (easy)


deeplook:
Please consider adding PyPy Trove classifiers, because apparently `black` works fine! At least on the recent PyPy (I tested PyPy3.6 v7.0 on macOS). I used version 18.9b0 of `black`. See https://pypy.org/download.html

zsol:
We should also add validation that this works as part of our CI pipeline.
Happy to review a pull request :)

pauloalem:
I was going to give this a shot but it seems travis does not yet have pypy 3.6 available.

hugovk:
Azure Pipelines are planning on adding PyPy: https://github.com/Microsoft/azure-pipelines-tasks/issues/8514.

(See https://github.com/ambv/black/pull/731 for adding Azure Pipelines with CPython to this repo.)

ambv:
I was actually targeting my PyPy sprint contributions in Dusseldorf with this in mind so I'm happy it works for you :)

To be fully frank, not everything works on PyPy yet, the asyncio parts are failing in particular.

alexchamberlain:
Just as an FYI: I tried configuring this on Travis (see https://github.com/alexchamberlain/black/blob/pypy3/.travis.yml#L36). `typed-ast` and `aiohttp` appear to be incompatible with `pypy3` if I've configured it correctly.

zsol:
if you drop `pip install -e '.[d]'` it will no longer pull in `aiohttp` (but `blackd` will not be tested). `typed-ast` is going away with python 3.8, so we could in theory strip away that dependency from that version on

paugier:
An update with pypy3.6 v7.2.0:

-  `aiohttp` can be installed without any problem.

- `black==19.3b0` works fine

- `black==19.10b0` can't be installed because of the `typed-ast` dependency

- `typed-ast` won't be compatible with PyPy for awhile (https://github.com/python/typed_ast/issues/111).

Wouldn't it be possible to install black without typed-ast on PyPy ?

With `typed-ast` has a hard dependency, it seems it won't be possible to install black until PyPy reaches 3.8 !

paugier:
For the record, the PR which introduced the `typed-ast` dependency: https://github.com/psf/black/pull/840 "Make --safe work for Python2.7 syntax, by using typed_ast".

ambv:
I **want** to support PyPy but this is going to be a tough sell.

### PyPy3 is 3.6
[...]

PR #2559, opened 2021-10-22
Allow install under pypy

olliemath:
### Description

This adds a minimal set of changes to use black under pypy. It's supposed to be a more acceptable version of https://github.com/psf/black/pull/1172 with the key point being there are zero changes for cpython users. The hope is that this will enough to close https://github.com/psf/black/issues/727

This is a draft - feedback appreciated - I plan to do all of the following (if needed):

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,6 @@
 - Add new `--workers` parameter (#2514)
 - Fixed feature detection for positional-only arguments in lambdas (#2532)
 - Bumped typed-ast version minimum to 1.4.3 for 3.10 compatiblity (#2519)
-- Add primer support for --projects (#2555)
 - Add experimental pypy support (#2559)
 
 ### _Blackd_
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 - Warn about Python 2 deprecation in more cases by improving Python 2 only syntax
   detection (#2592)
+- Add experimental pypy support (#2559)
 
 ## 21.10b0
 
@@ -18,7 +19,6 @@
 - Fixed a Python 3.10 compatibility issue where the loop argument was still being passed
   even though it has been removed (#2580)
 - Deprecate Python 2 formatting support (#2523)
-- Add experimental pypy support (#2559)
 
 ### _Blackd_
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@
 
 - Warn about Python 2 deprecation in more cases by improving Python 2 only syntax
   detection (#2592)
-- Add experimental pypy support (#2559)
+- Add experimental PyPy support (#2559)
 - Add partial support for the match statement. As it's experimental, it's only enabled
   when `--target-version py310` is explicitly specified (#2586)
 - Add support for parenthesized with (#2586)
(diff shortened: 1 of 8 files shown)


docs/faq.md
```diff
@@ -88,3 +88,9 @@ influence their behavior. While Black does its best to recognize such comments a
 them in the right place, this detection is not and cannot be perfect. Therefore, you'll
 sometimes have to manually move these comments to the right place after you format your
 codebase with _Black_.
+
+## Can I run black with PyPy?
+
+Yes, there is _experimental_ support for PyPy 3.7 and higher. You cannot format Python 2
```
JelleZijlstra, line 94: Is there any reason other than Python 2 support to call this experimental?

For what it's worth we're about to deprecate Python 2 support anyway.

docs/faq.md
```diff
@@ -88,3 +88,9 @@ influence their behavior. While Black does its best to recognize such comments a
...
+files under PyPy, because PyPy's inbuilt ast module does not support this. For the most
+stable user-experience, we recommend using CPython if possible.
```
JelleZijlstra, line 96:
```suggestion
stable user experience, we recommend using CPython if possible.
```
I'm also not too happy with this sentence. If we don't have concrete reasons to say PyPy is unstable, it feels like FUD.

JelleZijlstra (changes requested):
Thanks! I'm happy to add support for pypy, but have a few comments.

docs/faq.md
olliemath: My main thought on is that while it passes all of the tests and I've used it locally on files, there might still be edge cases where it fails while cpython doesn't.

I don't think we'll discover those edge cases without a lot more people trying it out. But yes, maybe that's not enough to say it's "experimental".

docs/faq.md
olliemath: I'll remove this sentence - if you're confident in your test suite then happy to just go with "it works" 🙂

docs/faq.md
olliemath: OK, I've removed these bits from the faq.

olliemath:
Hi @JelleZijlstra thanks for the review. I tightened up the FAQ based on your comments (although I couldn't apply your suggestion, due to the file being changed).

I'm not sure that I understand the CI failures - something to do with the hypothesis lib not being formatted correctly? I can't see how that would relate to my changes.

Anyhow, let me know if further changes are needed.

JelleZijlstra (approved):
The CI failure seems unrelated.

ichard26 (approved):
Given that we'll be dropping Python 2 support in 2-3 months I don't think we'll need to do much in terms of letting folks know PyPy doesn't support Python 2. Perhaps we should notice when our built-in parser (which has excellent cross-version syntax support; ignoring parenthesized with and pattern matching -> #2586) falls back to the Python 2 grammar and emit a friendly error but we can deal with that later as it isn't strictly related to PyPy compatibility.

ichard26:
> PyPy 3.8 works fine too - but it's still in beta - so I've parked it over here olliemath#3 pending a stable release

Hmm, to be fair usually it's good to test out development versions of Python implementations so we can root out issues early, but I don't have any experience with PyPy so I'll delegate the decision to someone who knows better ^^

JelleZijlstra (approved):

CHANGES.md
```diff
@@ -6,6 +6,7 @@
 
 - Warn about Python 2 deprecation in more cases by improving Python 2 only syntax
   detection (#2592)
+- Add experimental pypy support (#2559)
```
JelleZijlstra, line 9:
```suggestion
- Add experimental PyPy support (#2559)
```

PR #2562, opened 2021-10-22
Declare support for Python 3.10

hugovk:
### Description

Python 3.10 final is already tested, just needs the Trove classifier.

### Checklist - did you ...

- [n/a] Add / update tests if necessary?
- [n/a] Add new / update outdated documentation?

Is a changelog entry needed?

---

By the way, Black was first released in March 2018 and took the forward-thinking view to only support Python 3.6, which was the newest release, and therefore Black hasn't needed to drop any EOL versions yet.

Python 3.6 is EOL on 2021-12-23; will Black be dropping 3.6 before the (potentially stable) 2022 Q1 release?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 - Add partial support for the match statement. As it's experimental, it's only enabled
   when `--target-version py310` is explicitly specified (#2586)
 - Add support for parenthesized with (#2586)
+- Declare support for Python 3.10 (#2562)
 
 ## 21.10b0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@
 - Add partial support for the match statement. As it's experimental, it's only enabled
   when `--target-version py310` is explicitly specified (#2586)
 - Add support for parenthesized with (#2586)
-- Declare support for Python 3.10 (#2562)
+- Declare support for Python 3.10 for running Black (#2562)
 
 ## 21.10b0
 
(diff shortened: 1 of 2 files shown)


kukushking:
Hi folks, not quite sure this is the right place but shouldn't 3.10 be added to supported [target versions](https://github.com/psf/black/blob/b21c0c3d28d87bc944a1fdc979b30a0707b0df89/src/black/mode.py#L22) as part of this PR, too?

jack1142:
Classifier just means that Black can run on Python 3.10, not that it can parse Python 3.10 code (which it currently can't when it comes to new syntax like match statement and parenthesized with) which is what would constitute to adding 3.10 target version.

kukushking:
Yep, I understand - just have no idea how are you handling 3.10 story so thought I'd mention it here in case you missed it.

kukushking:
Just found a [PR](https://github.com/psf/black/pull/2586) that has that covered. Thanks guys 🙌

JelleZijlstra (changes requested):
Let's add this to the changelog

hugovk:
> Let's add this to the changelog

Added!

CHANGES.md
```diff
@@ -10,6 +10,7 @@
 - Add partial support for the match statement. As it's experimental, it's only enabled
   when `--target-version py310` is explicitly specified (#2586)
 - Add support for parenthesized with (#2586)
+- Declare support for Python 3.10 (#2562)
```
```suggestion
- Declare support for Python 3.10 for running Black (#2562)
```

PR #2573, opened 2021-10-29
Remove ignores in patch_click

nipunn1313:
May help unblock #2431

### Checklist - did you ...

(none of these necessary)


ichard26:
While I'm happy we have multiple workarounds for this messy issue I'll be going with the old workaround we've been using successfully. This cast hides the fact that this code doesn't type check right while the pre-existing one is more explicit that it's a temporary workaround.

thanks for investigating!

PR #2574, opened 2021-10-29
fix: allow tests to be run from the tests/ directory

dawnofmidnight:
### Description

Fixes the issues described [on Discord](https://canary.discord.com/channels/267624335836053506/846434317021741086/903457339976351744).

At the moment, tests break when they're run from a directory that is not the project root. Specifically, `tests/`. This makes changes to use the `PROJECT_ROOT`, `THIS_DIR`, and `DATA_DIR` constants defined in `tests.util` so that tests will be runnable from multiple locations.

### Checklist - did you ...

- [x] Add a CHANGELOG entry if necessary? (not necessary)
- [x] Add new / update outdated documentation? (not necessary afaik)

### DIFF
--- a/tests/test_ipynb.py
+++ b/tests/test_ipynb.py
@@ -1,4 +1,3 @@
-import pathlib
 from click.testing import CliRunner
 from black.handle_ipynb_magics import jupyter_dependencies_are_installed
 from black import (
@@ -8,11 +7,11 @@ from black import (
     format_file_contents,
     format_file_in_place,
 )
-import os
 import pytest
 from black import Mode
 from _pytest.monkeypatch import MonkeyPatch
 from py.path import local
+from tests.util import DATA_DIR
 
 pytestmark = pytest.mark.jupyter
 pytest.importorskip("IPython", reason="IPython is an optional dependency")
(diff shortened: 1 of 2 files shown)


tests/test_ipynb.py
```diff
@@ -342,23 +333,17 @@ def test_empty_string() -> None:
 
 
 def test_unparseable_notebook() -> None:
-    msg = (
-        r"File 'tests[/\\]data[/\\]notebook_which_cant_be_parsed\.ipynb' "
-        r"cannot be parsed as valid Jupyter notebook\."
-    )
+    path = DATA_DIR / "notebook_which_cant_be_parsed.ipynb"
+    msg = rf"File '{path}' cannot be parsed as valid Jupyter notebook\."
```
MarcoGorelli, line 337: this didn't pass when I tried running the CI checks in a fork, I had to use
```
'{re.escape(str(path))}'
```
instead of
```
'{path}'
```

https://github.com/MarcoGorelli/black/pull/1

ichard26 (approved):
Thank you so much for your contribution! You're awesome, @dawnofmidnight. (also that cat image is soo cute :3)

PR #2575, opened 2021-10-29
Use STDIN project in test_projects to ensure it runs quickly

nipunn1313:
Fixes @ichard's comment in https://github.com/psf/black/pull/2555#issuecomment-954284259
Runs in 0.4 seconds rather than 8 seconds previously (on my machine).

### Checklist - did you ...

(none needed)

### DIFF
--- a/tests/test_primer.py
+++ b/tests/test_primer.py
@@ -9,6 +9,7 @@ from io import StringIO
 from os import getpid
 from pathlib import Path
 from platform import system
+from pytest import LogCaptureFixture
 from subprocess import CalledProcessError
 from tempfile import TemporaryDirectory, gettempdir
 from typing import Any, Callable, Iterator, List, Tuple, TypeVar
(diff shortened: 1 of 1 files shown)


ichard26 (approved):
Still surprisingly slow at ~1.2s on my machine but this is good enough (and won't slow build + test cycles by any reasonable margin). Thank you for the quick turnaround!

FWIW while it's great you're working on black-primer, I'm currently writing a semi-replacement for the development tool: https://github.com/ichard26/diff-shades. Fundamentally it takes the idea of black-primer and extends it with whatever great ideas mypy-primer added (theirs was inspired by ours), notably diffing support between runs. It'll take a long time before diff-shades is at feature and quality parity, but I thought I'd let you know. 

Basically for now if there's any ideas you have for primer but that seem too complicated to implement in primer I'd be happy to investigate if it'd be a good fit for diff-shades.

cooperlees (approved):
Speed is good. I did have to take a few passes tho to see how this worked. Like @ichard26 I was amazed that pytest log magic and click's magic get along.

nipunn1313:
> Speed is good. I did have to take a few passes tho to see how this worked. Like @ichard26 I was amazed that pytest log magic and click's magic get along.

I was too. I actually couldn't get it to work any other way. `click.stderr` and `click.output` didn't have the relevant information. I was stumped momentarily until I figured out that pytest was eating the log lines.

> Basically for now if there's any ideas you have for primer but that seem too complicated to implement in primer I'd be happy to investigate if it'd be a good fit for diff-shades.

Good to know! Thank you. That would be really valuable! I've used mypy-primer and the diff mode is fantastic. If implemented in diff-shades then you can get valuable feedback running it against large projects that don't fully use black yet (eg django!), especially when trying to implement a stability policy.

PR #2577, opened 2021-10-30
Add Tesla to organizations list

rtdev-com:
### Checklist - did you ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Twisted, LocalStack, every Datadog Agent Integration, Home Assistant, Zulip, Ked
 many more.
 
 The following organizations use _Black_: Facebook, Dropbox, Mozilla, Quora, Duolingo,
-QuantumBlack.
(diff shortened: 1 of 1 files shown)


ichard26 (approved):
The more I review these sort of contributions the more I realize that we don't really have a good way to verify these claims are true. Seems unlikely that people would lie but it's always possible.

Anyway thank you!

rtdev-com:
> 
> 
> The more I review these sort of contributions the more I realize that we don't really have a good way to verify these claims are true. Seems unlikely that people would lie but it's always possible.
> 
> Anyway thank you!

You could ask a contributor to send you a email from work. Not a great solution nor would it work with former employees. Just an idea though.

cooperlees:
Also some profiles give it away - e.g. if you look at https://github.com/cooperlees it's very difficult to see who I work for ... I've looked here in the past.

cooperlees (approved):

PR #2579, opened 2021-10-30
Pin regex in docker to 2021.10.8

cooperlees:
- This is due to 2021.10.8 having arm wheels and newer versions not

I will go see if I can help restore arm build @ https://bitbucket.org/mrabarnett/mrab-regex/issues/399/missing-wheel-for-macosx-and-the-new-m1 soon.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 ### Integrations
 
 - Allow to pass `target_version` in the vim plugin (#1319)
+- Pin regex module to 2021.10.8 as it has arm wheels available (#2579)
 
 ## 21.9b0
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -22,6 +22,7 @@
 ### Integrations
 
 - Allow to pass `target_version` in the vim plugin (#1319)
+- Pin regex module to 2021.10.8 as it has arm wheels available (#2579)
```
JelleZijlstra, line 25: Let's clarify that this is only for Docker

ichard26 (approved):
hopefully this works :crossed_fingers:

cooperlees:
> hopefully this works 🤞

It built locally on my M1 (arm) Mac and didn't before the chagnge - What are you worried about? What could have I tested more? I felt it's pretty evident from testing this workaround will allow it to build again.

ichard26:
I totally forgot you tested this locally :facepalm:

PR #2580, opened 2021-10-30
Address mypy errors on 3.10 w/ asyncio loop parameter

ichard26:
### Description

Somehow we didn't notice that the loop parameter was removed in the various asyncio high-level APIs in 3.10 :p

### Checklist - did you ...


JelleZijlstra (approved):
ship it

Issue #2568, opened 2021-10-27
Docker upload CI is broken


cooperlees:
Haven't debugged yet but seems a build level bug inside the container. Need to see if it can be reproduced locally.

```
error: failed to solve: process "/bin/sh -c pip install --no-cache-dir --upgrade pip setuptools wheel     && apt update && apt install -y git     && cd /src     && pip install --no-cache-dir .[colorama,d]     && rm -rf /src     && apt remove -y git     && apt autoremove -y     && rm -rf /var/lib/apt/lists/*" did not complete successfully: exit code: 1
Error: buildx failed with: error: failed to solve: process "/bin/sh -c pip install --no-cache-dir --upgrade pip setuptools wheel     && apt update && apt install -y git     && cd /src     && pip install --no-cache-dir .[colorama,d]     && rm -rf /src     && apt remove -y git     && apt autoremove -y     && rm -rf /var/lib/apt/lists/*" did not complete successfully: exit code: 1
```

Example fail: https://github.com/psf/black/actions/runs/1391464711

vbarbaresi:
I opened this issue on regex project: https://bitbucket.org/mrabarnett/mrab-regex/issues/434/aarch64-build-is-failing
I didn't find the Travis build logs last time, but now I finally found them: https://app.travis-ci.com/bitbucket/mrabarnett/mrab-regex/builds
Now that I see the logs I guess that it's probably a fluke in the build

I saw that you pinned `regex` and that it solved the issue for now.

I tried to build after installing `build-essential` and that was enough to solve the issue:

https://github.com/vbarbaresi/black/compare/7bf233a..bb86cd#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

https://github.com/vbarbaresi/black/runs/4061992219?check_suite_focus=true
Though it took ~12 min to build instead of 4 min, so I don't think it's a good solution.
It would be good as a workaround in case some dependency is missing wheels again

cooperlees:
Thanks for trying this. I think long term we should definitely support building from source but will probably need a multi stage build so we can keep the image size around the current ~50MB ... 

If you/anyone have cycles I'd totally take a PR making us able to build from source via a multi stage build to keep our image size low.

vbarbaresi:
I took a stab at it https://github.com/psf/black/pull/2582

PR #2582, opened 2021-10-31
 install build-essential to compile dependencies and use multi-stage build 

vbarbaresi:
- Install build-essential to avoid build issues like #2568 when dependencies don't have prebuilt wheels available
- Removed `regex` pinned version: not necessary as we can build from source if wheels are missing
- Use multi-stage build instead of trying to purge packages and cache from the image
  Copying `/root/.local/` installs only black's built Python dependencies (< 20 MB).
  So the image is barely larger than python:3-slim base image

Tested locally and in my fork GitHub Actions: https://github.com/vbarbaresi/black/runs/4062490765?check_suite_focus=true

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,8 +24,6 @@
 ### Integrations
 
 - Allow to pass `target_version` in the vim plugin (#1319)
-- Pin regex module to 2021.10.8 in our docker file as it has arm wheels available
-  (#2579)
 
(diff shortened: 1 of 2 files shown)


cooperlees (approved):
Awesome, thanks! Nice use of --user to get all the artifacts in an easy place to copy.

PR #2586, opened 2021-11-03
black/parser: partial support for pattern matching

isidentical:
Partial implementation for https://github.com/psf/black/issues/2242. Only works when explicitly stated `-t py310`.

### DIFF
--- a/src/blib2to3/pgen2/driver.py
+++ b/src/blib2to3/pgen2/driver.py
@@ -28,17 +28,90 @@ from typing import (
     List,
     Optional,
     Text,
+    Iterator,
     Tuple,
+    TypeVar,
+    Generic,
     Union,
 )
+from dataclasses import dataclass, field
 
 # Pgen imports
 from . import grammar, parse, token, tokenize, pgen
 from logging import Logger
 from blib2to3.pytree import _Convert, NL
 from blib2to3.pgen2.grammar import Grammar
+from contextlib import contextmanager
 
 Path = Union[str, "os.PathLike[str]"]
+
+
+@dataclass
+class ReleaseRange:
+    start: int
+    end: Optional[int] = None
+    tokens: List[Any] = field(default_factory=list)
+
+    def lock(self) -> None:
+        total_eaten = len(self.tokens)
+        self.end = self.start + total_eaten
+
+
+class TokenProxy:
+    def __init__(self, generator: Any) -> None:
+        self._tokens = generator
+        self._counter = 0
+        self._release_ranges: List[ReleaseRange] = []
+
+    @contextmanager
+    def release(self) -> Iterator["TokenProxy"]:
+        release_range = ReleaseRange(self._counter)
+        self._release_ranges.append(release_range)
+        try:
+            yield self
+        finally:
+            # Lock the last release range to the final position that
+            # has been eaten.
+            release_range.lock()
+
+    def eat(self, point: int) -> Any:
+        eaten_tokens = self._release_ranges[-1].tokens
+        if point < len(eaten_tokens):
+            return eaten_tokens[point]
+        else:
+            while point >= len(eaten_tokens):
+                token = next(self._tokens)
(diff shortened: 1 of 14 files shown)


src/blib2to3/pgen2/driver.py
```diff
@@ -28,19 +28,92 @@
...
+    start: int
+    end: Optional[int] = None
+    tokens: List[Any] = field(default_factory=list)
+
+    def lock(self) -> None:
+        total_eaten = len(self.tokens)
+        self.end = self.start + total_eaten
+
+
+class TokenProxy:
```
isidentical, line 60: This class was an artifact of my work done for the parso, and from what I see we might not even need for the black. I probably should get rid of it in one of the upcoming patches.

src/blib2to3/pgen2/parse.py
```diff
@@ -137,11 +201,44 @@ def setup(self, start: Optional[int] = None) -> None:
...
+                if next_token_type == tokenize.OP:
+                    next_token_type = grammar.opmap[cast(str, next_token_value)]
+
+                recorder.add_token(next_token_type, next_token_value)
+                counter += 1
+
+            ilabel = cast(int, recorder.determine_route(next_token_value, force=force))
+            assert ilabel is not None
+
+        return self._addtoken(ilabel, type, value, context)
```
isidentical, line 237: This is a very complex machinery, which I think might deserve an inline explanation. But if anyone is curious, here is the details about the underlying implementation: https://tree.science/what-the-backtracking.html

ichard26:
FWIW I can't this valid code to parse under this PR:
```python
value = (1, 2)

match value:
    case x, y:
        pass
```

ichard26:
And of course CI is also broken to due recent changes in some third party packages, sigh.

isidentical:
I've not included the 3.10 grammar among the default ones, since it is pretty experimental at the moment (and might have other, unknown problems). But it can be activated through specifying `target_versions= {3.10}`. I also added some tests. Hopefully after fixing the CI (which I am not really sure why it is failing), this should be ready for a review.

Zac-HD:
> a recent change in hypothesis makes our primer configuration broken currently

Is this something I should know about as  Hypothesis dev?

(also I love pinning deps with `pip-compile`, happy to write up a detailed issue if that would be of interest)

ichard26:
@Zac-HD oh don't worry about it! We have a development tool called black-primer and its job to run Black over millions of lines of code (by formatting a large array of OSS projects) to assure it doesn't crash or do other weird stuff. Hypothesis happens to be one of the projects and it seems like one of your in-tree development tools was recently made Python 3.8<= which broke the primer workflow on 3.7 and below. I've already fixed by marking Hypothesis as a 3.8+ project in the primer config.

JelleZijlstra (approved):
This is exciting!

If I remember correctly this also makes parenthesized with statements work thanks to parser magic. Should we add tests for that?

Zac-HD:
> it seems like one of your in-tree development tools was recently made Python 3.8<= which broke the primer workflow on 3.7 and below. I've already fixed by marking Hypothesis as a 3.8+ project in the primer config.

Ah, yep, the [Array API standard](https://data-apis.org/) requires Python 3.8+ and therefore our new [extension](https://hypothesis.readthedocs.io/en/latest/numpy.html#array-api) for it does too.

isidentical:
> If I remember correctly this also makes parenthesized with statements work thanks to parser magic. Should we add tests for that?

Added tests for that, nice catch! I'll also work on some more patterns for the PEP 634 in the upcoming PRs, but hopefully as is this PR should cover some of the basic cases for both of them. (not sure if there is anything else needed for the context managers).

ichard26 (approved):
Awesome to see 3.9 and 3.10 compatibility truly realized, thank you!

Kurt-von-Laven:
Is more work needed here, or can this PR be merged?

PR #2592, opened 2021-11-06
Improve Python 2 only syntax detection

ichard26:
### Description

First of all this fixes a mistake I made in Python 2 deprecation PR
using token.* to check for print/exec statements. Turns out that
for nodes with a type value higher than 256 its numeric type isn't
guaranteed to be constant. Using syms.* instead fixes this.

Also add support for the following cases:

    print "hello, world!"

    exec "print('hello, world!')"

    def set_position((x, y), value):
        pass

    try:
        pass
    except Exception, err:
        pass

    raise RuntimeError, "I feel like crashing today :p"

    `wow_these_really_did_exist`

    10L

This should also fix the Python 2 related failures in #2586.

### Checklist - did you ...

- [x] Add new / update outdated documentation? -> /na

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1132,8 +1132,12 @@ def get_features_used(node: Node) -> Set[Feature]:  # noqa: C901
                 features.add(Feature.F_STRINGS)
 
         elif n.type == token.NUMBER:
-            if "_" in n.value:  # type: ignore
+            assert isinstance(n, Leaf)
+            if "_" in n.value:
                 features.add(Feature.NUMERIC_UNDERSCORES)
+            elif n.value.endswith("L"):
+                # Python 2: 10L
+                features.add(Feature.LONG_INT_LITERAL)
 
         elif n.type == token.SLASH:
             if n.parent and n.parent.type in {
@@ -1171,10 +1175,31 @@ def get_features_used(node: Node) -> Set[Feature]:  # noqa: C901
                         if argch.type in STARS:
                             features.add(feature)
 
-        elif n.type == token.PRINT_STMT:
+        # Python 2 only features (for its deprecation) except for 10L, see above
+        elif n.type == syms.print_stmt:
             features.add(Feature.PRINT_STMT)
-        elif n.type == token.EXEC_STMT:
+        elif n.type == syms.exec_stmt:
             features.add(Feature.EXEC_STMT)
+        elif n.type == syms.tfpdef:
+            # def set_position((x, y), value):
+            #     ...
+            features.add(Feature.AUTOMATIC_PARAMETER_UNPACKING)
+        elif n.type == syms.except_clause:
+            # try:
+            #     ...
+            # except Exception, err:
+            #     ...
+            if len(n.children) >= 4:
+                if n.children[-2].type == token.COMMA:
+                    features.add(Feature.COMMA_STYLE_EXCEPT)
+        elif n.type == syms.raise_stmt:
+            # raise Exception, "msg"
+            if len(n.children) >= 4:
+                if n.children[-2].type == token.COMMA:
+                    features.add(Feature.COMMA_STYLE_RAISE)
+        elif n.type == token.BACKQUOTE:
+            # `i'm surprised this ever existed`
+            features.add(Feature.BACKQUOTE_REPR)
 
     return features
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1135,9 +1135,14 @@ def get_features_used(node: Node) -> Set[Feature]:  # noqa: C901
             assert isinstance(n, Leaf)
             if "_" in n.value:
                 features.add(Feature.NUMERIC_UNDERSCORES)
-            elif n.value.endswith("L"):
+            elif n.value.endswith(("L", "l")):
                 # Python 2: 10L
                 features.add(Feature.LONG_INT_LITERAL)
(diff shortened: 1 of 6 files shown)


src/black/__init__.py
```diff
@@ -1132,8 +1132,12 @@ def get_features_used(node: Node) -> Set[Feature]:  # noqa: C901
                 features.add(Feature.F_STRINGS)
 
         elif n.type == token.NUMBER:
-            if "_" in n.value:  # type: ignore
+            assert isinstance(n, Leaf)
+            if "_" in n.value:
                 features.add(Feature.NUMERIC_UNDERSCORES)
+            elif n.value.endswith("L"):
```
JelleZijlstra, line 1138:
```suggestion
            elif n.value.endswith(("l", "L")):
```
It can be lowercase too.

Also, we could look for `010` (old octal syntax), which doesn't work in 3.0 any more. This is in https://docs.python.org/3/whatsnew/3.0.html#integers and PEP 3127. We could check that the first char is 0 and the second is a digit.

src/black/__init__.py
```diff
@@ -1171,10 +1175,31 @@ def get_features_used(node: Node) -> Set[Feature]:  # noqa: C901
...
+            # except Exception, err:
+            #     ...
+            if len(n.children) >= 4:
+                if n.children[-2].type == token.COMMA:
+                    features.add(Feature.COMMA_STYLE_EXCEPT)
+        elif n.type == syms.raise_stmt:
+            # raise Exception, "msg"
+            if len(n.children) >= 4:
+                if n.children[-2].type == token.COMMA:
+                    features.add(Feature.COMMA_STYLE_RAISE)
```
JelleZijlstra, line 1199: Could you make sure this doesn't trigger on something like `raise Exception("foo",)`, which is valid Python 3?

JelleZijlstra (commented):
Thanks!

src/black/__init__.py
ichard26: committed as part of https://github.com/psf/black/pull/2592/commits/78d32bf5f4c1d3fe0db9e61badba7895facb096b

src/black/__init__.py
ichard26: great catch, thank you!

JelleZijlstra (approved):

JelleZijlstra:
Thank you!

ichard26:
cc @Zac-HD 

```pytb
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesmith/syntactic.py", line 110, in draw_symbol
    compile(
SystemError: Negative size passed to PyUnicode_New

  File "/home/runner/work/black/black/fuzz.py", line 73, in <module>
    test_idempotent_any_syntatically_valid_python()
  File "/home/runner/work/black/black/fuzz.py", line 20, in test_idempotent_any_syntatically_valid_python
    max_examples=1000,  # roughly 1k tests/minute, or half that under coverage
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesis/core.py", line 1199, in wrapped_test
    raise the_error_hypothesis_found
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesmith/syntactic.py", line 89, in do_draw
    result = super().do_draw(data)
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesis/extra/lark.py", line 153, in do_draw
    self.draw_symbol(data, start, state)
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesmith/syntactic.py", line 107, in draw_symbol
    super().draw_symbol(data, symbol, draw_state)
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesis/extra/lark.py", line 181, in draw_symbol
    self.draw_symbol(data, e, draw_state)
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesmith/syntactic.py", line 107, in draw_symbol
    super().draw_symbol(data, symbol, draw_state)
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesis/extra/lark.py", line 181, in draw_symbol
    self.draw_symbol(data, e, draw_state)
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesmith/syntactic.py", line 107, in draw_symbol
    super().draw_symbol(data, symbol, draw_state)
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesis/extra/lark.py", line 181, in draw_symbol
    self.draw_symbol(data, e, draw_state)
  File "/home/runner/work/black/black/.tox/fuzz/lib/python3.9/site-packages/hypothesmith/syntactic.py", line 129, in draw_symbol
    raise type(err)(
SystemError: compile('delA\\ \n#\n', '<string>', 'single') raised SystemError: Negative size passed to PyUnicode_New
```

https://github.com/psf/black/runs/4184619816?check_suite_focus=true#step:5:85

Seems like the fuzz issue might be related to https://bugs.python.org/issue45738. Anyhow we can't do anything about it ourselves :)

Zac-HD:
Thanks, https://github.com/Zac-HD/hypothesmith/issues/16 is open and I guess I'm adding another workaround for an upstream bug.

(Wish they'd just [run it upstream](https://bugs.python.org/issue42109), but I'm not particularly optimistic about that anymore)

PR #2602, opened 2021-11-11
primer: Hypothesis now requires Python>=3.8

ichard26:
### Description

looks like their project dev tooling uses some newer syntax or something

### Checklist - did you ...

- [x] Add / update tests if necessary? does primer count?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,13 +7,18 @@ _build
 docs/_static/pypi.svg
 .tox
 __pycache__
+
+# Packaging artifacts
 black.egg-info
+black.dist-info
 build/
 dist/
 pip-wheel-metadata/
+.eggs
+
 src/_black_version.py
 .idea
-.eggs
+
 .dmypy.json
 *.swp
 .hypothesis/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -7,13 +7,18 @@ _build
 docs/_static/pypi.svg
 .tox
 __pycache__
+
+# Packaging artifacts
 black.egg-info
+black.dist-info
```
ichard26, line 13: I've been spending a lot time recently in the Python packaging land hence this addition :)

JelleZijlstra (approved):

ichard26:
Fuzz and tests are broken on Python 3.9 due to a change between CPython 3.9.7 and 3.9.8 removing a private API typed-ast depends on. Merging to unblock at least primer CI even if CI will still be red tonight :)

David8166 (commented):
K

cooperlees:
Maybe we could teach primer to parse out the classifiers in setup.* and respect them.

PR #2608, opened 2021-11-14
black/parser: support as-exprs within call args

isidentical:
I apparently forget to `as`-exprs when they happen inside call arguments, which have their own special mini-grammar to support stuff like walrus (higher precedence operators). This patch should resolve that issue.


JelleZijlstra:
Thanks!

PR #2610, opened 2021-11-14
Removed distutils import from autoload/black.vim (#2607)

### LINKED ISSUE #2607: Vim plugin fail with Python 3.10 due to importing distutils in autoload/black.vim
When you have vim builded with Python 3.10 the black plugin will fail when running :BlackUpgrade or :Black

> Error detected while processing /home/michal/.vim/bundle/black/autoload/black.vim:
> line  201:
> <string>:5: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives

I am able to fix the issue by removing `from distutils.util import strtobool` from **autoload/black.vim** and re-implementing `strtobool` as suggested in [PEP 632](https://www.python.org/dev/peps/pep-0632/).

515k4:
### Description

Fixed #2607 by removing distutils import from autoload/black.vim, which is deprecated in Python 3.10, and by re-implementing the distutils.strtobool as suggested in PEP 632.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,10 @@
 - Add partial support for the match statement. As it's experimental, it's only enabled
   when `--target-version py310` is explicitly specified (#2586)
 - Add support for parenthesized with (#2586)
-- Fixed Vim with Python 3.10 by removing deprecated distutils (#2610)
+
+### Integrations
+
+- Fixed vim plugin with Python 3.10 by removing deprecated distutils import (#2610)
 - Declare support for Python 3.10 for running Black (#2562)
 
 ## 21.10b0
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 - Add partial support for the match statement. As it's experimental, it's only enabled
   when `--target-version py310` is explicitly specified (#2586)
 - Add support for parenthesized with (#2586)
+- Fixed Vim with Python 3.10 by removing deprecated distutils (#2610)
 
 ## 21.10b0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,11 +10,11 @@
 - Add partial support for the match statement. As it's experimental, it's only enabled
   when `--target-version py310` is explicitly specified (#2586)
 - Add support for parenthesized with (#2586)
+- Declare support for Python 3.10 for running Black (#2562)
 
 ### Integrations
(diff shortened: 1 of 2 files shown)


autoload/black.vim
```diff
@@ -3,8 +3,13 @@ import collections
 import os
 import sys
 import vim
-from distutils.util import strtobool
 
+def strtobool(text):
+    if text.lower() in ['y', 'yes', 't', 'true' 'on', '1']:
```
JelleZijlstra, line 8: We seem to be using two-space indents in this file for some reason, so let's stay consistent with that.

Also, please use an f-string on line 12.

JelleZijlstra (changes requested):
Thanks for the PR!

Could you also add a changelog entry?

CHANGES.md
```diff
@@ -10,6 +10,7 @@
 - Add partial support for the match statement. As it's experimental, it's only enabled
   when `--target-version py310` is explicitly specified (#2586)
 - Add support for parenthesized with (#2586)
+- Fixed Vim with Python 3.10 by removing deprecated distutils (#2610)
```
```suggestion

### Integrations

- Fixed vim plugin with Python 3.10 by removing deprecated distutils import (#2610)
```

See 21.10b0 for previous vim changes, we should try to keep the wording consistent.

CHANGES.md
```diff
@@ -10,7 +10,10 @@
 - Add partial support for the match statement. As it's experimental, it's only enabled
   when `--target-version py310` is explicitly specified (#2586)
 - Add support for parenthesized with (#2586)
-- Fixed Vim with Python 3.10 by removing deprecated distutils (#2610)
+
+### Integrations
+
+- Fixed vim plugin with Python 3.10 by removing deprecated distutils import (#2610)
 - Declare support for Python 3.10 for running Black (#2562)
```
JelleZijlstra, line 17: This ended up in the wrong section, can you move it up?

515k4: But vim plugin is integration issue, isn't it?

JelleZijlstra: I'm referring to the Python 3.10 line

515k4: I see, I've moved it.

JelleZijlstra:
Thanks!

PR #2504, opened 2021-09-21
Pull Request for #2476 - Parse `skip_magic_trailing_comma` from pyproject.toml

nullromo:
### Description

This PR is in response to issue #2476. It allows the black vim plugin to parse the `skip_magic_trailing_comma` option from `pyproject.toml` and act accordingly.

### Checklist

- [x] Add a CHANGELOG entry if necessary?
- [ ] Add / update tests if necessary?
I'm not sure exactly how to do this, or if it's necessary. If it is, I can try to figure it out.
- [x] Add new / update outdated documentation?
Not necessary in this case. Black already claims to use `pyproject.toml`.

PR #2613, opened 2021-11-16
vim: Parse skip_magic_trailing_comma from pyproject.toml

### LINKED ISSUE #2476: Vim plugin not respecting `pyproject.toml`
**Describe the bug**

The vim plugin does not read or does not respect the options set in `pyproject.toml`.

**To Reproduce**

1. Create a file `test.py`:

    ```python
    def test(arg):
        print(arg)


    test(4,)
    ```

    There are 2 ways that black can format this. With the `--skip-magic-trailing-comma` option set to `false`, it formats like this (the **broken-line way**):

    ```python
    def test(arg):
        print(arg)


    test(
        4,
    )
    ```

    and with the `--skip-magic-trailing-comma` option set to `true`, it formats like this (the **connected-line way**):

    ```python
    def test(arg):
        print(arg)


    test(4)
    ```

2. Create a `pyproject.toml` file:
    ```toml
    [tool.black]
    skip-magic-trailing-comma = false
    ```
    Now run `python -m black test.py --check --diff` on the command line. It will show the **broken-line way**.

    Try running `:Black` on `test.py` in vim, and it will format to the **broken-line way**.

3. Change the `pyproject.toml` file to:
    ```toml
    [tool.black]
    skip-magic-trailing-comma = true
    ```
    Now run `python -m black test.py --check --diff` on the command line again. It will show the **connected-line way**.

    Try running `:Black` on `test.py` in vim again, and it will format to the **broken-line way**. This is incorrect. It should respect the `pyproject.toml` file's settings.

**Expected behavior**

I expect to be able to set the `--skip-magic-trailing-commas` option for the vim plugin. If it won't respect the `pyproject.toml` file, then it should at least have a `.vimrc`-settable option and this should be described in the README or documentation pages somewhere.

**Environment**

- Version: [main]
- OS and Python version: WSL (Linux)/Python 3.8.10

JelleZijlstra:
This is #2504 but I updated it locally to make sure the changelog ended up in the right place.

Fixes #2476. Closes #2504.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## _Unreleased_
+
+### _Black_
+
+- Warn about Python 2 deprecation in more cases by improving Python 2 only syntax
+  detection (#2592)
+
 ## 21.10b0
 
 - The vim plugin now parses `skip_magic_trailing_comma` from pyproject.toml (#2504)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Change Log
 
-- Parse `skip_magic_trailing_comma` from pyproject.toml (#2504)
+- The vim plugin now parses `skip_magic_trailing_comma` from pyproject.toml (#2504)
 
 ## 21.9b0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,12 +10,16 @@
 - Add new `--workers` parameter (#2514)
 - Fixed feature detection for positional-only arguments in lambdas (#2532)
 - Bumped typed-ast version minimum to 1.4.3 for 3.10 compatiblity (#2519)
-- Add primer support for --projects (#2555)
 
 ### _Blackd_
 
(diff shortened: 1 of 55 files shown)


CHANGES.md
```diff
@@ -15,6 +15,7 @@
 ### Integrations
 
 - Fixed vim plugin with Python 3.10 by removing deprecated distutils import (#2610)
+- The vim plugin now parses `skip_magic_trailing_comma` from pyproject.toml (#2504)
```
```suggestion
- The vim plugin now parses `skip_magic_trailing_comma` from pyproject.toml (#2613)
```

515k4:
Something went wrong with this merge, now I got this in vim 8.2 (with Python 3.10):

```
Error detected while processing function black#Black:
line   10:
  File "<string>", line 131, in Black
  File "<string>", line 189, in get_configs
  File "<string>", line 190, in <dictcomp>
vim.error: Vim:E121: Undefined variable: g:black_skip_magic_trailing_comma
```

I am not using the pyproject.toml.

PR #2615, opened 2021-11-16
fix vim plugin

JelleZijlstra:
- Define a global for magic trailing comma (this was missed in #2613)
- Add py310 as an option


JelleZijlstra:
@515k4 would you be able to help test this?

JelleZijlstra:
Confirmed locally that this works.

515k4:
> @515k4 would you be able to help test this?

Too busy too late but I also confirm. Now it works with my setup (Vim 8.2 + Python 3.10).

PR #2616, opened 2021-11-17
prepare release 21.11b0

JelleZijlstra:


ichard26 (approved):
:rocket: let's do this without explosive decomposition :fire: :)

PR #2617, opened 2021-11-17
release docs: slightly better example link

JelleZijlstra:


cooperlees (commented):
I was more hoping to remove the need to update to docs as well as the CHNAGES.md …

Maybe we should write a script for the releaser to run to change these versions everywhere … I was sad to learn about this last release.

JelleZijlstra:
It's not great but it's what we have for now, so it's better to document it. Writing a script to update it feels like overkill to me; it's still just three places to change.

felix-hilden (approved):
As long as it's documented, I wouldn't mind editing a few places. Seems reasonable.

Issue #2620, opened 2021-11-18
21.11b01 release broken with older versions of regex


rowanseymour:
Latest `21.11b01` release includes https://github.com/psf/black/commit/117891878e5be4d6b771ae5de299e51b679cea27#diff-431968a04535c5ee2570855fe3a3473527050ed0e6d5cb5a9544bcef254d0590 which references the `regex.Pattern` class.

But this was only added in `regex==2021.4.4` and black depends on `regex>=2020.1.8` so running with an older regex version gives..

```
AttributeError: module 'regex' has no attribute 'Pattern'
```

PR #2621, opened 2021-11-18
Bump regex dependency to 2021.4.4 to fix import of Pattern class

rowanseymour:
Fixes https://github.com/psf/black/issues/2620

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### _Black_
+
+- Bumped regex version minimum to 2021.4.4 to fix Pattern class usage (#2621)
+
 ## 21.11b0
 
 ### _Black_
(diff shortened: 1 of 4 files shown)


JelleZijlstra (approved):
Thanks!

JelleZijlstra:
Could you add a changelog entry? I'm leaning towards making a new release for this, so we should announce the change.

JelleZijlstra:
People on Discord brought up that we also need to update Pipfile.lock.

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## 21.11b1
```
JelleZijlstra, line 3:
```suggestion
## Unreleased
```
We'll change this when we actually release

rowanseymour:
@JelleZijlstra took a few goes to figure out pipenv locking but I think I have it right now..

ichard26 (approved):
Pipenv is a pain but this looks good enough? Thank you for noticing the original issue by the way @rowanseymour!

rowanseymour:
thanks for freeing us from worrying about how format python!

Issue #2197, opened 2021-05-04
Look into alternatives to using `regex`


JelleZijlstra:
We're using the `regex` third-party library to avoid a bug around certain Unicode symbols: see #1047 (PR) and #455 (issue). However, this dependency is costly because it's a C library and compiling it is difficult for some users. We should explore rewriting the relevant parts of the parser to not use a regex so we can avoid this dependency.

ichard26:
And with the merger of #2654 and #2644 the last thing to do is to remove references to regex in the documentation and setup.py, Pipfile[.lock] 🎉

PR #2624, opened 2021-11-18
Upgrade regex to latest to avoid import error

pablogamboa:
### Description

Upgrade regex to latest to avoid import error. Unfortunately 2021.11b1 didn't fix the issues present in 2021.11b0. Pipenv has updated a bunch of unrelated packages unfortuantely when updating the lock

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### _Black_
+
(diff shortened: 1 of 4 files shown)


ichard26:
Thanks for the PR but we are going to be nixing regex (GH-2197) soon so this won't be necessary at all :)

Issue #2627, opened 2021-11-19
Jupyter cell magics are not completely ignored


danielsparing:
**Describe the bug**

When a Jupyter cell is a cell magic, despite the [doc](https://black.readthedocs.io/en/latest/faq.html#why-is-my-jupyter-notebook-cell-not-formatted), sometimes the whole content is parsed and can throw an indentation error (even though it should be ignored). For example, the `%%writefile` magic should be able to contain arbitrary text regardless of indentation.

**To Reproduce**

For example, create a notebook `notebook.ipynb` containing the below code cell:

```
%%writefile e.txt
free text
    with arbitrary
  indentation
```

And run it with these arguments:

```sh
$ black notebook.ipynb
```

The resulting error is:

> error: cannot format notebook.ipynb: unindent does not match any outer indentation level (\<tokenize\>, line 4)
> Oh no! 💥 💔 💥
> 1 file failed to reformat.

**Expected behavior**

Cell magics should be ignored, in line with doc.

**Environment**

- Black's version: 21.11b1
- OS and Python version: Linux/Python 3.7.10

**Additional context**

The above error message seems to come from iPython itself, or at least similar to the error you get if you try to run this cell in the notebook _without the cell magic line_:

```
# %%writefile e.txt
free text
    with arbitrary
  indentation
```
```
  File "<tokenize>", line 4
    indentation
    ^
IndentationError: unindent does not match any outer indentation level
```

PR #2630, opened 2021-11-19
Return `NothingChanged` if non-Python cell magic is detected, to avoid tokenize error

danielsparing:
### Description

Fixes https://github.com/psf/black/issues/2627 , a non-Python cell magic such as `%%writeline` can legitimately contain "incorrect" indentation, however this causes `tokenize-rt` to return an error. To avoid this, `validate_cell` should early detect cell magics (just like it detects `TransformerManager` transformations).

Test added too, in the shape of a "badly indented" `%%writefile` within `test_non_python_magics`.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,9 @@
 
 ### _Black_
 
-- Fixed non-Python cell magics sometimes failing due to indentation (#2630)
+- Cell magics are now only processed if they are known Python cell magics. Earlier, all
+  cell magics were tokenized, leading to possible indentation errors e.g. with
+  `%%writefile`. (#2630)
 
 ## 21.11b1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### _Black_
+
+- Fixed non-Python cell magics sometimes failing due to indentation (#2630)
+
 ## 21.11b1
 
 ### _Black_
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -57,7 +57,7 @@ from black.handle_ipynb_magics import (
     remove_trailing_semicolon,
     put_trailing_semicolon_back,
     TRANSFORMED_MAGICS,
-    NON_PYTHON_CELL_MAGICS,
+    PYTHON_CELL_MAGICS,
     jupyter_dependencies_are_installed,
 )
 
@@ -962,7 +962,9 @@ def validate_cell(src: str) -> None:
     """
     if any(transformed_magic in src for transformed_magic in TRANSFORMED_MAGICS):
         raise NothingChanged
-    if any("%%" + cell_magic in src for cell_magic in NON_PYTHON_CELL_MAGICS):
+    if src[:2] == "%%" and src.split()[0] not in (
+        "%%" + magic for magic in PYTHON_CELL_MAGICS
+    ):
         raise NothingChanged
 
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -962,9 +962,7 @@ def validate_cell(src: str) -> None:
     """
     if any(transformed_magic in src for transformed_magic in TRANSFORMED_MAGICS):
         raise NothingChanged
-    if src[:2] == "%%" and src.split()[0] not in (
-        "%%" + magic for magic in PYTHON_CELL_MAGICS
-    ):
+    if src[:2] == "%%" and src.split()[0][2:] not in PYTHON_CELL_MAGICS:
         raise NothingChanged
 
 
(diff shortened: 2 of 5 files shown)


MarcoGorelli (changes requested):
Thanks @danielsparing for this PR!

I'm not a `black` maintainer, but I did make the Jupyter PR, so am leaving a couple of comments:

1 - does this mean that lines

https://github.com/psf/black/blob/9a73bb86db59de1e12426fec81dcdb7f3bb9be7b/src/black/handle_ipynb_magics.py#L233-L234

can be removed?

2. Perhaps this should be done the other way round? I.e. rather than checking if the magic name is in a blocklist, and skipping the cell if that's the case, `black` should check if the magic name is in an allow list, and only format the cell if it is. Thoughts on this?

danielsparing:
@MarcoGorelli Thanks for the quick review!

I agree that the allowlist is cleaner. This also necessitates something I should have done earlier, to exactly check for the first word in the cell for the cell magic.

This does mean that I needed to remove the test with `%%t` and make it a `%%timeit` as it is a custom magic, so shouldn't be in the allowlist.

I also removed the lines you mentioned.

What do you think?

src/black/handle_ipynb_magics.py
```diff
@@ -37,20 +37,13 @@
...
-        "ruby",
-        "script",
-        "sh",
-        "svg",
-        "writefile",
+        "capture",
+        "prun",
+        "pypy",
+        "time",
+        "timeit",
```
MarcoGorelli, line 46: I think `python` and `python3` should also be added?

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### _Black_
+
+- Fixed non-Python cell magics sometimes failing due to indentation (#2630)
```
MarcoGorelli, line 7: I think this can be made more explicit - can you make it clearer what the behaviour was before, and what it is now, so it's clear that `black` is now intentionally more timid about processing cell magics?

MarcoGorelli (changes requested):
Thanks for updating @danielsparing !

Generally looks good to me, tagging @JelleZijlstra and @ichard26 (who reviewed the original PR extensively) for further comments

src/black/handle_ipynb_magics.py
danielsparing: Yes my bad -- adding them, as well as `python2`.

CHANGES.md
danielsparing: addressed in https://github.com/psf/black/pull/2630/commits/1e8a0b1246e74406731981830a05203a35ad65bd

src/black/handle_ipynb_magics.py
```diff
@@ -37,20 +37,16 @@
...
-        "ruby",
-        "script",
-        "sh",
-        "svg",
-        "writefile",
+        "capture",
+        "prun",
+        "pypy",
+        "python",
+        "python2",
```
MarcoGorelli, line 46: Python2 support has been deprecated from `black`, I think this can be removed

MarcoGorelli (approved):
Awesome - just got a minor comment left, will handover to the `black` maintainers now

src/black/handle_ipynb_magics.py
danielsparing: Thanks, removed in https://github.com/psf/black/pull/2630/commits/744820647cc1728d0c4e82e20fcfbeaeb0fd6cad

JelleZijlstra (approved):

src/black/__init__.py
```diff
@@ -959,6 +962,10 @@ def validate_cell(src: str) -> None:
     """
     if any(transformed_magic in src for transformed_magic in TRANSFORMED_MAGICS):
         raise NothingChanged
+    if src[:2] == "%%" and src.split()[0] not in (
+        "%%" + magic for magic in PYTHON_CELL_MAGICS
+    ):
```
MarcoGorelli, line 967: really minor, but perhaps
```suggestion
    if src[:2] == "%%" and src.split()[0][2:] not in PYTHON_CELL_MAGICS:
```
is simpler / marginally more efficient?

(feel free to disregard this comment of course 😆 )

danielsparing: makes perfect sense, thanks for simplifying it!

ichard26:
Thank you so much for your contribution! This project is only possible by contributions like these :black_heart:. Seriously, the core team barely has any experience with Jupyter notebooks so we rely on third-party contributions to keep this feature of Black alive and well. You're awesome, @danielsparing - congrats on your first PR to psf/black :tada: 

I'm also going to take this moment to thank the defacto core dev on black's jupyter support: @MarcoGorelli!

Y'all are great :)

PR #2632, opened 2021-11-20
Change `cfg` to `ini` for text highlighting

AshIsbitt:


cooperlees (approved):
Agree this should look better / easier to parse. Thanks.

onerandomusername:
Curious, but what about `toml`?

AshIsbitt:
There was a mix of `ini` and `toml` throughout the document - I'm not sure which the devs/maintainers would prefer, but they're free to change it if preferred

JelleZijlstra:
These aren't actually toml files. Thanks for your contribution!

PR #2638, opened 2021-11-23
README: Add KeepTruckin to the list of orgs

kalbasit:
### Description

At KT, we used Black to format all Python code in our Mono-repo. Added it to the list of orgs using Black.


ichard26 (approved):
Wonderful, thank you very much!

PR #2639, opened 2021-11-24
grammar: accept open sequences on match subject

### LINKED ISSUE #2637: [3.10] Match formatting does not support multiple value
Hello and thanks for this library and the reactivity to support python 3.10 new syntaxes. 

**Describe the bug**

While running with the latest version (gotten from git),basic `match` on a single pattern is correctly supported <3 
But it seems that it does not support multiple match value.

I tried on a simple example after observing it on my real code:

**To Reproduce**

```python
a = 11
b = 12

match a, b:
    case _, 5:
        print("a5")
    case 11, 12:
        print("default")
    case _:
        print("badass")

```

Should be able to parse. It runs correctly:
```shell
python test_black.py 
default
```
But black fails :'( 

```shell
black --target-version py310  test_black.py
error: cannot format test_black.py: Cannot parse: 4:6: match a, b:
Oh no! 💥 💔 💥
1 file failed to reformat.
```

**Expected behavior**

Should format without errors

**Environment**

- Black's version: black, 21.11b2.dev3+g4075944 (compiled: no)
- OS and Python version: Linux[DevContainer]/3.10

**Additional context**

isidentical:
Resolves #2637

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 ### _Black_
 
 - Fixed Python 3.10 support on platforms without ProcessPoolExecutor (#2631)
-- Fixed `match` statements with open sequence subjects (#2639)
+- Fixed `match` statements with open sequence subjects, like `match a, b:` (#2639)
 
 ## 21.11b1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ### _Black_
 
 - Fixed Python 3.10 support on platforms without ProcessPoolExecutor (#2631)
+- Fixed `match` statements with open sequence subjects (#2639)
 
 ## 21.11b1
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ### _Black_
 
 - Fixed Python 3.10 support on platforms without ProcessPoolExecutor (#2631)
+- Fixed `match` statements with open sequence subjects (#2639)
```
```suggestion
- Fixed `match` statements with open sequence subjects, like `match a, b:` (#2639)
```
I feel like most users wouldn't understand what an "open sequence subject" is.

isidentical: Makes sense, thanks!

JelleZijlstra (approved):
Thank you!

ichard26:
Thank you @isidentical :black_heart:

PR #2643, opened 2021-11-25
fix regex in test_black.py

JelleZijlstra:
Trying to get rid of the regex dependency, but found instead that we were incorrectly using a character range in this regex.


ichard26 (approved):
Thank you!

PR #2654, opened 2021-11-29
Fix determination of f-string expression spans

hauntsaninja:
The original goal here was to get rid of the regex dependency, but it turns out the regex failed many of the test cases added in this PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 - Fixed `match` statements with open sequence subjects, like `match a, b:` (#2639)
 - Fixed assignment to environment variables in Jupyter Notebooks (#2642)
 - Add `flake8-simplify` and `flake8-comprehensions` plugins (#2653)
+- Fixed determination of f-string expression spans (#2654)
 
 ## 21.11b1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,7 @@
   times, like `match re.match()` (#2661)
 - Fix assignment to environment variables in Jupyter Notebooks (#2642)
 - Add `flake8-simplify` and `flake8-comprehensions` plugins (#2653)
-- Fixed determination of f-string expression spans (#2654)
+- Fix determination of f-string expression spans (#2654)
 - Fix parser error location on invalid syntax in a `match` statement (#2649)
 
 ## 21.11b1
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -982,6 +982,8 @@ def iter_fexpr_spans(s: str) -> Iterator[Tuple[int, int]]:
                 i += len(delim)
                 while i < len(s) and s[i : i + len(delim)] != delim:
                     i += 1
+                i += len(delim)
+                continue
         i += 1
 
 
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -942,6 +942,48 @@ class BaseStringSplitter(StringTransformer):
         return max_string_length
 
 
+def iter_fexpr_spans(s: str) -> Iterator[Tuple[int, int]]:
+    """
+    Yields spans corresponding to expressions in a given f-string.
+    Assumes the input string is a valid f-string.
+    """
+    stack = []  # our curly paren stack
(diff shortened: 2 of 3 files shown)


hauntsaninja:
Here's a basic property test:
```
import ast
import hypothesis
from hypothesis import strategies as st


@hypothesis.given(st.text(alphabet="{}' "))
@hypothesis.settings(max_examples=1000)
def test_parse(text):
    string = f'f"{text}"'
    try:
        tree = ast.parse(string)
    except SyntaxError:
        hypothesis.assume(False)
    print(string)

    str_bits = [c.value for c in tree.body[0].value.values if isinstance(c, ast.Constant)]
    expr_spans = list(iter_fexpr_spans(string))
    starts = [a for a, _ in expr_spans]
    ends = [b for _, b in expr_spans]
    expected_bits = [
        x for i, j in zip([2] + ends, starts + [-1])
        if (x := string[i:j].replace("{{", "{").replace("}}", "}"))
    ]
    assert str_bits == expected_bits, (str_bits, expected_bits)
```

src/black/trans.py
```diff
@@ -942,6 +942,53 @@ def _get_max_string_length(self, line: Line, string_idx: int) -> int:
         return max_string_length
 
 
+def iter_fexpr_spans(s: str) -> Iterator[Tuple[int, int]]:
+    """
+    Yields spans corresponding to expressions in a given f-string.
+    Assumes the input string is a valid f-string.
```
JelleZijlstra, line 948: Document that the yielded values are half-open ranges

src/black/trans.py
```diff
@@ -942,6 +942,53 @@ def _get_max_string_length(self, line: Line, string_idx: int) -> int:
...
+        # if we're in an expression part of the f-string, fast forward through strings
+        # note that backslashes are not legal in the expression portion of f-strings
+        if stack:
+            delim = None
+            if s[i : i + 3] in ("'''", '"""'):
+                delim = s[i : i + 3]
+            elif s[i] in ("'", '"'):
+                delim = s[i]
+            if delim:
+                i += len(delim)
```
JelleZijlstra, line 982: This seems wrong; triple-quoted string will only work if their length is a multiple of 3. I believe just `i += 1` is right.

src/black/trans.py
```diff
@@ -942,6 +942,53 @@ def _get_max_string_length(self, line: Line, string_idx: int) -> int:
...
+                while i < len(s) and s[i : i + len(delim)] != delim:
+                    i += 1
+        i += 1
+
+
+def fstring_contains_expr(s: str) -> bool:
+    return any(True for _ in iter_fexpr_spans(s))
```
JelleZijlstra, line 989:
```suggestion
    return any(iter_fexpr_spans(s))
```

JelleZijlstra (changes requested):
Thanks, this is great!

For the tests could you add `test_trans.py` with tests like the ones you list in the PR description.

src/black/trans.py
hauntsaninja: I'm not sure I understand :-( , do you have an example that fails? This is just skipping past the opening (triple) quote, we do `i += 1` in the loop.

src/black/trans.py
hauntsaninja: Sure, I guess `iter_fexpr_spans` only yields truthy things and that's unlikely to change :-)

src/black/trans.py
JelleZijlstra: You're right, I misread.

CHANGES.md
```diff
@@ -15,6 +15,7 @@
   times, like `match re.match()` (#2661)
 - Fix assignment to environment variables in Jupyter Notebooks (#2642)
 - Add `flake8-simplify` and `flake8-comprehensions` plugins (#2653)
+- Fixed determination of f-string expression spans (#2654)
```
```suggestion
- Fix determination of f-string expression spans (#2654)
```

JelleZijlstra (approved):
Thank you!

PR #2656, opened 2021-11-29
Improve black startup to format a single file. Fixes #2564

### LINKED ISSUE #2564: Black slow startup to format a single file
I'm a bit annoyed that it takes too much time to format a single file when using black to auto-save contents in my IDE (in this case it's Eclipse/PyDev, but I guess it should be the same for any place using the command line to format a single file).

-- In Eclipse/PyDev, the integration (I've done) is done so that for each new file to be formatted, black is executed and contents are piped in/out.

For doing this execution (in my machine), black takes `0.3s`, but out of that time, `0.255s` (`83%` of the time) is taken on doing imports, `0.025s` (`9%`) getting git info and only `0.020s` (`8%`) on actually doing the formatting...

In use case where a single file is passed is that it'll call `black.files.get_gitignore` even when a single `.py` is passed, which is pretty slow and it shouldn't be needed in this case (without it, the formatting time of `0.045s` goes to `0.02s`... this can be seen passing some dummy `--exclude` which has the side-effect of not loading the git ignore contents).

Some profiling numbers on doing a single run for a file in `black` from the command line:

image

So, some things I considered to improve that:

1. Move some of the bigger imports from top-level to inside methods so that it happens lazily (that way things unneeded such as the `asyncio` wouldn't be executed at all). Note: this would need to move some top-level imports to a `typing.TYPE_CHECKING` check too.
2. Make the `get_gitignore` called only if needed.

I guess another option could be using the black server in the integration, but given that it's not installed by default in user environments this is a bit of a hassle for end users (since it's not really polite to automatically mess user environments installing `black[d]` one needs to ask them to do so -- and it doesn't fix the issue where it's used in the command line).

So, I was considering providing a pull request for that -- to make some imports local / don't call `get_gitignore` unless needed -- is this something that seems ok? (for me it's a no-brainer, but I've seen projects adamant on using top-level imports even if it works to their detriment, so, I'd rather ask first than do the work and get rejected later).

fabioz:
### Make the startup of black faster when formatting a single file by importing less modules and skipping loading of `.gitignore`.

With this patch the time to format one file went from a baseline of:
(min /max time from doing many runs on my machine)
`0.29s -> 0.47s` to `0.17s -> 0.32s`

-- Note that this includes the time to format too, not just the startup time. Also, this depends a lot on the target machine (slower machines will probably have more to gain)... in my machine I'd say there's around a 50% increase in the startup speed. There's still room for improvement, but I'd say that the low-hanging fruits have been taken care of.

The number of imported modules for doing a run in this case went from `268` to `202`.

It went something as:

Modules imported (baseline):
268

Without asyncio:
235

Without concurrent.futures
222

Without pathspec:
214

Without multiprocessing:
202

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1,17 +1,13 @@
-import asyncio
+import typing
 from json.decoder import JSONDecodeError
 import json
-from concurrent.futures import Executor, ThreadPoolExecutor, ProcessPoolExecutor
 from contextlib import contextmanager
 from datetime import datetime
 from enum import Enum
 import io
-from multiprocessing import Manager, freeze_support
 import os
 from pathlib import Path
-from pathspec.patterns.gitwildmatch import GitWildMatchPatternError
 import regex as re
-import signal
 import sys
 import tokenize
 import traceback
@@ -28,11 +24,11 @@ from typing import (
     Sized,
     Tuple,
     Union,
+    Iterable,
 )
 
 import click
 from dataclasses import replace
-from mypy_extensions import mypyc_attr
 
 from black.const import DEFAULT_LINE_LENGTH, DEFAULT_INCLUDES, DEFAULT_EXCLUDES
 from black.const import STDIN_PLACEHOLDER
@@ -42,8 +38,7 @@ from black.linegen import transform_line, LineGenerator, LN
 from black.comments import normalize_fmt_off
 from black.mode import Mode, TargetVersion
 from black.mode import Feature, supports_feature, VERSION_TO_FEATURES
-from black.cache import read_cache, write_cache, get_cache_info, filter_cached, Cache
-from black.concurrency import cancel, shutdown, maybe_install_uvloop
+from black.cache import read_cache, write_cache, get_cache_info, Cache
 from black.output import dump_to_file, ipynb_diff, diff, color_diff, out, err
 from black.report import Report, Changed, NothingChanged
 from black.files import find_project_root, find_pyproject_toml, parse_pyproject_toml
@@ -64,6 +59,14 @@ from black.handle_ipynb_magics import (
 # lib2to3 fork
 from blib2to3.pytree import Node, Leaf
 from blib2to3.pgen2 import token
+
+if typing.TYPE_CHECKING:
+    # Don't import the pathspec module unless really needed (it's only needed
+    # when listing contents to format from a directory).
+    from pathspec.pathspec import PathSpec
+    from concurrent.futures import Executor
+    import asyncio
+
 
 from _black_version import version as __version__
 
@@ -451,21 +454,18 @@ def main(
             content=code, fast=fast, write_back=write_back, mode=mode, report=report
         )
     else:
-        try:
-            sources = get_sources(
-                ctx=ctx,
-                src=src,
-                quiet=quiet,
-                verbose=verbose,
-                include=include,
-                exclude=exclude,
-                extend_exclude=extend_exclude,
-                force_exclude=force_exclude,
-                report=report,
-                stdin_filename=stdin_filename,
-            )
-        except GitWildMatchPatternError:
-            ctx.exit(1)
+        sources = get_sources(
+            ctx=ctx,
+            src=src,
+            quiet=quiet,
+            verbose=verbose,
+            include=include,
+            exclude=exclude,
+            extend_exclude=extend_exclude,
+            force_exclude=force_exclude,
+            report=report,
+            stdin_filename=stdin_filename,
+        )
 
         path_empty(
             sources,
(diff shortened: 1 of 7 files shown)


cooperlees:
Thanks. I like the idea. I don't know if we (maintainers collective) want lazy imports all over the code base tho. I think it's worth it potentially, but for cases where latency really matters we recommended blackd.

Some of your import moving has broke tests ... I did expect more work to be needed in some of those places. This is the fun and games with moving imports for speed introduces.

Bringing in @ichard26 to ensure this does not break mypyc - I think it's time we add an action to build with mypyc @ichard26 - I'm happy to help here to avoid regressions there.

I am interested to hear others thoughts.

fabioz:
Ok, I've just fixed the tests and added back methods which had been removed from the public API.

Also, lazy imports are added very judiciously in the PR, so, I hope this would be OK (I personally nowadays prefer lazy imports project-wide, but this really needs to be a conscious decision from the project and I understand this isn't the case for black, so, I tried to leave changes to a minimum to make it work yet still bring a substantial benefit for the use case where those modules aren't used at all).

As for the tests which were failing due to the imports, I don't think that something as `black.ProcessPoolExecutor` would be expected as a part of its API, so, fixing the test seems fair... the issue here is more on using monkey-patching for that (it's easy to do, but it really does makes the tests a bit more brittle).

src/black/__init__.py
```diff
@@ -500,6 +505,18 @@ def main(
...
+def compute_exclude_and_gitignore(
+    exclude: Optional[Pattern[str]], root: Path
+) -> Tuple[Pattern[str], Optional["PathSpec"]]:
+    gitignore: Optional["PathSpec"]
+    if exclude is None:
+        exclude = re_compile_maybe_verbose(DEFAULT_EXCLUDES)
+        gitignore = get_gitignore(root)
+    else:
+        gitignore = None
+    return exclude, gitignore
```
felix-hilden, line 517: While I appreciate the idea of making small functions, I feel this is more indirection because the decision of computing the exclude pattern is moved to the computing function. And if the decision was moved back, this would be a two-line wrapper to two simple calls. But this is highly debatable. Thoughts?

This might be better if we get rid of the `exclude_and_gitignore` sentinel as I suggested in another comment, because we could unpack directly: `ex, ign = compute(ex, root)`.

src/black/__init__.py
```diff
@@ -484,6 +484,11 @@ def main(
                 report=report,
             )
         else:
+            # Load reformat_many as lazily as possible as loading it will
+            # require support for asyncio and concurrent.futures which
+            # are slow to import.
+            from black.reformat_many import reformat_many
```
felix-hilden, line 490: +1 for this, seems like a good win for essentially nothing.

felix-hilden (commented):
Overall, this must be a positive. All of these dynamic imports makes me a bit iffy style-wise, but if the performance gains are deemed worth it, then it's fine by me 👍 Some comments mostly about the ignore handling below.

src/black/__init__.py
```diff
@@ -1361,9 +1264,50 @@ def patch_click() -> None:
             module._verify_python_env = lambda: None  # type: ignore
 
 
+def shutdown(loop: "asyncio.AbstractEventLoop") -> None:
```
ichard26, line 1267: Are these used anywhere?

src/black/__init__.py
```diff
@@ -1361,9 +1264,50 @@ def patch_click() -> None:
...
+
+
+def cancel(tasks: Iterable["asyncio.Task[Any]"]) -> None:
+    """asyncio signal handler that cancels all `tasks` and reports to stderr."""
+    from black import concurrency
+
+    return concurrency.cancel(tasks)
+
+
+async def schedule_formatting(
```
ichard26, line 1281: Same thing, is this used anywhere? Right now Black doesn't have an official API (and even then I doubt multiprocessing code should be treated as public).

src/black/__init__.py
```diff
@@ -1361,9 +1264,50 @@ def patch_click() -> None:
...
-    freeze_support()
+    global __BLACK_MAIN_CALLED__
+    __BLACK_MAIN_CALLED__ = True
+    for arg in sys.argv:
+        if "multiprocessing" in arg:
+            # Don't import multiprocessing unless really needed.
+            from multiprocessing import freeze_support
+
+            freeze_support()
+            break
```
ichard26, line 1310: Is there documentation on that this is a valid way of using freeze_support on Windows? This feels exceedingly sketchy.

ichard26 (commented):
First review pass, nothing majorly wrong but I got quite a few questions.

I'll take a closer look at this tomorrow -- in particular I'll make sure this doesn't break compilation with mypyc which brings its own import time reductions :)

Thank you for working on this!

ichard26:
FYI I picked up your PR and pushed it to completion, hopefully you don't mind. Thanks for the PR in the first place!

fabioz:
@ichard26 thank you for taking up the PR to completion (I'm sorry I didn't have more time available to continue the work to get it to completion and I really appreciate that you were able to do that).

PR #2657, opened 2021-11-29
Remove hidden import from PyInstaller build

jalaziz:
### Description

The recent 2021.4 release of pyinstaller-hooks-contrib now contains a
built-in hook for platformdirs. Manually specifying the hidden import
arg should no longer be needed.

The latest release of pyinstaller-hooks-contrib is installed when installing
PyInstaller, so reverting #2466 is all that is needed.

### Checklist - did you ...

I did not add a changelog entry for this as there should be no change in behavior to the end user.

### DIFF
--- a/.github/workflows/upload_binary.yml
+++ b/.github/workflows/upload_binary.yml
@@ -16,17 +16,14 @@ jobs:
             pathsep: ";"
             asset_name: black_windows.exe
             executable_mime: "application/vnd.microsoft.portable-executable"
-            platform: windows
           - os: ubuntu-20.04
             pathsep: ":"
             asset_name: black_linux
             executable_mime: "application/x-executable"
-            platform: unix
           - os: macos-latest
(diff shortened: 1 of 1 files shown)


ichard26 (approved):
Thank you for keeping tabs on this mate! Always nice to simply the CD pipeline so it has less room to break in an already chaotic time.

cooperlees (approved):
Awesome - Thanks!

Did you test we got functioning executable binaries from this PR? Should just be able to run it on respective platforms right?

jalaziz:
> Awesome - Thanks!
> 
> Did you test we got functioning executable binaries from this PR? Should just be able to run it on respective platforms right?

I tested the macOS binary, but also confirmed that the hooks were being picked up on all the platforms based on the logs.

cooperlees:
Thanks. Sounds good to me then!

PR #2661, opened 2021-11-30
Fix line generation for `match match:` / `case case:`

### LINKED ISSUE #2660: match/case statements containing those soft-keywords gets wrongly separated
The following input;
```py

match match:
    case case:
        pass
```

Will be formatted as

```py

match
match:
    case
    case:
        pass
```

which is unfortunately invalid. Though the good thing is that, the parser works but apparently I choose to wrong visitor method on the initial implementation. So we need a separate one to handle soft-keywords better.

isidentical:
Fixes #2660

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -127,7 +127,7 @@ class LineGenerator(Visitor[Line]):
         """Visit a statement.
 
         This implementation is shared for `if`, `while`, `for`, `try`, `except`,
-        `def`, `with`, `class`, `assert`, `match`, `case` and assignments.
+        `def`, `with`, `class`, `assert`, and assignments.
 
         The relevant Python language `keywords` for a given statement will be
         NAME leaves within it. This methods puts those on a separate line.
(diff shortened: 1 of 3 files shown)


src/black/linegen.py
```diff
@@ -142,6 +142,14 @@ def visit_stmt(
 
             yield from self.visit(child)
 
+    def visit_match_case(self, node: Node) -> Iterator[Line]:
+        """Visit either a match or case statement."""
+        normalize_invisible_parens(node, parens_after=set())
+
+        yield from self.line()
```
isidentical, line 149: I am not sure if this is the correct solution; it originates from the assumptions that for each `match_statement`/`case_statement` node, there will be only one valid `match`/`case` keyword for the context. So we simply yield a line; and for all following encounters of child nodes, we process them as usual.

JelleZijlstra (approved):

tests/data/pattern_matching_extras.py
```diff
@@ -23,10 +23,10 @@
 
 def func(match: case, case: match) -> case:
     match Something():
-        case another:
-            ...
         case func(match, case):
             ...
+        case another:
```
isidentical, line 28: I've fixed some syntax errors in this file (e.g a name pattern can't be followed by anything else) to ensure everything is also parsable by the python parser.

isidentical:
@JelleZijlstra that was fast 🚀

PR #2663, opened 2021-12-02
remove regex dependency

JelleZijlstra:
We were no longer using it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### _Black_
 
+- Remove dependency on `regex` (#2663)
 - Cell magics are now only processed if they are known Python cell magics. Earlier, all
   cell magics were tokenized, leading to possible indentation errors e.g. with
   `%%writefile`. (#2630)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,12 +4,11 @@
 
 ### _Black_
 
-- Remove dependency on `regex` (#2663)
 - Cell magics are now only processed if they are known Python cell magics. Earlier, all
   cell magics were tokenized, leading to possible indentation errors e.g. with
   `%%writefile`. (#2630)
 - Fix Python 3.10 support on platforms without ProcessPoolExecutor (#2631)
-- Reduce usage of the `regex` dependency (#2644)
+- Remove dependency on `regex` (#2644) (#2663)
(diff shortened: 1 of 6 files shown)


ichard26:
I can handle the pipenv related changes as I suspect the changes might break compatibility with older versions.

CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 ### _Black_
 
+- Remove dependency on `regex` (#2663)
```
ichard26, line 7: Should we keep the "reduce usage of the `regex` dependency ..." entry. In my opinion it looks a little weird especially since it's _after_ we state we removed regex.

JelleZijlstra: Good point, done

JelleZijlstra:
> I can handle the pipenv related changes as I suspect the changes might break compatibility with older versions of Python.

Thanks! Feel free to push to this branch. Hopefully I got CI clean this time.

setup.py
```diff
@@ -101,7 +101,6 @@ def find_python_files(base: Path) -> List[Path]:
         "platformdirs>=2",
         "tomli>=0.2.6,<2.0.0",
         "typed-ast>=1.4.2; python_version < '3.8' and implementation_name == 'cpython'",
-        "regex>=2021.4.4",
```
cooperlees, line 104: So nice.

src/black/trans.py
```diff
@@ -453,7 +453,7 @@ def make_naked(string: str, string_prefix: str) -> str:
             # with 'f'...
             if "f" in prefix and "f" not in next_prefix:
                 # Then we must escape any braces contained in this substring.
-                SS = re.subf(r"(\{|\})", "{1}{1}", SS)
+                SS = re.sub(r"(\{|\})", r"\1\1", SS)
```
cooperlees, line 456: This will never have unicode to match right so we should be safe ... :D

cooperlees (approved):
Nice work all. This will make black a much lighter and reliable install.

src/black/trans.py
JelleZijlstra: As far as I can tell the missing unicode support is just about character classes, so some "word" characters don't match `\w` in re but do in regex. That doesn't apply here.

PR #2665, opened 2021-12-02
Ensure match/case are recognized as statements

### LINKED ISSUE #2664: py310 match: one-line case breaks
**Describe the bug**
In `python3.10` the `case` black can be written in the same line as the `case` keyword. However, this breaks `black`.

**To Reproduce**
Take this example code
```python
# example.py
x = 5
match x:
    case 5: print("it works")
```
It runs under `python3.10`:
```sh
python example.py
it works
```
However, `black` crashes:
```
black --target-version py310 example.py
error: cannot format example.py: INTERNAL ERROR: Black produced invalid code on pass 1: expected an indented block after 'case' statement on line 3 (<unknown>, line 4). Please report a bug on https://github.com/psf/black/issues.  This invalid output might be helpful: /var/folders/cz/v1lxwy5579502ksd81z14_t40000gp/T/blk_inxhkvdg.log
Oh no! 💥 💔 💥
1 file failed to reformat.
```
After manually reformatting the example as follows
```python
x = 5
match x:
    case 5:
        print("it works")
```
`black` stops crashing.

**Expected behavior**
`black` should not crash.

**Environment**
* macOS Big Sur 11.6
* `python 3.10.0`
* `black --version`: `black, 21.11b1 (compiled: no)`

isidentical:
Resolves #2664

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
   `match a, *b:` (#2639) (#2659)
 - Fix `match`/`case` statements that contain `match`/`case` soft keywords multiple
   times, like `match re.match()` (#2661)
+- Fix `case` statements with an inline body (#2665)
 - Fix assignment to environment variables in Jupyter Notebooks (#2642)
 - Add `flake8-simplify` and `flake8-comprehensions` plugins (#2653)
 - Fix determination of f-string expression spans (#2654)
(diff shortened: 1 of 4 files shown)


JelleZijlstra (approved):

JelleZijlstra:
Thank you!

PR #2667, opened 2021-12-03
Make star-expression spacing consistent in match/case

### LINKED ISSUE #2666: Black unnecessarily separating star from the expression on match statement
Input:
```py
match a, *b(), c:
    case d, *f, g:
        pass
```

Expected:
```py
match a, *b(), c:
    case d, *f, g:
        pass
```

Got:
```py
match a, * b(), c:
    case d, *f, g:
        pass
```

isidentical:
Resolves #2666


CHANGES.md
```diff
@@ -14,6 +14,7 @@
 - Fix `match`/`case` statements that contain `match`/`case` soft keywords multiple
   times, like `match re.match()` (#2661)
 - Fix `case` statements with an inline body (#2665)
+- Fix styling of starred expressions inside `match` subject (#2666)
```
onerandomusername: not sure if these are typically put in a specific order, from lowest pr to highest pr, as a result this would go below the next one

tests/data/pattern_matching_extras.py
```diff
@@ -77,3 +77,8 @@ def func(match: case, case: match) -> case:
 match match:
     case case:
         pass
+
+
+match a, *b(), c:
```
onerandomusername, line 82: perhaps also add a test case with `*a`, and no b?

CHANGES.md
ichard26: It honestly doesn't matter. We put 'em in no real order.

tests/data/pattern_matching_extras.py
isidentical: That is not a valid syntax.
```py
>>> match *a:
  File "<stdin>", line 1
    match *a:
            ^
SyntaxError: invalid syntax
```
For one regular starred and one normal mixed, see the same file. We have an example for that.

ichard26: Ah my bad, I should perhaps play with pattern matching for once haha.

onerandomusername: Well then, I need to play with pattern matching too.

JelleZijlstra (approved):

onerandomusername (approved):
lgtm!

Thanks for being patient with my suggested changes... which weren't even needed at all!

PR #2668, opened 2021-12-03
tell users to use -t py310

JelleZijlstra:


isidentical:
This is very cool! Just one question, I know it doesn't sound really good but considering the algorithm is still experimental, would it make sense to run `driver.parse_string()` on a separate thread with a fixed timeout (idk, `.join(timeout=3)`?) (in case it gets really really slow, unlikely but still, a nice precaution)?

ichard26 (approved):
Seems good to me, thanks!

JelleZijlstra:
@isidentical thanks! We discussed on Discord and decided to skip the extra timeout to avoid extra complexity.

PR #2671, opened 2021-12-05
blib2to3 can raise TokenError and IndentationError too

ichard26:
Fix the regression of GH-2317 caused by GH-2668. I explicitly made this PR from my fork of psf/black so the workflows are triggered from the PR event which should catch post-merge issues.

- [x] Add a CHANGELOG entry if necessary? -> the regression hasn't been released yet
- [x] Add / update tests if necessary? -> CI will test this


JelleZijlstra (approved):
Thank you!

PR #2676, opened 2021-12-06
chore: dump docs deps and pre-commit hooks

ichard26:
### Description

It's been a while since I last checked these pins.

### Checklist - did you ...

- [x] Add new / update outdated documentation? -> seems fine to me.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,7 @@ repos:
           - flake8-simplify
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.910
+    rev: v0.910-1
     hooks:
       - id: mypy
         exclude: ^docs/conf.py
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -41,7 +41,7 @@ repos:
           - flake8-simplify
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.910
+    rev: v0.910-1
```
JelleZijlstra, line 44: Kind of weird but it does seem to exist

JelleZijlstra (approved):

cooperlees (commented):
Should we add pre-commit + one of the requirements updaters to hit these files? (Can be separate PR)

.pre-commit-config.yaml
ichard26: I just ran `pre-commit autoupdate` :shrug:

MarcoGorelli:
A bunch of projects are using [pre-commit.ci](https://pre-commit.ci/) for this, which comes with a regular `autoupdate` built-in - for what my suggestion's worth, I really recommend it!

ichard26:
Yeah, we'll have to look into it as I'm not sure whether pre-commit.ci will play nicely with our local black hook. I'm down to add dependabot for docs/requirements.txt and test_requirements.txt though!

PR #2679, opened 2021-12-07
django and pandas don't need changes

JelleZijlstra:


src/black_primer/primer.json
```diff
@@ -95,7 +95,7 @@
     },
     "pandas": {
       "cli_arguments": ["--experimental-string-processing"],
-      "expect_formatting_changes": true,
+      "expect_formatting_changes": false,
```
JelleZijlstra, line 98:
```suggestion
      "expect_formatting_changes": true,
```
And now it needs changes again? Compare https://github.com/psf/black/runs/4446024299?check_suite_focus=true

cooperlees (approved):
Wow. Crazy of me not expecting changes here - I don't think django uses black.

ichard26:
Seems like this isn't the case anymore? Not sure what happened here to be honest.

JelleZijlstra:
Yeah, I guess they must have both been changed.

Issue #2680, opened 2021-12-07
#fmt: skip results in change in second pass error


MJohnson459:
**Describe the bug**

When using `#fmt: skip`, if there is no space between the `#` and `f` then black fails to format due to different result in the second pass.

**To Reproduce**

For example, take this code:
```python
def test():
    x = 1 #fmt: skip
```

https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ABwAFNdAD2IimZxl1N_WlbvK5V_5PM4Nceap1_OfA9LqGqzctuugz_qJKhKcm0xODvQpn1SGiQJFVPjQaZCfva5pKHycj-2VXHw1ESugA17W0JffVNXXrlGAAC1vzI706u2xQABb3Hpg-HlH7bzfQEAAAAABFla

The resulting error is:

```
INTERNAL ERROR: Black produced different code on the second pass of the formatter.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_9vf9y7f5.log

Mode(target_versions=set(), line_length=88, string_normalization=True, is_pyi=False, is_ipynb=False, magic_trailing_comma=True, experimental_string_processing=False)
--- source
+++ first pass
@@ -1,2 +1,6 @@
 def test():
     x = 1 #fmt: skip
+  # fmt: skip
+
+
+# fmt: skip
--- first pass
+++ second pass
@@ -1,6 +1,3 @@
 def test():
-    x = 1 #fmt: skip
-  # fmt: skip
-
-
-# fmt: skip
+x = 1 #fmt: skip
+  # fmt: skip  # fmt: skip
```

**Expected behavior**

It should format correctly. I assume this is due to the first pass it being an invalid `fmt` specified without the space and so one is added which makes it valid for the second pass.

**Environment**

Black's version: black, 21.12b0 (compiled: no)

andrerod22:
Replicated the issue on my machine, looking into the problem now.

felix-hilden:
I think we have two options here:
- Accept any spacing in the skip comment
- Format the comment but not the line

Personally I'd be inclined to accept any amount of spacing, since the comment is also part of the line 😄 Thoughts?

We should also have a look if this is an issue with `# fmt: off`.

andrerod22:
Ran the following with black and the #fmt: off was resolved with the space added
```
def test():
    #fmt: off
[...]

PR #2682, opened 2021-12-10
# fmt skip internal error fix (space between # and f)

andrerod22:
### Description
Bug report: https://github.com/psf/black/issues/2680
Test that replicates bug (see fmtskip7.py in tests/data): 
```py
a, b = 1, 2
c =   6 #fmt: skip
d = 5
```
Resolve: It's seen that of all the formatting options, the inline option, or # fmt skip causes an internal error for black, when the comment should be reformatted like # fmt off. A proposal for this issue was made in the  fmtskip_nospace branch which aims to hard code logic handling this bug within the make_comment function of comments.py. 

Note: This does NOT handle arbitrary spacing for the # fmt: skip which should be handled in another pull request. 

Issues with current pull request: Building on different platforms fails due to the unit test fmtskip7.py failing when running pre-config. An investigation is further needed to resolve this issue.
### Checklist
- [X] Add / update tests if necessary? See fmtskip7.py in tests/data


felix-hilden:
Thanks for the PR! For housekeeping, the list appears as checkboxes if you leave an empty space for "no" and put in an "x" for "yes".

Some points: did you run tests before submitting? Now your test is failing, at least in the CI. So I'm guessing it does not work locally either. Also for the record, I'd like to take care of the multi-space case while we're at it. But I'll be fine if other maintainers just want to push this in.

andrerod22:
Looking back at the unit test that failed, the error is within _assert_format_equal
Specifically, the expected output is the following:
image

And the actual output differs here 
image


Notice how the NAME variable in the expression statement is different. It seems as though, Black is expecting just the variable I had written on the same line and completely ignores the #fmt: skip comment on the same line. This is strange behavior and I will bring it up in the discord later.

JelleZijlstra:
Any updates here? I agree with @felix-hilden that it'd be good to handle other spacing edge cases at the same time.

JelleZijlstra:
It appears #2970 already fixed this crash in a more general way, so I'm going to close this PR. Thanks for your contribution!

PR #2686, opened 2021-12-12
Support as-expressions on dict items

### LINKED ISSUE #2685: As expressions on case statements are broken inside dicts
This is a left-over from my [initial comment](https://github.com/psf/black/issues/2242#issuecomment-952162467) at #2242. When you have a dict with the item as an `blabla as blabla`, black can't parse it;
```py
match something:
    case {
        "key": key as key_1,
        "password": PASS.ONE | PASS.TWO | PASS.THREE as password,
    }:
        pass
    case {"maybe": something(complicated as this) as that}:
        pass
```

isidentical:
Fixes #2685

### DIFF
--- a/src/blib2to3/Grammar.txt
+++ b/src/blib2to3/Grammar.txt
@@ -168,8 +168,8 @@ subscript: test [':=' test] | [test] ':' [test] [sliceop]
 sliceop: ':' [test]
 exprlist: (expr|star_expr) (',' (expr|star_expr))* [',']
 testlist: test (',' test)* [',']
-dictsetmaker: ( ((test ':' test | '**' expr)
-                 (comp_for | (',' (test ':' test | '**' expr))* [','])) |
+dictsetmaker: ( ((test ':' asexpr_test | '**' expr)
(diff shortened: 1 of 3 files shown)


src/blib2to3/Grammar.txt
```diff
@@ -168,8 +168,8 @@ subscript: test [':=' test] | [test] ':' [test] [sliceop]
 sliceop: ':' [test]
 exprlist: (expr|star_expr) (',' (expr|star_expr))* [',']
 testlist: test (',' test)* [',']
-dictsetmaker: ( ((test ':' test | '**' expr)
-                 (comp_for | (',' (test ':' test | '**' expr))* [','])) |
+dictsetmaker: ( ((test ':' asexpr_test | '**' expr)
```
isidentical, line 171: `asexpr_test` is simply `test ['as' test]`, so I do not anticipate any behavior change on regular dicts.

JelleZijlstra (approved):

ichard26:
Thank you @isidentical!

PR #2698, opened 2021-12-15
Standardize type comments

### LINKED ISSUE #2097: is_type_comment() is wrong
```
    t = leaf.type
    v = leaf.value
```

13Ducks:
### Description

Fixes #2097 and standardizes type comments to always have 1 space after the # and 1 space after the :

As per the issue, the functions `is_type_comment()` and `contains_pragma_comment()` have been modified to return true even with any extra whitespace. Type comments without non-breaking spaces are now stripped of any extra whitespace as well.

Waiting on confirmation from the issue that spaces before the : are acceptable (`#    type   :    something`).

### Checklist - did you ...

^ Is documentation needed for this change? If so, I can add it.


ichard26:
This was discussed on Python Discord but the summary of it is that we can't really normalize type comments until all of the major AST libraries that provide first-class type comment support are liberal in the whitespace they understand. It wouldn't be good to have a type comment that's ignored by typed-ast transformed into one that is noticed which then affects a type checker like mypy.

JelleZijlstra (changes requested):
I don't remember the discussion @ichard26 alludes to above. Could you provide some details on how existing tools (typed-ast, pyright, pyre, ...) interpret spaces in type comments?

ichard26:
Before I forget to link it again: https://discord.com/channels/267624335836053506/846434317021741086/921790074922864670

src/black/comments.py
```diff
@@ -271,7 +276,7 @@ def contains_pragma_comment(comment_list: List[Leaf]) -> bool:
         pylint).
     """
     for comment in comment_list:
-        if comment.value.startswith(("# type:", "# noqa", "# pylint:")):
+        if re.match(r"#\s*(type\s*:|pylint\s*:|noqa\s*)", comment.value):
```
CarliJoy, line 279:
```suggestion
        if re.match(r"#\s*(type\s*:|pylint\s*:|noqa\s*|nosec\s*)", comment.value):
```
Would be nice if [Bandit](https://bandit.readthedocs.io/en/latest/config.html#suppressing-individual-lines) is supported as well.

CarliJoy:
@13Ducks giving up on this MR ?

PR #2700, opened 2021-12-15
Unpacking on flow constructs (return/yield) now implies 3.8+

### LINKED ISSUE #2496: Star operator without enclosing within parentheses
**Describe the bug**

Returning a tuple containing star operator without enclosing with parentheses leads to a `Cannot parse` error.

Oh no! 💥 💔 💥

**To Reproduce**

```python
def apply_to_last(input, func):
    *header, last = input
    return *header, func(last)

# which is valid code doing this:
assert apply_to_last([1,2,3], lambda x: x+10) == (1, 2, 13)
```

And try to blackify...

**Expected behavior**

I expect it to not choke. 

There is a *workaround* -- enclosing in parentheses

```python
def apply_to_last(input, func):
    *header, last = input
    return (*header, func(last))
```

But doing this makes the linter complain that _"Remove redundant parentheses"._

- Version: `axblack 20210720`
- MacOS 10.15.7
- Python 3.8.6

isidentical:
Resolves #2496

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
   (#2686)
 - No longer color diff headers white as it's unreadable in light themed terminals
   (#2691)
+- Tuple unpacking on `return` and `yield` constructs now implies 3.8+ (#2700)
 
 ## 21.12b0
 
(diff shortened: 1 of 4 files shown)


ichard26 (approved):
Weee thank you!

PR #2701, opened 2021-12-16
Use 'python -m build' to build wheel and source distributions

mwtoews:
### Description

This PR uses modern approaches for building wheel and source distributions.

According to the [build docs](https://pypa-build.readthedocs.io/en/latest/differences.html):
> build is roughly the equivalent of `setup.py sdist bdist_wheel` but with [PEP 517](https://www.python.org/dev/peps/pep-0517/) support

So the modern equivalent is simply:
```
python -m build
```
which builds a `*.whl` and `*.tar.gz` file in `dist/`. Setuptools and wheel do not need to be installed, as these build dependancies are taken from pyproject.toml

### Checklist - did you ...

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -15,14 +15,14 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
 
-      - name: Install latest pip, setuptools, twine + wheel
+      - name: Install latest pip, build, twine
         run: |
-          python -m pip install --upgrade pip setuptools twine wheel
+          python -m pip install --upgrade --disable-pip-version-check pip
+          python -m pip install --upgrade build twine
 
-      - name: Build wheels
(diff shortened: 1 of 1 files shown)


ichard26 (approved):
Yay for more modern packaging standards and tooling :)

thanks for noticing and fixing this!

mwtoews:
Note that this PR does not touch any of the checks. There's only a few commands which could be double checked in a local venv.

cooperlees (approved):
Nice - Thanks. TIL ... Wonder where these things are announced. So simple.

Thanks for the PR! Surprised @hugovk didn't beat you since he's a contributor over there too.

hugovk:
Here's the long vesion of "don't invoke setup.py directly" with a handy table in the summary:

* https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

I don't see any other direct uses in Black. Thanks for the PR!

Issue #2681, opened 2021-12-07
Writing failed: [Errno 2] No such file or directory: '/home/user/.cache/black/21.11b1/tmpb3yqz52h'


avatar-lavventura:
When I update `black`, or have a clean install I keep seeing following message:

```bash
load_grammar()] Generating grammar tables from /home/alper/venv/lib/python3.8/site-packages/blib2to3/Grammar.txt
load_grammar()] Writing grammar tables to /home/alper/.cache/black/21.11b1/Grammar3.8.10.final.0.pickle
load_grammar()] Writing failed: [Errno 2] No such file or directory: '/home/alper/.cache/black/21.11b1/tmpnwk07_xq'
load_grammar()] Generating grammar tables from /home/alper/venv/lib/python3.8/site-packages/blib2to3/PatternGrammar.txt
load_grammar()] Writing grammar tables to /home/alper/.cache/black/21.11b1/PatternGrammar3.8.10.final.0.pickle
load_grammar()] Writing failed: [Errno 2] No such file or directory: '/home/alper/.cache/black/21.11b1/tmpb3yqz52h'
```

I always have to do: `mkdir /home/user/.cache/black/21.11b1`, where could it be done automatically by the `black`?

How can I prevent this error from showing up? Would it be possible to for black to force create `/home/user/.cache/black/21.*` folders after each upgrade?

michaelaye:
`black` should be allowed to do what it needs to do inside its own `.cache/black` folder, right?

avatar-lavventura:
@michaelaye Hm I believe `black` should be allowed to create files inside `.cache/black` folder. I think the problem was `cache grammar tables` are tried to generated before the 21.11b1` folder is created, which was causing the error messages.

I think this is resolved with the mentioned commit.

michaelaye:
Well #2702 was closed with the comment that the issue would not be picked up for a while, so it’s not resolved, it seems.

avatar-lavventura:
Ah you are right, I misread  it :-(

During the installation of `black` I tried to come up with this simple hack but the origin of the problem remains.

```bash
sudo chown $(logname) -R $HOME/.cache/black
black_version=$(pip freeze | grep black | sed 's|black==||g')
if [ "$black_version" != "" ]; then
    rm -rf $HOME/.cache/black/*
    mkdir -p $HOME/.cache/black/$black_version
fi
```

Issue #2058, opened 2021-03-23
Unable to disable logging for moldule blib2to3.pgen2.driver


vlurgio:
**Describe the bug** A clear and concise description of what the bug is.
There is no way to disable logging of `blib2to3.pgen2.driver` class as the logger is created in the `__init__` method and the optional logger isn't passed into the class from the top level black module. (driver is created on line 1130)
**To Reproduce** Steps to reproduce the behavior:

Anytime the format_str() function is called you get logging like
```
2021-03-23 18:52:48 blib2to3.pgen2.driver[86854] INFO Generating grammar tables from <filepath>
2021-03-23 18:52:48  blib2to3.pgen2.driver[86854] INFO Writing grammar tables to <filepath>
2021-03-23 18:52:48  blib2to3.pgen2.driver[86854] INFO Writing failed: [Errno 2] No such file or directory: '<filepath>
2021-03-23 18:52:48  blib2to3.pgen2.driver[86854] INFO Writing grammar tables to <filepath>
2021-03-23 18:52:48  blib2to3.pgen2.driver[86854] INFO Writing failed: [Errno 2] No such file or directory: <filepath>
```
The code I'm using is 
```python
from black import format_str, FileMode

format_str(out_string, mode=FileMode())
```
**Expected behavior** A clear and concise description of what you expected to happen.
using something like
```python
import logging
logger = logging.getLogger("black")
logger.setLevel(logging.ERROR)
```
or even
```python
import logging
logger = logging.getLogger("blib2to3")
logger.setLevel(logging.ERROR)
```
would disable logging
**Environment (please complete the following information):**

- Version: 19.10b0
- OS and Python version: Mac osx 3.6.9

**Does this bug also happen on master?** To answer this, you have two options:
The code still has logger being created in __init__ on master so this problem still exists. 

**Additional context** Add any other context about the problem here.

Issue #1143, opened 2019-11-06
Improve the the blib2to3 grammar caching mechanism  


EWouters:
I get an error upon launching Jupyter Lab with the [jupyterlab_code_formatter](https://github.com/ryantam626/jupyterlab_code_formatter) extension enabled:

```
[I 10:37:25.686 LabApp] Node v10.8.0

[I 10:37:26.075 LabApp] Build is up to date
Generating grammar tables from C:\Users\%USERNAME%\Anaconda3\lib\site-packages\blib2to3\Grammar.txt
Writing grammar tables to C:\Users\%USERNAME%\AppData\Local\black\black\Cache\18.9b0\Grammar3.7.1.final.0.pickle
Writing failed: [Errno 2] No such file or directory: 'C:\\Users\\%USERNAME%\\AppData\\Local\\black\\black\\Cache\\18.9b0\\Grammar3.7.1.final.0.pickle'
Generating grammar tables from C:\Users\%USERNAME%\Anaconda3\lib\site-packages\blib2to3\PatternGrammar.txt
Writing grammar tables to C:\Users\%USERNAME%\AppData\Local\black\black\Cache\18.9b0\PatternGrammar3.7.1.final.0.pickle
Writing failed: [Errno 2] No such file or directory: 'C:\\Users\\%USERNAME%\\AppData\\Local\\black\\black\\Cache\\18.9b0\\PatternGrammar3.7.1.final.0.pickle'
```

After manually creating the folder `C:\\Users\\%USERNAME%\\AppData\\Local\\black\\black\\Cache\\18.9b0` and relaunching Jupyter lab the error disappears (even though the folder seems to stay empty).

The error appears to originate from [this](https://github.com/psf/black/blob/3e60f6d454616a795acb1e3e2b43efa979de4f46/blib2to3/pgen2/driver.py#L202) line. It might be related to #192. I think [this](https://github.com/psf/black/blob/9b484d1bcc2e15dcd5544cddab729c76b4d1d2e9/blib2to3/pgen2/grammar.py#L112) line fails because the cache directory does not exist, but I don't understand why it doesn't exist.

Other users confirmed this happens on linux and with version 19.10b as well in [this](https://github.com/ryantam626/jupyterlab_code_formatter/issues/10) issue.

I tried to find the cause of this issue in the extension, but it doesn't seem to do anything out of the ordinary with the cache folder, so I was hoping someone here can tell me why the cache folder would not exist?

EWouters:
[...]

PR #2702, opened 2021-12-16
Cache grammar tables after finishing, not at startup

ichard26:
### Description

*and also get rid of the confusing log messages related to the grammar cache which just confuses everyone and leads to confused issues to be opened

Resolves GH-2681.
Resolves GH-2058.
Resolves GH-1143.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@
   (#2686)
 - No longer color diff headers white as it's unreadable in light themed terminals
   (#2691)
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
It looks like we'd still load them at import time due to the `pygram.initialize(CACHE_DIR)` line in `black/nodes.py`.

ichard26:
I won't be picking this up for a while.

Issue #2704, opened 2021-12-16
pytest 7: ModuleNotFoundError: No module named '_pytest.store'


hroncok:
Hello, in Fedora, we are testing the impact of pytest 7.0.0rc1 on our packages.

https://github.com/psf/black/blob/61fe8418cc868723759fb08d76adab1542bb7630/tests/optional.py#L24

Here we get:

```
+ /usr/bin/pytest -rs
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/_pytest/config/__init__.py", line 732, in import_plugin
    __import__(importspec)
  File "/usr/lib/python3.10/site-packages/_pytest/assertion/rewrite.py", line 171, in exec_module
    exec(co, module.__dict__)
  File "/builddir/build/BUILD/black-21.11b0/tests/optional.py", line 24, in <module>
    from _pytest.store import StoreKey
ModuleNotFoundError: No module named '_pytest.store'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/bin/pytest", line 8, in <module>
    sys.exit(console_main())
  File "/usr/lib/python3.10/site-packages/_pytest/config/__init__.py", line 188, in console_main
    code = main()
  File "/usr/lib/python3.10/site-packages/_pytest/config/__init__.py", line 146, in main
    config = _prepareconfig(args, plugins)
  File "/usr/lib/python3.10/site-packages/_pytest/config/__init__.py", line 325, in _prepareconfig
    config = pluginmanager.hook.pytest_cmdline_parse(
  File "/usr/lib/python3.10/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "/usr/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/usr/lib/python3.10/site-packages/pluggy/_callers.py", line 55, in _multicall
    gen.send(outcome)
  File "/usr/lib/python3.10/site-packages/_pytest/helpconfig.py", line 102, in pytest_cmdline_parse
    config: Config = outcome.get_result()
  File "/usr/lib/python3.10/site-packages/pluggy/_result.py", line 60, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/usr/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
[...]

PR #2705, opened 2021-12-17
Support pytest 7

hroncok:
### Description

Fixes https://github.com/psf/black/issues/2704

### Checklist - did you ...

- [x] Add a CHANGELOG entry if necessary? not necessary
- [x] Add / update tests if necessary? not necessary
- [x] Add new / update outdated documentation? not necessary

### DIFF
--- a/tests/optional.py
+++ b/tests/optional.py
@@ -21,7 +21,12 @@ import re
 from typing import FrozenSet, List, Set, TYPE_CHECKING
 
 import pytest
-from _pytest.store import StoreKey
+
+try:
+    from pytest import StashKey
+except ImportError:
+    # pytest < 7
+    from _pytest.store import StoreKey as StashKey
 
 log = logging.getLogger(__name__)
 
@@ -33,8 +38,8 @@ if TYPE_CHECKING:
     from _pytest.nodes import Node
 
 
-ALL_POSSIBLE_OPTIONAL_MARKERS = StoreKey[FrozenSet[str]]()
-ENABLED_OPTIONAL_MARKERS = StoreKey[FrozenSet[str]]()
+ALL_POSSIBLE_OPTIONAL_MARKERS = StashKey[FrozenSet[str]]()
+ENABLED_OPTIONAL_MARKERS = StashKey[FrozenSet[str]]()
 
 
 def pytest_addoption(parser: "Parser") -> None:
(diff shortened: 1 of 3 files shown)


tests/test_no_ipynb.py
```diff
@@ -22,14 +22,14 @@ def test_ipynb_diff_with_no_change_single() -> None:
     assert expected_output in result.output
 
 
-def test_ipynb_diff_with_no_change_dir(tmpdir: tmpdir) -> None:
+def test_ipynb_diff_with_no_change_dir(tmp_path: tmp_path) -> None:
```
bluetech, line 25: The `tmp_path` value is a `pathlib.Path`, so this should be

```suggestion
def test_ipynb_diff_with_no_change_dir(tmp_path: pathlib.Path) -> None:
```

hroncok:
The second commit is not necessary to make pytest 7 work, but it feels more complete this way.

tests/test_ipynb.py
```diff
@@ -371,52 +371,52 @@ def test_ipynb_diff_with_no_change() -> None:
...
     monkeypatch.setattr(
         "black.files.jupyter_dependencies_are_installed", lambda verbose, quiet: True
     )
-    result = runner.invoke(main, [str(tmpdir)])
+    result = runner.invoke(main, [str(tmp_path)])
     assert "reformatted" in result.output
 
 
-def test_ipynb_flag(tmpdir: local) -> None:
+def test_ipynb_flag(tmp_path: local) -> None:
```
bluetech, line 417:
```suggestion
def test_ipynb_flag(tmp_path: pathlib.Path) -> None:
```

tests/optional.py
```diff
@@ -21,7 +21,12 @@
 from typing import FrozenSet, List, Set, TYPE_CHECKING
 
 import pytest
-from _pytest.store import StoreKey
```
bluetech, line 24: Additionally it would be good to replace uses of `_store` with `stash`. The former still works but will probably be removed at some point.

hroncok: config._store with config.stash?

bluetech: Yes

ichard26 (approved):
Beautiful -- thank you!

PR #2714, opened 2021-12-21
Define is name token to resolve some `type: ignore`s

MarcoGorelli:
### Description

Gets rid of a few `# type: ignore`s by using `TypeGuard`

### Checklist - did you ...


isidentical (commented):
This is definitely a great usage of `TypeGuards`. Just a question, would it make sense to put these to `black/nodes.py`? (Since from my understanding `black.nodes` has a lot of this sort of utilities but `blib2to3.pytree` is a more private package for internal parser structs)

MarcoGorelli:
Thanks for your review @isidentical ! Yes, good suggestion, I've moved the definition to nodes.py

JelleZijlstra (approved):

PR #2716, opened 2021-12-21
Support multiple top-level as-expressions on case statements

### LINKED ISSUE #2715: Black does not accept multiple top-level as-expressions on a case
I've modified my script from [the original issue](https://github.com/psf/black/issues/2242#issuecomment-952162467) to include the whole `Lib/test/test_patma.py`. It uncovered one potential bug, but other than the good news is we are fully compatible (even though it takes a ridiculously large amount of time to parse that file, twice (due to the second pass)). I still haven't figured out the best way to optimize, but that is on my bucket list. 

The only example we can't parse is the following:
```py
def test_patma_238(self):
    x = ((0, 1), (2, 3))
    match x:
        case ((a as b, c as d) as e) as w, ((f as g, h) as i) as z:
            y = 0
```

Where we have 2 top-level as-expressions (`blabla as bla, blabla2 as bla2`) it fails to parse since it doesn't recognize as expressions on the individual unparenthesized case expressions.

isidentical:
Fixes #2715

### DIFF
--- a/src/blib2to3/Grammar.txt
+++ b/src/blib2to3/Grammar.txt
@@ -247,5 +247,5 @@ subject_expr: (namedexpr_test|star_expr) (',' (namedexpr_test|star_expr))* [',']
 # cases
 case_block: "case" patterns [guard] ':' suite
 guard: 'if' namedexpr_test
-patterns: pattern ['as' pattern]
-pattern: (expr|star_expr) (',' (expr|star_expr))* [',']
+patterns: pattern (',' pattern)* [',']
+pattern: (expr|star_expr) ['as' expr]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@
   underlying SyntaxError (#2693)
 - Fix mapping cases that contain as-expressions, like `case {"key": 1 | 2 as password}`
   (#2686)
+- Fix cases that contain multiple top-level as-expressions, like `case 1 as a, 2 as b`
+  (#2716)
 - No longer color diff headers white as it's unreadable in light themed terminals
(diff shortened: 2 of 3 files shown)


JelleZijlstra (approved):

src/blib2to3/Grammar.txt
```diff
@@ -247,5 +247,5 @@ subject_expr: (namedexpr_test|star_expr) (',' (namedexpr_test|star_expr))* [',']
 # cases
 case_block: "case" patterns [guard] ':' suite
 guard: 'if' namedexpr_test
-patterns: pattern ['as' pattern]
-pattern: (expr|star_expr) (',' (expr|star_expr))* [',']
+patterns: pattern (',' pattern)* [',']
```
wbolster, line 250: maybe i'm missing something, but this seems to allow a trailing comma?

it's legal syntax:

```python
>>> x = 1
>>> match x:
...     case 1,:
...         print('yep')
... 
```

…but the trailing `,` is part of the pattern, turning it into a tuple, as this example shows:

```python
>>> x = (1,)
>>> match x:
...     case 1,:
...         print('yep')
... 
yep
```

isidentical: I am unsure what do you mean by it is turning into a tuple? The grammar has no semantical meaning, it only defines what is accepted as is or not. Would you mind giving a bit more explanation? The trailing comma is a valid syntax AFAIK.

wbolster: nevermind, i though the possibility of a trailing might've been included already in the `expr` on the line below already, but that seems not the case.

PR #2717, opened 2021-12-21
Remove usage of Pipenv, rely on good ol' `pip` and `virtualenv` in docs

ambv:
Per team discussion, none of us uses Pipenv for developing Black anymore and it's become a burden to have to update both `Pipfile` and `setup.py` with each dependency change.


docs/contributing/the_basics.md
```diff
@@ -10,37 +10,27 @@ You can use any operating system.
...
-If you haven't used `pipenv` before but are comfortable with virtualenvs, just run
-`pip install pipenv` in the virtualenv you're already using and invoke the command above
-from the cloned _Black_ repo. It will do the correct thing.
-
-Non pipenv install works too:
-
-```console
-$ pip install -r test_requirements.txt
-$ pip install -e .[d]
+$ python3 -m venv .venv
```
JelleZijlstra, line 13:
```suggestion
$ python3 -m venv .venv
$ source .venv/bin/activate
```
(Or something else on Windows?)

ambv: Ah, having to activate manually... I use [`vf`](https://virtualfish.readthedocs.io/en/latest/) these days so forgot about it. Windows users probably know how to create and activate their venvs. And those that don't we probably don't want to encourage to use Black from `main` (at least not **yet**!).

felix-hilden: I think it would suffice to mention just to use any virtual environment. Any user can then see the commands for their specific stuff. The deps are anyways installed from requirements. Or is there a difference I'm not aware of?

felix-hilden (approved):
Looks good, some unimportant thoughts below.

PR #2719, opened 2021-12-21
Update contributing wording

felix-hilden:
Yo, quick update candidate to our contributing docs wording now that we don't have pipenv. Could be better to let contributors choose their tool since dependencies are not managed with any specific tool.

### DIFF
--- a/docs/contributing/the_basics.md
+++ b/docs/contributing/the_basics.md
@@ -7,7 +7,8 @@ An overview on contributing to the _Black_ project.
 Development on the latest version of Python is preferred. As of this writing it's 3.9.
 You can use any operating system.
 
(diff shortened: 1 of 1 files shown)


docs/contributing/the_basics.md
```diff
@@ -7,7 +7,7 @@ An overview on contributing to the _Black_ project.
 Development on the latest version of Python is preferred. As of this writing it's 3.9.
 You can use any operating system.
 
-Install all development dependencies using:
+Install development dependencies inside a virtual environment of your choice:
```
JelleZijlstra, line 10:
```suggestion
Install development dependencies inside a virtual environment of your choice, for example:
```
To make it clear that people can do something else if they want

PR #2726, opened 2021-12-26
Hug power operators if its operands are "simple"

ichard26:
### Description

Since power operators almost always have the highest binding power in expressions, it's often more readable to hug it with its operands. The main exception to this is when its operands are non-trivial in which case the power operator will not hug, the rule for this is the following:

> For power ops, an operand is considered "simple" if it's only a NAME, numeric CONSTANT, or attribute access (chained attribute access is allowed), with or without a preceding unary operator. 

Fixes GH-538.
Closes GH-2095.

diff-shades results: https://gist.github.com/ichard26/ca6c6ad4bd1de5152d95418c8645354b

### Checklist - did you ...

- [x] This needs inline developer documentation cause these functions are magical


felix-hilden (commented):
Looks good, but some nits and discussion. I'm still quite weak on the transform implementation, but I trust the tests, which are quite comprehensive 👍

src/black/trans.py
```diff
@@ -71,6 +71,88 @@ def TErr(err_msg: str) -> Err[CannotTransform]:
...
+    """A transformer which normalizes spacing around power operators."""
+
+    # Performance optimization to avoid unnecessary Leaf clones and other ops.
+    for leaf in line.leaves:
+        if leaf.type == token.DOUBLESTAR:
+            break
+    else:
+        raise CannotTransform("No doublestar token was found in the line.")
+
+    def is_simple_lookup(index: int, step: Literal[1, -1]) -> bool:
```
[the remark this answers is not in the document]
ichard26: So for some reason I'm starting to really like nested functions as part of larger functions. The nested functions aren't intended for generic reuse (for now) but are logically a bit more separate if that makes sense. I'm happy to change this bit as I understand this may be not a style most would like.

src/black/linegen.py
```diff
@@ -22,7 +22,12 @@
 from black.strings import get_string_prefix, fix_docstring
 from black.strings import normalize_string_prefix, normalize_string_quotes
 from black.trans import Transformer, CannotTransform, StringMerger
-from black.trans import StringSplitter, StringParenWrapper, StringParenStripper
+from black.trans import (
+    StringSplitter,
+    StringParenWrapper,
+    StringParenStripper,
+    hug_power_op,
+)
```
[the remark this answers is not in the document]
ichard26: Oh I completely missed the trans.py import above this line. I don't have a preference eitherway.

ichard26:
I'll fix the lint issues after another round of review is done.

docs/the_black_code_style/current_style.md
```diff
@@ -292,8 +292,8 @@ multiple lines. This is so that _Black_ is compliant with the recent changes in
 style guide, which emphasizes that this approach improves readability.
 
 Almost all operators will be surrounded by single spaces, the only exceptions are unary
-operators (`+`, `-`, and `~`), and power operators when both its operands are considered
-"simple". For power ops, an operand is considered "simple" if it's only a NAME, numeric
+operators (`+`, `-`, and `~`), and power operators when both operands are simple.
+For power ops, an operand is considered simple if it's only a NAME, numeric
 CONSTANT, or attribute access (chained attribute access is allowed), with or without a
 preceding unary operator.
```
felix-hilden, line 298: Reading it through once more, we could also say that the surrounding spaces are removed for unaries and simple powers. Now it just says that they are an exception to the rule.

src/black/trans.py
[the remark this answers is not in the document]
felix-hilden: Fair enough, my intuition would be to make private functions in that case. But style-wise I think it's a matter of taste, I was just more concerned about the potential performance hit with math-heavy code. Haven't tested it at all though 😄

src/black/linegen.py
```diff
@@ -342,9 +347,9 @@ def transform_line(
     ):
         # Only apply basic string preprocessing, since lines shouldn't be split here.
         if mode.experimental_string_processing:
-            transformers = [string_merge, string_paren_strip]
+            transformers = [string_merge, string_paren_strip, hug_power_op]
         else:
-            transformers = []
+            transformers = [hug_power_op]
     elif line.is_def:
         transformers = [left_hand_split]
```
JelleZijlstra, line 354: Should hug_power_op be in here? Maybe it's easier to just unconditionally add it at the end.

src/black/trans.py
```diff
@@ -71,6 +71,88 @@ def TErr(err_msg: str) -> Err[CannotTransform]:
...
+        # Brackets and parenthesises indicate calls, subscripts, etc. ...
```
JelleZijlstra, line 85:
```suggestion
        # Brackets and parentheses indicate calls, subscripts, etc. ...
```

tests/data/power_op_spacing.py
```diff
@@ -0,0 +1,65 @@
...
+i = funcs.f() ** 5
+j = super().name ** 5
+k = [(2**idx, value) for idx, value in pairs]
+l = mod.weights_[0] == pytest.approx(0.95**100, abs=0.001)
+m = [([2**63], [1, 2**63])]
+n = count <= 10**5
+o = settings(max_examples=10**6)
+p = {(k, k**2): v**2 for k, v in pairs}
+q = [10**i for i in range(6)]
+
```
JelleZijlstra, line 31: A fun case to test might be `1.0.real ** 5.0.real`. I'd say that should not hug but I don't feel strongly if it's easier to implement the other way.

JelleZijlstra: Also could use some tests with float literals instead of int literals.

src/black/linegen.py
ichard26: Ah yeah I forgot that power ops can exist as parameter default expressions, gah. Good catch!

tests/data/power_op_spacing.py
ichard26: I don't care either way but this feels like it goes against the allowance of attribute access in the rule stated above. We could stop hugging attribute access entirely but I'd argue it's more readable once you add other operators:
```python
f = bound = Decimal(10 ** field.max_digits - 1) / (10 ** field.decimal_places) 
g = a.b**c.d + factor * 2

# vs.

f = Decimal(10**field.max_digits - 1) / (10**field.decimal_places)
g = a.b ** c.d + factor * 2
```

For the record making this case not hug should be as simple as:
```diff
@@ -101,6 +101,10 @@ def hug_power_op(line: Line, features: Collection[Feature]) -> Iterator[Line]:
                 # of a comprehension.
                 return True
 
+            # Attribute access on numeric constants is also disallowed.
+            if current.type == token.NAME:
+                disallowed.add(token.NUMBER)
```

docs/the_black_code_style/current_style.md
ichard26: I'm a little confused, could you elaborate?

ichard26:
Another problem I just thought of is that this introduces inconsistency in another manner, right now only subscript have the notion of "simple" / "complex", but this PR introduces another meaning, they're similar but the power op's one is more permissive to what is considered simple. We might want to get these different meanings of simple in alignment.

hugovk:
https://pre-commit.ci/ can autofix lint things like this:

> **auto fixing pull requests:** if tools make changes to files during a pull request, pre-commit.ci will automatically fix the pull request.

ichard26:
Unfortunately pre-commit.ci doesn't support local hooks and I'd rather not keep both the GHA job and the pre-commit.ci one too. 

---

FWIW personally I don't like it when commits are automatically pushed to my branch, it requires me to sync up my local branch which could lead to conflicts. Although yeah `pre-commit.ci fix` or whatever the magical comment could be nice.

docs/the_black_code_style/current_style.md
felix-hilden: Oh sorry, I missed this: I meant to say that we don't actually say what happens to unaries and powers. We only say that they are an exception to the rule. Maybe something like: "[the only exceptions are unaries and simple pows], for which the spaces are removed."

ichard26:
Alright I'd appreciate re-reviews, if you're confident enough in your suggestions just please them for me as I don't have much time to resolve them fast enough. The main thing I'm worried about now is https://github.com/psf/black/pull/2726#issuecomment-1013374126.

src/black/linegen.py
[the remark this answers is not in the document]
ichard26: Resolved.

PR #2728, opened 2021-12-26
Speed up new backtracking parser

### LINKED ISSUE #2727: Speed up new backtracking parser
The current revision of the new backtracking parser (that gets enabled with `-t py310`) is kind of slow.

For regular inputs that only use `match`/`case` without using the pattern matching statements, it is somewhere between 3 to 4 X slower at the moment (going up to 6 or 7 on hand crafted inputs). On some inputs (like [`Lib/test/test_patma.py`](https://github.com/python/cpython/blob/main/Lib/test/test_patma.py)) that contain `match`/`case` hundreds of times, with new PEP 634 statements, it takes *multiple minutes* (`2 minutes 34 seconds` to be exact) to format these.

I don't know if there is any definite way to speed-up these, but I have an idea which I'll post shortly as a draft PR.

isidentical:
Resolves #2727 
## Core Idea
The current parser; even when it is backtracking, builds nodes which is a terrible waste of time since every time we see a `match` or `case`, we need to copy the whole stack over and over in order to preserve the actual structure and not do any modifications on top of it. This is obviously not cheap. 

The following implementation does something very similar to old school state machines, where the parser is temporarily converted into a structure that tells you whether it accepts an input or not. It doesn't build any trees, at all. Which cuts a lot of time (see the benchmarks below). We simply change the parser's mode to accept-only from accept-and-build-tree when we are backtracking, and once we have the rule to follow, we switch back. 

The implementation is obviously not really straightforward, because lib2to3's code is very old and it is kind of hard to move pieces around. I did not want to make much modification to regular functions, in order to not cause any additional breakage. We simply change `push`/`pop`/`shift` from functions that create / modify nodes to functions that only modify the parser's state/transition table. This way, we don't need to copy the new nodes on each modification.

## Benchmarks

Benchmarking script is available [here](https://gist.github.com/isidentical/325325766daf7d3588b50704ef7bf064) (thanks to @ichard26 for sharing it with me).

> It'd be really appreciated if someone can also re-test these and share their results as well, for further verification.

An example that doesn't contain any PEP 634 syntax, but contains a lot of `match`/`case` keywords: [example.py](https://gist.github.com/isidentical/76d8b5e1777541d871ef1c4016952681).
```
-t py39 (main):: Mean +- std dev: 7.21 ms +- 0.25 ms
-t py39 (this branch):: Mean +- std dev: 7.25 ms +- 0.27 ms

-t py310 (main):: Mean +- std dev: 41.7 ms +- 2.5 ms
-t py310 (this branch):: Mean +- std dev: 13.8 ms +- 0.6 ms
```
From what I can interpret, this means that for `-t py310` there is a speed-up of ~**4X**.

An example that contains 25 non-PEP 634 usages and 25 PEP 634 usages of `match/case`: [example_2.py](https://gist.github.com/isidentical/0bfc74c21d351a87ae933087fd36304d)
```
-t py310 (main):: Mean +- std dev: 1.22 sec +- 0.03 sec
-t py310 (this branch):: Mean +- std dev: 93.0 ms +- 6.8 ms
```

And another example is parsing the [`Lib/test/test_patma.py`](https://github.com/python/cpython/blob/e6497fe698f6e87344501a68ffdea106eafcb257/Lib/test/test_patma.py), which has all of the pattern matching statement variations known to the humans (I've only run `lib2to3_parse()` once since it takes so much time on the `main`, so assume this is `+/- 20 seconds`):
```
main: 2 minutes, 33 seconds
this branch: 2.499 seconds
```

## Verification

Obviously, a change like this, at the core of the parser, should be verified though I do not have many examples of pattern matching myself. My script to parse every variation of pattern matching statements (originating through `Lib/test/test_patma.py`) says it can parse all of the examples. 

I've also taken the diff of the following command and compared it, which yielded no difference in formatting between main/this branch (speed-wise it is like 50X faster):
```console
$ black -tpy310 --fast --diff ~/projects/cpython/Lib/test/test_patma.py
```

Other than that, I've also tried formatting the whole stdlib (including tests) and there is also no relevant changes. If you have other ideas to verify this, that would be wonderful.

### DIFF
--- a/src/blib2to3/pgen2/parse.py
+++ b/src/blib2to3/pgen2/parse.py
@@ -47,14 +47,14 @@ def lam_sub(grammar: Grammar, node: RawNode) -> NL:
 
 
 # A placeholder node, used when parser is backtracking.
-FAKE_NODE = (-1, None, None, None)
+DUMMY_NODE = (-1, None, None, None)
 
 
 def stack_copy(
     stack: List[Tuple[DFAS, int, RawNode]]
 ) -> List[Tuple[DFAS, int, RawNode]]:
     """Nodeless stack copy."""
-    return [(copy.deepcopy(dfa), label, FAKE_NODE) for dfa, label, _ in stack]
+    return [(copy.deepcopy(dfa), label, DUMMY_NODE) for dfa, label, _ in stack]
 
 
 class Recorder:
@@ -94,16 +94,12 @@ class Recorder:
         want to restore to the initial state as quick as possible, which
         can only be done by having as little mutatations as possible.
         """
-        original_functions = {}
-        for name in self.parser.STATE_OPERATIONS:
-            original_functions[name] = getattr(self.parser, name)
-            safe_variant = getattr(self.parser, name + "_safe")
-            setattr(self.parser, name, safe_variant)
+        is_backtracking = self.parser.is_backtracking
         try:
+            self.parser.is_backtracking = True
             yield
         finally:
-            for name, func in original_functions.items():
-                setattr(self.parser, name, func)
+            self.parser.is_backtracking = is_backtracking
 
     def add_token(self, tok_type: int, tok_val: Text, raw: bool = False) -> None:
         func: Callable[..., Any]
@@ -214,6 +210,7 @@ class Parser(object):
         self.grammar = grammar
         # See note in docstring above. TL;DR this is ignored.
         self.convert = convert or lam_sub
+        self.is_backtracking = False
 
     def setup(self, proxy: "TokenProxy", start: Optional[int] = None) -> None:
         """Prepare for parsing.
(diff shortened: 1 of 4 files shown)


ichard26:
Man, I wish GH-2725 was already merged so I could run diff-shades across this PR but I'll do it manually (assuming it supports the necessary features for such a stress test :crossed_fingers:),

isidentical:
> Man, I wish GH-2725 was already merged so I could run diff-shades across this PR but I'll do it manually (assuming it supports the necessary features for such a stress test crossed_fingers). Also don't mind the primer or test uvloop failures, those are happening on main too and I haven't able to fix the crash yet.

Thanks a lot @ichard26. If you want, I can try to rebase on top of it but I have no idea whether there are anything that would block me to do so.

ichard26:
It probably won't work as GitHub places restrictions on where workflows must be to trigger (IIRC it has to be on the default branch -- although I say that and I just realized I made that PR from my fork and not a branch upstream whoops).

isidentical:
I've just ran the [syntax_test_suite](https://github.com/isidentical/syntax_test_suite) with this patch, and @ichard26 ran the [diff-shades](https://github.com/ichard26/diff-shades) which as far as I can tell neither yielded any changes so I am marking this PR as ready for review.

src/blib2to3/pgen2/parse.py
```diff
@@ -54,21 +65,45 @@ def __init__(self, parser: "Parser", ilabels: List[int], context: Context) -> No
...
+
+        This saves us a ton of time when we are backtracking, since we
+        want to restore to the initial state as quick as possible, which
+        can only be done by having as little mutatations as possible.
+        """
+        original_functions = {}
+        for name in self.parser.STATE_OPERATIONS:
+            original_functions[name] = getattr(self.parser, name)
+            safe_variant = getattr(self.parser, name + "_safe")
+            setattr(self.parser, name, safe_variant)
```
JelleZijlstra, line 101: Is mypyc going to like this?

JelleZijlstra (commented):
Thanks, this is a great speed improvement. I think it's worth mentioning in the changelog.

src/blib2to3/pgen2/parse.py
isidentical: I was unsure about this one, but at worst we can simply merge the implementations of the safe/unsafe and guard the safe one with a flag like `parser.is_safe = True`. CC: @ichard26

isidentical:
Self-note: after talking with @JelleZijlstra, we decided to merge `_safe` function to the original definitions and guard the immutable implementations with a flag.

src/blib2to3/pgen2/parse.py
```diff
@@ -54,21 +65,41 @@ def __init__(self, parser: "Parser", ilabels: List[int], context: Context) -> No
...
+            self.parser.stack = self._points[ilabel]
+            try:
+                yield
+            except ParseError:
+                self._dead_ilabels.add(ilabel)
+            finally:
+                self.parser.stack = self._start_point
+
+    @contextmanager
+    def patch(self) -> Iterator[None]:
```
JelleZijlstra, line 86: Can you rename this to something like `with self.backtrack():`?

src/blib2to3/pgen2/parse.py
```diff
@@ -54,21 +65,41 @@ def __init__(self, parser: "Parser", ilabels: List[int], context: Context) -> No
...
+        """
+        Patch basic state operations (push/pop/shift) with node-level
```
JelleZijlstra, line 88: Should be reworded to remove mention of "patching"

src/blib2to3/pgen2/parse.py
```diff
@@ -317,30 +349,44 @@ def classify(self, type: int, value: Text, context: Context) -> List[int]:
             raise ParseError("bad token", type, value, context)
         return [ilabel]
 
+    STATE_OPERATIONS = ["shift", "push", "pop"]
```
JelleZijlstra, line 352: This is no longer used

JelleZijlstra (changes requested):
Thanks, I have a couple of nits and I'd like to also see a changelog entry.

isidentical:
Thanks @JelleZijlstra, should be fixed now.

JelleZijlstra (approved):

Issue #248, opened 2018-05-22
Add command line option to ignore cache


dgouldin:
I ❤️ Black!

It would be nice to have the option to ignore its local cache in order to force it to reevaluate all files. Perhaps a `--no-cache` flag?

Thanks!

ambv:
What's your use case for this?

ambv:
There are a number of things we could do to make developmen a bit easier but that's not a use case I'd like to prioritize.

A need to ignore the cache suggests there is something wrong with it that needs to be fixed.

tiran:
In the mean time and on Linux, you can set ``XDG_CACHE_HOME`` to an invalid path to suppress caching:

```
$ XDG_CACHE_HOME=/invalid python3 -m black black.py 
```

ambv:
X Desktop Group to the rescue.

dgouldin:
From the source, it looks like a cache hit is determined by a file's size and modified time. I can conceive of ways this could result in false positives, but admittedly it would be an unusual case.

I think it's reasonable to say you don't want to prioritize this feature. I hadn't seen it proposed yet, so I thought I would. Thankfully the cache storage path is well documented, so when I was uncertain about the cache hits I was seeing, I was able to `rm -rf` the cache itself.

ambv:
If this is reassuring to you, what Black is caching on is exactly what `git status` is using to determine changes. It would take a deliberate attempt or a broken file system to fool it.

Or a bug, which is why I'd like those exposed and fixed rather than worked around.

ambv:
I'll close this as this is not a planned feature.

sanmai-NL:
> There are a number of things we could do to make developmen a bit easier but that's not a use case I'd like to prioritize.
> 
> A need to ignore the cache suggests there is something wrong with it that needs to be fixed.

Suggests, but doesn't imply. Not all calls to disable a feature are due to (functional) correctness defects, it's also a matter of performance efficiency.

PR #2739, opened 2022-01-01
Allow setting custom cache directory on all platforms with environmen…

### LINKED ISSUE #2506: Custom cache directory to be thread/process safe

The cache files are not thread/process safe. I've checked multiple projects simultaneously in different processes but get permission errors when one process is writing to the cache while another is reading from it. On linux the fix is pretty easy: set a different temp directory for each process using the ``XDG_CACHE_HOME`` environment variable but there is no equivalent for Windows (there's somewhat an equivalent but you need a rather [specific setup](https://github.com/platformdirs/platformdirs/blob/main/src/platformdirs/windows.py#L157-L165) to use the environment variable).

The solution that will work for my use case (and I think is the least amount of changes) is to create a custom environment variable that overrides using platformdirs to get the cache directory. Think this could look like:
```python
CACHE_DIR = Path(os.environ.get("BLACK_CACHE_DIR", user_cache_dir("black", version=__version__)))
if not CACHE_DIR.exists():
    raise RuntimeError(f"{CACHE_DIR} does not exist")
```

1. Add the command line option to turn off cacheing (as this ticket as #248 asked for) (guess is this  more work)
2. Add command line option to set the cache directory (guess is this is more work but not a lot)
3. Make the cache dir thread/process safe (guess is this is more work)

percurnicus:
### Description
Fixes #2506

``XDG_CACHE_HOME`` does not work on Windows. To allow for users to set a custom cache directory on all systems I added a new environment variable ``BLACK_CACHE_DIR`` to set the cache directory. The default remains the same so users will only notice a change if that environment variable is set.

The specific use case I have for this is I need to run black on in different processes at the same time. There is a race condition with the cache pickle file that made this rather difficult. A custom cache directory will remove the race condition.

I created ``get_cache_dir`` function in order to test the logic. This is only used to set the ``CACHE_DIR`` constant.

### Checklist - did you ...

    - Added ``test_get_cache_dir`` to ensure behavior is as expected.
    - Updated ``file_collection_and_discovery.md`` to have users use the new environment variable rather than ``XDG_CACHE_HOME``. Still mention they can use ``XDG_CACHE_HOME`` on macOS and linux.
    - Updated ``reference_functions.rst`` to autodoc new function.


cooperlees (commented):
Thanks. This all makes sense to me and clean tested code, but, (and I didn't go looking) why can't we fix `XDG_CACHE_HOME` to work on Windows in platformdirs?

percurnicus:
Yeah great question! I can make an issue in appdirs if you want but it seems like XDG is rather specific to linux and a cross platform environment variable would be out of scope of that project, as the solution would be for projects like this one to use their own. But let me know if you want me to make an issue ticket to the appdirs project before merging!

cooperlees:
> Yeah great question! I can make an issue in appdirs if you want but it seems like XDG is rather specific to linux and a cross platform environment variable would be out of scope of that project, as the solution would be for projects like this one to use their own. But let me know if you want me to make an issue ticket to the appdirs project before merging!

I'd love to see us explore adding something into appdirs (sorry about wrong dep named above) as this is a legit need for Windows users in general and not just black. Would love to know why there isn't support - maybe just no one could test it? If they push back for a valid reason we can merge this. I just like to try and avoid new env vars / config if we can.

percurnicus:
No need to apologize, I got the wrong project! It is platformdirs. There's an open ticket https://github.com/platformdirs/platformdirs/issues/4 for this change already and seems like they are willing to [accept a PR to address it](https://github.com/platformdirs/platformdirs/pull/35#issuecomment-907518116). I can submit a PR to that project and close this one if it gains traction.

cooperlees:
Awesome. Yeah I think this is the correct route! Thanks for digging in, I think we all win fixing the root issue.

percurnicus:
https://github.com/platformdirs/platformdirs/issues/4#issuecomment-1003795762 doesn't look like they'll accept a PR to extend env var paths for windows. The other option is to create a no-cache mode as requested in #248 (I also see a minimal use case in CI where the CI system does not use docker and it's one more file to clean up - again, minimal use case).

cooperlees:
So, with platformdirs stating that their stance is that an application should handle this kind of edge case / cross platform normalization and also with @ambv 's (and my) disliking of disabling the cache (seen in #248), it leads us back to this plan of attack. That is, having our own all platform env var to change the cache directory per invocation via an env var. The only other thing we could maybe do is move it to the config toml like most of our other settings. But I think this makes sense to be an env var since that's kinda what platformdirs does for this, but not on all platforms. It would be nice if they had options for all platforms but that seems the two schools of thoughts colliding about how to handle this from my quick reading.

As it's not that invasive, and covers a workaround for all platforms spawning many black processes at a time, I see only minor overhead introducing this. I'll leave this sit here a few days to see if any other maintainer disagrees, but I think I'll merge this if no one does. Thanks @percurnicus for digging in and looking at all options. It's important to do so for a popular package like black.

docs/contributing/reference/reference_functions.rst
```diff
@@ -96,6 +96,8 @@ Caching
 
 .. autofunction:: black.cache.filter_cached
 
+.. autofunction:: black.cache.get_cache_dir
```
ichard26, line 99: I honestly forgot these docs existed, I should probably fix that haha

docs/usage_and_configuration/file_collection_and_discovery.md
```diff
@@ -22,10 +22,11 @@ run. The file is non-portable. The standard location on common operating systems
...
 
-To override the location of these files on macOS or Linux, set the environment variable
-`XDG_CACHE_HOME` to your preferred location. For example, if you want to put the cache
-in the directory you're running _Black_ from, set `XDG_CACHE_HOME=.cache`. _Black_ will
-then write the above files to `.cache/black/<version>/`.
+To override the location of these files on all systems, set the environment variable
+`BLACK_CACHE_DIR` to the preferred location. Alternatively on macOS and Linux, set
+`XDG_CACHE_HOME` to you your preferred location. For example, if you want to put the
+cache in the directory you're running _Black_ from, set `BLACK_CACHE_DIR=.cache/black`.
+_Black_ will then write the above files to `.cache/black`.
```
ichard26, line 29: It may be implied in how this is structured (in which case I'm totally missing it) but it'd be nice to also note that `BLACK_CACHE_DIR` takes priority over `XDG_CACHE_HOME`.

ichard26 (approved):
I don't see any problems with this, just one minor suggestion. Thank you for filing the issue, researching the options AND submitting a PR - truly a rockstar!

docs/usage_and_configuration/file_collection_and_discovery.md
percurnicus: Agreed! Think this will be a definite improvement!

percurnicus: addressed

JelleZijlstra:
Why does this use an env var instead of a command line/pyproject.toml option? I believe we don't use env vars for anything else.

percurnicus:
Great question! I opted for an env var as that was the documented way to set the cache directory, it just didn't work on Windows. Also, the implementation was very simple. I like the idea of using a command line argument but I'd have to look at how simple that would be to implement.

pvargas-sx:
@JelleZijlstra Looks like we could use a command line argument, would require adding another argument to `reformat_one`, `reformat_many`, `schedule_formatting`, `read_cache`, `write_cache`, and `get_cache_file`. A more pervasive change but could be worth it to keep configuration consistent. @cooperlees @ichard26 What do you prefer? If a new command line argument is preferred, then I'd rather remove documentation on using `XDG_CACHE_HOME` so people only use the command line (there could be two ways but prefer the "only one best way to do something" philosophy). I would also deny this PR and ticket and create new ones to keep the scopes focused.

percurnicus:
@JelleZijlstra  It appears a new command line argument would be pretty simple to implement. Would require adding arguments to multiple functions so would be more pervasive than the current implementation, but could be worth it to have configuration be consistent. @cooperlees @ichard26 what do you think? If this is the way to go I would remove documentation to use `XDG_CACHE_HOME` in favor of the command line argument/configuration. Would also make new ticket/PR to keep scopes focused.

cooperlees:
I thought about the config file/cli vs. env var and didn't bring it up as platformdirs already uses an environment variable and thought we should keep it the same since people playing with cache dirs might already be familiar with this approach.

I am happy either way. There are benefits to all config being in one spot and benefits to staying with the "standard" library that deals with this. As I said earlier, ideally platformdirs would handle this and clean it up but they have platform fun and games there in the way. So I vote for what ever the majority here wants.

JelleZijlstra (approved):
Let's just stay with the env var for now.

ichard26:
Congrats on your first PR here! Thank you so much for your contribution! This project is only possible by contributions like these :black_heart:. You're awesome, @percurnicus.

- Feel free to share any feedback on your experience here: #2238

PR #2744, opened 2022-01-04
Add support for custom python cell magics

### LINKED ISSUE #2742: Allow defining custom python magics for Jupyter support
Recent updates of black have introduced out-of-the-box support for Jupyter Notebooks, which is extremely helpful.  As of #2630 only *known* python cell magics are processed (see e.g. [here](https://github.com/psf/black/blob/8a84bebcfcabddfd5b82a8cff0b830a745999b6c/src/black/handle_ipynb_magics.py#L40)), which certainly makes sense.  We have the use case that we have several custom python cell magics that we would also like to process.  

This request is to allow configuration to support additional (user-defined) magics.  

Possible solutions:

I am aware that we could certainly call black from python directly and programmatically update `PYTHON_CELL_MAGICS` to include our custom python magics.  However, I'm not keen on doing this because it would depend on implementation details within black.   If this would be the preferred solution, then I would suggest to at least add a function that could be called to register the additional magics.

Another possibility would be to add a configuration flag/option (i.e. in `pyproject.toml`/CLI) that would allow the input of a list of additional python cell magic names.  (Using an environment variable would be a similar option, but I don't see any other place that this is done, other than with e.g. `XDG_CACHE_HOME` which is a more "global" configuration variable.)  This is likely my preferred solution, but, given that black makes a concerted effort *to avoid* configuration options, I could understand why this would not be acceptable.

Thanks in advance for any input/consideration.

mgmarino:
### Description

Fixes #2742.

This PR adds the ability to configure additional python cell magics.  This
will allow formatting cells in Jupyter Notebooks that are using custom (python)
magics.  

### Checklist - did you ...


felix-hilden (commented):
Thanks for submitting! The changes look good in general, but one discussion below.

Also, we'll need other maintainers to have look as well since they haven't voiced their opinions yet.

tests/test_ipynb.py
```diff
@@ -139,6 +140,17 @@ def test_cell_magic_with_magic() -> None:
...
 
+def test_cell_magic_with_custom_python_magic() -> None:
+    src = "%%custom_python_magic -n1 -n2\nx=2"
+    result = format_cell(
+        src,
+        fast=True,
+        mode=replace(JUPYTER_MODE, python_cell_magics={"custom_python_magic"}),
+    )
+    expected = "%%custom_python_magic -n1 -n2\nx = 2"
+    assert result == expected
```
MarcoGorelli, line 151: should there also be a test to check that if you pass some custom magic, then the default Python magics still get formatted?

MarcoGorelli (changes requested):
Nice work! Just got a comment about tests

tests/test_ipynb.py
felix-hilden: Probably, and that when not passing but trying to format custom it isn't formatted 👍

src/black/__init__.py
```diff
@@ -224,6 +224,16 @@ def validate_regex(
         "(useful when piping source on standard input)."
     ),
 )
+@click.option(
+    "--python-cell-magics",
+    multiple=True,
```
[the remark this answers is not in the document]
mgmarino: Agreed, though this is how `click` supports multiple options out of the box.  This is also how `target-version` takes multiple arguments.  I expect that we would need to introduce a custom callback to support multiple options, which might then also require custom logic when reading from the `pyproject.toml` file.

My expectation is that this will rather mostly be used in a configuration file and so perhaps the clunkiness on the command line is acceptable.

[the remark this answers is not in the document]
felix-hilden: That's fair!

tests/test_ipynb.py
mgmarino: No problem, pushed.  :-)

src/black/mode.py
```diff
@@ -183,5 +184,6 @@ def get_cache_key(self) -> str:
             str(int(self.is_ipynb)),
             str(int(self.magic_trailing_comma)),
             str(int(self.experimental_string_processing)),
+            ",".join(self.python_cell_magics),
```
JelleZijlstra, line 187:
```suggestion
            ",".join(sorted(self.python_cell_magics)),
```
Set iteration order is randomized, so this would make the cache key unstable

tests/test_ipynb.py
```diff
@@ -139,6 +162,41 @@ def test_cell_magic_with_magic() -> None:
...
+        pytest.param(
+            replace(JUPYTER_MODE, python_cell_magics={"custom_python_magic", "cust1"}),
+            "%%custom_python_magic -n1 -n2\nx = 2",
+            does_not_raise(),
+            id="Correctly change when cell magic registered",
+        ),
+    ],
+)
+def test_cell_magic_with_custom_python_magic(
+    mode: Mode, expected_output: str, expectation: ContextManager[Any]
```
JelleZijlstra, line 189:
```suggestion
    mode: Mode, expected_output: str, expectation: ContextManager[object]
```
Avoid Any if possible. `object` makes sense if you don't want to do anything with the object.

JelleZijlstra (approved):

src/black/mode.py
```diff
@@ -147,5 +148,6 @@ def get_cache_key(self) -> str:
             str(int(self.is_ipynb)),
             str(int(self.magic_trailing_comma)),
             str(int(self.experimental_string_processing)),
+            ",".join(sorted(self.python_cell_magics)),
```
ichard26, line 151: At this point I'm a little worried that we might hit some filesystem file name length limits. I assume that Python cell magics are also filesystem safe? can't contain slashes, colons, whatever isn't safe.

The cache should fail gracefully but I'm not entirely sure if it does so yeah. This can be resolved separately though.

ichard26 (approved):
Seems fine to me. We're going to have to discuss further how to manage the complexity of supporting more languages but that can happen independently. 

Thank you!

src/black/mode.py
JelleZijlstra: Good point. Maybe we can hash the value with some consistent hashing algorithm available in the stdlib.

JelleZijlstra: It doesn't need to be human-readable, though it's helpful for debugging if it's not too bad. The other components of the key should mostly just be ints, so I think we can get away with not hashing them.

JelleZijlstra: Could you make the change to hashing? Also, there are a few merge conflicts now.

mgmarino: Of course.  I've done that and resolved the merge conflicts.

src/black/__init__.py
```diff
@@ -401,6 +411,7 @@ def main(
     fast: bool,
     pyi: bool,
     ipynb: bool,
+    python_cell_magics: List[str],
```
JelleZijlstra, line 414:
```suggestion
    python_cell_magics: Sequence[str],
```
Seems like it's sometimes a tuple at runtime, which makes mypyc unhappy

PR #2748, opened 2022-01-05
Action: Support running in a docker container

joshowen:
`${{ github.action_path }}` doesn't point to the correct path when run in a docker container, you have to use the environmental variable `GITHUB_ACTION_PATH` 

https://github.com/actions/runner/issues/716

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,10 @@
 
 - All upper version bounds on dependencies have been removed (#2718)
 
+### Integrations
+
+- Update GitHub action to support containerized runs (#2748)
+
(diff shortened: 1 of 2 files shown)


cooperlees (approved):
I'm guessing this is for this: https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action

This breaks none of our CI so I see no reason to merge - will wait for @ichard26 to agree.

Can you just please add a Change Log entry please.

joshowen:
@cooperlees I've updated the changelog

PR #2752, opened 2022-01-07
Create --preview CLI flag

### LINKED ISSUE #2751: Create a --preview flag for disruptive changes between major updates
As discussed in #2394, we'd like to shove disruptive style changes between major updates behind a `--preview` flag. And I think it should be created and available even if such changes have not been implemented yet, so that our CLI is consistent.

Perhaps:
- add the CLI option
- create a separate style docs section for describing the style
- from now on announce these kinds of changes under a "preview" header in our changelog
- document a consistent way of writing preview tests without disrupting stable (I think just duplicating might be fine)

Thoughts, suggestions?

---
@ichard26 shall we add this to the "stable" project?

felix-hilden:
Closes #2751: this PR creates a `--preview` flag for our disruptive style changes. It includes documentation (changes described under "future style"), tests (now empty skeleton) and the flag itself as a CLI option and added to `Mode`.

The name of the flag was originally `--future` in the stability policy, but I changed it to `--preview` based on the original discussion. While I like it more, it isn't really important. Which would y'all prefer?

### Checklist


tests/test_format.py
```diff
@@ -78,6 +78,8 @@
     "parenthesized_context_managers",
 ]
 
+PREVIEW_CASES = []
```
felix-hilden, line 81: I guess `mypy` has some advanced analysis for the test case lists, since the hint isn't needed for other lists. Should I:

1) add hints for all case lists
2) add it just here and remove it when the first entry is added
3) add test infra later?

JelleZijlstra: I'd say add hints for all of them.

felix-hilden:
Primer / Django failing seems unrelated, although `main` succeeds 🤔

docs/the_black_code_style/future_style.md
```diff
@@ -40,3 +40,9 @@ Currently, _Black_ does not split long strings to fit the line length limit. Cur
 there is [an experimental option](labels/experimental-string) to enable splitting
 strings. We plan to enable this option by default once it is fully stable. This is
 tracked in [this issue](https://github.com/psf/black/issues/2188).
+
+## Preview style
+
+Potentially disruptive style changes between major releases are gathered under the
+`--preview` CLI flag. They will eventually become part of main functionality, but until
```
JelleZijlstra, line 47: "will" seems too strong a word; the intention is that the preview features will have time to incubate and become stable. At the end of each year, we'll have to make a call on whether to promote a feature to stable.

JelleZijlstra: Let's do that

docs/the_black_code_style/future_style.md
```diff
@@ -40,3 +40,9 @@ Currently, _Black_ does not split long strings to fit the line length limit. Cur
...
+then are described here.
```
JelleZijlstra, line 48:
```suggestion
Experimental, potentially disruptive style changes are gathered under the
`--preview` CLI flag. At the end of each year, these changes may be
adopted into the default style, as described in
[The Black Code Style](the_black_code_style/index.rst).
```
How about this wording

docs/the_black_code_style/future_style.md
```diff
@@ -40,3 +40,10 @@ Currently, _Black_ does not split long strings to fit the line length limit. Cur
 there is [an experimental option](labels/experimental-string) to enable splitting
 strings. We plan to enable this option by default once it is fully stable. This is
 tracked in [this issue](https://github.com/psf/black/issues/2188).
+
+## Preview style
+
+Experimental, potentially disruptive style changes are gathered under the
+`--preview` CLI flag. At the end of each year, these changes may be
+adopted into the default style, as described in
+[The Black Code Style](the_black_code_style/index.rst).
```
JelleZijlstra, line 49:
```suggestion
[The Black Code Style](./index.rst).
```

docs/the_black_code_style/future_style.md
```diff
@@ -40,3 +40,10 @@ Currently, _Black_ does not split long strings to fit the line length limit. Cur
 there is [an experimental option](labels/experimental-string) to enable splitting
 strings. We plan to enable this option by default once it is fully stable. This is
 tracked in [this issue](https://github.com/psf/black/issues/2188).
+
+## Preview style
+
+Experimental, potentially disruptive style changes are gathered under the
+`--preview` CLI flag. At the end of each year, these changes may be
+adopted into the default style, as described in
+[The Black Code Style](./index.rst).
```
JelleZijlstra, line 49:
```suggestion
Experimental, potentially disruptive style changes are gathered under the `--preview`
CLI flag. At the end of each year, these changes may be adopted into the default style,
as described in [The Black Code Style](./index.rst).
```

felix-hilden:
Thanks for the updates, they look a-ok! I'd still like to get the Mode / feature flag machinery in before merging though.

felix-hilden:
Here's the feature flag machinery! I hope it's not too cute. I aimed specifically for a certain syntax:

```python
from black.mode import Preview, Mode

def some_transform(..., mode: Mode):
    if Preview.EXPERIMENTAL_STRING_PROCESSING in mode:
        return awesomize_string(s)

def other_transform(..., mode: Mode):
    if Preview.MORE_PARENTHESES in mode:
        return ((((line))))
```

I think this is ever so slightly nicer than relying on some other Mode-like object. And it's easier to see where `Preview` is used in an IDE, rather than trying to find the attributes of a `Mode+`. But let me know what you think! I know not everyone is into this syntax trickery 😄

JelleZijlstra (approved):
Looks good to me.

We might need something a bit more complicated if we want to keep the separate flag for ESP in addition to the `--preview` flag.

PR #2758, opened 2022-01-10
Enable pattern matching by default

### LINKED ISSUE #2757: Enable pattern matching by default
Now that the speed-up PR landed, it should be safe for us to enable pattern matching by default as the last grammar case to try. This would still make us initially test the 3.7+ and 3.0-3.6 grammars first before we try to run with the backtracking.

isidentical:
Resolves #2757


tests/test_format.py
```diff
@@ -190,6 +190,12 @@ def test_python_310(filename: str) -> None:
     assert_format(source, expected, mode, minimum_version=(3, 10))
 
 
+def test_python_310_without_target_version() -> None:
+    source, expected = read_data("pattern_matching_simple")
+    mode = black.Mode()
+    assert_format(source, expected, mode, minimum_version=(3, 10))
+
+
 def test_patma_invalid() -> None:
```
isidentical, line 199: This should cover the exception choosing logic ^

ichard26:
I fixed the merge conflict mostly because I want to trial diff-shades on a PR :slightly_smiling_face:

src/black/parsing.py
```diff
@@ -99,20 +103,21 @@ def lib2to3_parse(src_txt: str, target_versions: Iterable[TargetVersion] = ()) -
...
+            errors[grammar.version] = InvalidInput(
+                f"Cannot parse: {lineno}:{column}: {te.args[0]}"
+            )
 
     else:
-        if pygram.python_grammar_soft_keywords not in grammars and matches_grammar(
-            src_txt, pygram.python_grammar_soft_keywords
-        ):
-            original_msg = exc.args[0]
-            msg = f"{original_msg}\n{PY310_HINT}"
```
JelleZijlstra, line 114: I'm guessing we can remove PY310_HINT

ichard26:
OK so the diff-shades-comment workflow is broken due to https://github.community/t/pull-request-attribute-empty-in-workflow-run-event-object-for-pr-from-forked-repo/154682/2 but here's a DIY diff-shades comment:
```
╭─────────────────────── Summary ────────────────────────╮
│ 0 projects & 0 files changed / 0 changes [+0/-0]       │
│                                                        │
│ ... out of 2 064 043 lines, 10 105 files & 23 projects │
╰────────────────────────────────────────────────────────╯

Nothing-changed.
```

src/black/parsing.py
ichard26: Resolved: https://github.com/psf/black/pull/2758/commits/3a3ac469b1bb9a56ee7ff73bb4bd4793a825b7b1.

isidentical:
Thanks for the fix @ichard26.

ichard26 (approved):
LGTM, thank you!

KotlinIsland:
@JelleZijlstra Hi!

> Enable Python 3.10+ by default, without any extra need to specify `--target-version=py310`. 

Does this mean it is redundant to include the `target-version` in the pyproject configuration?
```toml
[tool.black]
target-version = ["py310"]
```

JelleZijlstra:
I'd still recommend adding it for clarity. If you don't, you may sometimes get confusing errors from Black if there's a syntax error. But it's no longer needed to make Black support match statements.

KotlinIsland:
Thanks!

It sucks that you have to specify this in the config of each tool, it would be cool if pyproject had a 'global' section for "supported python versions".

PR #2760, opened 2022-01-10
Fix handling of standalone match/case with newlines/comments

### LINKED ISSUE #2759: Backtracking parser does not ignore newlines/comments
The regular parser ignores `token.NL`/`token.COMMENT` since the grammar assumes there can be any number of them during parsing. This is actually handled in the driver rather than parser, so when the parser got implemented initially we did not specifically instruct this. But seems like it is now needed.

https://github.com/pantsbuild/pants/blob/aa1d38d4f3b791509bff773a65c2b07490bacad3/src/python/pants/util/enums_test.py#L33-L41

This seem to only cover cases where a standalone `match` or `case` is used in a newline fashion:
```py
match(
     something
)
```

It does not include `re.match()` or other variants as far as I can see. The fix is should be really easy, just ignoring those tokens on the recorder.

isidentical:
Resolves #2759

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,8 +26,8 @@
   be formatted. (#2526)
 - Speed-up the new backtracking parser about 4X in general (enabled when
   `--target-version` is set to 3.10 and higher). (#2728)
-- Fix handling of standalone `match()` or `case()` when used there is a trailing newline
-  or a comment inside of the parentheses. (#2760)
+- Fix handling of standalone `match()` or `case()` when there is a trailing newline or a
+  comment inside of the parentheses. (#2760)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,8 @@
   be formatted. (#2526)
 - Speed-up the new backtracking parser about 4X in general (enabled when
   `--target-version` is set to 3.10 and higher). (#2728)
+- Fix handling of standalone `match()` or `case()` when used there is a trailing newline
+  or a comment inside of the parentheses. (#2760)
 
 ### Packaging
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -26,6 +26,8 @@
   be formatted. (#2526)
 - Speed-up the new backtracking parser about 4X in general (enabled when
   `--target-version` is set to 3.10 and higher). (#2728)
+- Fix handling of standalone `match()` or `case()` when used there is a trailing newline
```
JelleZijlstra, line 29:
```suggestion
- Fix handling of standalone `match()` or `case()` when there is a trailing newline
```

tests/data/pattern_matching_style.py
```diff
@@ -10,6 +10,60 @@
...
+
+match something:
+    case b(): print(1+1)
+    case c(
+        very_complex=True,
+        perhaps_even_loooooooooooooooooooooooooooooooooooooong=-   1
+    ): print(1)
+    case c(
+        very_complex=True,
+        perhaps_even_loooooooooooooooooooooooooooooooooooooong=-1
```
JelleZijlstra, line 46: Maybe add a trailing comma here just to see that it interacts correctly with the magic trailing comma
```suggestion
        perhaps_even_loooooooooooooooooooooooooooooooooooooong=-1,
```

isidentical: That'll probably need a change in the output as well, will commit separately. Thanks!

isidentical:
Thanks @JelleZijlstra! For editing the output of the test case, I've committed your suggestions locally.

JelleZijlstra (approved):

PR #2761, opened 2022-01-11
Change git url for pip installation in README

cbows:
Unauthenticated git protocol was disabled recently by Github and should not be used anymore.

https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git


cooperlees (approved):
Sure. Thanks for helping people not get confused here!

Issue #2762, opened 2022-01-11
Performance tracker: Backtracking Parser


isidentical:
The previous PRs successfully optimized the backtracking in the factor of 4X. It might make some sense to profile and see whether there are still other potential bottlenecks.

tanvimoharir:
Hi, 
I would like to work on this issue.
I have committed to this repository in the past but I would need some guidance to get started.
Thanks

isidentical:
You can benchmark the backtracking parser (with passing `-t py310`) and run the profiler to see the hot spots, if there is anything too obvious that might be the target of the optimization. Here is an example
https://github.com/psf/black/pull/2763#issuecomment-1010201541

tanvimoharir:
> 

Hi @isidentical,
Perhaps I'm missing some context, I was assuming that I need to run this with cprofile which is what I did:

```
bash-3.2$ python3 -m cProfile -s tottime -m src.blib2to3.pgen2.parse
         26469 function calls (25215 primitive calls) in 0.132 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       23    0.046    0.002    0.046    0.002 {method 'read' of '_io.BufferedReader' objects}
       10    0.034    0.003    0.034    0.003 {built-in method _imp.create_dynamic}
       23    0.005    0.000    0.005    0.000 {built-in method marshal.loads}
      157    0.005    0.000    0.005    0.000 {built-in method posix.stat}
    100/6    0.003    0.000    0.008    0.001 sre_parse.py:493(_parse)
       23    0.002    0.000    0.002    0.000 {built-in method io.open_code}
       53    0.002    0.000    0.005    0.000 {built-in method builtins.__build_class__}
       79    0.002    0.000    0.008    0.000 <frozen importlib._bootstrap_external>:1505(find_spec)
    194/6    0.001    0.000    0.003    0.000 sre_compile.py:71(_compile)
     1320    0.001    0.000    0.001    0.000 sre_parse.py:164(__getitem__)
```

However looking at your comment I'm guessing this isn't what is expected.
I looked at the related comment from https://github.com/psf/black/pull/2763#issuecomment-1010201541 
Are you using pyperf module? (I have started going through pyperf docs) 
I don't understand how to pass ``` -t py310``` or to be precise, is py310 a benchmark that should be created? 
Please let me know what I' missing.
Thanks again for helping me, appreciate it.

JelleZijlstra:
[...]

PR #2763, opened 2022-01-11
Don't make redundant copies of the DFA

isidentical:
`copy.deepcopy` is still the most expensive part (75/100 of the extra overhead introduced by the backtracking parser). I initially assumed it was mutable (since it was represented as a tuple of lists), but seems like we only use it as a basic table and don't mutate it anywhere in the code.

```
-t py39 (main):: Mean +- std dev: 7.21 ms +- 0.25 ms
-t py310 (this branch):: Mean +- std dev: 13.8 ms +- 0.6 ms
-t py310 (this branch):: Mean +- std dev: 7.98 ms +- 0.24 ms
```

This reduces the total overhead of the new backtracking parser from %91 to %11 (ran with tuned pyperf, but confirm please). 

We can even get rid of the `stack_copy` and just do a bare `parser.stack.copy()`, but I think it gives a nice abstraction and prevents leakage of any real nodes into the backtracking stack. In case anything goes wrong, and this does not seem to have an observable overhead.

### DIFF
--- a/src/blib2to3/pgen2/parse.py
+++ b/src/blib2to3/pgen2/parse.py
@@ -54,7 +54,7 @@ def stack_copy(
     stack: List[Tuple[DFAS, int, RawNode]]
 ) -> List[Tuple[DFAS, int, RawNode]]:
     """Nodeless stack copy."""
-    return [(copy.deepcopy(dfa), label, DUMMY_NODE) for dfa, label, _ in stack]
+    return [(dfa, label, DUMMY_NODE) for dfa, label, _ in stack]
 
 
 class Recorder:


isidentical:
> This reduces the total overhead of the new backtracking parser from %91 to %11 (ran with tuned pyperf, but confirm please).

Just as a FYI, the rest of the overhead seem to be happening in the `TokenProxy.__next__`. It would probably the next target of an optimization.
https://github.com/psf/black/blob/4efb795129b17b96bdf299eacaca4243d9af86d0/src/blib2to3/pgen2/driver.py#L79-L87

isidentical:
Related #2762

PR #2770, opened 2022-01-14
Use parentheses with equality check in walrus/assignment statements

### LINKED ISSUE #449: Use parentheses in: a = b == c
Some changes black wants to make to my code:

```diff
-on_windows = (os.name == "nt")
+on_windows = os.name == "nt"
```

```diff
-        currently_registered = (fd in self._registered)
+        currently_registered = fd in self._registered
```

The parentheses were originally there to aid readability, by emphasizing that RHS of the `=` is an unusual-but-valid self-contained expression and not, like, a chained assignment or comprehension or something. Perhaps black should preserve, or even add, parentheses in cases like this? I guess the `in` case is more debateable, but the `= ==` case is really really easy to misread.

Shivansh-007:
Closes #449

### Checklist - did you ...


JelleZijlstra:
Looking at the results of this change, I'm not sure it's really worth it. The diff shows that this is a pretty common pattern, so this would be a relatively disruptive change. To me, most of the changes don't seem like clear improvements, so I'd default to sticking to Black's existing style. (Relatedly, the original issue got more downvote reactions than upvotes.)

The parentheses seem especially unnecessary in cases where the left operand of the `==` isn't a valid assignment target, like this one in pandas:
```
-            is_last_column = len(columns) - 1 == i
+            is_last_column = (len(columns) - 1 == i)
```

(diff-shades commenting currently doesn't work yet, but you can manually download the artifacts from https://github.com/psf/black/actions/runs/1697124259.)

Shivansh-007:
Sorry, I was looking to edit my previous comment, but misclicked delete instead.

I agree with you but in cases where the expression is in form of `a = b == c` where `b` and `c` can only be a single leaf, like `system.platform()` and not `len(columns) - 1`. This is pretty helpful: 

```diff
 def test_hash_equal(a, b, expected):
-    result = a == b
+    result = (a == b)
```

But still, I am not sure about cases like, where they are in the above format but are looking better without the parentheses:

```diff
-        result = Series(a1, dtype=cat_type) == Series(a2, dtype=cat_type)
+        result = (Series(a1, dtype=cat_type) == Series(a2, dtype=cat_type))
```

I think this can still be discussed a bit more on the issue with the original commenters.

JelleZijlstra:
To be clear my current preference is just to never add these parentheses, it's just that the parens are more jarring in cases where the LHS of the == is complex. I'm OK with landing a version of this change if there's consensus for it though.

felix-hilden:
I tend to always add the parens, but this isn't the hill I'm willing to die on. A good middle ground could be using a definition of "complex" like in our power hugging endeavour. It's not so easy to determine if something can be used as an assignment target, but a combination of name-attribute-subscript is a good approximation. `foo().bar = 3` is just asking for trouble 😄

JelleZijlstra:
This will now need to go in the preview style only. (And my personal preference is still to just not change anything here, but the preview style does allow us to change our mind more easily.)

JelleZijlstra:
Why is this blocked? There are a lot of merge conflicts now. I vote that we close this and leave the style as is.

ichard26:
I don't remember well. I think I marked it as blocked because it wasn't even clear if the style change it was implementing was accepted (the PR is blocked until the style decision is made). Made sense in my head when I added the label, but I agree it's confusing in hindsight.

PR #2773, opened 2022-01-14
Fix and speedup diff-shades integration 

ichard26:
### Description

- Pins third party actions to a specific SHA to reduce supply chain risk
- Adds a concurrency group to diff-shades so quick pushes to PRs with little time in between doesn't end up running diff-shades only to overwrite the results soon after repeatedly
- Compiles the baseline/target revision w/ mypyc to hopefully see runtime reductions
- Fixes the PR comment bug (hence why this is from my fork and not a branch here)

### Checklist - did you ...

### DIFF
--- a/.github/mypyc-requirements.txt
+++ b/.github/mypyc-requirements.txt
@@ -1,0 +1,14 @@
+mypy == 0.920
+
+# A bunch of packages for type information
+mypy-extensions >= 0.4.3
+tomli >= 0.10.2
+types-typed-ast >= 1.4.2
+types-dataclasses >= 0.1.3
+typing-extensions > 3.10.0.1
+click >= 8.0.0
+platformdirs >= 2.1.0
+
+# And because build isolation is disabled, we'll need to pull this too
+setuptools-scm[toml] >= 6.3.1
+wheel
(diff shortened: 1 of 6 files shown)


ichard26:
OK, clearly something's still up as diff-shades ran on #2776 even when the path was ignored.

JelleZijlstra (approved):

JelleZijlstra:
Everything is failing with `  49. tests/test_preprocessing/test_randomadder.py: TypeError - str object expected; got None                      `. Unfortunately I can't find a stack trace.

ichard26:
Yeah main is currently broken under mypyc due as https://github.com/psf/black/blob/8c22d232b56104376a12d1e68eaf216d04979830/src/black/parsing.py#L248 isn't type annotated so mypy guesses it's of type `str`. Unfortunately this is not true as https://github.com/psf/black/blob/8c22d232b56104376a12d1e68eaf216d04979830/src/black/parsing.py#L250 assigns whatever value is which the result of a getattr call on the node object. This wasn't caught by mypy as `value` is typed as `Any`. This is what https://github.com/psf/black/pull/2773/commits/cdea55b2418ac466873293854ed00fc559c507bd resolves. 

I should probably log the stacetrace too now you mention it. diff-shades logs the log files but stacktraces would be helpful in non-logged errors.

There was one more thing I wanted to add to the PR but honestly I'm going to be very busy for the next week and a half so I won't be able to really do anything OSS related so yeah. Feel free to merge although I do feel like this is undertested still :/

Issue #2782, opened 2022-01-18
Complex decorators break safe mode


emfdavid:
Using decorator syntax with a chained function call breaks the AST validation in safe mode.
This is a pattern for [prometheus client](https://github.com/prometheus/client_python#three-step-demo)

Use black [playground](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AG-AOddAD2IimZxl1N_WlkPinBFoR3Cl8B_4kpdkic7jS1dbrWRqxezQBIp0NHTWOT5p6HLb5MYiZOZMQhBUo9v9HGxxTDZ6Jcf2XsfLszWV0uR1KuWpiBNz2CSAba_UuLYf64DaLxWUtyLz-FBC7i3ois4nMyISUR1pfMJeUpXrWj7PrIQCvZD5y3DlweYKnGOfqiitLs-A9iannW5V6DZIwd7SWKbVC-0_unHTkPxkTATwS2_n6innP-GvKr9ZoK9luBdOeEkBN1AHfQgHJL6m5ie9wXZ4DUj9gRP5l7VBHv6Qtp1TSz_mjI7VgAAaPyaf3_P58cAAYMCvwMAAF4B5maxxGf7AgAAAAAEWVo=) repro case

<img width="1435" alt="Screen Shot 2022-01-18 at 6 28 09 PM" src="https://user-images.githubusercontent.com/84335963/150035137-e9be53a7-c66c-4eb3-a647-61f059369f46.png">

If I comment out the chained labels - it works.
<img width="1431" alt="Screen Shot 2022-01-18 at 6 30 16 PM" src="https://user-images.githubusercontent.com/84335963/150035344-e1e448bb-91a4-4436-a1b3-a8807e94ebf7.png">

Currently my best work around is to ignore the files using this syntax in my config and run black manually with fast for those files.

JelleZijlstra:
We may never have properly added support for https://www.python.org/dev/peps/pep-0614/.

isidentical:
AFAIK the PEP 614 support is there:
https://github.com/psf/black/blob/98db4abc21477cbd247c8fbd4cf9e8d1cf61ca0f/src/blib2to3/Grammar.txt#L15

This seems like a problem with Black Playground, since they are still using Python 3.8. 

@emfdavid what is the version of the python interpreter you are running the black with?

emfdavid:
Excellent catch @isidentical 
[...]

PR #2786, opened 2022-01-20
Hint at likely cause of ast parsing failure in error message

emfdavid:
### Description

Add more descriptive error message to help prevent confusion when parsing source code AST fails as in #2782

### Checklist - did you ...

Nothing to do here. 
I don't think I need a change log entry for this - but happy to add if needed.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1315,7 +1315,9 @@ def assert_equivalent(src: str, dst: str, *, pass_num: int = 1) -> None:
         src_ast = parse_ast(src)
     except Exception as exc:
         raise AssertionError(
-            f"cannot use --safe with this file; failed to parse source file: {exc}"
+            f"cannot use --safe with this file; failed to parse source file AST: {exc}"
+            f"This could be caused by running black with an older python version "
+            f"that does not support new syntax used in your source file."
         ) from exc
 
     try:
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1315,7 +1315,8 @@ def assert_equivalent(src: str, dst: str, *, pass_num: int = 1) -> None:
         src_ast = parse_ast(src)
     except Exception as exc:
         raise AssertionError(
-            f"cannot use --safe with this file; failed to parse source file AST: {exc}\n"
+            f"cannot use --safe with this file; failed to parse source file AST: "
+            f"{exc}\n"
             f"This could be caused by running black with an older python version "
             f"that does not support new syntax used in your source file."
         ) from exc
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1315,7 +1315,7 @@ def assert_equivalent(src: str, dst: str, *, pass_num: int = 1) -> None:
         src_ast = parse_ast(src)
     except Exception as exc:
         raise AssertionError(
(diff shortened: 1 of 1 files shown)


src/black/__init__.py
```diff
@@ -1315,7 +1315,9 @@ def assert_equivalent(src: str, dst: str, *, pass_num: int = 1) -> None:
         src_ast = parse_ast(src)
     except Exception as exc:
         raise AssertionError(
-            f"cannot use --safe with this file; failed to parse source file: {exc}"
+            f"cannot use --safe with this file; failed to parse source file AST: {exc}"
+            f"This could be caused by running black with an older python version "
+            f"that does not support new syntax used in your source file."
```
JelleZijlstra, line 1320: This will all get mushed into one line, which probably won't look great.

Not sure if it's worth it, but maybe we can raise a custom exception class here, and in the top-level code that displays errors to the user format it nicely.

emfdavid: I agree it isn't ideal.
That is more than I can take on at this time though(Covid/childcare problems).

src/black/__init__.py
```diff
@@ -1315,7 +1315,9 @@ def assert_equivalent(src: str, dst: str, *, pass_num: int = 1) -> None:
         src_ast = parse_ast(src)
     except Exception as exc:
         raise AssertionError(
-            f"cannot use --safe with this file; failed to parse source file: {exc}"
+            f"cannot use --safe with this file; failed to parse source file AST: {exc}"
```
isidentical, line 1318: Perhaps a simpler solution to the problem @JelleZijlstra noticed would be:
```suggestion
            f"cannot use --safe with this file; failed to parse source file AST: {exc}\n"
```

isidentical (approved):

src/black/__init__.py
```diff
@@ -1315,7 +1315,9 @@ def assert_equivalent(src: str, dst: str, *, pass_num: int = 1) -> None:
         src_ast = parse_ast(src)
     except Exception as exc:
         raise AssertionError(
-            f"cannot use --safe with this file; failed to parse source file: {exc}"
+            f"cannot use --safe with this file; failed to parse source file AST: {exc}\n"
+            f"This could be caused by running black with an older python version "
```
JelleZijlstra, line 1319:
```suggestion
            f"cannot use --safe with this file; failed to parse source file AST: "
            f"{exc}\n"
            f"This could be caused by running black with an older python version "
```
Line too long for lint

emfdavid:
Thanks folks - hope this helps prevent other people filing bugs.

ichard26 (approved):
I like it, this could make the situation where this parse fails when our main parser accepts the input code (it's more liberal in what it accepts so we can support 3.10 without replacing it with a new parser design) a bit more confusing, but I'm not sure what we could do given it should be clear enough something is wrong with the source file. I'm just trying to cover all of our bases before merge :)

Thank you for the PR!

ichard26:
Thank you so much for your contribution! This project is only possible by contributions like these :black_heart:.  You're awesome, @emfdavid. Better error messages are indeed in vogue (for good reason! they're important) right now as CPython continues to gain better error reporting :)

Feel free to share any feedback having contributed for the first time :tada:  to this repo: #2238

PR #2787, opened 2022-01-20
Pretty Terminal Output

Shivansh-007:
This introduces a style helper for styling your logs to get a bit *pretty* output in the terminal. 

It adds an `OutputLevels` enum which can be used to define the level of the message "log" you are *outputting*. You can still use the old manual styles by passing `fg`, `bold`, `dim`, etc. individually to the function `black.output.out()`, but if they are specified alongside with `style=` kwarg (which specifies the `OutputLevels`) the manual styles would over-write the style specified by the `style=` kwarg. An example use case is:

```py
out(
	error_msg if report.return_code else "All done! ✨ 🍰 ✨",
	style=OutputLevels.success,
	bold=False,
)
# The final style would be {"fg": "green", "bold": False} 
# (as we are overwriting it by manually passing `bold=False`
```

The current terminal output looks like this:

Screenshot from 2022-01-20 15-35-17

This PR needs discussion regarding the colour styling, currently, I just picked up what I think would be okay for the level.

### Checklist - did you ...
- [ ] Add new / update outdated documentation? -> n/a


src/black/const.py
```diff
@@ -1,4 +1,19 @@
...
+from typing import Any
+
 DEFAULT_LINE_LENGTH = 88
 DEFAULT_EXCLUDES = r"/(\.direnv|\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|venv|\.svn|_build|buck-out|build|dist)/"  # noqa: B950
 DEFAULT_INCLUDES = r"(\.pyi?|\.ipynb)$"
 STDIN_PLACEHOLDER = "__BLACK_STDIN_FILENAME__"
+
+
+class OutputLevels(Enum):
+    null: Any = dict()
```
felix-hilden, line 11: As discussed, I prefer literals. This way the flake8 config doesn't have to change as well. I'll include it here as well so that others may comment.

felix-hilden: Also `none` could be a more pythonic name.

src/black/output.py
```diff
@@ -8,34 +8,40 @@
 from mypy_extensions import mypyc_attr
 import tempfile
 
-from click import echo, style
+from click import echo, style as _style
```
felix-hilden, line 11: Is this renamed because of the function parameter shadowing? We could change the parameter name instead.

src/black/output.py
```diff
@@ -8,34 +8,40 @@
...
-        message = style(message, **styles)
-    echo(message, nl=nl, err=True)
+def out(
+    message: Optional[str] = None,
+    nl: bool = True,
+    style: OutputLevels = OutputLevels.null,
+    **_styles: Any,
+) -> None:
+    styles = style.value.copy()
+    if style:
```
felix-hilden, line 31: To me this is redundant because `.update` with an empty dict does nothing. Also discussed before.

src/black/output.py
```diff
@@ -8,34 +8,40 @@
...
 
 def err(message: Optional[str] = None, nl: bool = True, **styles: Any) -> None:
-    _err(message, nl=nl, **styles)
+    if isinstance(styles, dict):
+        if not styles.get("fg"):
+            styles["fg"] = "red"
+    elif len(styles) == 0:
+        styles = OutputLevels.error.value
+
+    _out(message, nl, **styles)
```
felix-hilden, line 44:
```suggestion
    return out(message, nl, OutputLevels.error, **styles)
```
I feel like this could be condensed a bunch as well. And surely we always want the error level style as a base.

src/black/report.py
```diff
@@ -55,7 +62,7 @@ def failed(self, src: Path, message: str) -> None:
 
     def path_ignored(self, path: Path, message: str) -> None:
         if self.verbose:
-            out(f"{path} ignored: {message}", bold=False)
+            out(f"{path} ignored: {message}", style=OutputLevels.warning)
```
felix-hilden, line 65: I feel like this should be some other level, like `info`.

src/black/const.py
```diff
@@ -1,4 +1,19 @@
+from enum import Enum
+from typing import Any
+
 DEFAULT_LINE_LENGTH = 88
 DEFAULT_EXCLUDES = r"/(\.direnv|\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|venv|\.svn|_build|buck-out|build|dist)/"  # noqa: B950
 DEFAULT_INCLUDES = r"(\.pyi?|\.ipynb)$"
 STDIN_PLACEHOLDER = "__BLACK_STDIN_FILENAME__"
+
+
+class OutputLevels(Enum):
```
felix-hilden, line 10: Minor thing, but the name could be improved if you also prefer something like `OutStyle` or `OutputLevel` if you're adamant about alluding to logging levels. Anyways a singular form always sounds nicer in enums and module names etc. at least to me.

felix-hilden (changes requested):
All in all, I like the change especially for the easy access to predefined styles. Some comments below.

src/black/const.py
Shivansh-007: `OutputStyleLevel`?

src/black/report.py
Shivansh-007: I wanted to highlight them, I could use dim yellow though, by changing to `OutputLevels.info` and `fg="yellow"`.

src/black/output.py
Shivansh-007: Yeah, but I wanted to keep the `.update()` change explicit hence keeping it in an `if` block.

src/black/output.py
Shivansh-007: If change the parameter name, it would change how we specify the `OutputLevel` to the `black.output.out` function, and since `click.style` is only used once in the whole file, I think this is fine.

src/black/const.py
felix-hilden: Perhaps not 😅

src/black/const.py
Shivansh-007: `None` would be more pythonic agreed, but I would need to add some additional logic to `_out` function, I wanted to keep it simple hence using am using an empty `dict` here.

Regarding the `dict` literals, I am using function calls here as I found them more readable compared to literals.

Shivansh-007:
I don't think a `skip news` label is needed, I just haven't added the changelog yet. This change will be affecting the user side of the project, so a changelog would be good 👍🏻

src/black/report.py
felix-hilden: Any particular reason as to why you'd want to highlight them? Don't get me wrong, it could be nice. But this is why we define the styles right? 😄

src/black/const.py
Shivansh-007: I would go with `OutputLevel` then 👍🏻

src/black/report.py
Shivansh-007: I just wanted them to be easily recognizable from the crowd of "wasn't modified logs".

src/black/const.py
felix-hilden: I meant `none: dict = {}`!

JelleZijlstra: We should either use dict literals here or decide that we globally don't care for this flake8 setting and disable it.

Shivansh-007: > I meant `none: dict = {}`!

Ah, that would work 👍🏻 (the name, mypy complains about `: dict` as it needs the parameters)

src/black/const.py
ichard26: how about `LogLevel(s)?`?

src/black/const.py
```diff
@@ -1,4 +1,23 @@
...
 DEFAULT_INCLUDES = r"(\.pyi?|\.ipynb)$"
 STDIN_PLACEHOLDER = "__BLACK_STDIN_FILENAME__"
+
+
+class Empty(TypedDict):
+    pass
+
+
+class LogLevel(Enum):
+    none: Empty = {}
```
felix-hilden, line 15:
```suggestion
    none: Dict[str, Any] = {}
```
How about simply this instead?

felix-hilden (commented):
Looking better, but let's address all the comments!

src/black/const.py
felix-hilden: The Flake8 rule is definitely useful. So maybe we'll sacrifice a bit of readability and use literals here.

src/black/const.py
Shivansh-007: But it isn't entirely correct though.

src/black/const.py
Shivansh-007: I am not sure, either way is fine. We could even safely ignore the rule since very few people use functional calls over literals for normal use cases (which can be caught in reviews), otherwise, they are only used for some specific reasons like readability/explicitness (this case).

src/black/output.py
JelleZijlstra: This seems wrong, doesn't it mean we ignore `_styles` if `style == LogLevel.none`?

Also, the variable names `style`, `styles`, and `_styles` here are pretty confusing.

Shivansh-007: Ah yeah, I forgot to make that change.

Screenshot from 2022-01-22 10-43-07

JelleZijlstra:
Screen Shot 2022-01-21 at 9 12 56 PM
Here's how it looks in my terminal (top is this PR, bottom is current main).

Feedback:
- We need to worry about making this readable for everyone, which is hard because there are so many different configurations. In your screenshot the "wasn't modified" lines are kind of hard to read.
- The stuff at the top in blue is easier to read on current main where it's bolded.
- I like the color difference between the ignored and well-formatted lines on your branch.

src/black/output.py
Shivansh-007: Umm, can't think of a better name for `_styles`, these are the "styles" which the user explicitly passes as kwargs to overwrite the style defined by the `LogLevel`.

Shivansh-007:
That's a lot of bright green 👀 Anyways, regarding your second point, do you want to make all `info` level logs `bold` blue or just those? I would go with the former.

src/black/output.py
JelleZijlstra: I'd suggest `style_overrides`.

JelleZijlstra:
I don't think you should change anything just yet, would be better to get more feedback first.

JelleZijlstra:
This now has heavy merge conflicts. Should we still pursue this change?

hauntsaninja:
Given lack of activity and merge conflicts, I say we close. Very few users will run Black with `--verbose` anyway

Issue #2766, opened 2022-01-12
Replace CHANGES.md with something less prone to merge conflicts


JelleZijlstra:
As @Shivansh-007 noted in https://github.com/psf/black/issues/2238#issuecomment-1010922392, our current CHANGES.md mechanism for the changelog is very prone to merge conflicts. Another problem is that when a release happens between the time a PR is opened and when it is merged, it's possible to have the changelog entry go into the wrong section.

I haven't looked in detail in this space, but I know that possibilities include:
- towncrier
- blurb, which is used by CPython
- writing our own; I'd like to avoid that unless we really have unusual needs that existing tools can't support

onerandomusername:
towncrier only supports rst, afaik. I haven't been able to make it work with an md template.

hauntsaninja:
I've never really liked towncrier, blurb or their ilk. The dumb solution is to just add a bunch of bullet points with a note saying "please fill in a random bullet point". When you release, remove the empty ones.

cooperlees:
I think the simplicity of what we have to ensure we get change log entries far outweighs the complexity of all alternatives and make release time much easier on the releaser.

To be fair too, these have to be the easiest merge conflicts in the universe to fix and deal with compared to source code. It’s a list of scentences.

ichard26:
I'm open to using towncrier assuming if I can get it to handle Markdown reasonably well and the amount of code necessary to integrate it isn't that high. Note if I _do_ bring over towncrier I'd also bring over changelog update automation that pip (I consider its [changelog](https://pip.pypa.io/en/stable/news/) to be very nice and I'd like to model it if we do choose a more advanced changelog solution). 

Except! ... @hauntsaninja's solution is also pretty appealing to me as well, though I'd make the modification to introduce some more sectioning of our entries grouping bug fixes, enhancements, etc. separately. Adding sections would also have the nice side-effect of making merge conflicts less likely as all of the changelog additions aren't concentrated at the end of the list causing all of these conflicts.

I am indeed a bit wary of adding more complexity to this domain.

felix-hilden:
[...]

PR #2790, opened 2022-01-21
reorganize release notes for 22.1.0

JelleZijlstra:


CHANGES.md
```diff
@@ -2,49 +2,59 @@
 
 ## Unreleased
 
-### _Black_
+### Highlights
 
+- At long last, _Black_ is no longer a beta product! This is the first non-beta release
```
felix-hilden, line 7: Instead of an item this could be text under the top heading.

felix-hilden (approved):
Looks good! Just one opinion below.

We could use these headings as a template for future changelogs to reduce conflicts until we decide on which tool to use in #2766.

CHANGES.md
```diff
@@ -2,49 +2,59 @@
...
+  and the first release covered by our new stability policy.
```
```suggestion
At long last, _Black_ is no longer a beta product! This is the first non-beta release
and the first release covered by our new stability policy.

### Highlights

```
Sure

CHANGES.md
```diff
@@ -2,49 +2,60 @@
 
 ## Unreleased
 
-### _Black_
+At long last, _Black_ is no longer a beta product! This is the first non-beta release
```
JelleZijlstra, line 5:
```suggestion

At long last, _Black_ is no longer a beta product! This is the first non-beta release
```

isidentical (approved):

JelleZijlstra:
I'll hold off on merging this for a while to keep the merge conflict situation manageable.

JelleZijlstra:
> It feels weird to have a section for such a minor change

I don't feel strongly about this, just made the section because the changelog entry was there and didn't fit anywhere else. Plus I'd generally err on the side of documenting more. Especially with sections like this, it's easy for people to skip over things they're not interested in.

CHANGES.md
```diff
@@ -2,49 +2,60 @@
...
   `values: Tuple[int, ...] = 1, 2, 3`) now implies 3.8+ (#2708)
-- Text coloring added in the final statistics (#2712)
-- For stubs, one blank line between class attributes and methods is now kept if there's
-  at least one pre-existing blank line (#2736)
-- Verbose mode also now describes how a project root was discovered and which paths will
-  be formatted. (#2526)
 - Speed-up the new backtracking parser about 4X in general (enabled when
   `--target-version` is set to 3.10 and higher). (#2728)
 - Fix handling of standalone `match()` or `case()` when there is a trailing newline or a
   comment inside of the parentheses. (#2760)
```
JelleZijlstra, line 35:
```suggestion
- Fix handling of standalone `match()` or `case()` when there is a trailing newline or a
  comment inside of the parentheses. (#2760)

### Performance

- Speed-up the new backtracking parser about 4X in general (enabled when
  `--target-version` is set to 3.10 and higher). (#2728)
```
This section can also be used for mypyc

ichard26 (approved):
This looks so much better, thank you!

We should add an entry about mypyc though, not sure whether to put in under highlights or performance (I'd lean towards performance but mypyc could break things although I sure hope not). 
> \- \_Black\_  is now compiled with \[mypyc](https://github.com/mypyc/mypyc) for an overall 2x speed-up. (#1009, #2431)

CHANGES.md
```diff
@@ -2,54 +2,66 @@
...
 - Unparenthesized tuples on annotated assignments (e.g
   `values: Tuple[int, ...] = 1, 2, 3`) now implies 3.8+ (#2708)
+- Fix handling of standalone `match()` or `case()` when there is a trailing newline or a
+  comment inside of the parentheses. (#2760)
+
+### Performance
+
+- Speed-up the new backtracking parser about 4X in general (enabled when
+  `--target-version` is set to 3.10 and higher). (#2728)
+
```
ichard26, line 39:
```suggestion
- _Black_ is now compiled with [mypyc](https://github.com/mypyc/mypyc) for an overall 2x
  speed-up. (#1009, #2431)

```
for ease of application

JelleZijlstra: Agree, it should be under "performance". Maybe we should say something about the OSs for which we'll have wheels?

ichard26: Updated.

JelleZijlstra: Thanks!

PR #2799, opened 2022-01-23
Use parentheses on method access on float and int literals

### LINKED ISSUE #2034: Use parentheses for method access on integer literals
**Describe the style change**

When code tries to access a method on integer literals, the "space before dot" way should be reformatted to the "surround the integer literal with parentheses" way.

```python
count = 5 .bit_count()
```

**Desired style**

```python
count = (5).bit_count()
```

**Additional context** 

This is Twitter Driven Development btw :laughing: 

designrequestviatwitter

Shivansh-007:
Closes #2034 

### Checklist - did you ...

- [ ] Add new / update outdated documentation? -> n/a


JelleZijlstra:
Thanks! This definitely needs a changelog entry since it changes formatter output.

Haven't looked at the PR in detail but I like the idea.

Shivansh-007:
Yeah, I will be adding it 👍🏻

src/black/linegen.py
```diff
@@ -197,6 +197,14 @@ def visit_decorators(self, node: Node) -> Iterator[Line]:
             yield from self.line()
             yield from self.visit(child)
 
+    def visit_power(self, node: Node) -> Iterator[Line]:
+        for idx, leaf in enumerate(node.children[:-1]):
+            next_leaf = node.children[idx + 1]
+            if leaf.type == token.NUMBER and next_leaf.type == syms.trailer:
```
isidentical, line 203: AFAIK `trailer` also includes the subscript and the call, e.g `1(duh)` or `1[duh]` would be also parenthesized (even though there is no such use case). I wonder whether we should make a more stricter check and ensure we are in an attribute trailer 🤔

CHANGES.md
```diff
@@ -38,6 +38,7 @@
   `--preview` (#2789)
 - Enable Python 3.10+ by default, without any extra need to specify
   `--target-version=py310`. (#2758)
+- Use parenthesis on method access on number integerals (float and int) (#2799)
```
JelleZijlstra, line 41:
```suggestion
- Use parenthesis on method access on float and int literals (#2799)
```

Shivansh-007: that was super fast, thanks 👍🏻

tests/data/use_braces_in_int_method_calls.py
```diff
@@ -0,0 +1,24 @@
+count = 5 .bit_count()
```
JelleZijlstra, line 1: What about complex numbers? They don't require parens (`10j.imag` is valid), so it would be good to test that we don't add it.

Similarly `0x10.imag` is valid, so let's test that we don't add unnecessary parens.

Shivansh-007: It even works on floats, but on floats, the multiple "dot operators" look a bit confusing, so I decided to keep it on floats
```py
>>> 10.1.is_integer()
False
```

Right will do, `10j.imag` are `0x10.imag` are fine with the parentheses 👍🏻

src/black/linegen.py
Shivansh-007: Sorry missed this, yeah I noticed that when I was going through the grammar file but subscript and calls on `token.NUMBER` isn't valid python syntax, so I left that case.

src/black/linegen.py
```diff
@@ -197,6 +197,24 @@ def visit_decorators(self, node: Node) -> Iterator[Line]:
...
+    def visit_power(self, node: Node) -> Iterator[Line]:
+        for idx, leaf in enumerate(node.children[:-1]):
+            next_leaf = node.children[idx + 1]
+            if (
+                isinstance(leaf, Leaf)
+                and leaf.type == token.NUMBER
+                and next_leaf.type == syms.trailer
+                # Integers prefixed by 0x are represented under the hood in just
+                # the same way as integers that aren't. This means that to Python,
+                # 0xFF is the same as 255, and there's no way to tell them apart
```
JelleZijlstra, line 209: If we do this we also have to exclude binary and octal literals.

src/black/linegen.py
```diff
@@ -197,6 +197,24 @@ def visit_decorators(self, node: Node) -> Iterator[Line]:
...
             yield from self.visit(child)
 
+    def visit_power(self, node: Node) -> Iterator[Line]:
+        for idx, leaf in enumerate(node.children[:-1]):
+            next_leaf = node.children[idx + 1]
+            if (
+                isinstance(leaf, Leaf)
+                and leaf.type == token.NUMBER
+                and next_leaf.type == syms.trailer
+                # Integers prefixed by 0x are represented under the hood in just
```
JelleZijlstra, line 207: I'm not sure why this comment is relevant; at this point the ints are represented as strings.

tests/test_black.py
```diff
@@ -137,6 +137,8 @@ def invokeBlack(
 
 
 class BlackTestCase(BlackBaseTestCase):
+
+    maxDiff = None
```
JelleZijlstra, line 141: Why do you need this?

CHANGES.md
```diff
@@ -40,8 +40,11 @@
 - Enable Python 3.10+ by default, without any extra need to specify
   `--target-version=py310`. (#2758)
 - Make passing `SRC` or `--code` mandatory and mutually exclusive (#2804)
+- Don't add whitespaces after hex-decimals and exclusive complex numbers (ending "j") on
+  method access (#2799)
 - Work around bug that causes unstable formatting in some cases in the presence of the
   magic trailing comma (#2807)
+- Use parentheses on method access on float and int literals (#2799)
```
JelleZijlstra, line 47:
```suggestion
- Use parentheses for attribute access on decimal float and int literals (#2799)
```
It's about attributes, not methods, and only affects literals represented as decimals.

tests/data/use_braces_in_int_method_calls.py
```diff
@@ -0,0 +1,30 @@
...
+is_integer = 10.5 .is_integer()
+
+if 10 .real: ...
+
+for number in range(1, 10):
+    if number.imag: ...
+
+exclusive_complex = 10j.imag
+
+hex_decimal = 0x10.real
```
JelleZijlstra, line 12: Should also test all the other kinds of numeric literals:
- octal
- binary
- scientific notation

CHANGES.md
```diff
@@ -40,8 +40,11 @@
 - Enable Python 3.10+ by default, without any extra need to specify
   `--target-version=py310`. (#2758)
 - Make passing `SRC` or `--code` mandatory and mutually exclusive (#2804)
+- Don't add whitespaces after hex-decimals and exclusive complex numbers (ending "j") on
+  method access (#2799)
```
JelleZijlstra, line 44:
```suggestion
- Don't add whitespace for attribute access on hexadecimal, binary, octal, and complex literals (#2799)
```

JelleZijlstra (changes requested):

tests/test_black.py
Shivansh-007: Forgot to remove this, when running the tests in pycharm, it wasn't showing the diff for the `.diff` tests as the diff between the required diff and actual diff was more than the `maxDiff`. To view the diff, I had to set it to `None`.

src/black/linegen.py
```diff
@@ -197,6 +197,30 @@ def visit_decorators(self, node: Node) -> Iterator[Line]:
...
+            value = leaf.value.lower()
+            if (
+                leaf.type == token.NUMBER
+                and next_leaf.type == syms.trailer
+                # Shouldn't wrap octal literals
+                and (not value.startswith("0o"))
+                # Shouldn't wrap binary literals
+                and (not value.startswith("0b"))
+                # Shouldn't wrap hexadecimal literals
+                and (not value.startswith("0x"))
```
JelleZijlstra, line 216: You can combine all these `value.startswith(("0x", "0b", "0o"))`

JelleZijlstra (approved):

src/black/lines.py
```diff
@@ -62,11 +62,21 @@ def append(self, leaf: Leaf, preformatted: bool = False) -> None:
...
             del self.leaves[-2:]
         if self.leaves and not preformatted:
-            # Note: at this point leaf.prefix should be empty except for
-            # imports, for which we only preserve newlines.
-            leaf.prefix += whitespace(
-                leaf, complex_subscript=self.is_complex_subscript(leaf)
-            )
+            grandparent_leaf = leaf.parent.parent if leaf.parent else None
+
+            # It shouldn't add whitespaces after dot operators in expressions
```
JelleZijlstra, line 67: Looking at this again, shouldn't the logic you're adding be handled in the whitespace() function?

What if you remove the NUMBER check from nodes.py line 311, `if not prevp or prevp.type != token.NUMBER:`?

JelleZijlstra (approved):

src/black/linegen.py
```diff
@@ -197,6 +197,26 @@ def visit_decorators(self, node: Node) -> Iterator[Line]:
...
+
+            if not isinstance(leaf, Leaf):
+                continue
+
+            value = leaf.value.lower()
+            if (
+                leaf.type == token.NUMBER
+                and next_leaf.type == syms.trailer
+                # It shouldn't wrap hexadecimal, binary and octal literals
+                and (not value.startswith(("0x", "0b", "0o")))
```
felix-hilden, line 212:
```suggestion
                and not value.startswith(("0x", "0b", "0o"))
```
NIT: let's not have the extra parens

Shivansh-007: I find the brackets to be more readable than not having them.

src/black/linegen.py
isidentical: It is actually a valid syntax, just not valid semantically. So we should probably handle it (check the first child of the trailer, if it is a dot, then it is an attribute access).

isidentical: @Shivansh-007 could you also add a test to verify this behavior? (both for subscript and calls)

src/black/linegen.py
isidentical: I think it would make sense to follow the existing convention as @felix-hilden just said, e.g:
```console
$ grep -rni 'and (not' src/black
$ grep -rni 'and not' src/black
src/black/trans.py:346:            and not has_triple_quotes(string_leaf.value)
src/black/trans.py:682:                and not is_empty_rpar(LL[next_idx])
src/black/trans.py:1228:                and not csplit.has_prefix
src/black/trans.py:1416:                and not breaks_unsplittable_expression(i)
src/black/trans.py:1421:        while is_valid_index(break_idx - 1) and not passes_all_checks(break_idx):
src/black/trans.py:1431:            while is_valid_index(break_idx + 1) and not passes_all_checks(break_idx):
src/black/trans.py:1458:        if "f" in prefix and not fstring_contains_expr(string):
src/black/__init__.py:92:        if check and not diff:
src/black/__init__.py:139:    if target_version is not None and not isinstance(target_version, list):
src/black/__init__.py:586:            if p.suffix == ".ipynb" and not jupyter_dependencies_are_installed(
src/black/__init__.py:971:    if not fast and not mode.is_ipynb:
src/black/__init__.py:1200:            if len(n.children) > 1 and not is_simple_decorator_expression(
src/black/linegen.py:332:        and not line.should_split_rhs
src/black/linegen.py:333:        and not line.magic_trailing_comma
src/black/linegen.py:338:        and not (line.inside_brackets and line.contains_standalone_comments())
[...]
```

ichard26 (approved):
and now we can claim twitter driven development is a viable methodology :bird: :)

although minor note do we need to change the style docs or are we good? This feels minor so it probably doesn't need to be called out but then again completeness wouldn't hurt imo.

JelleZijlstra (approved):

Shivansh-007:
I am not sure how the style docs work (what are included, what are not), but if it is needed I will them 👍🏻

PR #2802, opened 2022-01-23
Use `magic_trailing_comma` for `FileMode` in `fuzz`

sobolevn:
Without it fuzz tests will always use the default value. Quick demo:

```python
>>> import hypothesis
>>> class Some:
...    def __init__(self, arg: bool = True):
...      print(arg)
... 
>>> hypothesis.strategies.builds(Some).example()
True
<__main__.Some object at 0x10c462490>
>>> hypothesis.strategies.builds(Some).example()
True
<__main__.Some object at 0x10c411850>
>>> hypothesis.strategies.builds(Some).example()
True
```

CC @Zac-HD

### DIFF
--- a/fuzz.py
+++ b/fuzz.py
@@ -32,6 +32,7 @@ from blib2to3.pgen2.tokenize import TokenError
         black.FileMode,
         line_length=st.just(88) | st.integers(0, 200),
         string_normalization=st.booleans(),
+        preview=st.booleans(),
         is_pyi=st.booleans(),
         magic_trailing_comma=st.booleans(),
     ),
--- a/fuzz.py
+++ b/fuzz.py
@@ -33,6 +33,7 @@ from blib2to3.pgen2.tokenize import TokenError
         line_length=st.just(88) | st.integers(0, 200),
         string_normalization=st.booleans(),
         is_pyi=st.booleans(),
+        magic_trailing_comma=st.booleans(),
     ),
 )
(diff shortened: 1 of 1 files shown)


fuzz.py
```diff
@@ -33,6 +33,7 @@
         line_length=st.just(88) | st.integers(0, 200),
         string_normalization=st.booleans(),
```
JelleZijlstra, line 34:
```suggestion
        string_normalization=st.booleans(),
        preview=st.booleans(),
```
While we're here let's add this too.

sobolevn:
I've actually decided to take a look at black's internals after going on your profile page 🙂

JelleZijlstra (approved):
Thanks! Please stalk and improve any other projects of mine that you can find :)

sobolevn:
Sounds a bit creepy now 😆

Zac-HD (approved):
LGTM!

We should add experimental string processing to this before it's considered stable, too, but maybe not yet.

sobolevn:
Looks like it was deprecated: https://github.com/psf/black/blob/6e3677f3f0c0542f858f7fc06d20cca5fab59348/src/black/mode.py#L150-L156

Now `preview` does the same thing. (This is the first time I read `black`'s code, so I might be wrong)

JelleZijlstra:
@sobolevn is right, preview now includes experimental string processing. For now the flags are exactly equivalent.

PR #2803, opened 2022-01-24
Add `--colour/--no-colour` alias to `--color/--no-color`

Shivansh-007:
`color` is the preferred spelling in American English while in British English, `colour` is preferred. A lot of people use `colour` (me being one of them) and there is no reason to not support `--colour` and make them use `--color`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,6 +38,7 @@
   `--preview` (#2789)
 - Enable Python 3.10+ by default, without any extra need to specify
   `--target-version=py310`. (#2758)
+- Add `--colour/--no-colour` alias to `--color/--no-color` (#2803)
 
 ### Packaging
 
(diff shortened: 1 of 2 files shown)


isidentical:
Are there any other console tools which alias them together? (just looked a up a few, like pytest which also has `--color` but no `--colour`, or pylint which has `colorized` but not `colourized`).

felix-hilden (commented):
Thanks! I think this is nice, but I'd like it even more if the aliases would be hidden because it will reduce some clutter and still helps with the typo to silently accept both. But I'm not strong on that. Also not strong on including this if other maintainers disagree, but I think it's a nice touch.

Jason's screen shot from the new help:

help

Shivansh-007:
> Are there any other console tools which alias them together? (just looked a up a few, like pytest which also has `--color` but no `--colour`, or pylint which has `colorized` but not `colourized`).

Not any that I know of.

JelleZijlstra:
I'd rather not do this. It complicates the CLI and it's not like "color" won't be understood by people who prefer the British spelling.

PR #2804, opened 2022-01-24
Make SRC or code mandatory and mutually exclusive (#2360)

### LINKED ISSUE #2360: Make SRC/CODE mandatory and mutually exclusive
From #2356, which made (assuming it got merged) SRC appear as a mandatory CLI parameter: Strictly speaking, SRC is optional both in the way that Black accepts no arguments (and then prints "nothing to do") and with `--code ""`. I think accepting SRC and then skipping it when `-c` is passed is a bit confusing.

I'll propose we make `SRC` and `--code` mutually exclusive and require one of them.

So, something like this:
```
> black
BLACK IS MAD: specify one of SRC / --code

> black --code "" .
BLACK IS FURIOUS: are you trying to confuse me?
```

I wonder if Click has some way of documenting call modes. Maybe something like:

```
> black --help

black [opts] src
black [opts] --code CODE

Options:
...
```

The "nothing to do" message could be emitted when the supplied path doesn't contain any valid files, like it is currently.

*Ideally* though, we would have a flag to turn the source to code I think. If I'm thinking about it from a documentation perspective, it would be much easier to explain:

```
> black --help

black [opts] src

SRC     Sources to be formatted as directories or files.
        If --code was passed, treat SRC as Python text.

Options:
code    Treat SRC as Python text.
```

I don't know what multiple sources of Python text would do then. Maybe different lines, or we could require only one source in that case. Anyways, this is kind of an addon to my proposition, not the main thing.

**Additional context**
Also, relevant discussion in #861: SRC is currently undocumented on the CLI help. We should definitely document it while we're at it (if it doesn't get documented earlier).

felix-hilden:
### Description
Closes #2360: I'd like to make passing SRC or `--code` mandatory and the arguments mutually exclusive. This will change our (partially already broken) promises of CLI behavior, but I'll comment below.

### Checklist - did you ...

### DIFF
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -4,11 +4,11 @@ Foundational knowledge on using and configuring Black.
 
 _Black_ is a well-behaved Unix-style command-line tool:
 
-- it does nothing if no sources are passed to it;
+- it does nothing if it finds no sources to format;
 - it will read from standard input and write to standard output if `-` is used as the
   filename;
 - it only outputs messages to users on standard error;
-- exits with code 0 unless an internal error occurred (or `--check` was used).
+- exits with code 0 unless an internal error occurred or a CLI option prompted it.
 
 ## Usage
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -431,6 +431,17 @@ def main(
 ) -> None:
     """The uncompromising code formatter."""
     ctx.ensure_object(dict)
+
+    if src and code is not None:
+        out(
+            main.get_usage(ctx)
+            + "\n\n'SRC' and 'code' cannot be passed simultaneously."
+        )
+        ctx.exit(1)
+    if not src and code is None:
+        out(main.get_usage(ctx) + "\n\nOne of 'SRC' or 'code' is required.")
+        ctx.exit(1)
+
     root, method = find_project_root(src) if code is None else (None, None)
     ctx.obj["root"] = root
 
@@ -569,7 +580,6 @@ def get_sources(
 ) -> Set[Path]:
     """Compute the set of files to be formatted."""
     sources: Set[Path] = set()
-    path_empty(src, "No Path provided. Nothing to do 😴", quiet, verbose, ctx)
 
     if exclude is None:
         exclude = re_compile_maybe_verbose(DEFAULT_EXCLUDES)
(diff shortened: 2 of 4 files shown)


docs/usage_and_configuration/the_basics.md
```diff
@@ -4,11 +4,11 @@ Foundational knowledge on using and configuring Black.
 
 _Black_ is a well-behaved Unix-style command-line tool:
 
-- it does nothing if no sources are passed to it;
+- it does nothing if it finds no sources to format;
```
felix-hilden, line 7: New with this PR. With previous discussions about passing no sources I think it should be considered user error. Finding no files to format still passes and returns success.

docs/usage_and_configuration/the_basics.md
```diff
@@ -4,11 +4,11 @@ Foundational knowledge on using and configuring Black.
...
 - it will read from standard input and write to standard output if `-` is used as the
   filename;
 - it only outputs messages to users on standard error;
-- exits with code 0 unless an internal error occurred (or `--check` was used).
+- exits with code 0 unless an internal error occurred or a CLI option prompted it.
```
felix-hilden, line 11: We already had `--check`, `--require-version` and pyi/ipynb. Now we also have the source arg check.

src/black/__init__.py
```diff
@@ -569,7 +580,6 @@ def get_sources(
 ) -> Set[Path]:
     """Compute the set of files to be formatted."""
     sources: Set[Path] = set()
-    path_empty(src, "No Path provided. Nothing to do 😴", quiet, verbose, ctx)
```
felix-hilden, line 572: Shouldn't happen anymore because the check is done above.

tests/test_black.py
```diff
@@ -1229,12 +1234,14 @@ def test_invalid_cli_regex(self) -> None:
...
-            ["--required-version", black.__version__], exit_code=0, ignore_config=True
+            ["--required-version", black.__version__, "-c", "0"],
+            exit_code=0,
+            ignore_config=True,
         )
 
     def test_required_version_does_not_match_version(self) -> None:
         self.invokeBlack(
-            ["--required-version", "20.99b"], exit_code=1, ignore_config=True
+            ["--required-version", "20.99b", "-c", "0"], exit_code=1, ignore_config=True
```
felix-hilden, line 1244: Technically still would succeed without the dummy code, but the exit code can't tell the source. Should we assert on error message?

felix-hilden: On a second thought, the code is required, because otherwise the error message will be the source requirement. But we could still assert on the error to make sure to avoid mistakes in the future.

JelleZijlstra (approved):

PR #2806, opened 2022-01-25
    Fix crash on some power hugging cases

JelleZijlstra:
Found by the fuzzer. Repro case:
    
            python -m black -c 'importA;()<<0**0#'
    
Unfortunately the resulting code still crashes due to an
apparently unrelated stability bug. I'd like to fix the crash
first.


JelleZijlstra:
Skipping news because this fixes a crash that was just introduced and never released.

JelleZijlstra:
Confirmed that with #2807 the repro case doesn't hit stability issues.

tests/test_format.py
```diff
@@ -256,3 +256,10 @@ def test_python38() -> None:
 def test_python39() -> None:
     source, expected = read_data("python39")
     assert_format(source, expected, minimum_version=(3, 9))
+
+
+def test_power_op_newline() -> None:
+    # requires line_length=0
+    source, expected = read_data("power_op_newline")
+    # fast=True because this exposes a different stability bug
```
JelleZijlstra, line 264:
```suggestion
```

cooperlees (approved):
Makes sense + makes CI pass.

Anything blocking this from being merged?

ichard26 (commented):
I'm a little worried about the `bracket_depth` attribute usage as it's not copied over but doing that breaks tests so yeah. I inspected one use of it and it's fine but it's really hard to inspect all of the usecases finding an input that hits the right control flow construct.

PR #2572, opened 2021-10-28
Fix instability due to trailing comma logic

nipunn1313:
### Description

Fixes #2488 and #2518

Consistently apply paren omission rules regardless of existence of magic trailing comma. Previously, the first run would expand + add a magic comma, and the second run would get a different paren omission rule, causing black to be unstable. We had some hax in black/__init__ to work around this which this diff removes.

This changes the way black formats (adds more parens in some cases). Based on the black release stability policy, we'd want this to go into the next major release. Something like #2549 could be a bandaid until that point (or we can live with the bug and recommend users to run black twice in the meantime).

For validation: After removing the double-run hax in the stability checker, black-primer crashed on django (for being unstable). With this diff, it no longer crashes. Additionally we were able to re-enable three tests that covered such cases.

### Checklist - did you ...

- [x] Add a CHANGELOG entry if necessary?
- [x] Add / update tests if necessary?
- [ ] Add new / update outdated documentation? (not sure what is necessary)

PR #2549, opened 2021-10-20
Run second pass of formatter

nipunn1313:
### Description

Run the formatter twice to ensure stability.
Fixes #2488 and #2518

Admittedly, it doesn't fix the underlying issue. There are three
expected-failure tests in test_black.py, which prove the issue
(test_trailing_comma_optional_parens_stability{1,2,3})

This is a (somewhat terrible) bandaid which runs the entire formatter
twice to work around this problem. This bandaid actually already existed
in the stability-checking-validation code, but resulted in a non-stable
black executable. The problem appears preexisting and known
based on the disabled tests and the comments, so this feels like a
reasonable step-in-the-right-direction.

Follow up tasks may want to consider removing this hack and actually
fixing the underlying bug that forces us to run the formatter twice. I tried
to dig into it, but it was pretty tough to figure out. There are some notes in
#2488. The result of `can_omit_invisible_parens` changes during pass1, causing
pass2 to do new work that wasn't considered in pass1.

### Checklist - did you ...

- [x ] Add a CHANGELOG entry if necessary?
- [x ] Add / update tests if necessary?
- [x ] Add new / update outdated documentation?

PR #2807, opened 2022-01-25
properly run ourselves twice

JelleZijlstra:
The previous run-twice logic only affected the stability checks but not the output. Now, we actually output the twice-formatted code.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1367,7 +1367,10 @@ def assert_equivalent(src: str, dst: str, *, pass_num: int = 1) -> None:
 
 def assert_stable(src: str, dst: str, mode: Mode) -> None:
     """Raise AssertionError if `dst` reformats differently the second time."""
-    newdst = format_str(dst, mode=mode)
+    # We shouldn't call format_str() here, because that formats the string
+    # twice and may hide a bug where we bounce back and forth between two
+    # versions.
+    newdst = _format_str_once(dst, mode=mode)
     if dst != newdst:
         log = dump_to_file(
             str(mode),
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -968,17 +968,7 @@ def check_stability_and_equivalence(
     content differently.
     """
     assert_equivalent(src_contents, dst_contents)
-
-    # Forced second pass to work around optional trailing commas (becoming
-    # forced trailing commas on pass 2) interacting differently with optional
-    # parentheses.  Admittedly ugly.
-    dst_contents_pass2 = format_str(dst_contents, mode=mode)
-    if dst_contents != dst_contents_pass2:
-        dst_contents = dst_contents_pass2
-        assert_equivalent(src_contents, dst_contents, pass_num=2)
-        assert_stable(src_contents, dst_contents, mode=mode)
-    # Note: no need to explicitly call `assert_stable` if `dst_contents` was
-    # the same as `dst_contents_pass2`.
+    assert_stable(src_contents, dst_contents, mode=mode)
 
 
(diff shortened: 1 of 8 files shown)


JelleZijlstra:
The fuzz failures is just #2806.

felix-hilden:
Related: #2549 and #2572. I'll repeat what I said there: I don't much care for going further into masking instability issues with additional runs. I feel like we'll never fix the issues after they are gone with this 😅 But then again, this does fix the issue so maybe it's enough for now.

JelleZijlstra:
It's definitely not great, but we basically already decided to do it this way—we just didn't implement it properly.

felix-hilden:
Do you mean that we always wanted to run twice and return that result? I think there's a big difference between that and running twice to check for instability, which I'm sure you get as well. I just might be a bit confused about this and what you mean by proper implementation 😅

felix-hilden:
Yeah, quick discussion [on Discord](https://discord.com/channels/267624335836053506/846434317021741086/935564572872552448), I did indeed misunderstand the matter. And this is the way to go.

felix-hilden (approved):

src/black/__init__.py
```diff
@@ -968,17 +968,7 @@ def check_stability_and_equivalence(
...
-    # forced trailing commas on pass 2) interacting differently with optional
-    # parentheses.  Admittedly ugly.
-    dst_contents_pass2 = format_str(dst_contents, mode=mode)
-    if dst_contents != dst_contents_pass2:
-        dst_contents = dst_contents_pass2
-        assert_equivalent(src_contents, dst_contents, pass_num=2)
-        assert_stable(src_contents, dst_contents, mode=mode)
-    # Note: no need to explicitly call `assert_stable` if `dst_contents` was
-    # the same as `dst_contents_pass2`.
+    assert_stable(src_contents, dst_contents, mode=mode)
```
ichard26, line 971: Does assert_stable really need to call `format_str` now?

ichard26 (approved):
This will hurt the runtime for files where changes are made (without trailing comma shenanigans) by 50% which sucks especially as this will undo 50% of the mypyc gains in these cases, but I can't think of anyway of working around this without changing the informal APIs which is a no-go.

Consider this approved barring one minor issue / question. Thanks.

src/black/__init__.py
JelleZijlstra: I suppose we could call `_format_str_once` instead, but I like the idea of that being a private function that nobody else should call.

We still need to call some variant of format_str here in case it's still not stable after the second pass.

ichard26: This is already a bit of stretch but I was concerned about `assert_stable` hiding a stability bug chain of two. I suppose it doesn't matter at this point.

JelleZijlstra: Actually, it might hide a cycle, where we move back and forth between two versions deterministically.

JelleZijlstra:
```
=================================== FAILURES ===================================
_________ BlackTestCase.test_trailing_comma_optional_parens_stability1 _________
Unexpected success
_________ BlackTestCase.test_trailing_comma_optional_parens_stability2 _________
Unexpected success
_________ BlackTestCase.test_trailing_comma_optional_parens_stability3 _________
Unexpected success
=============================== warnings summary ===============================
```
Looks like I fixed too much

isidentical (approved):

JelleZijlstra:
I'm pretty sure I didn't break anything only on Ubuntu 3.9 and Windows 3.10 (while Ubuntu 3.10 and Windows 3.9 work). I already retried four times. Time to land this and say goodbye to black-primer.

PR #2809, opened 2022-01-26
black-primer: stop running it

JelleZijlstra:
At the moment, it's just a source of spurious CI failures and busywork
updating the configuration file.

Unlike diff-shades, it is run across many different platforms and
Python versions, but that doesn't seem essential. We already run unit
tests across platforms and versions.

I chose to leave the code around for now in case somebody is using it,
but CI will no longer run it.


JelleZijlstra:
Fuzz finds #2806 as usual.

Shivansh-007:
Might want to remove the following also:

- [ ] The [primer test](https://github.com/psf/black/blob/main/tests/test_primer.py)
- [ ] The primer status badge in the [readme](https://github.com/psf/black/blame/d0b3d38493a4687be801f98f35630d37f9c1958b/README.md#L7)
- [ ] Black primer package data in [setup.py](https://github.com/psf/black/blob/e0c572833a3e2b42cd45237c26a67c6f5be4b09d/setup.py#L95)

JelleZijlstra:
Thanks! I'd like to keep the setup.py stuff until we fully remove it. Removed the badge.

.github/workflows/uvloop_test.yml
```diff
@@ -38,8 +38,12 @@ jobs:
...
 
-      - name: Primer uvloop run
+      - name: Install test requirements
         run: |
-          black-primer
+          python -m pip install -r test_requirements.txt
+
+      - name: Run unit tests
+        run: |
+          pytest tests
```
cooperlees, line 49: Will these even hit patched_main() and use uvloop ... I don't think so.

cooperlees (commented):
Remove away. I feel we're regressing our test suite a little by missing some coverage primer gave us that diff shades does not, but it seems that's been deemed not useful to black contributors.

The left uvloop_test..yml is wasting cycles as it's not testing uvloop at all I think. So might as well remove it too and state we don't test uvloop.

.github/workflows/uvloop_test.yml
JelleZijlstra: Oh, good point. I guess uvloop only kicks in if we use parallelism, so even if I made the test suite use uvloop it wouldn't help much.

ichard26 (commented):
For the record when I said we could replace primer with diff-shades I meant we could replace the primer command / step in the pre-existing primer workflows for `diff-shades analyze` (and then also `diff-shades show-failed --check`), but I won't object to removing the workflows altogether if we deem the platform and Python version coverage not necessary.

.github/workflows/uvloop_test.yml
JelleZijlstra: I'll add a basic test that just runs black on src/, to validate that uvloop mostly works.

cooperlees (approved):

sciyoshi:
👏

PR #2817, opened 2022-01-28
Fix arithmetic stability issue

hauntsaninja:
It turns out "simple_stmt" isn't that simple: it can contain multiple
statements separated by semicolons. Invisible parenthesis logic for
arithmetic expressions only looked at the first child of simple_stmt.
This causes instability in the presence of semicolons, since the next
run through the statement following the semicolon will be the first
child of another simple_stmt.

I believe this along with #2572 fix the known stability issues.


hauntsaninja:
I'm not sure there's a good way to test this other than manually applying:
```
index 769e693..1037030 100644
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1132,6 +1132,7 @@ def format_str(src_contents: str, *, mode: Mode) -> str:
     # Forced second pass to work around optional trailing commas (becoming
     # forced trailing commas on pass 2) interacting differently with optional
     # parentheses.  Admittedly ugly.
+    return dst_contents
     if src_contents != dst_contents:
         return _format_str_once(dst_contents, mode=mode)
     return dst_contents

```
and running `pytest -k test_power_op_newline`

JelleZijlstra (approved):

JelleZijlstra:
Thank you!

PR #2819, opened 2022-01-28
Elaborate on Python support policy

### LINKED ISSUE #2251: [Python 2 has been deprecated] - Formalise Python support policy

felix-hilden:
### Description
Closes #2251: I've included a short Python support policy to our FAQ. TL;DR: we promise to support all non-EOL versions, but don't promise to keep old 3+ versions laying around if we decide to remove them at some point.

### Checklist - did you ...

### Some discussion:
- Is the content fine? It's basically what we do now, or at least the least strict and most convenient version of it. We could promise a deprecation period as well.
- Should we relocate the text? The Python 2 support text was nice as a FAQ entry, but this could be better off somewhere else.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -84,6 +84,7 @@ and the first release covered by our new stability policy.
 - Change HTML theme to Furo primarily for its responsive design and mobile support
   (#2793)
 - Deprecate the `black-primer` tool (#2809)
+- Document Python support policy (#2819)
 
 ## 21.12b0
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -71,9 +71,13 @@ readability because operators are misaligned. Disable W503 and enable the
 disabled-by-default counterpart W504. E203 should be disabled while changes are still
 [discussed](https://github.com/PyCQA/pycodestyle/issues/373).
 
-## Does Black support Python 2?
+## Which Python versions does Black support?
 
-Support for formatting Python 2 code was removed in version 22.0.
+_Black_ supports at least all Python versions that have not reached their end of life.
+This is the case for both running _Black_ and formatting code. Support for formatting
+Python 2 code was removed in version 22.0. While we've made no plans to stop supporting
+older Python 3 minor versions immediately, their support might also be removed some time
+in the future without a deprecation period.
 
 ## Why does my linter or typechecker complain after I format my code?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -73,11 +73,14 @@ disabled-by-default counterpart W504. E203 should be disabled while changes are 
 
 ## Which Python versions does Black support?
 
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -84,6 +84,7 @@ and the first release covered by our new stability policy.
 - Change HTML theme to Furo primarily for its responsive design and mobile support
   (#2793)
 - Deprecate the `black-primer` tool (#2809)
+- Document Python support policy (#2819)
```
ichard26, line 87: Arguably this could count as a highlight

cooperlees (approved):
Nice. Agree with it all. Thanks!

CHANGES.md
felix-hilden: I hesitated, because it doesn't really affect anything currently 😄 but I can change it if you'd like

cooperlees: Can't hurt to have ti here ... might make more people come read it advertising it here ...

felix-hilden: Not sure if I misunderstood you Cooper, but I think Richard was asking if it should be put in the "highlights" section at the very top instead

docs/faq.md
```diff
@@ -71,9 +71,13 @@ readability because operators are misaligned. Disable W503 and enable the
 disabled-by-default counterpart W504. E203 should be disabled while changes are still
 [discussed](https://github.com/PyCQA/pycodestyle/issues/373).
 
-## Does Black support Python 2?
+## Which Python versions does Black support?
 
-Support for formatting Python 2 code was removed in version 22.0.
+_Black_ supports at least all Python versions that have not reached their end of life.
```
JelleZijlstra, line 76: This FAQ entry doesn't actually answer the question: we currently support 3.6 through 3.10.

"Supporting" a Python version in Black can mean two things: running Black under that version, or formatting all code that's valid for that version. Those two meanings are the same right now, but they weren't in the past (we supported formatting Python 2 but not running under Python 2; we supported running under 3.10 before we could fully format 3.10) and they might not be again in the future. Maybe we can make this more explicit too.

JelleZijlstra (approved):

PR #2820, opened 2022-01-28
Treat blank lines in stubs the same inside top-level `if` statements

### LINKED ISSUE #2784: Blank lines at top level code in stubs should be the same inside an if
**Describe the style change**

Blank lines should be treated the same in top-level code under an if-statement as in top-level code proper

```python
class A:
    def f(self): ...

class B:
    def f(self): ...

if sys.version_info >= (3, 8):
    class C:
        def f(self): ...
    class D:
        def f(self): ...
```

**Desired style**

```python
class A:
    def f(self): ...

class B:
    def f(self): ...

if sys.version_info >= (3, 8):
    class C:
        def f(self): ...

    class D:
        def f(self): ...
```

**Additional context**

[Playground link](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AD8AIRdAD2IimZxl1N_WlXnON2nzNApqtITJJee2SMStFF7KJQ8b5tLWLTPF97YwrFIbsO2CJkraBrsaGIFFxZyS8cW-bxi-KzgtfNZt0bUoUnR71oEBfdQhSZgYLHjXKPdI-W9GBft7c-sf-zudh3LWYfAsj7EDjUuebHXI_AiVpD5OuhIJSCPsAC6Ox7BianITQABoAH9AQAAzkjnkrHEZ_sCAAAAAARZWg==)

Similar to / discussed in https://github.com/psf/black/issues/2783

hauntsaninja:
Resolves #2784

While that's the primary motivation, the behaviour changes aren't limited to inside top-level `if` statements. Part of the difficulty is that adding blank lines is mostly context free: we mostly only know if we have a non-zero depth, not why we have a non-zero depth.
That said, I've added new test cases for behaviour changes. I added these tests in their own commit with the previous behaviour, so if you drill down to the individual commit view, it should be easier to see what these changes do.

Additionally tested against typeshed with `black --diff ~/dev/typeshed`.


tests/data/stub.pyi
```diff
@@ -32,6 +32,42 @@ def g():
...
+    def f(): ...
+
+class Nested:
+    class dirty: ...
+    class little: ...
+    class secret:
+        def who_has_to_know(self): ...
+    def verse(self): ...
+
+class CondMethod:
```
JelleZijlstra, line 59: What about a conditional nested class?

JelleZijlstra:
I like it! diff-shades also shows the typeshed changes.

hauntsaninja:
Added tests for a conditional nested class!

PR #2822, opened 2022-01-29
Add a test case to torture.py

### LINKED ISSUE #2563: different code on the second pass of the formatter.
produce error

error: cannot format /Users/SOROKAN6/projects/idse-pqsar-api/tests/test_ellipsize.py: INTERNAL ERROR: Black produced different code on the second pass of the formatter.

run black for the file and it will show the error.
but there is impossible to attach py-file

black, version 21.8b0
Python 3.7.6
MacOS 11.6

No

**Additional context**

[blk_9gtrwz8w.log](https://github.com/psf/black/files/7402245/blk_9gtrwz8w.log)

hauntsaninja:
Resolves #2563

### DIFF
--- a/tests/data/torture.py
+++ b/tests/data/torture.py
@@ -22,6 +22,12 @@ def test(self, othr):
             (name, description, self.default, self.selected, self.auto_generated, self.parameters, self.meta_data, self.schedule) ==
             (name, description, othr.default, othr.selected, othr.auto_generated, othr.parameters, othr.meta_data, othr.schedule))
 
+
+assert (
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):
Thanks!

(I realize I was wrong on Discord earlier, I was confusing this case with the other one. Sorry for that!)

hauntsaninja:
Yeah, I don't understand that one out yet. The "pylint" comment is a red herring, any comment of the same length produces the same stability issue.

I'm not sure what black's promises re: pylint are, but if black does make promises then there's also an argument to be made for a second bug, since black moves the pylint comment to a place where it no longer disables the no-member warning.

PR #2823, opened 2022-01-29
Update classifiers to reflect stable

ichard26:
### Description

Even people browsing PyPI using its search filter shall know Black is now stable 😎 

### Checklist - did you ...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -114,7 +114,7 @@ setup(
     },
     test_suite="tests.test_black",
     classifiers=[
-        "Development Status :: 4 - Beta",
+        "Development Status :: 5 - Production/Stable",
         "Environment :: Console",
         "Intended Audience :: Developers",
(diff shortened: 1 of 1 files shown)


felix-hilden (approved):
🎉

JelleZijlstra (approved):

setup.py
```diff
@@ -114,7 +114,7 @@ def find_python_files(base: Path) -> List[Path]:
     },
     test_suite="tests.test_black",
```
JelleZijlstra, line 115: Unrelatedly, isn't this wrong now? We have more test files than this.

cooperlees: This is deprecated anyways and does not help pytest. So it can be removed I’m pretty sure.

cooperlees (approved):
The time has come. Long time since @ambv showed me black working on his laptop @ work in late 2017 or 2018?

PR #2824, opened 2022-01-29
Remove test suite from setup.py

JelleZijlstra:


cooperlees (approved):
Be gone deprecated option!

Issue #2831, opened 2022-01-30
Doing a soft comparison with --required-version instead of a hard one.


freud14:
**Is your feature request related to a problem? Please describe.**

Since Black just made it as stable, it would be nice to be able to set the `--required-version` argument to something else other than a strict equality. For instance, something like 22 would encompass 22.1.0, 22.1.1, 22.2.0, etc.

**Describe the solution you'd like**

The solution would essentially be to modify this line to a more soft comparison:
https://github.com/psf/black/blob/d038a24ca200da9dacc1dcb05090c9e5b45b7869/src/black/__init__.py#L477

felix-hilden:
Fair request! We break the version down into its parts and only compare until the specification ends. So e.g. `22.1` would accept `22.1.1` as well, and `22` would accept everything this year.

JelleZijlstra:
Yes, this is a good idea.

I think we only need to accept the year, since that lets you use the stable style. I don't think `22.1` makes much sense, as we provide no guarantee that releases that happen to fall within the same month are more compatible.

felix-hilden:
Oh right, there's not really a point in going for more detail when the intention of the option was to require a certain style.

freud14:
Juste made PR #2832. Let me know what you think.

ichard26:
Oh, by the way, you can use special keywords in the PR description so GitHub will automatically close an issue upon merge. You can read more here: https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword. This makes sure any issue that should be closed are in fact closed (and I won't have to go proactively hunt for such issues!), keeping our very chaotic issue tracker ever so slightly less chaotic and scary :)

PR #2832, opened 2022-01-30
Soft comparison of --required-version

freud14:
### Description

Add change suggested in #2831.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 ### Configuration
 
-- Add support for specifying stable version with `--required-version`.
+- Add support for specifying stable version with `--required-version` (#2832).
 
 ## 22.1.0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### Configuration
+
+- Add support for specifying stable version with `--required-version`.
+
 ## 22.1.0
 
 At long last, _Black_ is no longer a beta product! This is the first non-beta release
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -291,8 +291,8 @@ def validate_regex(
     type=str,
     help=(
         "Require a specific version of Black to be running (useful for unifying results"
-        " across many environments e.g. with a pyproject.toml file). It can be only an"
-        " integer representing a major version or an exact version."
+        " across many environments e.g. with a pyproject.toml file). It can be"
+        " either a major version number or an exact version."
     ),
 )
 @click.option(
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -474,7 +474,9 @@ def main(
                 out(f"Using configuration in '{config}'.", fg="blue")
 
     error_msg = "Oh no! 💥 💔 💥"
-    if required_version and required_version != __version__:
+    if required_version and (
+        required_version != __version__ or required_version != __version__.split(".")[0]
+    ):
         err(
             f"{error_msg} The required version `{required_version}` does not match"
             f" the running version `{__version__}`!"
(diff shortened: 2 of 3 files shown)


JelleZijlstra (commented):
Thanks! This behavior should also be documented in the `@click.option` help string for `--required-version`, which gets mirrored in https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#command-line-options.

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### Configuration
+
+- Add support for specifying stable version with `--required-version`.
```
```suggestion
- Add support for specifying stable version with `--required-version` (#2832).
```

src/black/__init__.py
```diff
@@ -474,7 +474,9 @@ def main(
                 out(f"Using configuration in '{config}'.", fg="blue")
 
     error_msg = "Oh no! 💥 💔 💥"
-    if required_version and required_version != __version__:
+    if required_version and (
+        required_version != __version__ or required_version != __version__.split(".")[0]
```
jack1142, line 478: This is going to always fail when a required version is specified:
```
if True and (
    "22" != "22.01"  # True
    or "22" != "22"  # False
):
```
```
if True and (
    "22.01" != "22"        # False
    or "22.01" != "22.01"  # True
):
```

freud14: Lol right.

freud14:
Alright, now it should be ok. There is only test_skip_magic_trailing_comma that doesn't pass on my machine but I suspect it is a matter of configuration.

felix-hilden (changes requested):
Thanks for submitting! I'd also want to consider the option of not allowing for detailed control at all. Meaning we'd only pass an integer for the major version. But regardless of what we decide there, I'd like to see a test where the major is correct but the minor is different, which currently I'd expect to fail.

jack1142:
I think that both exact and major-only matching should be supported as I would say it's quite common to pin formatters and linters to exact versions to ensure the repeatable experience.

freud14:
@felix-hilden I'm not sure what you meant but I just added a new test. Let me know if it's what you meant.

src/black/__init__.py
```diff
@@ -291,7 +291,8 @@ def validate_regex(
     type=str,
     help=(
         "Require a specific version of Black to be running (useful for unifying results"
-        " across many environments e.g. with a pyproject.toml file)."
+        " across many environments e.g. with a pyproject.toml file). It can be only an"
+        " integer representing a major version or an exact version."
```
felix-hilden, line 295:
```suggestion
        " across many environments e.g. with a pyproject.toml file). It can be"
        " either a major version number or an exact version."
```
This would be a bit clearer to me.

freud14:
For the debate about keeping the detailed version, I personally don't care very much. If you tell me you don't want to keep the detailed version, I will make the changes.

JelleZijlstra:
Let's keep the detailed version too.

felix-hilden (commented):
I think allowing for the full version is okay 👍

tests/test_black.py
```diff
@@ -1198,6 +1198,20 @@ def test_required_version_matches_version(self) -> None:
...
         )
 
+    def test_required_version_matches_partial_version(self) -> None:
+        self.invokeBlack(
+            ["--required-version", black.__version__.split(".")[0], "-c", "0"],
+            exit_code=0,
+            ignore_config=True,
+        )
+
+    def test_required_version_does_not_matches_on_minor_version(self) -> None:
```
felix-hilden, line 1208:
```suggestion
    def test_required_version_does_not_match_on_minor_version(self) -> None:
```

felix-hilden (approved):
Looks good now! Final nit below. Thanks a bunch for the quick updates!

tests/test_black.py
freud14: Right.

JelleZijlstra (approved):

PR #2839, opened 2022-01-30
Document each configuration option in more detail

### LINKED ISSUE #2835: Improve documentation for configuration options
Currently, our config options are documented only in a collapsed-by-default text block in https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#command-line-options. This is not very discoverable and makes it hard to give more detailed documentation, such as examples.

Instead, we should have a docs page with a separate section for each option. We can start with the existing descriptions, and extend them as needed for options with more complicated behavior.

JelleZijlstra:
The documentation is partly copied from `--help`, partly from other parts of the docs, partly new.

Fixes #2835


JelleZijlstra:
I couldn't figure out how to get the anchor links to work properly yesterday, I'll have to give it another shot later.

PR #2853, opened 2022-02-01
correct Vim integration code

benknoble:
- use `Black` directly: the commands an autocommand runs are Ex commands, so no
  execute or colon is necessary.
- use an `augroup` (best practice) to prevent duplicate autocommands from
  hindering performance.

https://vi.stackexchange.com/questions/26699/how-to-format-files-on-save-using-black-with-neovim-and-coc/36663?noredirect=1#comment65844_36663

### Checklist - did you ...

Does this need a CHANGELOG entry?

### DIFF
--- a/docs/integrations/editors.md
+++ b/docs/integrations/editors.md
@@ -189,10 +189,13 @@ If you need to do anything special to make your virtualenv work and install _Bla
 example you want to run a version from main), create a virtualenv manually and point
 `g:black_virtualenv` to it. The plugin will use it.
 
-To run _Black_ on save, add the following line to `.vimrc` or `init.vim`:
+To run _Black_ on save, add the following lines to `.vimrc` or `init.vim`:
 
 ```
-autocmd BufWritePre *.py execute ':Black'
+augroup black_on_save
+  autocmd!
+  autocmd BufWritePre *.py Black
+augroup end
 ```
 
 To run _Black_ on a key press (e.g. F9 below), add this:


felix-hilden:
Thanks for submitting! Unfortunately, IIRC we don't really have Vim expertise in the Black maintainer team, the most active sector of it anyway. I'm willing to take your word for it (and thanks for the discussion link), but a proper review might take a while 😄 Do we have any Vim users around?

Or would you care to explain just a bit more: I had a quick look at the autocmd documentation, and seems that `autocmd!` deletes past commands defined in the group. Does that mean previous lines in the vim config, or perhaps aborting previous unfinished runs on the same file?

Otherwise this seems pretty simple.

benknoble:
Some more explanation:

1. As mentioned, `execute ':Black'` is a bit redundant in the case of `autocmd`. The `{cmd}` argument in `:help :autocmd` is an Ex command, so `Black` is legal and correct.
2. The `autocmd!` is actually scoped to the enclosing `augroup`. Further down in the help is the following block:

```
`:autocmd` adds to the list of autocommands regardless of whether they are
already present.  When your .vimrc file is sourced twice, the autocommands
will appear twice.  To avoid this, define your autocommands in a group, so
that you can easily clear them: >

	augroup vimrc
	  " Remove all vimrc autocommands
	  autocmd!
	  au BufNewFile,BufRead *.html so <sfile>:h/html.vim
	augroup END
```

This has lead to a best practice of grouping autocommands so that re-running the file defining the autocommands doesn't cause duplicates to pile up. For example, if I run `:autocmd BufEnter * echom 'hi'` 3 times, I will see 3 `hi` messages when I enter a buffer. But if I enclose in the `augroup` with `autocmd!` spec, running that whole block only leads to one autocommand (and, in this case, only one invocation of `Black`, which is better from a performance perspective).

Further reading:
- https://vi.stackexchange.com/q/9455/10604
- esp. https://vi.stackexchange.com/a/9465/10604

benknoble:
So, I suppose a direct answer to your question is: the `autocmd!` (in this case) deletes all autocommands in the `black_on_save` group, effectively clearing it.

cooperlees:
I wonder if we can write an action with a shell script to load and install this plugin only when the VIM specific files are touched so us non vim extension using people can have more confidence in changes working. Should we open an issue for an action like this?

(The old Unix BOFH in me is ashamed to admit, but I write a big chucks of code exclusively in vscode these days - with the vim plugin of course - the GitHub web vscode can even install the vim plugin!)

JelleZijlstra:
With @benknoble's excellent explanation I'm confident enough to merge this.

PR #2854, opened 2022-02-01
Isolate command line tests for notebooks from user-level config

freud14:
### Description

Following https://github.com/psf/black/pull/2851 but for tests on Jupyter Notebooks.

### Checklist - did you ...

- [X] Add a CHANGELOG entry if necessary? -> n/a
- [X] Add / update tests if necessary? -> n/a

### DIFF
--- a/tests/test_ipynb.py
+++ b/tests/test_ipynb.py
@@ -24,6 +24,8 @@ pytest.importorskip("tokenize_rt", reason="tokenize-rt is an optional dependency
 
 JUPYTER_MODE = Mode(is_ipynb=True)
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra (commented):
This looks good, but I think @ichard26 wanted to put up a more general solution in the test framework. This solution is fragile because we could easily forget the argument in some future test.

freud14:
> This looks good, but I think @ichard26 wanted to put up a more general solution in the test framework. This solution is fragile because we could easily forget the argument in some future test.

Yes indeed. Just close this if you decide that way.

JelleZijlstra:
Let's just merge this for now and change it in the future if we go with the more general solution.

PR #2874, opened 2022-02-08
Copy over comments when hugging power ops 

ichard26:
### Description

Otherwise they'd be deleted which was a regression in 22.1.0 (oops! my bad!). Also type comments are now tracked in the AST safety check on all compatible platforms to error out if this happens again.

Overall the line rewriting code has been rewritten to do "the right thing (tm)", I hope this fixes other potential bugs in the code (fwiw I got to drop the bugfix in blib2to3.pytree.Leaf.clone since now bracket metadata is properly copied over).

cc @henryiii


JelleZijlstra:
This needs to go into `--preview`: "The same code, formatted with the same options, will produce the same output for all releases in a given calendar year."

ichard26:
ohhhh for goodness sake why does this bugfix have to change formatting, cool I'll have to lock most of subfixes under preview which is frustrating.

JelleZijlstra (approved):
Thanks for figuring this out! I think we should merge this despite the stability policy, but add a note to the stability policy docs about it. I can do that in a separate PR.

JelleZijlstra:
A principled way to defend this under the stability policy would be to distinguish between two cases: running Black on a codebase that's already Black-formatted (under an earlier version), and running Black on a non-formatted codebase. This PR only affects the second case, but stability is most important for the first case.

JelleZijlstra:
Done in #3155

JelleZijlstra (approved):

PR #2879, opened 2022-02-11
Allow `for`'s target expression to be starred

### LINKED ISSUE #2878: Black cannot parse multiple unpacking operations in one line
**Describe the bug**

Black cannot parse an expression with multiple unpacking operations on one line that executes fine with Python.

**To Reproduce**

For example, take this code:

```python
a = [1, 2]
b = ["3", "4"]

for x in *a, *b:
    print(x)

c = *a, *b
print(c)

```

```sh
$ black file.py --target-version py39
```

The resulting error is:

> error: cannot format test.py: Cannot parse: 4:9: for x in *a, *b:

**Expected behavior**

Essentially, nothing. The file is already formatted fine. But black shouldn't raise an error ;)

**Environment**

- Black's version: 22.1.0
- OS and Python version: Python 3.9.1, Red Hat 4.8.5-44

**Additional context**

Running the file produces the following output:

```
$ python file.py
1
2
3
4
(1, 2, '3', '4')
```

Microsoft's Pylance also doesn't understand this (and the second to last line where `c` is assigned). See https://github.com/microsoft/pylance-release/issues/2371

isidentical:
Fixes #2878

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,7 +41,7 @@
 
 ### Parser
 
-- Black can now parse starred expressions in the target of the `for`/`async for`
+- Black can now parse starred expressions in the target of `for` and `async for`
   statements, e.g `for item in *items_1, *items_2: pass` (#2879).
 
 ### Performance
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,7 +41,8 @@
 
 ### Parser
 
-<!-- Changes to the parser or to version autodetection -->
+- Black can now parse starred expressions in the target of the `for`/`async for`
+  statements, e.g `for item in *items_1, *items_2: pass` (#2879).
 
 ### Performance
 
--- a/tests/test_format.py
+++ b/tests/test_format.py
@@ -62,6 +62,7 @@ SIMPLE_CASES: List[str] = [
 ]
 
 PY310_CASES: List[str] = [
+    "starred_for_target",
     "pattern_matching_simple",
     "pattern_matching_complex",
     "pattern_matching_extras",
--- a/src/blib2to3/Grammar.txt
+++ b/src/blib2to3/Grammar.txt
@@ -109,7 +109,7 @@ compound_stmt: if_stmt | while_stmt | for_stmt | try_stmt | with_stmt | funcdef 
 async_stmt: ASYNC (funcdef | with_stmt | for_stmt)
 if_stmt: 'if' namedexpr_test ':' suite ('elif' namedexpr_test ':' suite)* ['else' ':' suite]
 while_stmt: 'while' namedexpr_test ':' suite ['else' ':' suite]
-for_stmt: 'for' exprlist 'in' testlist ':' suite ['else' ':' suite]
+for_stmt: 'for' exprlist 'in' testlist_star_expr ':' suite ['else' ':' suite]
 try_stmt: ('try' ':' suite
            ((except_clause ':' suite)+
 	    ['else' ':' suite]
--- a/tests/data/starred_for_target.py
+++ b/tests/data/starred_for_target.py
@@ -1,0 +1,27 @@
+for x in *a, *b:
+    print(x)
+
+for x in a, b, *c:
+    print(x)
+
+for x in *a, b, c:
+    print(x)
+
+for x in *a, b, *c:
+    print(x)
+
+async for x in *a, *b:
+    print(x)
+
+async for x in *a, b, *c:
+    print(x)
+
+async for x in a, b, *c:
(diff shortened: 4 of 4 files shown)


tests/test_format.py
```diff
@@ -62,6 +62,7 @@
 ]
 
 PY310_CASES: List[str] = [
+    "starred_for_target",
```
JelleZijlstra, line 65: I don't think this needs to be 310-specific

CHANGES.md
```diff
@@ -41,7 +41,8 @@
 
 ### Parser
 
-<!-- Changes to the parser or to version autodetection -->
+- Black can now parse starred expressions in the target of the `for`/`async for`
```
JelleZijlstra, line 44:
```suggestion
- Black can now parse starred expressions in the target of `for` and `async for`
```

JelleZijlstra (changes requested):
Thanks, some minor feedback.

Also, I noticed that `[x for x in *a, *b]` isn't allowed. Just curious, do you know why that is?

isidentical:
@JelleZijlstra this seems to be something that was introduced during the PEG parser merge, but there is no official reference in anywhere beside the grammar. I have no knowledge on the comprehensions, but perhaps @pablogsal can shed more light in that.

tests/test_format.py
isidentical: Since the PEP 617 was opt-out in 3.9, the given examples are rejected by the official AST when the user starts the python interpreter with `-Xoldparser`. It is why I've added tests to 3.10 group.

JelleZijlstra (approved):

pablogsal:
> @JelleZijlstra this seems to be something that was introduced during the PEG parser merge, but there is no official reference in anywhere beside the grammar. I have no knowledge on the comprehensions, but perhaps @pablogsal can shed more light in that.

Hummm, maybe I am misunderstanding this, but this syntax has never been allowed:

```
Python 3.8.12 (default, Nov  1 2021, 12:21:59)
>>> [y for y in *x, *y]
  File "<stdin>", line 1
    [y for y in *x, *y]
                ^
SyntaxError: invalid syntax
```

```
Python 3.10.2 (main, Jan 17 2022, 12:00:35)
>>> [y for y in *x, *y]
  File "<stdin>", line 1
    [y for y in *x, *y]
                ^
SyntaxError: invalid syntax
```

pablogsal:
This is In the same spirit as `[x,y for x,y in range(10)]` being invalid. We have an error for this in 3.10:

```
>>> [x,y for x,y in x]
  File "<stdin>", line 1
    [x,y for x,y in x]
     ^^^
SyntaxError: did you forget parentheses around the comprehension target?
```

JelleZijlstra:
My thinking process was just "hey, I didn't know this was allowed, I wonder where else it could be allowed". But it does make sense that we're a bit stricter in comprehensions than in standalone for loops. Thanks for the extra information!

isidentical:
@pablogsal no, I have meant the regular `for` syntax:
```
$ python -c "for a in *[], *[]: pass"
$ python -Xoldparser -c "for a in *[], *[]: pass"
  File "<string>", line 1
    for a in *[], *[]: pass
             ^
SyntaxError: invalid syntax
```

The git bisect tells me that this behavior initially got introduced in the PEP 617 merge.

pablogsal:
> The git bisect tells me that this behavior initially got introduced in the PEP 617 merge.

Ah, sorry, I misunderstood your comment. Yes, that was introduced on PEP 617 merge but I am not actually sure that was on purpose as I cannot find any previous discussion. 

Maybe we need to revert that :(

isidentical:
> Maybe we need to revert that :(

It might be a bit too late, no? (it was part of both 3.9/3.10 stable). But yea, let's create a bpo ticket to discuss further.

pablogsal:
> It might be a bit too late, no? (it was part of both 3.9/3.10 stable). But yea, let's create a bpo ticket to discuss further.

It may be, but if we treat it like a bug, you are technically not allowed to rely on the bug :)

pablogsal:
Opened https://bugs.python.org/issue46725

JelleZijlstra:
Let's hold off on merging this until the BPO is resolved. If CPython's decision is to remove the syntax, I think it makes sense for Black to disallow it too.

ichard26 (approved):
Has been documented in https://github.com/python/cpython/pull/31481. Thanks for fixing this Batuhan!

AIvinAt169922K (commented):
Good luck

PR #2884, opened 2022-02-17
switch license to GPL-V3

libererer:


isidentical:
I don't think you can simply re-license the whole repo without any valid reason / asking each non-trivial contributor for their permission.

cooperlees:
Interesting the contributor joined GitHub 10 hours ago too …

Issue #2150, opened 2021-04-26
Latest `black` changes ASTs


godlygeek:
**Describe the bug**

The README says:

> Also, as a temporary safety measure, Black will check that the reformatted code still produces a valid AST that is equivalent to the original. This slows it down. If you're feeling confident, use `--fast`.

But after #1740, docstrings beginning with a double quote have an extra space prepended to them. This produces a different AST than the file had before, but `--safe` doesn't complain about this.

Is Black no longer promising to produce identical ASTs?

**To Reproduce**

```console
$ cat test.py
def foo():
    '''"something" is what this does.'''


print(foo.__doc__)
$ python test.py
"something" is what this does.
$ black --safe test.py
reformatted test.py
All done! ✨ 🍰 ✨
1 file reformatted.
$ cat test.py
def foo():
    """ "something" is what this does."""


print(foo.__doc__)
$ python test.py
 "something" is what this does.
$
```

**Expected behavior**:

I believe that `black` is never supposed to change the AST, and therefore rewriting the docstring to contain this extra space is not a valid transformation for `black` to perform.

**Environment:**

- Version: `black, version 21.4b0`
- OS and Python version: Linux, Python 3.8.8

**Does this bug also happen on master?**

Yes

**Additional context**

```console
$ python -m ast test.py >ast1.out
$ black test.py
reformatted test.py
All done! ✨ 🍰 ✨
1 file reformatted.
$ python -m ast test.py >ast2.out
$ diff -U0 ast1.out ast2.out
--- ast1.out    2021-04-26 19:05:58.174110800 -0400
+++ ast2.out    2021-04-26 19:08:57.324110800 -0400
@@ -13 +13 @@
-               value=Constant(value='"something" is what this does.'))],
+               value=Constant(value=' "something" is what this does.'))],
```

JelleZijlstra:
Black does change the AST in a few places. This was so even before the most recent release; the previous release already made some changes to docstrings. We should update that sentence in the documentation.

godlygeek:
That's disappointing. The fact that I could trust Black to never change the behavior of a program was one of its greatest selling points, and made me feel comfortable applying Black indiscriminately without needing to check its work.

Now that I know it may change my program's behavior, I'll be much more cautious applying it to new projects.

JelleZijlstra:
[...]

PR #2885, opened 2022-02-17
Add improved docstring processing

TomFryers:
### Description

One of the things discussed in issue #144 was quote placement. Reading through the discussion, there seemed (to me, at least) to be general support for normalising this.

This does two things. For very short docstrings,

```python
def foo():
    """
    Why is this on its own line?
    """
```

becomes
```python
def foo():
    """Why is this on its own line?"""
```
and for docstrings of more than one line, a single newline at the start and (pre-indent) end are enforced, so
```python
def foo():
    """Words

    More words
    """
```
and
```python
def foo():
    """
    
    Words

    More words

    """
```
both become
```python
def foo():
    """
    Words

    More words
    """
```
which seemed to be the more popular style (compared with having no newline line at the start) in the #144 thread.

At the moment, any docstring with only one line of content will be converted to the one-line form. This may be problematic if it is very long, but I'm not sure what the line-length limit should be if this were changed: the normal line length limit, or 72 characters as specified by PEP 8 (which I'm not sure actually applies to single-line docstrings), or something else?

I changed some tests files where the docstring formatting didn’t seem relevant, so as not to have two copies of too many test files.

I didn’t reformat Black’s own source code with this change. There are parts of Black’s source that are already affected by `--preview`, so maybe this should change at the same time.
### Checklist - did you ...


src/black/strings.py
```diff
@@ -60,7 +60,7 @@ def lines_with_leading_tabs_expanded(s: str) -> List[str]:
     return lines
 
 
-def fix_docstring(docstring: str, prefix: str) -> str:
+def fix_docstring(docstring: str, prefix: str, preview: bool) -> str:
```
JelleZijlstra, line 63: Can you make this a keyword-only argument?

JelleZijlstra (dismissed):
Thanks for writing this up and putting it under the preview style.

I'm not yet convinced we should do this, since it could be disruptive. diff-shades currently gives us results only for pyanalyze (which happens to be my code), because it runs black with `preview = True`. As you see, I currently use a different docstring style :)

I'd be curious to see the result on more open-source code. Could you push a change to enable the new style without preview mode, so that we can see the diff-shades result? (Obviously we should turn it back to preview-only before merging.)

CHANGES.md
```diff
@@ -10,6 +10,8 @@
 
 <!-- Changes that affect Black's style -->
 
+- Format docstrings to have consistent quote placement (#2885)
```
JelleZijlstra, line 13: We should make it clear that this only affects the preview style. Probably we should have a separate section for the preview style.

JelleZijlstra: I opened #2890 to add a new CHANGELOG section for the preview style. (And also #2889 to format Black itself in preview mode, which also affects this PR.)

TomFryers: I merged the previewstyle branch into this one, which puts this bullet point into the preview section. Hopefully that wasn’t an awful idea (I never know with Git branches!).

TomFryers:
[Here’s](https://github.com/psf/black/runs/5278179500?check_suite_focus=true#step:11:1) diff-shades’ report when the new style is set as the default (rather than hiding it behind `--preview`). It’s +15 274 / ‒18 914 out of 2 136 881 lines.

TomFryers:
The most recent commit addresses an example of the single-line docstring line-length issue. It’s a bit complicated.

Re-reading PEP 8, it seems the 72-character limit does not apply to docstring summary lines as they’re not ‘flowing long blocks of text with fewer structural restrictions’. I think this means the most sensible limit to apply would be the normal length limit (i.e. usually 88 characters). However, PEP 257 says that, for one-line docstrings, ‘the closing quotes are on the same line as the opening quotes.’ Combined, these directives result in a peculiar docstring-line-length limbo, in which a docstring such as
```python
       """Test that the code option finds the pyproject.toml in the current directory."""
```
is prohibited. It must either violate line-length, violate single-line-docstring-quotes-on-same-line, or be shortened, or, oddly, *lengthened*, since adding body text would allow the quotes to move.

I can think of three ways Black can handle this:
* The current behaviour (of master). This means PEP 257’s requirements won’t be enforced, even for very short docstrings.
* The current behaviour (of this PR). This means occasionally producing illegally-long lines. This has the disadvantage (as the need for 8adc896 showed) that Flake8 does not like it. I think Black sometimes breaks line-length limits anyway, most commonly with string literals pre-`--preview`, but I think doing this to docstrings may be worse.
* Only move the docstring onto a single line if doing so keeps it under the line length. This may be a good compromise.

I don’t think the articles in the `--code` test docstrings were particularly valuable, so I just removed them, but maybe changing the behaviour is a better idea.

One advantage of the style of having quotes on their own line at the start of multi-line docstrings is that changing code to this style doesn’t cause another copy of this issue. (The other reasons I had for choosing this style were that it seemed more popular in the #144 thread (e.g. [this](https://github.com/psf/black/issues/144#issuecomment-628119309) poll), and that it’s consistent with the way Black places brackets over multiple lines.)

felix-hilden:
Thanks for the PR! For the record, this is the style I like best too. But since we're not sure, we should weigh in the fact that this is yet another case for changing AST (#2150) and the various https://github.com/psf/black/labels/F%3A%20strings%20%2F%20docstrings issues.

JelleZijlstra (commented):
A few thoughts:

- This is going to make even more people upset for similar reasons as #2150. And I can't blame them; it's easy to imagine a change like this breaking some docstring processing system, and Black's AST safety guarantee is an important feature. Perhaps we should bite the bullet and add an option to disable docstring processing entirely. Such an option would disable both this PR and the existing docstring changes. New options aren't great, but an option that allows users to opt out of one part of Black's style is more defensible.
- The one specific choice in this PR that I'm having trouble with is putting the opening quotes on their own line. It's clear that in Black itself previously we mostly made the opposite choice. So it may be less disruptive to put the first line right after the opening quotes.

TomFryers:
[diff-shades with the same-line opening quotes style](https://github.com/psf/black/runs/5388767567?check_suite_focus=true#step:11:1) is +13 296 / ‒29 637 out of 2 147 073. Testing this lead me to discover an issue with this quote-placement style:

```python
def foo():
    """Heading:
        content that is supposed to be indented
    """
```
gets converted to
```python
def foo():
    """Heading:
    content that is supposed to be indented
    """
```
by Black’s current docstring indentation logic.

JelleZijlstra (commented):
The code looks solid but I'd still rather not put a blank line at the beginning of each docstring. I'd be curious to hear other maintainers' opinions.

TomFryers:
The tests seem to have stopped working. I don’t really have time to fix them right now, but probably will in a month or so.

felix-hilden:
Yo Tom, thanks for keeping this up to date!

I think before this is merged, we'll have to decide our approach with docstrings, which has been recently touched again in #3087. The discussion is about whether this should be in scope for Black at all, among other things. Again, I think the style is solid, although Jelle disagrees, and there are some thumbs down on the PR. So there's a very realistic chance that we won't accept this in Black. But I'd also appreciate feedback from others!

felix-hilden:
Alright, we've decided to go ahead with docstring processing, and I'm very keen on bringing this in!

Whether intro lines end up in the first or second line is secondary to me (and pydocstyle is similarly ambivalent). The fact that we'd be collapsing short docs into one line is huge 👍 So once again, I prefer to start multiline docs on the second line (as this PR does currently), since then all the text is on the same level. @JelleZijlstra are you adamant about it being the other way around? Other maintainers?

TomFryers:
A summary of the points I can see in favour of each docstring start style:
### Summary line has opening quotes
* Uses less vertical space
* Easier to edit to/from single-line docstring manually
* Smaller docstring-line-length-limbo window
### Summary line after opening quotes
* Uses less horizontal space
  - Can convert to this style from the other one without risking exceeding line length. The reverse isn't the case.
* More consistent with the way Black places (normal/square/curly) brackets
* Tends to work better with indentation
  - It's impossible to tell the difference between
```python
"""Heading for indented section
    Indented section text.
"""
```
and
```python
"""Oops! The body is misindented.
    Misindented body.
"""
```
With the opening quotes on their own line, relative indentation can just be maintained.

TomFryers:
It still says ‘1 change requested’ with a menacing red document symbol. I’m not sure if this means that I’m supposed to request a review again, or if there’s something that I’ve missed. I’ve gone ahead and requested one, so I apologise for the spam if it’s the latter.

onerandomusername:
> It still says ‘1 change requested’ with a menacing red document symbol. I’m not sure if this means that I’m supposed to request a review again, or if there’s something that I’ve missed. I’ve gone ahead and requested one, so I apologise for the spam if it’s the latter.

it means that .@JelleZijlstra requested changes and they have to either dismiss their review or make a new review. Github leaves it up until they take an action, nothing you can do.

JelleZijlstra (commented):
The code is solid (and thanks for persisting with this PR!) but as before I am not happy with the new lines inserted into multiline docstrings. The change is disruptive enough in Black itself that the PR becomes hard to read.

onerandomusername:
> The one specific choice in this PR that I'm having trouble with is putting the opening quotes on their own line. It's clear that in Black itself previously we mostly made the opposite choice. So it may be less disruptive to put the first line right after the opening quotes.

@JelleZijlstra as a person who's used both styles I really can't decide between them. Putting the quotes on their own line sometimes makes sense, but I feel as if the docstring can be easier to read when text follows the quotes on the same line.

Ideally I'd like if we didn't solve this question and just made it togglable, but I understand that there'd rather not be a configuration option :^)

TomFryers:
Hmm… as often seems to be the case with Git, I managed to get myself confused. While trying to update this PR, I seem to have ended up overwriting your a583a32 – I’m sorry about that. Hopefully it’s either OK as is, or is at least reversible. I probably ought to be more careful around things with ‘force’ in the name!

cobaltt7:

PR #2889, opened 2022-02-21
Format ourselves in preview mode

JelleZijlstra:
This makes us a little more aware of what the preview style will do.


ichard26 (approved):
Thanks!

PR #2890, opened 2022-02-21
separate CHANGELOG section for preview style

JelleZijlstra:


ichard26 (approved):
Have we documented how there's two classes of formatting changes? We should clarify we'll try to limit non-preview style changes to bugfixes which don't change anything for already formatted codebases.

JelleZijlstra:
We should document that, yes. And I'd probably be a bit stricter than your statement; a bug has to be pretty bad for us to change the stable style at all. I'm hoping we won't have any for the vast majority of releases.

PR #2891, opened 2022-02-21
Overhaul diff-shades integration to fail on stable changes and preview changes for all projects 

ichard26:
### Description
    
Now PRs will run two diff-shades jobs, "preview-changes" which formats all projects with preview=True, and "assert-no-changes" which formats all projects with preview=False. The latter also fails if any changes were made.
    
Also the workflow_dispatch feature was dropped since it was complicating everything for little gain.

Example run: https://github.com/ichard26/black/pull/15

### Checklist - did you ...

### DIFF
--- a/.github/workflows/diff_shades.yml
+++ b/.github/workflows/diff_shades.yml
@@ -3,54 +3,61 @@ name: diff-shades
 on:
   push:
     branches: [main]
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
It would also be useful to have a non-preview run so we can make sure the stable style doesn't get changed.

ichard26:
Superseded by #2951.

PR #2898, opened 2022-02-26
Pycln QA tool as a pre-commit hook.

hadialqattan:
### Description

I saw `pycln` (which is a formatter for finding and removing unused import statements) a good addition to the current pre-commit hooks which I believe would ensure a high code quality standard besides `black` itself. 

BTW, I'm the author of `pycln`.

Anyway, please feel free to ask any questions and/or reject/close this PR.

### Checklist - did you ...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
 exclude: ^(src/blib2to3/|profiling/|tests/data/)
 repos:
   - repo: https://github.com/hadialqattan/pycln
-    rev: v1.2.3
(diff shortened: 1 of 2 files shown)


hadialqattan:
Please add a 'skip news' label because this change does not require a changelog.

JelleZijlstra:
Thanks, but I don't think this is necessary. flake8 already tells us about unused imports. Your tool may make it a bit easier to remove them, but I don't think that justifies the cost of an additional dev dependency that may break or behave unexpectedly.

PR #2903, opened 2022-03-02
Use tomllib on Python 3.11

hauntsaninja:


JelleZijlstra:
The CPython PR hasn't even been merged yet :D

hauntsaninja:
That's the only reason it's a draft PR :-)

ichard26 (approved):
Is CPython 3.11-dev available on GHA? if so we should add that (in a separate PR) ... although to be fair it'll fail until tomllib hits a prerelease ..

PR #2916, opened 2022-03-13
Format hex code in unicode escape sequences in string literals

### LINKED ISSUE #2067: Format the hex codes in Unicode/hex escape sequences (`\U`, `\u`, `\x`) in string literals
Currently, one can write either `"\U0001f977"` or `"\U0001F977"` which are equivalent but they look differently. Similarly, one can write `"\u200b"` or `"\u200B"` which also are equivalent but they look differently (do mind that `\U` and `\u` are NOT equivalent though; unless we also want to talk about shortening `"\U0000200b"` to `"\u200b"` which I guess would make sense but is probably a separate issue).

Right now, I'm forced to think whether I should use uppercase or lowercase letters as Black doesn't enforce it.

I think it would make sense to have these be consistent in some way, possibly in the same way as the numeric literals, although personally, I think it would make more sense to have it all uppercase in case of `\U0001F977` and all lowercase in case of `\u200b`. I definitely think that the `\u200b` should not be `\u200B` but I don't have a strong opinion on the `\U0001f977` vs `\U0001F977`.

Alternatives were not considered.

**Additional context**
None.

### LINKED ISSUE #2828: Format \N unicode escapes
**Describe the style change**

Black could have an opinion of the correct case for `\N` unicode escapes. For examples, formatting `\N{ox}` to `\N{OX}`.

```python
# This is not nice. All oxen should look the same.
print("\N{ox}\N{OX}")
```
**Desired style**

```python
# Now the oxen will be happier.
print("\N{OX}\N{OX}")
```

Upper-case is standard for these, I believe. (For example, in the [Python docs](https://docs.python.org/3/howto/unicode.html).)

This is related to [issue number 2067](https://github.com/psf/black/issues/2067).

Shivansh-007:
Closes #2067 
Closes #2828 

### Checklist - did you ...

- [X] Add new / update outdated documentation? -> n/a


src/black/linegen.py
```diff
@@ -255,6 +260,10 @@ def visit_factor(self, node: Node) -> Iterator[Line]:
         yield from self.visit_default(node)
 
     def visit_STRING(self, leaf: Leaf) -> Iterator[Line]:
+        if Preview.hex_codes_in_unicode_sequences in self.mode:
+            # Preview style only
```
felix-hilden, line 264:
```suggestion
```
No biggie, but the line above is pretty clear

src/black/strings.py
```diff
@@ -236,3 +246,30 @@ def normalize_string_quotes(s: str) -> str:
         return s  # Prefer double quotes
 
     return f"{prefix}{new_quote}{new_body}{new_quote}"
+
+
+def normalize_unicode_escape_sequences(leaf: Leaf) -> None:
+    """Replace hex codes in Unicode escape sequences with lowercase representation."""
```
felix-hilden, line 252: This will have to be thought out still, as [this comment](https://github.com/psf/black/issues/2067#issuecomment-1024993814) points out. My two cents: I prefer upper case, and since Black formats hex numbers to upper already I think it would be consistent. The Python repr argument is solid too, but we should think about changing hex literals as well then.

felix-hilden (commented):
Thanks for the PR once again! A comment and some nits below 👍 Let's discuss.

src/black/strings.py
```diff
@@ -236,3 +246,31 @@ def normalize_string_quotes(s: str) -> str:
...
+    text = leaf.value
+    prefix = get_string_prefix(text)
+
+    def replace(m: Match[AnyStr]) -> str:
```
JelleZijlstra, line 256:
```suggestion
    def replace(m: Match[str]) -> str:
```
Seems like you only handle strs here, not bytes. You can also remove a bunch of the str() calls below.

src/black/strings.py
```diff
@@ -18,6 +20,14 @@
     r"^([" + STRING_PREFIX_CHARS + r"]*)(.*)$", re.DOTALL
 )
 FIRST_NON_WHITESPACE_RE: Final = re.compile(r"\s*\t+\s*(\S)")
+UNICODE_RE = re.compile(
```
JelleZijlstra, line 23: Should probably also make this Final.

And using `re.MULTILINE` might be nice so you can put comments inside the regex.

src/black/strings.py
```diff
@@ -18,6 +20,14 @@
...
+    r"(\\+)("
+    r"(u([a-zA-Z0-9]{4}))"  # Formatting 16-bit unicodes i.e. \uxxxx
```
JelleZijlstra, line 25: Can remove the word "formatting", and "unicodes" isn't a thing, just say "characters" like https://docs.python.org/3.10/reference/lexical_analysis.html#string-and-bytes-literals

src/black/strings.py
JelleZijlstra: I'd rather not change hex numbers, we already changed our mind there a few times.

test.py
```diff
@@ -0,0 +1,10 @@
+x = "\x1f"
```
JelleZijlstra, line 1: stray file?

Shivansh-007: Yep, forgot to remove it before commiting.

src/black/strings.py
```diff
@@ -18,6 +20,15 @@
...
 )
 FIRST_NON_WHITESPACE_RE: Final = re.compile(r"\s*\t+\s*(\S)")
+UNICODE_RE: Final = re.compile(
+    r"(\\+)("
+    r"(u([a-zA-Z0-9]{4}))"  # Character with 16-bit hex value xxxx
+    r"|(U([a-zA-Z0-9]{0,8}))"  # Character with 32-bit hex value xxxxxxxx
+    r"|(x([a-zA-Z0-9]{2}))"  # Character with hex value hh
+    r"|(N\{([a-zA-Z0-9]{2})\})"  # Character named name in the Unicode database
+    r")",
+    re.MULTILINE,
```
JelleZijlstra, line 30: Sorry, I actually meant re.VERBOSE (https://docs.python.org/3/library/re.html#re.VERBOSE), which allows you to put comments and whitespace in the regex itself. re.MULTILINE isn't useful.

Shivansh-007: You actually got me there once originally. Done 👍🏻

CHANGES.md
```diff
@@ -12,7 +12,7 @@
 
 ### Preview style
 
-<!-- Changes that affect Black's preview style -->
+- Format hex code in unicode escape sequences in string literals (#2916)
```
JelleZijlstra, line 15:
```suggestion
<!-- Changes that affect Black's preview style -->

- Format hex code in unicode escape sequences in string literals (#2916)
```
Let's keep the comment

JelleZijlstra (approved):

src/black/strings.py
felix-hilden: So if we're not changing numbers (which I agree with), do y'all share the concern for consistency?

src/black/strings.py
```diff
@@ -236,3 +247,31 @@ def normalize_string_quotes(s: str) -> str:
         return s  # Prefer double quotes
 
     return f"{prefix}{new_quote}{new_body}{new_quote}"
+
+
+def normalize_unicode_escape_sequences(leaf: Leaf) -> None:
+    """Replace hex codes in Unicode escape sequences with lowercase representation."""
+    text = leaf.value
+    prefix = get_string_prefix(text)
```
ichard26, line 255: why not return early if this contains a `r` ?

```python
if "r" in get_string_prefix(text).lower():
    return
```

src/black/strings.py
```diff
@@ -236,3 +247,31 @@ def normalize_string_quotes(s: str) -> str:
...
+
+    def replace(m: Match[str]) -> str:
+        groups = m.groups()
+        back_slashes = groups[0]
+
+        if len(back_slashes) % 2 == 0 or prefix == "r":
```
ichard26, line 261: I'd prefer if you did not instead of comparing to zero :)

ichard26, line 261: This raw string check is too strict, you need to check for a `r` in a lowercased prefix, otherwise both of these cases crash with an AST safety error:

```python
no_touch = R"\xA1"
```

```python
no_touch = fr"\xA1"
```

src/black/strings.py
```diff
@@ -18,6 +20,15 @@
     r"^([" + STRING_PREFIX_CHARS + r"]*)(.*)$", re.DOTALL
 )
 FIRST_NON_WHITESPACE_RE: Final = re.compile(r"\s*\t+\s*(\S)")
+UNICODE_RE: Final = re.compile(
```
ichard26, line 23: Please rename this variable more specific, maybe `UNICODE_ESCAPE_RE`?

ichard26 (commented):
I won't comment on the actual formatting style, but I got quite a few other suggestions. Not sure if this is too minor, but I'd recommend checking this is covered in the Black code style documentation!

Thanks again!

src/black/strings.py
JelleZijlstra: For what it's worth I prefer `== 0` over relying on implicit boolean behavior.

ichard26: That's fair!

ichard26 (dismissed):
I forgot to mark my review as "request changes" which is relevant since this PR can still crash.

src/black/strings.py
ichard26: I'll note that this would be unnecessary if the other r-prefix related suggestion was implemented.

ichard26:
Hi @Shivansh-007, are you still able to and interested in working on this PR? If not, just lemme know and I'd be happy to pick it up!

felix-hilden:
Also the exact direction we'll be going towards is yet to be decided ([here](https://github.com/psf/black/pull/2916#discussion_r826165374)).

Shivansh-007:
Yeah sure @ichard26, sorry have been busy with school, let me know if something about my code is unclear.

src/black/strings.py
felix-hilden: My comments read a bit ambiguously. So to be clear, I'm proposing that we switch the formatting to be upper case to be consistent with hex numbers. Y'all in?

ichard26:
So it's been two months without any updates and that's because I'm not that interested on working on this PR to be honest. It's stale and I have a bunch of other things I'd like/need to work first. In interest of being a good maintainer by delegating tasks, I've remarked this PR as "up for grabs" (a term I stole from Python Discord's projects). Anyone who wants to pick up this PR and fix it up and finish it is totally welcome to.

I haven't looked at this PR enough to even know what needs to be done to get it review-ready, but I can think of these off the top of my head:

- Address merge conflicts
- Address review comments
  - Specifically decide whether we want to reformat escapes in lowercase or uppercase

felix-hilden:
Up-for-grabs seems like a neat idea, nice 👍

I think no other maintainers have yet expressed their opinion about lower vs. upper case. @ichard26 one way or the other?

src/black/strings.py
JelleZijlstra: Done

src/black/strings.py
JelleZijlstra: Indeed, removed this branch

src/black/strings.py
JelleZijlstra: Done

JelleZijlstra:
I brought this PR up to date, applied @ichard26's review suggestions, and fixed a few more things I noticed. I think this PR is now good to go unless we change our mind to go with uppercase (#2067).

JelleZijlstra:
I determined the legal characters in `\N` escapes by doing something like `[unicodedata.name(chr(i)) for i in range(65536)]` (but ignoring invalid characters) and taking the set of all characters in the output. The length of the names ranged from 3 to 83. However, `\N` also accepts aliases and I'm not sure how to get a list of all of those; the Python docs point to https://www.unicode.org/Public/14.0.0/ucd/NameAliases.txt but that doesn't include the "ox" alias for 🐂. I manually verified that there are no one-character aliases.

Jackenmen:
> However, `\N` also accepts aliases and I'm not sure how to get a list of all of those; the Python docs point to [unicode.org/Public/14.0.0/ucd/NameAliases.txt](https://www.unicode.org/Public/14.0.0/ucd/NameAliases.txt) but that doesn't include the "ox" alias for :ox: 

"ox" is the base name for :ox: so it's returned by `unicodedata.name()`.

JelleZijlstra:
Ah thanks, I should have gone past 65536 to include astral characters. That increases the length range from 2 to 88 but doesn't add more characters to the set of characters that appear in names.

JelleZijlstra:
Also the longest names are
```
In [13]: [n for n in names if len(n) > 80]
Out[13]: 
['ARABIC LIGATURE UIGHUR KIRGHIZ YEH WITH HAMZA ABOVE WITH ALEF MAKSURA ISOLATED FORM',
 'ARABIC LIGATURE UIGHUR KIRGHIZ YEH WITH HAMZA ABOVE WITH ALEF MAKSURA INITIAL FORM',
 'BOX DRAWINGS LIGHT DIAGONAL UPPER CENTRE TO MIDDLE LEFT AND MIDDLE RIGHT TO LOWER CENTRE',
 'BOX DRAWINGS LIGHT DIAGONAL UPPER CENTRE TO MIDDLE RIGHT AND MIDDLE LEFT TO LOWER CENTRE',
 'BOX DRAWINGS LIGHT DIAGONAL UPPER CENTRE TO MIDDLE RIGHT TO LOWER CENTRE TO MIDDLE LEFT',
 'BOX DRAWINGS LIGHT DIAGONAL UPPER CENTRE TO MIDDLE LEFT TO LOWER CENTRE TO MIDDLE RIGHT',
 'BOX DRAWINGS LIGHT DIAGONAL MIDDLE LEFT TO UPPER CENTRE TO MIDDLE RIGHT TO LOWER CENTRE',
 'BOX DRAWINGS LIGHT DIAGONAL MIDDLE RIGHT TO UPPER CENTRE TO MIDDLE LEFT TO LOWER CENTRE']
```

Issue #2193, opened 2021-05-04
Primer: Fix RuntimeError with a closed Event Loop with subprocesses


cooperlees:
From time to time it seems `pandas` causes an EventLoop `RuntimeError` error.

Let's debug and see if there is something we're not cleaning up / waiting for with subprocesses to catch this and exit cleanly / handle the issue.

- Example fail run: https://github.com/psf/black/runs/2503011226

Seen in both 3.6 and 3.9 the most.

```
-- primer results 📊 --
[2021-05-04 18:14:53,570] INFO: Analyzing results (lib.py:356)

17 / 19 succeeded (89.47%) ✅
2 / 19 FAILED (10.53%) 💩
 - 1 project disabled by config
 - 0 projects skipped due to Python version
 - 0 skipped due to long checkout

Failed projects:

## pandas:
 - Returned 0
 - stdout:
Expected formatting changes but didn't get any!

Exception ignored in: <bound method BaseSubprocessTransport.__del__ of <_UnixSubprocessTransport closed pid=2174 returncode=-9 stdout=<_UnixReadPipeTransport closing fd=6 open>>>
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/asyncio/base_subprocess.py", line 132, in __del__
    self.close()
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/asyncio/base_subprocess.py", line 106, in close
    proto.pipe.close()
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/asyncio/unix_events.py", line 423, in close
    self._close(None)
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/asyncio/unix_events.py", line 452, in _close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/asyncio/base_events.py", line 595, in call_soon
    self._check_closed()
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/asyncio/base_events.py", line 381, in _check_closed
Error:     raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
Error: Process completed with exit code 2.
```

If anyone has seen this PRs welcome!

cooperlees:
O, second look, it seems we expect formatting change but don't get any ... Wonder why not all primer fails then.

[...]

PR #2924, opened 2022-03-15
Farewell black-primer, it was nice knowing you

### LINKED ISSUE #2821: Primer deprecation plan
I removed black-primer from CI in #2809 because it was causing annoying and useless CI failures, but kept the code around to avoid unnecessary changes around the stable release. I'd like to keep it around for at least one release that marks it as formally deprecated.

How long should we keep it? Is there any reason to keep it around for longer?

Also, removing the primer tests lost us some platform coverage, because black-primer used to run on lots of OS and Python version combinations and diff-shades doesn't. I'm not too convinced that was helpful, because we already run our unit tests on lots of configurations, but if we wanted we could add back a bigger test matrix.

ichard26:
### Description

Enjoy your retirement at https://github.com/cooperlees/black-primer

Resolves #2821 and closes https://github.com/psf/black/issues/2193.

### Checklist - did you ...


JelleZijlstra (approved):

cooperlees (approved):
Good bye old warrior. You helped bring back black development to life!

Issue #1948, opened 2021-01-29
Black fails to parse parenthesized `(manager() as x)` in with statements


altendky:
**Describe the bug** A clear and concise description of what the bug is.

Black fails to parse `with` statements that are both parenthesized and use `as` such as `with (open("a", "w") as a):`.  I found this while noting that black would not reformat an overly long line of the form `with a as b, x as y:`.

**To Reproduce** Steps to reproduce the behavior:

https://black.now.sh/?version=master&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADxAGhdAD2IimZxl1N_WmufDs0LpGEw9emYfCBakMKmw1q3jXUj3wtHo6NCOupRzg7ydqs0H1WF8HT1qPymFuq8UolBICiBSQevQsn3TFo9kaHrTJtsFiiPLpaMZ8_MAWjg5A9X2Ar4c0sX0RkzAApiLbCr57chAAGEAfIBAAC4H7lWscRn-wIAAAAABFla

```python
with open("a", "w"):
    pass

with open("a", "w") as a:
    pass

with (open("a", "w")):
    pass

with (open("a", "w") as a):
    pass
```

```
Cannot parse: 10:21: with (open("a", "w") as a):
```

**Expected behavior** A clear and concise description of what you expected to happen.

I expect that black does not fail to parse and also reformats as needed when the line is too long.  Reformatting would include adding parentheses and wrapping such as below (for actually too-long lines).

```python
with (
    open("c", "w") as a,
    open("d", "w") as b,
):
```

**Environment (please complete the following information):**

- Version: 20.8b1 and 71117e730c4f62458b30af820f51890487b458e4
- OS and Python version: Locally with Linux and Python 3.9.1, also whatever black.now.sh is running right now

**Does this bug also happen on master?** To answer this, you have two options:

yes

**Additional context** Add any other context about the problem here.

Perhaps I should enter two bugs.  One for the parsing failure and one for the lack of willingness to add parentheses and reformat.  Or perhaps they are sufficiently intertwined...  let me know if I need to file a second issue.

JelleZijlstra:
This is new syntax in 3.9, right? We'll just have to update our parser.

altendky:
I didn't realize it was new, but I did just confirm that you are correct.  I really thought I had, at some point, done some multi-line `with` with parentheses etc.  And I rarely run exclusively 3.9.  Oh well...

ichard26:
[...]

PR #2926, opened 2022-03-15
Remove unnecessary parentheses from `with` statements

### LINKED ISSUE #2921: Drop redundant parenthesis in `with` statement
**Describe the bug**

Black removes redundant parenthesis in most contexts, but this does not apply to the `with` statement yet.

**To Reproduce**
[Playground link](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ACXAFtdAD2IimZxl1N_Wlws4SR5ySw_sBpDkhU93hOCZPT-hFkjq5Stk1KilvNEQNrc4EaZ8ufi_zEpnhZViAkrMAiyAFGllzQQ2rS3TSp7D0gVbLYWtVvdCKgKqXub2lwAAPSQHVnPZQFHAAF3mAEAAAAQ4riMscRn-wIAAAAABFla)

```python
if (open("bla.txt")):
    pass

with (open("bla.txt")):
    pass
```

Output:
```python
if open("bla.txt"):    # <-- parenthesis removed
    pass

with (open("bla.txt")):    # <-- parenthesis retained
    pass
```

Adding `--target-version=py310` or `--preview` does not change this behavior.

**Environment**

- Black's version: 22.1.0 and main branch
- OS and Python version: Python 3.10

### LINKED ISSUE #2952: Removing multiple unnecessary parenthesizes for for assignment targes is unstable
**Describe the bug**

Under preview, the removal of unnecessary parenthesizes can be unstable if there's multiple -- each pass removes only one pair.

**To Reproduce**

For example, take this code:

```python
for ((x, y)) in enumerate(range(10)):
    pass
```

```sh
$ black file.py --preview
```

The resulting error is:

```
error: cannot format test.py: INTERNAL ERROR: Black produced different code on the second pass of the formatter.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_473_benq.log

Oh no! 💥 💔 💥

Mode(target_versions={<TargetVersion.PY38: 8>, <TargetVersion.PY37: 7>, <TargetVersion.PY36: 6>}, line_length=88, string_normalization=True, is_pyi=False, is_ipynb=False, magic_trailing_comma=True, experimental_string_processing=False, python_cell_magics=set(), preview=True)
--- source
+++ first pass
@@ -1,2 +1,2 @@
-for ((x, y)) in enumerate(range(10)):
+for (x, y) in enumerate(range(10)):
     pass
--- first pass
+++ second pass
@@ -1,2 +1,2 @@
-for (x, y) in enumerate(range(10)):
+for x, y in enumerate(range(10)):
     pass
```

**Expected behavior**

All of the unnecessary parenthesizes being removed at once ^^

**Environment**

- Black's version:  14d84ba2e96c5ca1351b8fe4d0d415cc148f4117
- OS and Python version:  CPython 3.8.5

jpy-git:
### Description
Closes #2921 and closes #2952. Aims to remove redundant parentheses from `with` statements:
Screen-Recording-2022-03-15-at-17 20 39

I currently have one failing test [tests/data/parenthesized_context_managers.py](https://github.com/psf/black/blob/main/tests/data/parenthesized_context_managers.py) since this code now removes unnecessary parentheses around one-line `with` statements, so think I'll just need to update the test to reflect this. 
I tried looking in the contributing docs but couldn't see any info on how the tests work so please could you advise on where I should add this?

This is my first time contributing to black so let me know if there's anything I've missed 😄 

Thanks!


jpy-git:
thanks, I've got the tests working now and will refactor slightly to account for the preview mode

src/black/linegen.py
```diff
@@ -857,24 +878,34 @@ def normalize_invisible_parens(
             elif not (isinstance(child, Leaf) and is_multiline_string(child)):
                 wrap_in_parentheses(node, child, visible=False)
 
-        check_lpar = isinstance(child, Leaf) and child.value in parens_after
+        comma_check = child.type == token.COMMA if preview else False
+
+        check_lpar = (
+            isinstance(child, Leaf) and child.value in parens_after or comma_check
```
JelleZijlstra, line 884:
```suggestion
            child.value in parens_after if isinstance(child, Leaf) else comma_check
```

jpy-git: @JelleZijlstra I don't think this change is right. A comma is a `Leaf` as well so the comma wouldn't get checked here. The comma check doesn't need the `Leaf` check since it uses `.type` so I can do either of these if it's more clear?
```python
isinstance(child, Leaf) and (child.value in parens_after or comma_check)
```
or
```python
comma_check or (isinstance(child, Leaf) and (child.value in parens_after or comma_check)
```

JelleZijlstra: Thanks! Yes I assumed that `x and y or z` is equivalent to `y if x else z`, but that's not always true. Parens would make it more clear.

JelleZijlstra (approved):

ichard26 (commented):
This case is unstable unfortunately:

```python
with (((open("test"))) as a):
    pass
```

```diff
Mode(target_versions={<TargetVersion.PY37: 7>, <TargetVersion.PY38: 8>, <TargetVersion.PY36: 6>}, line_length=88, string_normalization=True, is_pyi=False, is_ipynb=False, magic_trailing_comma=True, experimental_string_processing=False, python_cell_magics=set(), preview=True)
--- source
+++ first pass
@@ -1,2 +1,2 @@
-with (((open("test"))) as a):
+with ((open("test"))) as a:
     pass
--- first pass
+++ second pass
@@ -1,2 +1,2 @@
-with ((open("test"))) as a:
+with open("test") as a:
     pass
```

And yes, we probably have to fix cases like these even if they're a bit out there :)

jpy-git:
Classic, people spend ages requesting brackets around context managers and then they go and add the ability to add as many brackets as you want! 😋 

Will take a look tomorrow/this weekend 👍

src/black/linegen.py
```diff
@@ -866,38 +884,62 @@ def normalize_invisible_parens(
...
         or is_one_tuple(node)
         or (is_yield(node) and parent.type != syms.expr_stmt)
-        or (max_delimiter_priority_in_atom(node) >= COMMA_PRIORITY and for_stmt_check)
+        or (
+            # This condition tries to prevent removing non-optional brackets
+            # around a tuple, however, can be a bit overzealous so we provide
+            # and option to skip this check for `for` and `with` statements.
+            not remove_brackets_around_comma
+            and max_delimiter_priority_in_atom(node) >= COMMA_PRIORITY
+        )
```
jpy-git, line 942: So the reason I had the `for_stmt_check`/`with_stmt_check` initially was to disable this check for the reason in the above comment (basically it's hard to cleanly detect a tuple).

The issue was that nested brackets are nested atoms so when we recursed into the second level of this function the parent would no longer be `for_stmt` or `with_stmt` so my check would fail and require a second run. I've added an argument to disable this check and then we can just use this when we're removing brackets from `for`/`with` in the main loop.

src/black/linegen.py
```diff
@@ -840,11 +843,26 @@ def normalize_invisible_parens(
...
+                and isinstance(child.prev_sibling, Leaf)
+                and child.prev_sibling.type == token.NAME
+                and child.prev_sibling.value == "for"
+            ):
+                if maybe_make_parens_invisible_in_atom(
+                    child,
+                    parent=node,
+                    remove_brackets_around_comma=True,
+                ):
+                    wrap_in_parentheses(node, child, visible=False)
```
jpy-git, line 859: `for` fix is now very simple. Detect the parent is a `for` loop and then remove the brackets using the option to remove around commas as well.

src/black/linegen.py
```diff
@@ -866,38 +884,62 @@ def normalize_invisible_parens(
...
+                remove_with_parens(child, node)
+    elif node.type == syms.asexpr_test and not any(
+        leaf.type == token.COLONEQUAL for leaf in node.leaves()
+    ):
+        if maybe_make_parens_invisible_in_atom(
+            node.children[0],
+            parent=parent,
+            remove_brackets_around_comma=True,
+        ):
+            wrap_in_parentheses(parent, node.children[0], visible=False)
```
jpy-git, line 917: `with` is slightly more complicated to achieve in one pass so I've split this into it's own function.

Basically the different variations of bracketed with statements give pretty different parse trees so we need to recursively solve this.

`with (open("test.txt")) as f: ...`: this is an `asexpr_test`
`with (open("test.txt") as f): ...`: this is an `atom` containing an `asexpr_test`
`with (open("test.txt")) as f, (open("test.txt")) as f: ...`: this is `asexpr_test`, `COMMA`, `asexpr_test`
`with (open("test.txt") as f, open("test.txt")) as f: ...`: this is an `atom` containing a `testlist_gexp` which then contains multiple `asexpr_test`s

jpy-git (commented):
@JelleZijlstra @ichard26 I've fixed both `for` and `with` bracket removals to work in one pass now 😄

jpy-git:
One thing that I've noticed this PR doesn't currently do is add optional brackets if there aren't already some there.

So we don't correct
```python
with ksjdhfdskfhdskfhdskjhfdsakjhfdskjhfdskjhfdskhfdskjhfdkjshfsdakjhf, ksjdhfdskfhdskfhdskjhfdsakjhfdskjhfdskjhfdskhfdskjhfdkjshfsdakjhf:
    pass
```
to
```python
with (
    ksjdhfdskfhdskfhdskjhfdsakjhfdskjhfdskjhfdskhfdskjhfdkjshfsdakjhf,
    ksjdhfdskfhdskfhdskjhfdsakjhfdskjhfdskjhfdskhfdskjhfdkjshfsdakjhf,
):
    pass
```

The reason is due to the slightly odd way that `with` is parsed when there are no brackets vs other similar cases
```python
with foo, bar:
    pass
```
`[Leaf(NAME, 'with'), Leaf(NAME, 'foo'), Leaf(COMMA, ','), Leaf(NAME, 'bar'), Leaf(COLON, ':'), Node(suite, [Leaf(NE...ENT, '')])]`
whereas in similar cases (e.g. `for foo, bar in d.items(): ...`) we would expect the comma separated list to be grouped as children of some node (e.g. `exprlist`). Therefore we can't add these brackets within [this](https://github.com/psf/black/blob/bd1e98034907463f5d86f4d87e89202dc6c34dd4/src/black/linegen.py#L831) loop and have to put invisible parens around multiple children.

It's not too bad to add these brackets, I've got some code tweaks locally that are working, but this does lead me to ask:
- If I did add the ability to add these brackets in and therefore fix long `with` lines, this would break `python<3.9` code since these brackets weren't introduced until `3.9`. Is there a way to have some logic occur only if we know the code being fixed is `>=3.9`?

The current behaviour without this additional change is basically "only fix brackets if they are already there since we know it must be `>=3.9`" so there's no risk of breaking code atm, but wanted to get your opinion on whether we could/should handle the no bracket case as well?

JelleZijlstra:
We should add these parentheses, but probably in a separate PR. There is already a framework for doing this based on versions; that's why we have the `--target-version` config option.

jpy-git:
Fab thanks, just wanted to check/make you aware whilst reviewing this PR. Will follow up with a separate PR to add those in once we've finished this one 😄

ichard26:
This has been discussed in https://github.com/psf/black/issues/664 and https://github.com/psf/black/issues/1948 FYI.

tests/data/remove_with_brackets.py
```diff
@@ -0,0 +1,79 @@
+with (open("bla.txt")):
```
JelleZijlstra, line 1: Could you add test cases with multiple levels of parentheses here?

(And probably also foor `for`, in `remove_for_brackets.py`)

jpy-git: I've added the multi-paren examples for `with` in that last commit 👍 

FYI I've noted as well that only the context manager can have multiple paren levels so that's why these examples below aren't included in those tests 😄 
```python
# This is valid syntax
with (open("bla.txt") as f):
    pass

# This is bad syntax
with ((open("bla.txt") as f)):
    pass

# This is valid syntax
with (open("bla.txt") as f, open("bla.txt") as f):
    pass

# This is bad syntax
with ((open("bla.txt") as f, open("bla.txt") as f)):
    pass
```

jpy-git: will add the `for` ones as well 👍

jpy-git: Got the `for` examples in as well

JelleZijlstra (approved):
Thanks! @ichard26 did you want to do another review?

ichard26 (approved):
Great clean work once again, thank you so much! #NewBlack

JelleZijlstra (approved):

PR #2928, opened 2022-03-16
Remove power hugging formatting from preview 

Shivansh-007:


Shivansh-007:
Could any of the maintainers add https://github.com/psf/black/labels/skip%20news

src/black/mode.py
```diff
@@ -163,7 +162,9 @@ def __contains__(self, feature: Preview) -> bool:
         """
         if feature is Preview.string_processing:
             return self.preview or self.experimental_string_processing
-        return self.preview
+        # TODO: Remove type ignore comment once preview contains more features
+        #  than just ESP
+        return self.preview  # type: ignore
```
ichard26, line 167: Hmmm? what's up with mypy?

JelleZijlstra: It thinks this code is unreachable because there's only one enum member.

Shivansh-007: Yeah ^

ichard26 (approved):

PR #2931, opened 2022-03-16
Update pylint config docs

### LINKED ISSUE #2914: pylint checks seem to no longer exist?
Using black with other tools says for [pylint](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#pylint), that one should switch off `disable = C0330, C0326`, however I find the checks on neither [here](https://pylint.pycqa.org/en/latest/technical_reference/features.html) nor [here](https://pylint.pycqa.org/en/latest/technical_reference/extensions.html).

If I'm not missing anything and they really don't exist anymore, that could be removed from the documentation.

jpy-git:
### Description
Closes #2914. I've confirmed that `C0330` & `C0326` checks no longer exist in pylint, therefore we can remove them from the config docs.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,7 +30,7 @@
 ### Documentation
 
 <!-- Major changes to documentation and policies. Small docs changes
- don't need a changelog entry. -->
+     don't need a changelog entry. -->
      
 - Update pylint config documentation (#2931)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,7 +31,7 @@
 
 <!-- Major changes to documentation and policies. Small docs changes
      don't need a changelog entry. -->
-     
+
 - Update pylint config documentation (#2931)
 
 ### Integrations
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,8 +29,7 @@
 
 ### Documentation
 
-<!-- Major changes to documentation and policies. Small docs changes
-     don't need a changelog entry. -->
+- Update pylint config documentation (#2931)
 
 ### Integrations
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,9 @@
 
 ### Documentation
 
+<!-- Major changes to documentation and policies. Small docs changes
+ don't need a changelog entry. -->
+     
 - Update pylint config documentation (#2931)
 
 ### Integrations
--- a/docs/guides/using_black_with_other_tools.md
+++ b/docs/guides/using_black_with_other_tools.md
@@ -218,6 +218,9 @@ max-line-length = 88
 Pylint should be configured to only complain about lines that surpass `88` characters
 via `max-line-length = 88`.
 
+If using `pylint<2.6.0`, also disable `C0326` and `C0330` as these are incompatible with
+_Black_ formatting and have since been deprecated.
+
 #### Formats
 
 <details>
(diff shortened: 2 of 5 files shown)


CHANGES.md
```diff
@@ -29,8 +29,7 @@
 
 ### Documentation
 
-<!-- Major changes to documentation and policies. Small docs changes
```
Shivansh-007, line 32: This comment shouldn't be removed, in case you saw me removing it from my PR and then did the same, I did that by mistake and will be reverting it.

docs/guides/using_black_with_other_tools.md
```diff
@@ -210,41 +210,20 @@ mixed feelings about _Black_'s formatting style.
...
-`C0330: Wrong hanging indentation before block (add 4 spaces)` warnings. Since _Black_
-isn't configurable on this style, Pylint should be told to ignore these warnings via
-`disable = C0330`.
-
-Also, since _Black_ deals with whitespace around operators and brackets, Pylint's
-warning `C0326: Bad whitespace` should be disabled using `disable = C0326`.
-
-And as usual, Pylint should be configured to only complain about lines that surpass `88`
-characters via `max-line-length = 88`.
+Pylint should be configured to only complain about lines that surpass `88` characters
```
Shivansh-007, line 218: Might be worth adding a small line/note that this configuration is only applicable latest versions, or to be more precise `>= 2.6.0`.

Shivansh-007 (changes requested):

JelleZijlstra:
When were these codes removed? We may still want to mention them in case people use an older version of pylint or pylint changes its mind.

docs/guides/using_black_with_other_tools.md
jpy-git: Makes sense, can see it in the [2.6.0](https://pylint.pycqa.org/en/latest/whatsnew/changelog.html?highlight=max-line-length#what-s-new-in-pylint-2-6-0) release notes.

Given there have been a good number of releases since 2.6.0, do you think this can just be something like
```
N.B. if using `pylint < 2.6.0`, also disable `C0326` and `C0330`.
```
or would you rather keep the longer descriptions?

JelleZijlstra: I think a short note like that is good. We can skip the details.

jpy-git: Added the old error code callout in now 😄

jpy-git:
@Shivansh-007 `C0326` was removed in https://github.com/PyCQA/pylint/pull/3577/files (same release as `CO330` removal)

docs/guides/using_black_with_other_tools.md
```diff
@@ -210,41 +210,23 @@ mixed feelings about _Black_'s formatting style.
...
-isn't configurable on this style, Pylint should be told to ignore these warnings via
-`disable = C0330`.
-
-Also, since _Black_ deals with whitespace around operators and brackets, Pylint's
-warning `C0326: Bad whitespace` should be disabled using `disable = C0326`.
-
-And as usual, Pylint should be configured to only complain about lines that surpass `88`
-characters via `max-line-length = 88`.
+If using `pylint<2.6.0`, also disable `C0326` and `C0330` as these are incompatible with
+_Black_ formatting and have since been deprecated.
```
JelleZijlstra, line 222:
```suggestion
_Black_ formatting and have since been removed.
```
Deprecate means saying you'll remove it later. What matters is that they actually removed these codes.

PR #2939, opened 2022-03-18
Remove unnecessary parentheses from `except` clauses

### LINKED ISSUE #2937: Remove redundant brackets from `except` clause

The except clause can accept a parenthesised tuple of exceptions. However, in the case of only a single exception the brackets should be removed.

Black currently leaves both of these examples unchanged:
```
a = 1
try:
    a.something
except (AttributeError) as err:
    raise err
```
and
```
a = 1
try:
    a.something
except (AttributeError,) as err:
    raise err
```

I would expect both examples to be corrected to:
```
a = 1
try:
    a.something
except AttributeError as err:
    raise err
```

I could see an argument to leave the tuple example (with the trailing comma) as is, maybe the user wants to add more exceptions in the future. Easy either way on that one, but we should at least fix the first (non-tuple) example.

**Additional context**

I actually came across this when someone assumed the `except` was a function in a PR. Black added the space before the brackets, but also removing the brackets would really help highlight that it's a keyword not a function 😄 
```
try:
    a.something
except(AttributeError) as err:
    raise err
```

jpy-git:
### Description

Closes #2937. As described in the issue we want to remove redundant parentheses from `except` clauses.

### Checklist - did you ...

### DIFF
--- a/src/black/mode.py
+++ b/src/black/mode.py
@@ -127,6 +127,7 @@ class Preview(Enum):
     """Individual preview style features."""
 
     string_processing = auto()
+    remove_except_parens = auto()
 
 
 class Deprecated(UserWarning):
@@ -162,9 +163,7 @@ class Mode:
         """
         if feature is Preview.string_processing:
             return self.preview or self.experimental_string_processing
-        # TODO: Remove type ignore comment once preview contains more features
-        #  than just ESP
-        return self.preview  # type: ignore
+        return self.preview
 
     def get_cache_key(self) -> str:
         if self.target_versions:
--- a/src/black/mode.py
+++ b/src/black/mode.py
@@ -127,7 +127,7 @@ class Preview(Enum):
     """Individual preview style features."""
 
     string_processing = auto()
-    remove_except_parens = auto()
+    remove_redundant_parens = auto()
 
 
 class Deprecated(UserWarning):
--- a/tests/data/remove_except_parens.py
+++ b/tests/data/remove_except_parens.py
@@ -18,6 +18,22 @@ try:
 except (AttributeError, ValueError) as err:
     raise err
 
+# Test long variants.
+try:
+    a.something
+except (some.really.really.really.looooooooooooooooooooooooooooooooong.module.over89.chars.Error) as err:
+    raise err
+
+try:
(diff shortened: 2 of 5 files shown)


src/black/mode.py
```diff
@@ -127,6 +127,7 @@ class Preview(Enum):
     """Individual preview style features."""
 
     string_processing = auto()
+    remove_except_parens = auto()
```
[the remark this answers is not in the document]
jpy-git: ^ Going to combine these

tests/data/remove_except_parens.py
```diff
@@ -0,0 +1,40 @@
+# These brackets are redundant, therefore remove.
+try:
+    a.something
+except (AttributeError) as err:
```
JelleZijlstra, line 4: Can you add a test case where the line is too long, like:

```
except (
    some.really.long.module.over89.chars.Error
) as really_long_name:
```

I worry that we'll run into instability issues in cases like this.

tests/data/remove_except_parens.py
```diff
@@ -0,0 +1,40 @@
...
+# we do not remove brackets to preserve AST.
+try:
+    a.something
+except (AttributeError,) as err:
+    raise err
+
+# This is a tuple of exceptions. Do not remove brackets.
+try:
+    a.something
+except (AttributeError, ValueError) as err:
```
JelleZijlstra, line 18: And here add a case where the tuple is too long to fit in one line.

JelleZijlstra (changes requested):

tests/data/remove_except_parens.py
jpy-git: I've added these tests and all works as expected 😄

JelleZijlstra (approved):

src/black/mode.py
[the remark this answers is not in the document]
ichard26: I agree with combing 'em :)

ichard26 (approved):
I feel a bit uneasy merging formatting changes without a fully operational diff-shades integration, but this looks fine to me :slightly_smiling_face: 

Thanks again for all of your work!

PR #2940, opened 2022-03-18
stub style: remove some possible future changes

### LINKED ISSUE #2938: Future style for stub file docstrings that use C/C++ extensions
Hello Black community. 
I write to you today to discuss the style outlined in the documentation for stub files (Python files meant for typing purposes with a `.pyi` extension). In the _Black Code Style_ section of the documentation there is a section describing future style guidelines for the `.pyi` files: [Typing Stub Files](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#typing-stub-files).

**Description**

The point I wish to argue against today is the line: "do not use docstrings". I imagine this is referring to a Python stub file that is providing type hints for another Python file, like so: 
_add.py_
```Python
def add(a, b):
    " here's an existing docstring"
    return a + b
```
_add.pyi_
```python
def add(a: int, b: int) -> int:
    " here's another stub file docstring"
    ...
```
In which case, it would make sense to either include the docstring in one file or the other. 

However, the edge case I wish to bring to your attention is when using C/C++ extensions for Python code. In particular I will discuss the use of [PyBind11](https://pybind11.readthedocs.io/en/stable/basics.html). PyBind11 produces a shared library with a name such as `add.cpython-38-x86_64-linux-gnu.so`, which contains inside it Python objects that you can import as any other Python object. However, modern Intellisense servers like [Pylance](https://github.com/microsoft/pylance-release) do not extract information from binary files by [design](https://github.com/microsoft/pylance-release/issues/70), and therefore you _cannot_ get any documentation from a binary file, be this docstrings or type information. 

Therefore, a stub file can be used to make up for the design limitation, but you need to include docstrings in the stub file in order to get useful information about a function written with PyBind11. 

_add_pybind11.py_
```python
# import the add function from a PyBind11 produced binary
import add from _add_pybind11

# the above will generate no intellisense due to the design limitation of tools like pylance
```

_add_pybind11.pyi_
```python
# define the docstring and type information for the C/C++ bound object
def add(a: int, b: int) -> int:
    " This function was generated by pybind11 and therefore needs a stub file to use with intellisense"
    ...
```

**Conclusion**
I'd like to offer the example of using stub files with binary libraries, such as those produced by PyBind11, as a reason to allow docstrings in stub files in future versions of the formatter. This can be with a flag or configuration, but rest assured having Black delete my docstrings for my PyBind11 objects would cause headaches. 

image

JelleZijlstra:
Fixes #2938.

All of these suggested future changes are out of scope for an autoformatter and should be handled by a linter instead.

### DIFF
--- a/docs/the_black_code_style/current_style.md
+++ b/docs/the_black_code_style/current_style.md
@@ -399,16 +399,11 @@ recommended code style for those files is more terse than PEP 8:
 _Black_ enforces the above rules. There are additional guidelines for formatting `.pyi`
 file that are not enforced yet but might be in a future version of the formatter:
 
-- all function bodies should be empty (contain `...` instead of the body);
-- do not use docstrings;
 - prefer `...` over `pass`;
-- for arguments with a default, use `...` instead of the actual default;
 - avoid using string literals in type annotations, stub files support forward references
   natively (like Python 3.7 code with `from __future__ import annotations`);
 - use variable annotations instead of type comments, even for stubs that target older
-  versions of Python;
-- for arguments that default to `None`, use `Optional[]` explicitly;
-- use `float` instead of `Union[int, float]`.
+  versions of Python.
 
 ## Pragmatism
 


ichard26 (approved):
None of the remaining guidelines sound AST-safe, but this is certainly better as is.

Thanks!

PR #2948, opened 2022-03-23
Github now supports .git-blame-ignore-revs

henrikhorluck:
### Description

Now reflects current state of the world.

### Checklist - did you ...

### DIFF
--- a/docs/guides/introducing_black_to_your_project.md
+++ b/docs/guides/introducing_black_to_your_project.md
@@ -43,8 +43,10 @@ call to `git blame`.
 $ git config blame.ignoreRevsFile .git-blame-ignore-revs
 ```
 
-**The one caveat is that GitHub and GitLab do not yet support ignoring revisions using
-their native UI of blame.** So blame information will be cluttered with a reformatting
-commit on those platforms. (If you'd like this feature, there's an open issue for
-[GitLab](https://gitlab.com/gitlab-org/gitlab/-/issues/31423) and please let GitHub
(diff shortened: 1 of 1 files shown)


felix-hilden (changes requested):
Cool! Thanks for submitting a PR right away 👌 Although there still is a warning about it being beta and possibly changing. So perhaps we should wait a bit? But I'm not totally opposed to changing it right away.

Most likely Prettier will complain about line lengths, but otherwise this looks good.

felix-hilden (approved):
Thanks for the quick fix!

henrikhorluck:
> Cool! Thanks for submitting a PR right away 👌 Although there still is a warning about it being beta and possibly changing. So perhaps we should wait a bit? But I'm not totally opposed to changing it right away.

I mean 🤷, I think it could be useful for someone to see that it's here at least, however limited or in beta it is. I think it only makes sense to either keep the documentation up to date, or not have this note in the documentation at all. I can make a new PR if they remove the support :P

ichard26 (approved):
I'll just merge this one manually. Thanks for keeping the docs up to date! Congrats on your first PR to psf/black 🎉

ichard26:
https://github.com/psf/black/commit/062b54931dc3ea35f673e755893fe28ff1f5a889 done -- although I completely managed to forgot to attach attribution to you, sorry! >.<

cooperlees:
Finally. This is awesome news!

PR #2953, opened 2022-03-24
Remove for brackets: Extra unit test

jpy-git:
### Description

Investigating #2952. Really weird but I'm not seeing this issue and this unit test works fine for me 😅 

### Checklist - did you ...


ichard26:
Sigh, it turns out this is my fault as I had black's source code edited when reviewing this change (I was debugging another PR). This issue still exists, but it's hidden by the second pass we force:

https://github.com/psf/black/blob/14d84ba2e96c5ca1351b8fe4d0d415cc148f4117/src/black/__init__.py#L1151-L1157

If you remove this hack for magic trailing commas then the instability shows up.

jpy-git:
> If you remove this hack for magic trailing commas then the instability shows up.

Ah thanks was wondering why I wasn't seeing the issue!

ichard26:
I'd be fine with relying on this hack for now, but we want to remove it eventually so that probably just kicks the can down the road.

jpy-git:
Makes sense, I've added a note to this unit test so at least we know the hack works for now. Will aim to fix this as part of #2926 since I'm pretty sure it'll be the same issue!

jpy-git:
Going to close this as I've figured out how to fix this for both the `for` and `with` case and will address this as part of #2926 😄

PR #2970, opened 2022-03-28
Correctly handle fmt skip comments without internal spaces

### LINKED ISSUE #2917: `#fmt:skip` causes NTERNAL ERROR: Black produced different code on the second pass of the formatter
**Describe the bug**

A comment `#fmt:skip` with no internal space causes `INTERNAL ERROR: Black produced different code on the second pass of the formatter`

Also the same code, if run with `--fast` creates incorrectly formatted code with multiple copies of the comment on the next line.

**To Reproduce**

For example, take this code:

```python
import re

line = "some dummy data"
log_search = re.search('CLOCK:\s*(\[.*\])--(\[.*\])\s*=>\s*(\d*:\d*)', line)  #fmt:skip
```

```sh
$ black fmt_skip.py .
```

The resulting error is:

error: cannot format fmt_skip.py: INTERNAL ERROR: Black produced different code on the second pass of the formatter.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /var/folders/cs/p77vt_fn271bcw6gw69d_2y40000gn/T/blk_i64a7rxi.log

ALSO

```sh
$ black fmt_skip.py . --fast
```

The resulting error is that the output is:
```python
import re

line = "some dummy data"
log_search = re.search('CLOCK:\s*(\[.*\])--(\[.*\])\s*=>\s*(\d*:\d*)', line)  #fmt:skip
  # fmt:skip  # fmt:skip
```

**Expected behavior**

Black should not change the line format as it is marked to be skipped.

**Environment**

* MacOS 12.2.1
* Python 3.8.5
* black 22.1.0

Also reproducible on the Black Playground: https://black.vercel.app/?version=main

siuryan:
### Description

Hi, first time contributor here! This PR fixes #2917 . The issue was that fmt skip comments were being handled inconsistently. A formatted version of the comment (e.g. `#fmt:skip` -> `# fmt:skip` in `comment.value`) was being compared to the original comment in the code (e.g. `#fmt:skip` in `leaf.prefix`), which would always be False.

Please let me know if there's anything I missed. I left a comment below explaining one of the changes. Also, does this change require a changelog entry or documentation change?

### Checklist - did you ...


src/black/comments.py
```diff
@@ -214,8 +214,11 @@ def generate_ignored_nodes(
     container: Optional[LN] = container_of(leaf)
     if comment.value in FMT_SKIP:
         prev_sibling = leaf.prev_sibling
-        if comment.value in leaf.prefix and prev_sibling is not None:
-            leaf.prefix = leaf.prefix.replace(comment.value, "")
+        # Need to properly format the leaf prefix to compare it to comment.value,
+        # which is also formatted
+        comments = list_comments(leaf.prefix, is_endmarker=False, preview=preview)
+        if comments and comment.value == comments[0].value and prev_sibling is not None:
+            leaf.prefix = ""
```
siuryan, line 221: Is setting `leaf.prefix` to empty string here okay? It appears `leaf.prefix` can only ever contain the comment with `# fmt:skip`, but I could be wrong about this.

siuryan:
Hi, just wanted to follow up on this PR. Would a maintainer be able to take a look at these changes? Thanks!

JelleZijlstra:
Unfortunately I triggered another CI failure because I merged main, I'll review the PR fixing that now.

JelleZijlstra (approved):
Thanks!

It occurs to me that maybe we shouldn't be formatting `# fmt: skip` comments at all, even adding the space. But I am not going to worry about that until and unless users actually complain.

ichard26 (approved):
LGTM, thank you and congrats on your first contribution to psf/black @siuryan! -- always nice to see some new faces :)

Issue #2971, opened 2022-03-29
Address breakage of our test suite *before* aiohttp 4.0.0 goes out of alpha


ichard26:
So thanks to the various dependency related fires we've had lately, I've created [ichard26/black-deps-ci](https://github.com/ichard26/black-deps-ci) which runs Black's test suite using the latest development versions for most of our dependencies (exceptions being subdependencies and most of the test dependencies) \*daily\*

Turns out there's a lot of overrides necessary for CI to be green. Python 3.6 support being dropped is responsible for the vast majority of overrides, but on the other hand, aiohttp breaks our test suite as the `aiohttp.test_utils.unittest_run_loop` decorator was removed :upside_down_face: 

https://github.com/ichard26/black-deps-ci/blob/4e25ac3dea340a956e35dab37b2ac5c4177c105e/generate.py#L9-L24

```python
overrides = {}
overrides["aiohttp"] = [
    ("aiohttp>=3.7.4 ; python_version == '3.6'", "aiohttp 4.x requires Python 3.7+"),
    ("aiohttp<4", "aiohttp 4+ breaks our test suite as @unittest_run_loop was deleted"),
]
overrides["click"] = [("click>=8.0.0 ; python_version == '3.6'", "click 8.1.0+ dropped Python 3.6 support")]
overrides["ipython"] = [
    (
        "ipython>7.8.0 ; python_version <= '3.7'",
        "ipython 7.17+ dropped Python 3.6 and ipython 8+ dropped Python 3.7 support",
    )
]
overrides["pytest"] = [("pytest>6.1.1 ; python_version == '3.6'", "pytest 7.1.0+ dropped Python 3.6 support")]
overrides["platformdirs"] = [("platformdirs>=2 ; python_version == '3.6'", "platformdirs 2.5.1< dropped Python 3.6 support")]
overrides["tokenize-rt"] = [("tokenize-rt>3.2.0 ; python_version == '3.6'", "tokenize-rt 4.2.1< dropped Python 3.6 support")]
overrides["tomli"] = [("tomli>=1.1.0 ; python_version == '3.6'", "tomli 2.0.0+ dropped Python 3.6 support")]
``` 

And FYI, we've been warned about this in CI: https://github.com/psf/black/runs/5727523727?check_suite_focus=true#step:6:325

---

The best solution is probably to allow the import to fail replacing `unittest_run_loop` with just a no-op decorator. Although I wouldn't object to bumping up our aiohttp lower bound.

[...]

PR #2974, opened 2022-03-29
Keep tests working w/ upcoming aiohttp 4.0.0

ichard26:
### Description

aiohttp.test_utils.unittest_run_loop was deprecated since aiohttp 3.8 and aiohttp 4 (which isn't a thing quite yet) removes it. To maintain compatibility with the full range of versions we declare to support, test_blackd.py will now define a no-op replacement if it can't be imported.

Also, mypy is painfully slow to use without a cache, let's reenable it.

Resolves GH-2971.

### Checklist - did you ...

### DIFF
--- a/tests/test_blackd.py
+++ b/tests/test_blackd.py
@@ -1,4 +1,5 @@
 import re
+from typing import Any
 from unittest.mock import patch
 
 from click.testing import CliRunner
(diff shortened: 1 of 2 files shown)


tests/test_blackd.py
```diff
@@ -8,12 +9,18 @@
 
 try:
     import blackd
-    from aiohttp.test_utils import AioHTTPTestCase, unittest_run_loop
+    from aiohttp.test_utils import AioHTTPTestCase
     from aiohttp import web
+except ImportError as e:
+    raise RuntimeError("Please install Black with the 'd' extra") from e
```
ichard26, line 15: The old `has_blackd_deps` logic was doing absolutely nothing useful so instead let's error out with an informative exception.

JelleZijlstra (approved):

PR #2979, opened 2022-03-30
Remove click pin in diff-shades workflow

JelleZijlstra:
Click 8.1.1 was released with a fix for pallets/click#2227.


JelleZijlstra:
Oh, I guess this doesn't work because there's no type ignore on master yet. So I'll first have to make a PR adding the type ignore on master, then another dropping the pin.

PR #2981, opened 2022-03-30
Add # type: ignore for click._unicodefun import

JelleZijlstra:
See #2979


ichard26 (approved):
Can confirm this unbreaks mypyc locally with click==8.1.1 🎉

Issue #2983, opened 2022-03-31
Python 3.11 ModuleNotFoundError: No module named 'tomllib'


PerchunPak:
**Describe the bug**

In Python 3.11 it is raising exception with `ModuleNotFoundError`.

**To Reproduce**

Install Python 3.11

```sh
$ black .
```

The resulting error is:

```py
Traceback (most recent call last):
  File "/home/runner/work/autodonate/autodonate/.venv/bin/black", line 5, in <module>
    from black import patched_main
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/autodonate/autodonate/.venv/lib/python3.11/site-packages/black/__init__.py", line 52, in <module>
    from black.files import (
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/autodonate/autodonate/.venv/lib/python3.11/site-packages/black/files.py", line 25, in <module>
    import tomllib
    ^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'tomllib'
```

**Environment**

- Black's version: 22.3.0
- OS and Python version: Github Actions Ubuntu, Python 3.11

**Additional context**

[Logs from my CI.](https://github.com/fire-squad/autodonate/runs/5764464577?check_suite_focus=true#step:7:18)
Also, why you haven't did any CI on 3.11?

SamuelMarks:
Also this did work on 3.11.0-alpha.5 just not 3.11.0-alpha.6

SamuelMarks:
Also this did work on 3.11.0-alpha.5 with black-22.1.0-py3-none-any.whl just not 3.11.0-alpha.6 with black-22.3.0-py3-none-any.whl.

SamuelMarks:
Also this did work on `3.11.0-alpha.5` with `black-22.1.0-py3-none-any.whl` just not `3.11.0-alpha.6` with `black-22.3.0-py3-none-any.whl`.

pawamoy:
Would it be possible to change to a try/except instead of a condition on the Python version, at least until tomllib is available in Python 3.11 (alpha or not)? I understand 3.11 is still alpha, but I believe many users want to test their code and run their CI tools (i.e. Black) on Python 3.11. Of course I can pin Black to `<22.3` for Python 3.11 only (with markers), but that's a bit cumbersome to do in every project :confused:

JelleZijlstra:
@pawamoy that's fair, opened #2987

PR #2987, opened 2022-04-01
try-except tomllib import

JelleZijlstra:
See #2983

I left the version check in place because mypy doesn't generally like try-excepted imports. I didn't try directly though.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,7 +39,7 @@
 
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
-* Use `tomli` instead of `tomllib` on Python 3.11 builds where `tomllib` is not
+- Use `tomli` instead of `tomllib` on Python 3.11 builds where `tomllib` is not
   available (#2987)
 
 ### Parser
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,8 +39,8 @@
 
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
-* Use `tomli` instead of `tomllib` on Python 3.11 builds where `tomllib`
-  is not available (#2987)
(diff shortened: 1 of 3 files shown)


pawamoy:
Thanks :slightly_smiling_face: Yeah leaving the version check makes sense. We'll need to remove the marker on `tomli` as well in setup.py.

ichard26:
Is there a way to only select the alphas using environment markers? I'd like to avoid having 3.11 stable users also pull in tomli unnecessarily.

JelleZijlstra:
Maybe we should just leave the setup.py markers alone and assume users testing the alphas are sophisticated enough to install tomli themselves. Or we could just remove the markers for a few months until the 3.11 betas are widely available.

pawamoy:
Based on PEP 508, you could use the `python_full_version` marker: https://peps.python.org/pep-0508/#environment-markers

> Maybe we should just leave the setup.py markers alone and assume users testing the alphas are sophisticated enough to install tomli themselves.

That would defeat the purpose of this PR, which is to avoid forcing users to change their dependencies specification (adding tomli or a marker on Black).

> Is there a way to only select the alphas using environment markers? I'd like to avoid having 3.11 stable users also pull in tomli unnecessarily.

That's understandable. Though that would only be the case for versions released until 3.11 includes tomllib, not the next ones.

ichard26:
> Or we could just remove the markers for a few months until the 3.11 betas are widely available.

And yeah that's fair. I guess if any one complains about the extra dependency we could just issue a newer build number to retroactively "fix" older releases. Not perfect since the source distribution can't be updated or overridden AFAIK but oh well.

ichard26 (approved):
Can confirm 3.11.0a7 should contain tomllib. Thanks!

pawamoy:
Thanks a lot for your very prompt response, PR and merge! This is very appreciated :slightly_smiling_face:

PR #2989, opened 2022-04-02
Fix broken link in README.md

jpy-git:
### Description

Spotted a broken link in the README so updating 

### Checklist - did you ...


ichard26 (approved):
The more I look at our docs, I wish we had prettier URLs that used dashes and had no file extensions, but that would break a lot of links so that's probably not feasible (unless we set up an ridiculous amount of redirects).

Thanks anyway!

PR #2990, opened 2022-04-02
Return annotation brackets

### LINKED ISSUE #2041: Redundant parentheses around return annotation
```python
def infer_encoding_trajectory(encoding: EncodingElement,
                              observation: PoolingMessage,
                              prediction: PredictionMessage,
                              weights: hk.Params,
                              rng: Generator,
                              iterations: int) -> (
                                  tuple[EncodingState, EncodingTrajectory]):
```
became
```python
def infer_encoding_trajectory(
    encoding: EncodingElement,
    observation: PoolingMessage,
    prediction: PredictionMessage,
    weights: hk.Params,
    rng: Generator,
    iterations: int,
) -> (tuple[EncodingState, EncodingTrajectory]):
```
The parentheses were only there to allow wrapping and should have been removed.  In general, black doesn't need to preserve parentheses around an entire expression, including a type annotation.

### LINKED ISSUE #2699: Line too long: String in return annotation
**To Reproduce**

Given input:
```python
class ThisIsTrulyUnreasonablyExtremelyLongClassName:
    pass


def frobnicate() -> "ThisIsTrulyUnreasonablyExtremelyLongClassName | list[ThisIsTrulyUnreasonablyExtremelyLongClassName]":
    pass
```

Black does not split the long type hint of the function return type

Even if I explicitly attempt to split the string:
```python
def frobnicate() -> "ThisIsTrulyUnreasonablyExtremelyLongClassName | " \
                    "list[ThisIsTrulyUnreasonablyExtremelyLongClassName]":
    pass
```

then Black collapses it back together :(
```python
def frobnicate() -> "ThisIsTrulyUnreasonablyExtremelyLongClassName | " "list[ThisIsTrulyUnreasonablyExtremelyLongClassName]":
    pass
```

**Workaround**

Only by parenthesizing the type hint I can convince Black to keep it reasonable.

```python
def frobnicate() -> (
    "ThisIsTrulyUnreasonablyExtremelyLongClassName | "
    "list[ThisIsTrulyUnreasonablyExtremelyLongClassName]"
):
    pass
```

**Environment**

- Black's version: 21.12b0
- OS and Python version: macOS, Python 3.10.1

jpy-git:
### Description

Closes #2041 and #2699. Allows us to better control placement of return annotations.

a) removes redundant parens:
```python
def foo() -> (int): ...
```
becomes
```python
def foo() -> int: ...
```

b) moves very long type annotations onto their own line:
```python
def foo() -> intsdfsafafafdfdsasdfsfsdfasdfafdsafdfdsfasdskdsdsfdsafdsafsdfdasfffsfdsfdsafafhdskfhdsfjdslkfdlfsdkjhsdfjkdshfkljds:
    return 2
```
becomes
```python
def foo() -> (
    intsdfsafafafdfdsasdfsfsdfasdfafdsafdfdsfasdskdsdsfdsafdsafsdfdasfffsfdsfdsafafhdskfhdsfjdslkfdlfsdkjhsdfjkdshfkljds
):
    return 2
```

### Checklist - did you ...


CHANGES.md
```diff
@@ -14,6 +14,8 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Optional brackets around return anotations (#2990)
```
JelleZijlstra, line 17:
```suggestion
- Parentheses around return annotations are now managed (#2990)
```
Not sure I like the "managed" wording, but we should say what changed

tests/data/return_annotation_brackets.py
```diff
@@ -0,0 +1,134 @@
+# Control
+def double(a: int) -> int:
+    return 2*a
+
+# Remove the brackets
+def double(a: int) -> (int):
```
JelleZijlstra, line 6: Also add an example with multiple sets of parens. And maybe for additional fun, what if there is more whitespace between the parentheses?

JelleZijlstra (commented):
Thanks! Some small feedback

tests/data/return_annotation_brackets.py
```diff
@@ -53,6 +53,27 @@ def foo(a: int, b: int, c: int,) -> intsdfsafafafdfdsasdfsfsdfasdfafdsafdfdsfasd
 def foo(a: int, b: int, c: int,) -> int:
     return 2
 
+# Deeply nested brackets
+# with *interesting* spacing
```
JelleZijlstra, line 57: love it

jpy-git: I pray I never meet a developer who likes to write their brackets like these examples 😅

JelleZijlstra (approved):

tests/data/return_annotation_brackets.py
JelleZijlstra: You missed an opportunity to write ASCII art using weirdly spaced parentheses

jpy-git: I've got a few sad faces `):` in there

ichard26 (approved):
Excellent work @jpy-git :tada:

PR #2995, opened 2022-04-05
Top PyPI Packages: Use 30-days data, 365 is no longer available

hugovk:
### Description

I needed to remove the 365-day data from https://hugovk.github.io/top-pypi-packages/ because it was using more quota than available, so let's just use the 30-day one instead.

Also use the minified version for a bit smaller file.

Re: https://github.com/hugovk/top-pypi-packages/pull/21, https://github.com/hugovk/top-pypi-packages/pull/20, https://github.com/hugovk/top-pypi-packages/issues/19.

### Checklist - did you ...

I don't think any of these are needed?

I didn't see the days param documented anywhere (including `gallery/README.md`).

### DIFF
--- a/gallery/gallery.py
+++ b/gallery/gallery.py
@@ -10,10 +10,9 @@ from argparse import ArgumentParser, Namespace
 from concurrent.futures import ThreadPoolExecutor
 from functools import lru_cache, partial
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):
Thanks!

Issue #2994, opened 2022-04-05
610faff656c4cfcbb4a3__mypyc


mrclary:
This is not a bug, but a general question.
Sorry, I could not ask this question on Discord as it keeps telling me I'm in an unclaimed account even though I've logged in.

`black` has listed in its distribution meta-data `top_level.txt` the module `610faff656c4cfcbb4a3__mypyc`. I'm wondering how the name of this module is generated as it is not found in `src/`.

JelleZijlstra:
It's relate to our usage of mypyc (https://mypyc.readthedocs.io/en/latest/). I am not sure exactly how the name is generated, but apparently that's how mypyc works.

ichard26:
So yeah we use mypyc to compile Black into a fast Python C Extension. We compile everything as one single massive compilation unit so all the code of all modules are packed in a single autogenerated extension module. It's faster this way compared to compiling each module as their own separate compilation unit. There's a little more information about this in the mypyc docs: https://mypyc.readthedocs.io/en/latest/compilation_units.html

ichard26:
If you look at the Pure Python wheel we distribute along side the optimized platform specific wheels, you'll see how it doesn't contain the autogenerated extension module. It's completely optional and you aren't losing anything by using the pure python wheel. 

Now I think about it, it's probably worth a FAQ entry about our usage of mypyc (especially as I'd imagine the "compiled: yes/no" in `black --version` could be very confusing).

mrclary:
Okay, thanks for the insight.
The reason for my question is that [spyder-ide/spyder](https://github.com/spyder-ide/spyder) uses [py2app](https://github.com/ronaldoussoren/py2app) to build standalone applications and needs to ensure that all of `black`'s packages are included in the application bundle. Given that `610faff656c4cfcbb4a3__mypyc` may or may not exist (or its name could be different) depending on the type of installation, I suppose it is best to inspect the `egg-info/top_level.txt` for all necessary packages.

ichard26:
If you're aiming to package standalone applications, it may be best to use the Pure Python wheel to have the best chance of compatibility, but yes looking at .dist-info's or .egg-info's top_level.txt is a good idea.

Unless you have more questions, I'll close this since there's nothing left to action here (well except for adding an entry in the FAQ 'bout our use of mypyc -- I'll do that right now). Have a great day!

PR #3002, opened 2022-04-09
Explain our use of mypyc in the FAQ

ichard26:
### Description

While responding to GH-2994 I realized we don't have a FAQ entry about this, let's change that so `compiled: yes/no` doesn't surprise as many people :) 

### Checklist - did you ...

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -122,10 +122,12 @@ for 64-bit versions of Windows, Linux (via the manylinux standard), and macOS ac
 supported CPython versions.
 
 Platforms including musl-based and/or ARM linux distributions, and ARM Windows are
-currently **not** supported. These platforms will fallback to the (slower) Pure Python
+currently **not** supported. These platforms will fallback to the slower pure Python
 wheel available on PyPI.
 
-If you are experiencing exceptionally weird issues or even segfaults, you can try the
-Pure Python wheel by passing `--no-binary black` to your pip install invocation.
+If you are experiencing exceptionally weird issues or even segfaults, you can try
+passing `--no-binary black` to your pip install invocation. Note this flag excludes all
+wheels (including the pure Python wheel) so this will involve building the sdist, this
+may be a tad more unreliable.
 
 [mypyc]: https://mypyc.readthedocs.io/en/latest/
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -126,8 +126,9 @@ currently **not** supported. These platforms will fall back to the slower pure P
 wheel available on PyPI.
 
 If you are experiencing exceptionally weird issues or even segfaults, you can try
-passing `--no-binary black` to your pip install invocation. Note this flag excludes all
-wheels (including the pure Python wheel) so this will involve building the sdist, this
-may be a tad more unreliable.
+passing `--no-binary black` to your pip install invocation. This flag excludes all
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -113,3 +113,19 @@ _Black_ is an autoformatter, not a Python linter or interpreter. Detecting all s
...
 CPython doesn't accept.
+
+## What is `compiled: yes/no` all about in the version output?
+
+While _Black_ is indeed a pure Python project, we use [mypyc] to compile _Black_ into a
+C Python extension, usually doubling performance. These compiled wheels are available
+for 64-bit versions of Windows, Linux (via the manylinux standard), and macOS across all
+supported CPython versions.
+
+Platforms including musl-based and/or ARM linux distributions, and ARM Windows are
```
JelleZijlstra, line 124:
```suggestion
Platforms including musl-based and/or ARM Linux distributions, and ARM Windows are
```

docs/faq.md
```diff
@@ -113,3 +113,19 @@ _Black_ is an autoformatter, not a Python linter or interpreter. Detecting all s
...
+currently **not** supported. These platforms will fallback to the (slower) Pure Python
```
JelleZijlstra, line 125:
```suggestion
currently **not** supported. These platforms will fall back to the slower pure Python
```

docs/faq.md
```diff
@@ -113,3 +113,19 @@ _Black_ is an autoformatter, not a Python linter or interpreter. Detecting all s
...
+wheel available on PyPI.
+
+If you are experiencing exceptionally weird issues or even segfaults, you can try the
+Pure Python wheel by passing `--no-binary black` to your pip install invocation.
```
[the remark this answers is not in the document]
JelleZijlstra: hm won't that use the sdist instead?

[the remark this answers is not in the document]
ichard26: ah shoot you're correct, I think it's still worth keeping this but I will note that this uses the sdist (since they're a bit more unreliable to install)

docs/faq.md
```diff
@@ -113,3 +113,21 @@ _Black_ is an autoformatter, not a Python linter or interpreter. Detecting all s
...
+supported CPython versions.
+
+Platforms including musl-based and/or ARM Linux distributions, and ARM Windows are
+currently **not** supported. These platforms will fall back to the slower pure Python
+wheel available on PyPI.
+
+If you are experiencing exceptionally weird issues or even segfaults, you can try
+passing `--no-binary black` to your pip install invocation. Note this flag excludes all
+wheels (including the pure Python wheel) so this will involve building the sdist, this
+may be a tad more unreliable.
```
JelleZijlstra, line 131:
```suggestion
passing `--no-binary black` to your pip install invocation. This flag excludes all
wheels (including the pure Python wheel), so this command will use the sdist.
```
I don't think vague wording like "a tad more unreliable" is helpful. Maybe we can link to some packaging documentation about sdists?

ichard26: Resolved in https://github.com/psf/black/pull/3002/commits/2805958c2ee874f30c5b19474ef3df061c639c05

JelleZijlstra (approved):

ichard26:
Thanks for the review @JelleZijlstra !

PR #3006, opened 2022-04-11
Quote black[jupyter] and black[d] in installation docs

JelleZijlstra:
We just got someone on Discord who was confused because the command as written caused their shell to try to do command expansion.


felix-hilden:
Let the first timer have it? 😄

ichard26 (approved):
Thank you!

PR #3008, opened 2022-04-11
Make ipynb tests compatible with ipython 8.3.0+

MarcoGorelli:
### Checklist - did you ...

- [ ] Add a CHANGELOG entry if necessary? (this isn't user-facing, so I don't think it's necessary)

xref https://github.com/ichard26/black-deps-ci/actions/runs/2105909618 - one test is failing due to changes in IPython

In the latest release of IPython:
```
>>> TransformerManager().transform_cell('a = b??')
"get_ipython().set_next_input('a = b');get_ipython().run_line_magic('pinfo2', 'b')\n"
```

In the development version:
```
>>> TransformerManager().transform_cell('a = b??')
"get_ipython().run_line_magic('pinfo2', 'b')\n"
```

The latest version looks more correct to me, because the `a = ` part will be ignored anyway if you type that line into an IPython shell:
```
In [2]: object??
Init signature: object()
Docstring:     
The base class of the class hierarchy.

When called, it accepts no arguments and returns a new featureless
instance that has no instance attributes and cannot be given any.
Type:           type
Subclasses:     type, weakref, weakcallableproxy, weakproxy, int, bytearray, bytes, list, NoneType, NotImplementedType, ...

In [3]: a = object??
Init signature: object()
Docstring:     
The base class of the class hierarchy.

When called, it accepts no arguments and returns a new featureless
instance that has no instance attributes and cannot be given any.
Type:           type
Subclasses:     type, weakref, weakcallableproxy, weakproxy, int, bytearray, bytes, list, NoneType, NotImplementedType, ...
```

So I think I'd suggest to update the test, and skip it if the IPython version is less than 8.3

### DIFF
--- a/tests/test_ipynb.py
+++ b/tests/test_ipynb.py
@@ -1,3 +1,4 @@
+import contextlib
 from dataclasses import replace
 import pathlib
 import re
@@ -18,6 +19,8 @@ from black import Mode
 from _pytest.monkeypatch import MonkeyPatch
 from tests.util import DATA_DIR
 
+with contextlib.suppress(ModuleNotFoundError):
+    import IPython
 pytestmark = pytest.mark.jupyter
 pytest.importorskip("IPython", reason="IPython is an optional dependency")
 pytest.importorskip("tokenize_rt", reason="tokenize-rt is an optional dependency")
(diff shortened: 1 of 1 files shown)


tests/test_ipynb.py
```diff
@@ -18,6 +19,8 @@
 from _pytest.monkeypatch import MonkeyPatch
 from tests.util import DATA_DIR
 
+with contextlib.suppress(ModuleNotFoundError):
```
JelleZijlstra, line 22: This won't work well because line 146 will fail if the module isn't available.

MarcoGorelli: on line 25 there's
```
pytest.importorskip("IPython", reason="IPython is an optional dependency")
```
so if `IPython` isn't available, line 146 is never reached

ichard26 (approved):
Finally I can stop getting notifications because black-deps-ci's CI is failing daily :D

PR #3016, opened 2022-04-15
PEP 654 syntax

### LINKED ISSUE #3012: Add support for PEP 654 (`except*`)
Python 3.11 adds the new `except*` syntax in PEP 654 (https://peps.python.org/pep-0654/#except). We should support it. I think this should be a straightforward change to the grammar.

We should also normalize formatting, e.g. change `except *E:` to `except* E:`. (There was a long python-dev thread about how to format it and I believe that was the consensus outcome.)

@isidentical are you interested in this as our resident parser expert?

isidentical:
Resolves #3012

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -53,6 +53,9 @@
 
 ### Parser
 
+- [PEP 654](https://peps.python.org/pep-0654/#except) syntax (for example,
+  `except *ExceptionGroup:`) is now supported (#3016)
+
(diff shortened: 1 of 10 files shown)


JelleZijlstra (approved):
Looks good, thanks for the quick fix!

ichard26 (approved):
Thank you for the amazingly quick patch! :tada:

PR #3025, opened 2022-04-21
Fix strtobool function

defntvdm:
### Description

'true' 'on' in python is 'trueon'


JelleZijlstra (commented):
Thanks, that's embarrassing.

Since this fixes a user-facing bug, could you add a news entry?

defntvdm:
@JelleZijlstra Did you mean to add row into CHANGELOG? I did it :)

CHANGES.md
```diff
@@ -62,6 +62,10 @@
 
 <!-- Changes that improve Black's performance. -->
 
+### Vim Plugin
+
+- Fixed strtobool function. It didn't parse true/on/false/off.
```
cooperlees, line 67:
```suggestion
- Fixed strtobool function. It didn't parse true/on/false/off. (#3025)
```
The grep check is stricter ...

cooperlees (changes requested):

cooperlees (approved):
Thankyou for fixing this typo!

PR #3032, opened 2022-04-24
follow python typing export requirements

danpf:
### Description

Black's` __init__.py` file does not explicitly alias imported modules as themselves which is required for type-checking.

I grepped the code for things that were directly called from `black.` and added `X as X` for anything mentioned in the code + anything I thought seemed related.

This isn't really very important, I was just annoyed by some of the pyright warnings I was getting when using `black.Mode`.

Happy to alter the PR in anyway, if you think more or less should be exported from `__init__.py`

https://github.com/python/typing/blob/master/docs/source/stubs.rst#imports

I don't think it needs a mention in the changelog.

### Checklist - did you ... 

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -37,7 +37,6 @@ from click.core import ParameterSource
 from dataclasses import replace
 from mypy_extensions import mypyc_attr
 
-from . import const as const, mode as mode
 from black.const import (
     DEFAULT_LINE_LENGTH as DEFAULT_LINE_LENGTH,
     DEFAULT_INCLUDES,
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -68,7 +68,7 @@ from black.output import (
     dump_to_file as dump_to_file,
     ipynb_diff,
     diff as diff,
-    color_diff,
+    color_diff as color_diff,
     out as out,
     err as err,
 )
@@ -90,7 +90,7 @@ from black.files import (
     wrap_stream_for_windows,
 )
 from black.parsing import (
-    InvalidInput as InvalidInput,
+    InvalidInput as InvalidInput,  # noqa F401
     lib2to3_parse as lib2to3_parse,
     parse_ast,
     stringify_ast,
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -72,7 +72,11 @@ from black.output import (
     out as out,
     err as err,
 )
-from black.report import Report as Report, Changed, NothingChanged as NothingChanged
+from black.report import (
+    Report as Report,
+    Changed as Changed,
+    NothingChanged as NothingChanged,
+)
 from black.files import (
     find_project_root as find_project_root,
     find_pyproject_toml as find_pyproject_toml,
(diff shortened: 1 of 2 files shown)


jack1142:
I would say that `__all__` would be the preferred option here (and is supported by both type checkers and various linters which is not necessarily the case with `X as X` syntax).

The `X as X` makes more sense in stub files where `__all__` should only be defined *if* the real package actually defines it.

JelleZijlstra:
I'm a bit hesitant about this, because at the moment Black is only an application, not a library. We have a longstanding issue about defining a public API, but I hesitate to mark anything as explicitly part of the API until we finish that issue.

danpf:
>  would say that __all__ would be the preferred option here

Can change it if you'd like.  I only prefer this way because I find it's easier to maintain. I'm not attached.

> I'm a bit hesitant about this

That's fine, if you don't think it's a good fit for the library or the library-as-it-is-now feel free to close the issue. no hard feelings.

I guess to combine the two comments:

Even after reading the documentation/best-practices/SO-posts w/r/t this area I find it all slightly murky tbh so I'll do whatever you both suggest.  I have personally decided that `__all__ =` is for *hard* exports that are integral to an API (changes require major version bump), while I use `X as X` for things that are more fluid but convenient at the time.

Under this logic I make sure to never version anything I develop past `beta` so I don't have to use `__all__` :p

felix-hilden:
Quick question, since this is new for me: looking at the link you provided, I see only a mention of stub files (it being a doc for stub files, duh). But this PR would apply it to our source. So does the same aliasing distinction indeed apply to ordinary sources as well?

I'd gravitate towards `__all__` as well, or even cleaning up the top level namespace to only include the things that we want to export in the first place. But let's discuss that with the public API.

danpf:
> So does the same aliasing distinction indeed apply to ordinary sources as well?

I can't answer your question without any doubt, but _in my experience_ since inline typing has increased in popularity the line between stub files and actual python code has gotten increasingly blurry.

... 20 minutes later ...

From what I can tell, it appears that the current intent is: if a library has a py.typed file, the .py files can be treated as .pyi files and can be held to a higher typing standard.  This interpretation seems to be mostly spread by the pyright developer Eric Traut, who has discussed (or at least mentioned) it with the powers that be.

Both pyright and mypy seem to somewhat follow this interpretation, and since mypy is the basis of the python typing system, I think it's fair to just do what they do.


Relevant literature:

https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport
https://github.com/python/mypy/issues/8754
https://github.com/pytest-dev/pytest/pull/7593
https://mail.python.org/archives/list/typing-sig@python.org/thread/YLJPWECBNPD2K4TRIBRIPISNUZJCRREY/#C2ZZKEM3EESWMDUF3PYZB4ZG2L42XUZT

not really sure where that leaves us today, but thats the skinny.

ichard26:
I know this PR was (implicitly) rejected, but I wanted to say thank you to you @danpf for opening this PR anyway. We now know to take this into account when we come around to defining a stable API for black.

PR #3034, opened 2022-04-25
Read simple data cases automatically

saroad2:
### Description

When testing black formatting, the simple use cases where defined hard-coded in a list named `SIMPLE_CASES` in *test_format*.
This is a shame since many times people add a new data case and forget to add it to this list and then it does not run!
Now, one can read all the available cases in a given directory using a method named `all_data_cases`. This method searches for all the cases in a subdirectory of *data* and return them as names.

### Checklist - did you ...

### Moving forward
If you approve this PR of simple cases, we can do the same for the other hard coded cases lists in *test_format.py*
Let me know if you want me to change those too.


tests/test_black.py
```diff
@@ -1168,7 +1168,7 @@ def test_reformat_one_with_stdin_and_existing_path(self) -> None:
             report = MagicMock()
             # Even with an existing file, since we are forcing stdin, black
             # should output to stdout and not modify the file inplace
-            p = Path(str(THIS_DIR / "data/collections.py"))
+            p = Path(str(THIS_DIR / "data" / "simple_cases" / "collections.py"))
```
felix-hilden, line 1171:
```suggestion
            p = THIS_DIR / "data" / "simple_cases" / "collections.py"
```
I know this is not your doing, but I'm pretty sure we don't need a double conversion if `THIS_DIR` is a path already 😅

felix-hilden (commented):
Hi! Thanks for the PR. I like the idea, since it's easier to just put a file in a folder rather than having to edit a list somewhere else. We could definitely do this for all the test categories. I'd be okay with doing it in a future PR, but thought I'd mention it already.

EDIT: Unless we have the same file in multiple locations!

The changes themselves look alright to me. And to the best of my abilities I tried to verify that the correct files were indeed moved 😄 Nicely done!

saroad2:
Hey @felix-hilden and @JelleZijlstra , thank you for the review!

Once this PR is approved, I will do the same for the other test cases.

felix-hilden (approved):
Yep, this looks fine! And thanks for considering taking on the whole task 🙏

saroad2:
@felix-hilden, would you like me to make the change for all cases in this PR, or in a separate one?
If the latter, Please merge this PR and I will create a new one.

felix-hilden:
A separate PR works just fine! But we like to have at least another maintainer have a look at things before merging so that we don't miss anything. Please stand by 😄

JelleZijlstra (approved):

CHANGES.md
```diff
@@ -62,6 +62,12 @@
 
 <!-- Changes that improve Black's performance. -->
 
+### Testing
+
+<!-- Changes that improve Black's testing process. -->
+
+- Simple cases now read automatically from specified subdirectory (#3034)
+
```
ichard26, line 70: Do we need this? it's not user facing ...

ichard26 (commented):
Would this kind of API for read_data be more developer-friendly?
```python
source, expected = read_data("simple_cases/expression.py")
```
vs what is currently proposed ...
```python
source, expected = read_data("expression.py", cases_dir="simple_cases")
```
Anyway, this PR looks great to me regardless.

felix-hilden:
Certainly works too! If I had to come up with one, and given that soon we won't have any files in the top directory, we could have:

```python
def read(subdir: str, file: str):
    data_file = (TOP / subdir / file).with_suffix(".py")
```

or whatever. That would work around hypothetical slash compatibility, would be naturally before the file name, *and* the fact that writing `cases_dir` all the time is a bit awkward.

CHANGES.md
saroad2: The PR won't pass the CI process if this is missing.
If you want, and you will force-push the PR

JelleZijlstra: We can add a label to skip that check. I think it's fine to keep the news entry though, so we give credit to your contribution.

cooperlees (approved):
Just want to share my thanks, I am a HUGE fan of removing static manual lists like this!

#techdebtCleanup

saroad2:
Hey everyone, thanks for the review

@ichard26's suggested API will work just fine at the moment without any change to code:

 ```python
 source, expected = read_data("simple_cases/expression.py")
 ```

However, I like @felix-hilden suggestion more since in the end there will be no cases in the top directory, but all should be placed in subdirectories according to their use-case:

```python
def read(subdir: str, file: str, ...):
    data_file = (TOP / subdir / file).with_suffix(".py")
```

My suggestion is this:
1. In this PR I should use the API suggested by @ichard26.
2. In the next PR I will move all cases to subdirectories and change the API to @felix-hilden suggestion.

Everyone agrees?

Two more things:
* Please let me know how to pass the CHANGES.md file check in order for me to remove this PR from the changelog. I agree with @ichard26 that since this is an internal change that is not user-facing it doesn't belong there.
*  @cooperlees , thanks for the complement! cleaning code and tests is one of my favorite things 😊

felix-hilden:
That works! And we make non-changelog changes happen using a label which I applied just now 👍

saroad2:
Great! I made all the changes we've discussed.

@felix-hilden , I really don't need that entry in the CHANGES.md to feel that I get credited for my work.
In my other PR I added myself to the AUTHORS.md, and that's all the credit I need for now.

*Black* is a tool that's really close to my heart and I'm happy to contribute for it whenever I have the time.

felix-hilden (approved):

JelleZijlstra (approved):

PR #2996, opened 2022-04-05
Standardise newlines after module-level docstrings

jpy-git:
### Description

Closes #1872.

Does 2 things:
a) Refactor `EmptyLineTracker` for more consistent look-back logic. Simplifies logic by removing complex interactions between `before` and `after` for various rules.

b) Standardise newlines after module-level docstrings. We now enforce 1 newline after both single and multi line docstrings.
e.g.
```python
"""docstring"""

import foo
```
```python
"""docstring

blah blah blah
"""

import foo
```

### Checklist - did you ...

- [X] Add a CHANGELOG entry if necessary?
- [X] Add / update tests if necessary?
- [X] Add new / update outdated documentation?

PR #3035, opened 2022-04-25
Remove newline after code block open

### LINKED ISSUE #902: Black should enforce zero empty leading lines in function body
Operating system: playground
Python version: playground
*Black* version: master

Consider following code snippet:
```python
class X:
    
    
    def f(self, x):
        
        
        return x**2
```

When class definition is formatted, empty lines between `class X:` line and first node of the body is forced to be zero.
When function definition is formatted, empty lines between `def f():` line and first node of the body is normalised to either 0 (if input code had 0 empty lines) or 1 (if input code had 1+ empty lines).

While it's preferable to preserve some of the empty lines of input code, doing this for *leading* empty lines of node body makes no sense, as this empty line cannot have any mental meaning in such context.

saroad2:
### Description

Fixes #902

Newlines are great. They help us organize our code and leave room for the coders and reviewers to breath.
However, too many newlines can make simple code gigantic and massive. This is unnecessary in a lot of cases.

I want to argue that having a newline right after `def` is redundant and Black should remove it.
**EDIT**: We decided to remove trailing newlines for *all code blocks*, including `def`

Now, the following code:
```python
def f():

    print("No newline above me!")

    print("There is a newline above me, and that's OK!")
```

will be changed by *Black* to:
```python
def f():
    print("No newline above me!")

    print("There is a newline above me, and that's OK!")
```

### Checklist - did you ...


saroad2:
hey @JelleZijlstra, Thanks for the review!

I made the new feature into a preview feature.
Let me know if any other changes are necessary or if you think another there are other test cases that I didn't add.

jpy-git:
I'd actually already fixed this but was waiting on #2996 to get reviewed first

Based on [this](https://github.com/psf/black/issues/902#issuecomment-1055606617) comment my approach was to just add an `opens_block` property to `Line`

```python
@property
def opens_block(self) -> bool:
    """Does this line open a new level of indentation."""
    try:
        last_leaf = self.leaves[-1]
    except IndexError:
        return False

    return last_leaf.type == token.COLON
```

Then the logic for all blocks (including `def`) is:
```python
if (
    Preview.remove_def_trailing_newline in current_line.mode
    self.previous_line.opens_block
):
    return 0, 0
```

saroad2:
Hey @jpy-git , I'm sorry for the work duplication. I didn't know you are working on it.

For now, I implemented the logic only for `def` statements because I wanted to see how people react to that change. Removing all trailing newlines for all code blocks seems to me like a little bit too big of a change and I thought maybe some users might have a good idea why it is not good to remove trailing newlines.

Let's wait for the maintainers to react to this PR and see which approach they prefer. If they choose to remove all trailing newlines from all code blocks, I'll change the implementation to what I suggested :)

Thanks!

felix-hilden:
As discussed on the issue, I'd be open to removing all newlines at the beginning of blocks. I don't see a situation where it would be better. Others?

saroad2:
Thank for the feedback @felix-hilden .

Let's see what other reviewers say and if all agree, I will make the change for all blocks using @jpy-git suggestion.

saroad2:
Hey everyone, checking in on this PR.

What did we decide? make the change for all code blocks, or keeping it just for `def` statements for now?

@cooperlees , @ichard26 , I would love to hear your opinion on this.

JelleZijlstra:
I'm fine with always removing them.

ichard26:
> What did we decide? make the change for all code blocks, or keeping it just for def statements for now?

I'm in favour of making the change for all syntactic blocks. Just make sure that code style documentation covers this change. (if changes are needed, please edit the future style documentation)

saroad2:
It seems like there is a consensus on removing new lines after all code blocks opening, so I'll go ahead and update the PR 🎉 .

One small request, if I might: I would love if you could merge #3062 first so I won't have to add the new test cases manually to `PREVIEW_CASES`. This is not a must, but it would be nice 😄

saroad2:
Well, THAT was fast!

Thanks @JelleZijlstra !

saroad2:
Hey everyone!

I'm glad to share that I made the change for ALL CODE BLOCKS OPENING. I added unit tests for every code block and added the new feature to future_style.md.

Please let me know if there's something else that needs to be done here.

I won't to give a special thanks to @jpy-git for the implementation suggestion. I changed it a little bit, but the core idea is yours.

felix-hilden (commented):
Nice! I quite like the idea of having atomic test cases in each file, but it's different from the current setup where a category of tests is contained in a single file (or a few files). I don't feel strongly about it, but perhaps it would be worth considering just to be consistent.

The doc is also good, although in my personal opinion we could keep it terse since we have similar generic newline speak in the ["current style" doc](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#empty-lines) already. I don't see a reason to motivate the reduction of newlines here, or at the very least when we make this default behavior and merge the docs.

Also, as a side note, since we already remove newlines in class bodies (and this code never handles class bodies or defs as a result) we could try to merge the logic eventually. Reading below, seems like pyi files are the only exception.

saroad2:
After looking at the results of `diff-shades`, I think I found an exception to the rule:

At the moment, *Black* will change the following code:
```python
class A:
    
    def hello(a):
        print(f"{a} is here")
```
To:
```python
class A:
    def hello(a):
        print(f"{a} is here")
```

However, this code:
```python
class A:

    my_name = "Inigo Montoya"

    def hello(a):
        print(f"{a} is here")
```

Will not be changed. If we apply the new feature in this PR it **will be changed** to:
```python
class A:
    my_name = "Inigo Montoya"

    def hello(a):
        print(f"{a} is here")
```

In my opinion, the previous style is better in this case.

Therefore, I suggest **excluding `class` from this rule**. That means that the change will be applied to all code blocks opens, except `class`.

felix-hilden:
For the record, I wouldn't mind removing the newline in this case also unless there's a specific reason for this exemption. Buuut, no biggie.

saroad2:
@felix-hilden , as for the tests, in my humble opinion the current way the data cases of *Black* are written is **wrong**.

Tests should be as small as possible and should each check only one behavior. When creating a case data file which is so big that many changes occurs in it, one could not isolate a problem in the formatting when he or she faces one.

A better way is to write more cases data files with each case containing as few lines as possible. In that way, when one line formatting is incorrect, you don't need to read huge files in order to track the error, but simply look at which test was failed, and see what line it failed in.

Now that the tests are dynamically read from the directory, you don't have to add those test cases to a long, tedious lists of cases. This allows you to write as many test cases as you want. Just keep them in the right directory and *Pytest* will read all of them automatically :).

felix-hilden:
I tend to agree, my comment was indeed only for consistency. The best way in my opinion would be to have subfolders for each category of tests so that scrolling and looking around would be minimised 👍

saroad2:
@felix-hilden , You're suggestion is very logical to me.

Maybe we could change the `all_data_cases` function to *search also in subdirectories of the given directory*. In that way we could use the current syntax:
```python
@pytest.mark.parametrize("filename", all_data_cases("preview"))
def test_preview_format(filename: str) -> None:
    check_file("preview", filename, black.Mode(preview=True))
```

But having inside the `preview` directory a subdirectory for each new feature with all of its cases.

saroad2:
@felix-hilden , as much as it bugs me to do so, I'll do it.

saroad2:
OK @felix-hilden , I merged the test cases and trimmed down the description of the new feature in the Future Style document.

The only thing left is for other maintainers to review this PR and decide what to do about `class` blocks.

src/black/mode.py
```diff
@@ -148,6 +148,7 @@ class Preview(Enum):
     one_element_subscript = auto()
     annotation_parens = auto()
     long_docstring_quotes_on_newline = auto()
+    remove_def_trailing_newline = auto()
```
felix-hilden, line 151:
```suggestion
    remove_block_trailing_newline = auto()
```
NIT: If we want to be fully, pedantically accurate 😄

felix-hilden (approved):
Provided that we agree to remove newlines in class bodies, I think this would be good to go!

saroad2:
Hey @JelleZijlstra , thanks for the review!

Fixed everything you've mentioned.

docs/the_black_code_style/future_style.md
```diff
@@ -49,3 +49,28 @@ plain strings. User-made splits are respected when they do not exceed the line l
...
+
+```python
+def my_func():
+    print("The line above me will be deleted!")
+
+    print("But the line above me won't!")
+```
+
+This new feature will be applied to **all code blocks**: `def`, `class`, `if`, `for`,
+`while` and `with`.
```
JelleZijlstra, line 76:
```suggestion
`while`, `with`, and `match`.
```

JelleZijlstra (approved):

saroad2:
Hey @ichard26 ,
I believe this PR is waiting for your review. If there's something missing from this PR, please let me know.

I would love to see it merged!

AUTHORS.md
```diff
@@ -148,6 +148,7 @@ Multiple contributions by:
 - [Rishikesh Jha](mailto:rishijha424@gmail.com)
 - [Rupert Bedford](mailto:rupert@rupertb.com)
 - Russell Davis
+- [Sagi Shadur](mailto:saroad2@gmail.com)
```
ichard26, line 151: This AUTHORS list is so out of date, gosh. Good for you adding yourself!

tests/test_black.py
```diff
@@ -1455,7 +1455,6 @@ def test_newline_comment_interaction(self) -> None:
         black.assert_stable(source, output, mode=DEFAULT_MODE)
 
     def test_bpo_2142_workaround(self) -> None:
-
```
ichard26, line 1458: Haha, I like it!

docs/the_black_code_style/future_style.md
```diff
@@ -49,3 +49,28 @@ plain strings. User-made splits are respected when they do not exceed the line l
 limit. Line continuation backslashes are converted into parenthesized strings.
 Unnecessary parentheses are stripped. The stability and status of this feature is
 tracked in [this issue](https://github.com/psf/black/issues/2188).
+
+### Removing trailing newlines after code block open
```
ichard26, line 53: Oh lovely, I've been worried lately our style documentation is out of date with the current state of affairs, but this quells my worries with this PR ^^

ichard26 (approved):
I still haven't looked into the test infrastructure improvements, but everything else looks great! I'm so sorry I took so long to review this, thank you so much for waiting!

felix-hilden:
With three approvals, I think it's high time we get this in! Thanks for the contribution, responsiveness and patience 😃 Also, I don't remember if this has been linked to you since this isn't your first PR, but Rich has been gathering feedback from contributors in #2238 We'd appreciate any feedback there!

elviskahoro:
Is there a configuration setting for disabling this? @JelleZijlstra @saroad2 @felix-hilden 

I use Vim and rely on the jump to next paragraph feature quite a bit. Having a line break between my function signature and the subsequent docstring is super helpful for quickly navigating across a file. How can I prevent Black from removing these line breaks?

Here's an example of what our functions often look like:

```py
def test(
    a: int,
    a: int,
    a: int,
) -> Tuple[str, str]:

"""
Long docstring
.
.
.
.
.
.
A number of lines
"""

if True:
   rest_of_function

lots_of_code = ""

return "", ""
```

ichard26:
We will not be adding a toggle for this. I'm still not convinced that this should be reverted, however, I realize this is a disruptive change. The recommended workflow is to pin Black to `~= 22.0` so you continue to only get the 2022 stable code style.

If you find this unacceptable and want style editions to be configurable, please see https://github.com/psf/black/issues/3539.

elviskahoro:
Thanks for the insight @ichard26

bzoracler:
I might have missed something, but any reason why this new style isn't enforced for *docstrings* in class bodies?

```python
class A:

    """
    This docstring doesn't shift upwards in black v23.1
    """
```

Other kinds of statements seem to work

```python
class A:

    a = 1  # This shifts upwards
```

PR #3038, opened 2022-04-27
Add script to ease migration to black

hbrunn:
### Description

I was looking for a script to update existing feature branches automatically when introducing black in the main branch, but couldn't find any

To test, you need to blackify your main branch, commit that, switch to a feature branch and then run the script. You'll end up with a copy of your branch blackified and rebased on the main branch 

### Checklist - did you ...

I actually did none of the above but happily will update the docs if you think it's a good idea to have such a script here in the first place.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,7 +38,7 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
-- Add script migrate-black.py to ease migration to black
+- Add migrate-black.py script to ease migration to black formatted git project (#3038)
 
 ### Output
 


cooperlees:
Thanks for the PR. The idea is great, but I feel if we were to add this we might need to think about it supporting some more configuration and support all Operating Systems - e.g. Windows. So maybe this could be written in Python using subprocess + we add some testing.

What do others think?

hbrunn:
I'm not others, but I think you're right the script should work on whatever platform black supports, and didn't think of windows (while taking care this is going to work on any unixoids, ie this should work on Macos, BSD, etc even though I use some Linux).

Converting this to pure Python is simple enough, but will take time as that's then my private fun and not customer's time.

I'm not very familiar with current windows stuff, can we expect a `patch` command there? Or should this pull another python library then to do it?

What kind of configuration do you have in mind?

hbrunn:
I just pushed a pure python version of this, also replacing `patch` with `git apply`

hbrunn:
I just pushed a pure python version of this, also replacing `patch` with `git apply`

jack1142:
I could probably check but do you know if this script retains the original author (and maybe committer?) date? And I guess the author/committer themselves too.

hbrunn:
that's what `git commit -C $commit` does

scripts/migrate-black.py
```diff
@@ -0,0 +1,83 @@
...
+    for commit in commits:
+        git("branch", "-qD", "%s-black" % commit)
+
+
+if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("base_branch")
+    parser.add_argument("--black_command", default="black -q .")
```
cooperlees, line 81: I'd recommend not using `-q` and piping black's output to a log incase something goes wrong and then we have logs to debug.

scripts/migrate-black.py
```diff
@@ -0,0 +1,83 @@
...
+# check out every commit added by the current branch, blackify them,
+# and generate diffs to reconstruct the original commits, but then
+# blackified
+import os
+import sys
+from subprocess import check_output, run, Popen, PIPE
+
+
+def git(*args):
+    return check_output(["git"] + list(args)).decode("utf8").strip()
```
cooperlees, line 11: Why are you stripping newlines here? Couldn't that messup output?

scripts/migrate-black.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def blackify(base_branch, black_command):
+    current_branch = git("branch", "--show-current")
+
+    if not current_branch or base_branch == current_branch:
+        print("You need to check out a feature brach to work on")
```
cooperlees, line 18: Generally logging message I recommend using a logger from the logging module and writing to stderr - But we can clean that up in a follow up diff.

scripts/migrate-black.py
```diff
@@ -0,0 +1,83 @@
...
+    if not merge_base:
+        print("Could not find a common commit for current head and %s" % base_branch)
+        return 1
+
+    commits = git(
+        "log", "--reverse", "--pretty=format:%H", "%s~1..HEAD" % merge_base
+    ).split()
+    for commit in commits:
+        git("checkout", commit, "-b" "%s-black" % commit)
+        check_output(black_command, shell=True)
```
cooperlees, line 35:
```suggestion
        import shlex
        check_output(shlex.split(black_command))
```
Shouldn't need shell=True here ...

cooperlees (changes requested):
Thanks for this. This is much friendlier to all operating systems.

Only things I think we could do here:

- Should we check that we have a new enough git or are all these arguments been around a long time?
- Lets log black output to a file with the logfile from an option argparse argument
- Maybe also specify you need git + black in your PATH for this to work
  - Optionally we could check for that
- Lets type annotate - Again - If you don't have time I can do once we merge ...

scripts/migrate-black.py
hbrunn: the idea was to allow things like `black .; isort .` here

scripts/migrate-black.py
hbrunn: no, but we need to get rid of the whitespace some versions of git add before/after its output

StefanRijnhart (commented):
As a member of @hbrunn's team, I just tested this script on some of our open PR branches and it works really well.

@hbrunn please check the linting of the script itself, and it looks like you'll want to add that changelog entry indeed.

hbrunn:
thanks for the reviews, now correctly formatted, using logging and with a changelog entry

jack1142:
I think you should probably use an underscore (`_`) instead of a hyphen (`-`) in the name of the script (`migrate-black.py`) since `migrate-black` is not a valid module name (not much of an issue for a script but still) and it's consistent with what's already in that folder.

Not sure if it's a particularly great name for the script since I personally would find it useful for reformatting PRs I made that are incorrectly formatted, not just for migrating existing feature branches during migration.

hbrunn:
> Not sure if it's a particularly great name for the script since I personally would find it useful for reformatting PRs I made that are incorrectly formatted, not just for migrating existing feature branches during migration.

I'll be happy to rename this, what do you propose?

CHANGES.md
```diff
@@ -38,6 +38,8 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- Add script migrate-black.py to ease migration to black
```
cooperlees, line 41:
```suggestion
- Add migrate-black.py script to ease migration to black formatted git project (#3038)
```

scripts/migrate-black.py
```diff
@@ -0,0 +1,92 @@
...
+# check out every commit added by the current branch, blackify them,
+# and generate diffs to reconstruct the original commits, but then
+# blackified
+import logging
+import os
+import sys
+from subprocess import check_output, run, Popen, PIPE
+
+
+def git(*args):
```
cooperlees, line 11: `mypy` CI is failing - Can you please type annotate the code since we have very high type coverage elsewhere.

cooperlees (changes requested):
I feel the name is fine. Lets just type annotate to get CI passing and we can merge this to have a starting place and let others improve it as people find time to do so.

e.g. unittests ...

scripts/migrate-black.py
hbrunn: done

scripts/migrate-black.py
```diff
@@ -0,0 +1,96 @@
...
+# and generate diffs to reconstruct the original commits, but then
+# blackified
+import logging
+import os
+import sys
+from collections import abc
+from subprocess import check_output, run, Popen, PIPE
+
+
+def git(*args: abc.Iterable[str]) -> str:
```
cooperlees, line 12:
```suggestion
def git(*args: Sequence[str]) -> str:
```

scripts/migrate-black.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python3
+# check out every commit added by the current branch, blackify them,
+# and generate diffs to reconstruct the original commits, but then
+# blackified
+import logging
+import os
+import sys
+from collections import abc
+from subprocess import check_output, run, Popen, PIPE
```
cooperlees, line 9:
```suggestion
from subprocess import check_output, run, Popen, PIPE
from typing import Sequence
```

cooperlees (changes requested):
Close, but one last error ... Lets prefer to use types from the `typing` library :)

It's easy to run `mypy` locally or run `pre-commit` to run the CI

scripts/migrate-black.py
```diff
@@ -0,0 +1,95 @@
...
+# check out every commit added by the current branch, blackify them,
+# and generate diffs to reconstruct the original commits, but then
+# blackified
+import logging
+import os
+import sys
+from subprocess import check_output, run, Popen, PIPE
+
+
+def git(*args: str) -> str:
```
cooperlees, line 11: Wow, this must be new support from mypy :|

cooperlees (approved):
Thanks for the persistent effort here. I hope people find it useful.

I guess the only way people might find it is docs - so maybe we should add a parapgraph here (with usage examples): https://github.com/psf/black/blob/main/docs/guides/introducing_black_to_your_project.md

Happy for this to be in another PR.

cooperlees:
This script is not included in the black distribution nor do I think it should be. I think we should just document it as an best effort supported helper script in our documentation that lives in the repo here.

As it has no externals deps, install is just - `wget https://github.com/psf/black/blob/main/scripts/migrate-black.py && chmod +x migrate-black.py` providing a python3 executable is in path ...

PR #3043, opened 2022-05-03
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/upload_binary.yml
+++ b/.github/workflows/upload_binary.yml
@@ -10,7 +10,7 @@ permissions:
 jobs:
   build:
     permissions:
-      contents: write  # for actions/upload-release-asset to upload release asset
+      contents: write # for actions/upload-release-asset to upload release asset
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/.github/workflows/upload_binary.yml
+++ b/.github/workflows/upload_binary.yml
@@ -4,8 +4,13 @@ on:
   release:
     types: [published]
 
+permissions:
+  contents: read
+
 jobs:
   build:
+    permissions:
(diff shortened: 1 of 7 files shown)


ichard26:
Hmm, it probably shouldn't be saying differences were found, I'll need to look into this 🤔

.github/workflows/upload_binary.yml
```diff
@@ -4,8 +4,13 @@ on:
   release:
     types: [published]
 
+permissions:
+  contents: read
+
 jobs:
   build:
+    permissions:
+      contents: write  # for actions/upload-release-asset to upload release asset
```
JelleZijlstra, line 13:
```suggestion
      contents: write # for actions/upload-release-asset to upload release asset
```
fix lintt

JelleZijlstra:
> Hmm, it probably shouldn't be saying differences were found, I'll need to look into this 🤔

Yes, weird. The report doesn't actually show any changes.

ichard26:
Thanks for the contribution @naveensrinivasan ! Congrats on your PR 🎉 

And yes we've locked down the default token permissions to read-only.

Issue #1632, opened 2020-08-26
Docstrings reformatted to be too long


mlissner:
**Describe the bug** When there's a docstring that's near the max line length with the `"""` on the next line, the new version of Black moves the `"""` to the first line, making the line too long.

**To Reproduce** Steps to reproduce the behavior:

1. Create a file with this content:

    ```
    def my_func():
            """Remove punctuation and return cleaned string plus its length in tokens.
            """
            pass
    ```

2. Reformat it

3. Note that it now has the docstring on one line and that that line has 82 chars:

    ```
    def my_func():
            """Remove punctuation and return cleaned string plus its length in tokens."""
            pass
    ```

**Expected behavior** 

The docstring shouldn't get wrapped.

**Environment (please complete the following information):**

- Version: [e.g. master] 20.8b1

- OS and Python version: 
Linux/Python 3.8.2

**Does this bug also happen on master?** To answer this, you have two options:

Yes: https://black.now.sh/?version=master&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AC3AIldAD2IimZxl1N_WlbvK5V-ga8lyrVzgK48mW1g5y7ZvjbGhEN7nG-8hLA9oQ3I4FUxivNyD3Gy0_4fbH0Mg7-k28ZpTZzLozFMvMFSuDJT_0ZQmFpNZw6AAvoHjtKEQjGQksHXW1WBInIzl_HhTgRlCs-04wcVCEBY4OInuM5XhKBPO4ZIAzTqLcpeAAAAADHge2pSp2OmAAGlAbgBAAByceBuscRn-wIAAAAABFla

ichard26:
Hello @mlissner.

The default line length limit is `88` so I don't see what is wrong with Black wrapping the docstring as it still fits. Do you happen to have a `pyproject.toml` configuration file?

mlissner:
Sorry, should have mentioned we have it set to 79. I think this link has that config and you can still see it:

[...]

PR #3044, opened 2022-05-03
Docstring too long (issue #1632 fix)

idorrington92:
### Description

Fixing this issue https://github.com/psf/black/issues/1632 where some docstrings were reformatted to be too long

### Checklist - did you ...

- [x] Add a CHANGELOG entry if necessary? Done
- [x] Add new / update outdated documentation? Don't think this is necessary


felix-hilden:
Thanks for submitting! Some tests to validate this behaviour would be in order, and we should add a changelog entry and point to this PR.

@JelleZijlstra should we allow even these smallest and clearly unintended bugs to be fixed in the default style, or should we let this marinate in `--preview`?

If preview, then we have to add a flag for it in `black.mode.Preview` and use it to enable this (example PR #2770). I'd suggest a flag called `bugfixes` or something similar, under which we'll gather all the small things we definitely want to add to the default style.

JelleZijlstra:
Yes, this should be in the preview style.

CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fixed bug where docstrings with triple quotes could exceed max line length (#3044)
```
JelleZijlstra, line 5: Move this to the "Preview style" section

felix-hilden:
In addition to the flag, we have to use it to enable the change. This would currently change the default behaviour. So something like `if Preview.doc_line_length in mode: bugfix`.

idorrington92:
> In addition to _having_ the flag, we have to use it to enable the change. This would currently change the default behaviour. So something like `if Preview.doc_line_length in mode: bugfix`. Also my bad for not noticing the change log, I was probably just looking at the first commit

Ah, I think I understand now. I didn't really understand what you meant about the preview but I think I've got it now

src/black/linegen.py
```diff
@@ -329,7 +329,24 @@ def visit_STRING(self, leaf: Leaf) -> Iterator[Line]:
 
             # We could enforce triple quotes at this point.
             quote = quote_char * quote_len
-            leaf.value = prefix + quote + docstring + quote
+
+            if Preview.long_docstring_quotes_on_newline in self.mode:
+                leaf.value = prefix + quote + docstring + quote
+            else:
```
felix-hilden, line 335: This should be the other way around, no? The idea is that if preview is enabled, we do the new thing.

src/black/linegen.py
```diff
@@ -305,9 +305,9 @@ def visit_STRING(self, leaf: Leaf) -> Iterator[Line]:
             quote_len = 1 if docstring[1] != quote_char else 3
             docstring = docstring[quote_len:-quote_len]
             docstring_started_empty = not docstring
+            indent = " " * 4 * self.current_line.depth
 
             if is_multiline_string(leaf):
-                indent = " " * 4 * self.current_line.depth
```
felix-hilden, line 310: This should most likely also have the preview check, or I've missed some context.

idorrington92: It doesn't make any difference to the output, as if it is a multiline string then it's identical, and otherwise indent doesn't get used. I can put it in the preview if you like? Is the idea of the preview to keep all the code the same or unless previewing or just keep the formatting the same?

src/black/linegen.py
idorrington92: I've fixed this now. Changed a test to have the old incorrect formatting as well

src/black/linegen.py
felix-hilden: Right, that was the missing context then 😄 It's a-ok! Having all refactors in preview would be.. not good. Just the formatting 👍

src/black/linegen.py
felix-hilden: Please add the tests to a dedicated file and name that file [here](https://github.com/psf/black/blob/main/tests/test_format.py#L80) so we can test with preview on!

idorrington92: Do you want the tests in the new file and the old file?

idorrington92: I assume you want the old tests as well so I've kept them in

src/black/linegen.py
```diff
@@ -329,7 +329,24 @@ def visit_STRING(self, leaf: Leaf) -> Iterator[Line]:
...
+            if Preview.long_docstring_quotes_on_newline in self.mode:
+                # Put closing quotes on new line if max line length exceeded
+                last_line_length = len(docstring.splitlines()[-1]) if docstring else 0
+
+                # Make the docstring apart from the closing quotes, which happen below
+                docstring = prefix + quote + docstring
+                if (
+                    len(indent) + len(prefix) + 2 * len(quote) + last_line_length
+                    <= self.mode.line_length
+                ):
```
felix-hilden, line 342: I've tried to think this through, but I'm not 100 % still. Wouldn't this be incorrect if we had a multiline docstring whose last line was just acceptable? We would sum the prefix and quote length to it and refuse to add quotes at the end.

```python
def foo():
    """first line---------------------------------------------------------------

    second line--------------------------------------------------------------"""
```

In any case, a test about this would be nice too.

felix-hilden (commented):
Still some comments, hang in there! And thanks for your patience 😄

src/black/linegen.py
felix-hilden: Yeah that works, can't have too many 👍

src/black/linegen.py
idorrington92: You were right. There were two issues with my original code, the last_line_length actually included the indent for docstrings of more than one line. I fixed this issue by using lstrip first. The other issue is that if the docstring has more than one line then we only have one set of quotes so we shouldn't multiply it by two. The logic is a bit more complicated now, but I've added the example above as a test and it works fine now. 

Thanks for the good spot :p

felix-hilden: Nice 👍 looking at the new code: are you sure that the prefix doesn't affect the calculation similarly? Should we instead only calculate the line length once? Since we only really care about the last one, perhaps something like:

```python
lines = docstring.splitlines()
last_line_len = lines[-1]
if len(lines) == 1:
    last_line_len += quote_len + len(prefix)

if last_line_len + quote_len > max_len:
    # add newline
...
```

I'm not sure how the indentation matters if we strip it only to add it back. Maybe that could simplify it a bit. But I might be disregarding some details.

idorrington92: I was pretty tired when working on this last night, that's my excuse for missing the prefix :p 
I think it's all sorted now. I've added some tests with prefixes now just to be sure. I've also refactored it how you suggest and added in some (possibly excessive) comments to explain exactly why we're doing all this.

Thanks for your help :)

felix-hilden (approved):
The comments are a nice touch 👌 I think I've convinced myself that this is working now, so LGTM! 😄

src/black/linegen.py
felix-hilden: No worries! Thanks for being so quick with the edits 👍

JelleZijlstra (approved):

ichard26:
Congrats on your first merged PR to black @idorrington92! 🎉

jensguballa:
I am still having problems: 

[https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADpAJ9dAD2IimZxl1N_WlXnON2nzPJ4souZqeMv9or-qWkdsKldYBXHRtFaSeHXTIoFVb8OVPJ6gxK1EZIccJrnnx6tDCX10kz5Nu2DOyfZf9FgwMdCmz4KGGVA4LIrx3tTw9B_MUu234i1BEVsjEPHourJoZoKcgwi_cYFC4ffNJa3fHSgvIzJHe_dSVoXi8Fbs4E6s2x8CJVs92SrX-EOfN3sAAAAEIIlY9ypyagAAbsB6gEAALDVODaxxGf7AgAAAAAEWVo=](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADpAJ9dAD2IimZxl1N_WlXnON2nzPJ4souZqeMv9or-qWkdsKldYBXHRtFaSeHXTIoFVb8OVPJ6gxK1EZIccJrnnx6tDCX10kz5Nu2DOyfZf9FgwMdCmz4KGGVA4LIrx3tTw9B_MUu234i1BEVsjEPHourJoZoKcgwi_cYFC4ffNJa3fHSgvIzJHe_dSVoXi8Fbs4E6s2x8CJVs92SrX-EOfN3sAAAAEIIlY9ypyagAAbsB6gEAALDVODaxxGf7AgAAAAAEWVo=)

Line length is set to 88. After the reformatting the line produced by black has a length of 89 characters, which causes flake8 to complain.

Black version: 22.12.0 ( I tried 22.6.0 as well, same result)

So I would regard this issue as not fixed, unless I am missing something.

idorrington92:
@jensguballa This was fixed, then this PR https://github.com/psf/black/pull/3430 was merged that causes the issue again for one line docstrings. The issue that being fixed by the PR is here https://github.com/psf/black/issues/3320

I don't know much about how black works beyond this PR, but I don't think it reformats docstrings, so I think if you have a one line docstring that goes over the line limit with the closing, then you should just manually make it a multiline docstring. 

Maybe it should show a warning when it's doing breaking the line limit?

jensguballa:
Hi @idorrington92, thanks for pointing me to #3430.

After going over it I am a little bit puzzled now, as I am not really sure what the intended behavior should be.

My assumption was that black should not apply reformattings which will introduce new line length violations. In my opinion this sounds reasonable, given that you might have a processing pipeline where e.g. flake8 is doing its job hereafter and would fail (like in my case). But it looks like #3430 is going in another direction.

My view: For the example I have provided [above](https://github.com/psf/black/pull/3044#issuecomment-1364710654) black needs to take a decision: either reformatting to a one-line string and violating the maximum line length, or, leaving the doc string as it is and keeping the conformity.

My preference is clearly to keep the conformity, but if I understand #3430 correctly, reformatting to a one-line string is given priority. I am not in favour of this as special crafting of the doc string by the developer just to work around this issue would go against the philosophy of black.

So would it be reasonable not to reformat the doc string for the example I provided?

PR #3045, opened 2022-05-04
chore: Enable codeql action

naveensrinivasan:
This action runs GitHub's industry-leading semantic code analysis engine, CodeQL, against a repository's source code to find security vulnerabilities.

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql

https://github.com/ossf/scorecard/blob/main/docs/checks.md#sast

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
I'm not sure this will help us much; Black is not a particularly security-sensitive tool.

Maybe we should document that `blackd` shouldn't be run as a publicly accessible server though.

ichard26:
I echo Jelle's thoughts, I don't believe CodeQL is worth the CI resources for Black as we're a developer tool that doesn't do much that could be exploited. The main concerns would be supply-chain attacks, a compromise of someone on the Core Team, or a leaked API token (leading to malicious code being published to PyPI).

Thanks for the PR, but I am rejecting this PR on these grounds. Have a great day/night.

PR #3046, opened 2022-05-06
Move imports of `ThreadPoolExecutor` into `reformat_many()`, allowing Black-in-the-browser

Zac-HD:
This is a slight perf win for use-cases that don't invoke `reformat_many()`, but more importantly to me today it means I can use Black in pyscript or pyodide - i.e. *client-side in the browser*.  

I admit that I wouldn't use it for https://black.vercel.app/ - the heavier initial load is probably worse than some server round-trips - but it's indispensable for demos like https://zhd.dev/ghostwriter/ which allow the user to run arbitrary code.  In fact, that demo is using a custom wheel from exactly this patch 😁 

(I'd be inclined to avoid adding this to the changelog unless "works in the browser" becomes part of the commited public API)

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1,7 +1,6 @@
 import asyncio
 from json.decoder import JSONDecodeError
 import json
-from concurrent.futures import Executor, ThreadPoolExecutor, ProcessPoolExecutor
 from contextlib import contextmanager
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


JelleZijlstra (commented):
I like it! Let's make sure it doesn't break mypyc or anything though.

PR #3062, opened 2022-05-09
Remove hard coded test cases

saroad2:
### Description

This PR is a continuation of #3034.
The end goal of thus PR is to move all cases in the *tests/data* directory into subdirectories and remove all hardcoded cases lists.

### Checklist - did you ...


tests/test_format.py
```diff
@@ -15,38 +15,6 @@
...
-    "percent_precedence",
-    "remove_except_parens",
-    "remove_for_brackets",
-    "one_element_subscript",
-    "remove_await_parens",
-    "return_annotation_brackets",
-    "docstring_preview",
-]
-
 SOURCES: List[str] = [
```
felix-hilden, line 18: Should we also walk the source directories here and get all of the Python files?

tests/test_ipynb.py
```diff
@@ -252,9 +252,7 @@ def test_empty_cell() -> None:
 
 
 def test_entire_notebook_empty_metadata() -> None:
-    with open(DATA_DIR / "notebook_empty_metadata.ipynb", "rb") as fd:
-        content_bytes = fd.read()
-    content = content_bytes.decode()
+    content = read_jupyter_notebook("notebook_empty_metadata.ipynb")
```
felix-hilden, line 255: Very nice refactors all over 👍

tests/test_format.py
```diff
@@ -164,15 +132,16 @@ def test_remove_with_brackets() -> None:
...
+@pytest.mark.parametrize("filename", all_data_cases("py_310"))
 def test_python_310(filename: str) -> None:
     source, expected = read_data(filename)
     mode = black.Mode(target_versions={black.TargetVersion.PY310})
     assert_format(source, expected, mode, minimum_version=(3, 10))
 
 
-def test_python_310_without_target_version() -> None:
-    source, expected = read_data("pattern_matching_simple")
+@pytest.mark.parametrize("filename", all_data_cases("py_310"))
```
felix-hilden, line 142: Just noting that this changed, but I guess it can't hurt to test all the cases!

felix-hilden (commented):
HUGE! 🎉

Some comments below. Do you think changing the `read_data` interface (and I guess that would mean moving notebooks as well) would still be relevant and better? Honestly, not a big deal to me.

tests/test_format.py
saroad2: I believe that the right thing to do is to add the following step/job to the CI process:

```shell
pip install -e .
black --check src
```

**This should not be a unit test in my opinion**, but another step/job in the CI process.

If you want an example, see what I did for my tool [Statue](https://github.com/saroad2/statue) in the CI process.
In the end of the CI I added a step called "Run on self" that checks that the code of statue pass the static code analysis ran by statue.

felix-hilden: True, it appears we only check each file with default params, so that would work. I don't have a strong preference for it being in tests or CI. Perhaps [here](https://github.com/psf/black/blob/main/.github/workflows/test.yml#L53) before publishing test results, so that the whole version matrix is tested. Other opinions?

saroad2:
This PR is now ready for review :)

felix-hilden (commented):
Thanks again! Overall, I like the categorisation and it's nice that most individual test functions are now gone!

Having a directory for a single `fast` is a bit awkward 😄 how about misc? No biggie tho, feel free to ignore if you think it's more natural there. Also, it could be nice to use the automatic `.py` appending consistently.

tests/test_format.py
```diff
@@ -95,25 +62,33 @@ def patch_dump_to_file(request: Any) -> Iterator[None]:
...
-@pytest.mark.parametrize("filename", PREVIEW_CASES)
+@pytest.mark.parametrize("filename", all_data_cases("preview"))
 def test_preview_format(filename: str) -> None:
-    check_file(filename, black.Mode(preview=True))
+    check_file("preview", filename, black.Mode(preview=True))
+
+
+@pytest.mark.parametrize("filename", all_data_cases("preview_39"))
+def test_preview_minimum_python_39_format(filename: str) -> None:
+    source, expected = read_data("preview_39", filename)
```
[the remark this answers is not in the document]
saroad2: We could, but I think it might cause more changes in code.
It's probably worth another PR.

saroad2:
> Having a directory for a single `fast` is a bit awkward 😄 how about misc?

I'm pretty sure there will be more in the future, but I can move it to misc if you think that this is the thing that should be done.

> Also, it could be nice to use the automatic .py appending consistently.

Maybe in a different PR :) This one is too big already.

saroad2:
Hey, checking in on this PR.

I've fixed and responded to all the comments here. @felix-hilden, @ichard26, @cooperlees , is there anything here that needs further attention?

JelleZijlstra (approved):

PR #3069, opened 2022-05-11
use is_number_token insteaed of assertion

MarcoGorelli:
### Checklist - did you ...

Apologies if this is too trivial, the `assert` was just kinda bothering me

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -40,7 +40,7 @@ from mypy_extensions import mypyc_attr
 from black.const import DEFAULT_LINE_LENGTH, DEFAULT_INCLUDES, DEFAULT_EXCLUDES
 from black.const import STDIN_PLACEHOLDER
 from black.nodes import STARS, syms, is_simple_decorator_expression
-from black.nodes import is_string_token
+from black.nodes import is_string_token, is_number_token
 from black.lines import Line, EmptyLineTracker
 from black.linegen import transform_line, LineGenerator, LN
 from black.comments import normalize_fmt_off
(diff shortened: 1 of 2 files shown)


cooperlees (approved):
Totally agree here. Thanks!

JelleZijlstra (approved):

src/black/__init__.py
```diff
@@ -1245,8 +1245,7 @@ def get_features_used(  # noqa: C901
             if value_head in {'f"', 'F"', "f'", "F'", "rf", "fr", "RF", "FR"}:
                 features.add(Feature.F_STRINGS)
 
-        elif n.type == token.NUMBER:
-            assert isinstance(n, Leaf)
```
felix-hilden, line 1249: So is `TypeGuard` is enforced with mypyc? If not, we're missing an assert. (Just wanting to make sure, since I'm not familiar with this)

MarcoGorelli: I'm not familiar with mypyc, so handling over to @ichard26 for this one

felix-hilden: As I understood [the doc](https://mypy.readthedocs.io/en/latest/type_narrowing.html), `TypeGuard` is not equivalent to `isinstance` at runtime 🤔 Or are we saying that we don't really need it?

MarcoGorelli: I don't think it's needed, IIUC the `assert` was only there originally as a hint to `mypy`, it's not a condition one would expect to fail at runtime

felix-hilden: Oh, I see. If this is the case then consider it resolved!

ichard26: Yeah I don't think it's even possible for this assertion to fail as the CST types representing numbers are always of type `Leaf`. TypeGuard is a cleaner way to tell type checkers that we can conditionally narrow a type.

felix-hilden (approved):

ichard26 (approved):

PR #3070, opened 2022-05-12
Add more examples to exclude files in addition to the defaults

ysk24ok:
### Description

I believe it would benefit users if this doc had an example on how to specify multiple files/directories.

### Checklist - did you ...


docs/usage_and_configuration/the_basics.md
```diff
@@ -259,10 +259,14 @@ expressions by Black. Use `[ ]` to denote a significant space character.
...
 target-version = ['py37']
 include = '\.pyi?$'
+# 'extend-exclude' excludes files or directories in addition to the defaults
 extend-exclude = '''
 # A regex preceded with ^/ will apply only to files and directories
 # in the root of the project.
-^/foo.py  # exclude a file named foo.py in the root of the project (in addition to the defaults)
+(
+  ^/foo.py    # exclude a file named foo.py in the root of the project
+  | *.py.bak  # exclude a file with '.py.bak' suffix within the project
```
cooperlees, line 268: Does black by default try to format these? If so, should we exclude this?

cooperlees (approved):
Thanks - I like the example, but I feel .py.bak we should ignore by default ... So do we want another example?

docs/usage_and_configuration/the_basics.md
ysk24ok: In this example, `'\.pyi?$'` is specified and `*.py.bak` is excluded by default.
So in 9a6f541 I changed this example to `*_pb2.py` which is a generated stub file by Protocol Buffer.

docs/usage_and_configuration/the_basics.md
```diff
@@ -259,10 +259,14 @@ expressions by Black. Use `[ ]` to denote a significant space character.
...
 target-version = ['py37']
 include = '\.pyi?$'
+# 'extend-exclude' excludes files or directories in addition to the defaults
 extend-exclude = '''
 # A regex preceded with ^/ will apply only to files and directories
 # in the root of the project.
-^/foo.py  # exclude a file named foo.py in the root of the project (in addition to the defaults)
+(
+  ^/foo.py    # exclude a file named foo.py in the root of the project
+  | *_pb2.py  # exclude an auto-generated file by Protocol Buffer within the project
```
JelleZijlstra, line 268:
```suggestion
  | *_pb2.py  # exclude autogenerated Protocol Buffer files anywhere in the project
```

PR #3071, opened 2022-05-12
Implement support for PEP 646

### LINKED ISSUE #3014: Add support for PEP 646 syntax
PEP 646 ads two new pieces of syntax in 3.11 (https://peps.python.org/pep-0646/#grammar-changes):

- `x[a, *b]`
- `def f(*args: *T): pass`

We should support both and normalize the syntax to have no space after the `*` (also cc @isidentical).

isidentical:
Fixes #3014


felix-hilden:
Yay for diff-shades! Just for an easier view, it reported this case, which seems not so good:
```
-    expected = df.loc[[False] * len(df.index), :]
+    expected = df.loc[[False] *len(df.index), :]
```

isidentical:
Indeed, it is a very nice catch!

JelleZijlstra (approved):

PR #3077, opened 2022-05-19
Use the Vanity Discord Invite

Pysics:
### Description

This PR replaces the normal Discord invite, with the vanity one.

### Checklist - did you ...


ichard26:
Hey thanks for opening a pull request! but ... is this necessary as the current invite is attached to the #black-formatter text channel specifically as far as I know, so in a way using the vanity link would be a downgrade. Are there any other benefits with the vanity URL that I am not aware of?

Pysics:
Yeah, I agree.

Issue #3081, opened 2022-05-21
Replace f-strings with regular strings


KAction:
Often as result of refactoring I end up with f-string with no
`{substitutions}`, and my editor (integrated with flake8) keeps
reminding me about it.

With some stretch one can argue that this is actually formatting issue
and as such within scope of black(1). Would you be interested in patch
that implement this proposal?

```python
print(f"Hello")
```

to become

``` python
print("Hello")
```

JelleZijlstra:
Preview mode does this in some cases but not here. I agree that it's worthwhile, although it does change Python's AST.

PBNMH22:
Hi I am new to this, I was searching for this but the code base is huge. 
Request you to please navigate me to desired file or folder. 
I will try my best.

KAction:
I think it is as simple as dropping `drop_useless_fstring` conditional. I'd like to grab lock on this issue for coming 72 hours. I'll do my best to come with PR within this timeframe.

SUDDSDUDDS:
Hi ur happy est being with out no one knows

JelleZijlstra:
@graingert (responding to your comment on #3087, where I'd like the focus to stay on the broader propsal). This does change the AST:

```
>>> ast.dump(ast.parse('f"x"'))
"Module(body=[Expr(value=JoinedStr(values=[Constant(value='x')]))], type_ignores=[])"
>>> ast.dump(ast.parse('"x"'))
"Module(body=[Expr(value=Constant(value='x'))], type_ignores=[])"
```

JelleZijlstra:
Duplicate of #1156

PR #3082, opened 2022-05-21
Strip pointless f-string

KAction:
### Description

As promised, I came with patch that mostly does what is discussed in #3081.
Code quality is back-of-envelope, and not all tests are converted. I will fix
it once we decide what to do with the most major issue:

```
Black produced code that is not equivalent to the source
```

I can't argue that it is very good and valuable sanity check, and just turning
it off seems reckless. Ideas?

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,8 @@
+with (import <nixpkgs> { });
+mkShell {
+  buildInputs = with python3.pkgs; [ python pytest pathspec click mypy-extensions typing-extensions platformdirs tomli aiohttp ];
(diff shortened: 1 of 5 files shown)


KAction:
Oh, clearly I based on old revision of master. Not really relevant at this point.

JelleZijlstra:
Thanks for your PR @KAction! Are you still interested in improving this?

Per our stability policy, the feature should initially be enabled only in `--preview` mode.

KAction:
Sorry, I feel this change is bigger and requires more resources than I have right now.

PR #3084, opened 2022-05-23
Introduce Pants to the repo, as a proof of concept.

benjyw:
### Description

This change introduces the [Pants build system](https://www.pantsbuild.org/) to the Black repo, on an experimental basis.

It does not change any of the existing build mechanisms for running tests, formatting, linting, 
typechecking, building distributions etc. Developers in this repo can remain unaware of Pants if they choose. 
Others may be  interested in playing around with it.

For context: While chatting with @ambv at PyCon 2022 he expressed interest in seeing what using Pants 
in the Black repo would look like. This change is a proof-of-concept as a result of that conversation.

Notes to reviewers: 
- This PR consists of several commits that, while not fully independent, are useful to review individually. Each commit's message describes in more detail what it contains. 
- My apologies for the size of the diff (8298 lines!). However it's not what it seems: 
  - 7232 lines are generated lockfiles. 
  - 418 lines are standard Pants runner scripts. 
  - 253 lines are "BUILD file" boilerplate and metadata, much of it also generated.
  - 244 lines are the new docs/contributing/pants.md documentation file, worth reading for more detail.
  - 151 lines are miscellaneous changes, and these are worth the greatest line-by-line scrutiny.
  - The commit messages will help distinguish which is which.
- Thanks for taking the time to review this. I'm happy to answer any questions, and the [Pants community](https://www.pantsbuild.org/docs/getting-help) is a good place for any questions or conversation outside the scope of this PR, if needed.

##### What is Pants, and why might you want to use it? 

Pants is a modern build system with a strong focus on Python. It installs, configures and orchestrates 
the underlying tools, such as pytest, flake8, setuptools, mypy, and, of course, Black itself. Pants provides 
a uniform CLI interface for build actions. It aims to make builds repeatable, stable, fast and secure. 
It achieves these goals via features such as: dependency inference, fine-grained invalidation, caching, 
concurrency, lockfiles, hermetic execution and more.

### Checklist - did you ...

My understanding is that this change doesn't require a CHANGELOG entry, since it doesn't modify
Black itself in any way, and is not user-visible. But please let me know if I am wrong about this!


JelleZijlstra:
Thanks, but we recently switched to Hatch and I don't see a reason to switch.

PR #3098, opened 2022-06-04
Memory error

ashgrantham:
memory errer dectetord


ichard26:
`even-more-platforms` is a branch I am working on. It is currently stalled to upstream incompatibilities with CPython 3.11-dev; there are no conflicts so it doesn't need main merged into it. Also, referring to memory error makes no sense here soo I'll be rejecting this PR. I hope you understand. Thank you.

PR #3109, opened 2022-06-06
Document pre-commit with --force-exclude

### LINKED ISSUE #3015: Document using pre-commit and how --force-exclude can help

--- 

Ttibsi:
### Description

Closes #3015 

### Checklist - did you ...

- [/] Add a CHANGELOG entry if necessary? (I don't believe this is required - please let me know if it is and I can amend my PR with an update)
- [/] Add / update tests if necessary?
- [/] Add new / update outdated documentation?

### DIFF
--- a/docs/integrations/source_version_control.md
+++ b/docs/integrations/source_version_control.md
@@ -29,3 +29,15 @@ onwards).
 [black-tags]: https://github.com/psf/black/tags
 [pre-commit-mutable-rev]:
   https://pre-commit.com/#using-the-latest-version-for-a-repository
+
+Pre-commit collects all filepaths recursively throughout the repo to pass to Black. You
+can use the `--force-exclude` flag to ignore specific files or directories, which
+pre-commit takes as an argument. See the partial example below ignoring a `tests`
+directory:
+
+```yaml
+hooks:
+  - id: black
+    args:
+      - --force-exclude=tests/
+```


cooperlees (approved):
Makes sense and works for me.

Only enhancement I could think of is maybe showing or at least stating how to specify multiple directories - i.e. is it two `--force-exclude` or is it comma separated paths etc.

docs/integrations/source_version_control.md
```diff
@@ -29,3 +29,15 @@ onwards).
...
+Pre-commit collects all filepaths recursively throughout the repo to pass to Black. You
+can use the `--force-exclude` flag to ignore specific files or directories, which
+pre-commit takes as an argument. See the partial example below ignoring a `tests`
+directory:
+
+```yaml
+hooks:
+  - id: black
+    args:
+      - --force-exclude=tests/
```
MarcoGorelli, line 42: is this the best place to put this? if you're using pre-commit, you might as well use pre-commit's own `exclude`

I think the more common use-case for `force-exclude` is for putting it in `pyproject.toml`, so that this config stay synced between when running `black` via `pre-commit` or via the command-line

ichard26: I agree with @MarcoGorelli, using `--force-exclude` in pre-commit-config.yaml is a bit idiosyncratic. Either you use it in `pyproject.toml` or you use pre-commit's own exclude configuration. I'd personally recommend pre-commit's exclude first, and then --force-exclude, as the latter can be overkill.

cooperlees (changes requested):
Thanks for this. Reading others comments, I agree people should just use pyproject.toml black config to exclude directories rather than the CLI. Here is an example of how we even do it with black on the black repo: https://github.com/psf/black/blob/main/pyproject.toml#L12

I think due to this should we close this PR as we already have this documented in black docs. If you feel it could be better I'd take a PR improving that.

mscheltienne:
> Thanks for this. Reading others comments, I agree people should just use pyproject.toml black config to exclude directories rather than the CLI. Here is an example of how we even do it with black on the black repo: https://github.com/psf/black/blob/main/pyproject.toml#L12

Aren't the entries in `exclude` and in `extend-exclude` ignored by `pre-commit`? And then you now also need to define `force-exclude=exclude + extend-exclude` in your `pyproject.toml`?

PR #3116, opened 2022-06-09
Added option to manually specifiy project root

### LINKED ISSUE #3040: black does not exclude files in git submodules
**Describe the bug**

black does not exclude files in git submodules when full filename is passed

**To Reproduce**

I created an example project with a git submodule to test. The super-project has a pyproject.toml which excludes the subproject:
```
[tool.black]
force-exclude = 'subproject'
```
Commands to clone and reproduce bug:
```bash
git clone --recurse-submodules git@github.com:janjachnik-dyson/black-submodule-test-super.git
cd black-submodule-test-super
black --check subproject/file.py
```
The output:
```
Identified `/tmp/black-submodule-test-super/subproject` as project root containing a .git directory.
Sources to be formatted: "file.py"
subproject/file.py already well formatted, good job.

All done! ✨ 🍰 ✨
```

**Expected behavior**

`subproject/file.py` would be ignored because its path matches the force-exclude regex in pyproject.toml

**Environment**

- Black's version: 22.3.0
- OS and Python version: Ubuntu 20.04, Python 3.8.10

**Additional context**

We use black from within Visual Studio Code, where it is always called with the full filename. If black is manually run on the root folder of the super-project, the excludes are handled correctly. Within VSCode we pass the `--config` argument to specify the correct configuration file of the super-project.

**Possible solution**

The problem occurs because black decides the project root by recursively searching parent directories for `.git`, `.hg` and `pyproject.toml` files. In this case, the project root becomes the subproject folder (due to .git) and the exclude regex is compared against the path relative to the project root, which no longer contains the `subproject` string.

As a possible fix, I added the option to manually specify the project root:
https://github.com/janjachnik-dyson/black/commit/6fc33683f17ecbb689add9c722ba96be0428d78c

Then, the following command works to properly exclude the file:
```
black --project-root . subproject/file.py
```

If this sounds like a valid fix/workaround, I'll raise a PR. Or is there a better way to handle excludes with git submodules?

janjachnik-dyson:
### Description

This is one possible fix for: https://github.com/psf/black/issues/3040

I have added the `--project-root` command line argument which can be used to force black to evaluate paths relative to that root. Otherwise, black will pick up things like `.git` folders in submodules and use the submodule path as the project root instead.

We want to auto-format our code in Visual Studio Code but want to exclude 3rd-party git submodules by specifying them in pyproject.toml. This is needed to make that work.

### Checklist - did you ...


cobaltt7:

PR #3119, opened 2022-06-12
Remove trailing spaces in f-string expressions

saroad2:
### Description

Added a new feature in preview mode: *Black* now knows how to trim trailing whitespaces inside f-strings.

For example the following code:
```python
print(f"there are { 2 } unnecessary spaces in this f string")
print(f"there are {   3} unnecessary spaces on the left")
print(f"there are {3   } unnecessary spaces on the right")
print(f"you should trim { 'me' } and {   'me!'  }")
print(f"trimming is easy as {  1 + 1   } == {2}")
```

will transform to:
```python
print(f"there are {2} unnecessary spaces in this f string")
print(f"there are {3} unnecessary spaces on the left")
print(f"there are {3} unnecessary spaces on the right")
print(f"you should trim {'me'} and {'me!'}")
print(f"trimming is easy as {1 + 1} == {2}")
```

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 - Parentheses around return annotations are now managed (#2990)
 - Remove unnecessary parentheses from `with` statements (#2926)
 - Remove trailing newlines after code block open (#3035)
+- Remove trailing whitespaces inside f-string expressions (#3119)
 
 ### _Blackd_
 
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -74,3 +74,26 @@ def my_func():
 
 This new feature will be applied to **all code blocks**: `def`, `class`, `if`, `for`,
 `while`, `with`, `case` and `match`.
+
+### Removing trailing whitespaces in f-string expressions
+
+_Black_ will remove trailing whitespaces inside f-string expressions. That means that
+the following code:
+
+```python
+print(f"there are { 2 } unnecessary spaces in this f string")
+print(f"there are {   3} unnecessary spaces on the left")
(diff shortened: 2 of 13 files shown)


zsol:
Hey, thanks for taking a stab at this. I have two concerns:
1. This only solves the problem partially. prefix and postfix spaces will be stripped, but the rest of the expression won't be formatted
2. I think the PR is not quite ready yet. Consider `f'ok { expr = !r: aosidjhoi } end'`, or `f'{  bar : { baz } }'`. Both of these cause Black to fail on this PR because of unstable formatting

saroad2:
Hey @zsol .

Thanks for the feedback! I think I've fixed everything you've mentioned, but let me know if I missed something.

JelleZijlstra:
This has a lot of merge conflicts now.

saroad2:
I'm feeling conflicted about this PR.

On the one hand, this PR does what it says it should do. solving the merge conflicts should not be an issue.
On the other hand, I think that a better approach is to finally add f-string parsing to blib2to3. It all seems to much of a hack to me.

If we want to take the second approach, this PR need to be closed and splitted to 2 new PRs:
1. Adding f-string parsing to blib2to3
2. Using this new node to parse elements inside f-string proprly.

What do you think we should do? should we proceed with this PR and do a makeover later, or should we close this PR and do it again, the right way this time?

JelleZijlstra:
I would prefer to do proper full formatting of f-string contents (which doesn't necessarily require blib2to3 changes; we could do it inside black). I don't think there is enough value in piecemeal changes to fix minor aspects of f-string formatting.

saroad2:
Ok.
Closing this PR in favor of a more thorough change.

PR #3136, opened 2022-06-24
Update preview style docs to include recent changes

felix-hilden:
### Description
I went through the commits after the last release to see are the style docs outdated, and here's an update for #2926, #2990, #2991 and #3035. Conveniently, they share a theme, so I wrote quite a terse doc and made another section shorter too. Feel free to elaborate if you see the need, I'll have to catch a train now 😄

### Checklist - did you ...

- [x] N/A Add a CHANGELOG entry if necessary?
- [x] N/A Add / update tests if necessary?

### DIFF
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -50,9 +50,10 @@ limit. Line continuation backslashes are converted into parenthesized strings.
 Unnecessary parentheses are stripped. The stability and status of this feature is
 tracked in [this issue](https://github.com/psf/black/issues/2188).
 
-### Removing trailing newlines after code block open
+### Removing newlines in the beginning of code blocks
 
-_Black_ will remove trailing newlines after code block openings. For example:
+_Black_ will remove newlines in the beginning of new code blocks, i.e. when the
+indentation level is increased. For example:
 
 ```python
 def my_func():
@@ -72,17 +73,45 @@ This new feature will be applied to **all code blocks**: `def`, `class`, `if`, `
 
 ### Improved parentheses management
 
-_Black_ will format parentheses around return annotations, `await` expressions and
-`with` statements similarly to other sets of parentheses. For example:
+_Black_ will format parentheses around return annotations similarly to other sets of
+parentheses. For example:
+
+```python
+def foo() -> (int):
+    ...
+```
+
+will be changed to:
+
+```python
+def foo() -> int:
+    ...
+```
+
+```python
+def foo() -> looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong:
+    ...
+```
+
+will be changed to:
+
+```python
+def foo() -> (
+    looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong
+):
+    ...
+```
+
+Extra parentheses in `await` expressions and `with` statements are removed.
 
 ```python
 with ((open("bla.txt")) as f, open("x")):
-    pass
+    ...
 ```
 
 will be changed to:
 
 ```python
 with open("bla.txt") as f, open("x"):
-    pass
+    ...
 ```
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -52,25 +52,37 @@ tracked in [this issue](https://github.com/psf/black/issues/2188).
 
 ### Removing trailing newlines after code block open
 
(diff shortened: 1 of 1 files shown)


docs/the_black_code_style/future_style.md
```diff
@@ -52,25 +52,37 @@ tracked in [this issue](https://github.com/psf/black/issues/2188).
...
-
-    print("But the line above me won't!")
 ```
 
 This new feature will be applied to **all code blocks**: `def`, `class`, `if`, `for`,
 `while`, `with`, `case` and `match`.
+
+### Improved parentheses management
+
+_Black_ will format parentheses around return annotations, await expressions and with
```
JelleZijlstra, line 75:
```suggestion
_Black_ will format parentheses around return annotations, `await` expressions and `with`
```

docs/the_black_code_style/future_style.md
```diff
@@ -52,25 +52,37 @@ tracked in [this issue](https://github.com/psf/black/issues/2188).
...
 def my_func():
     print("The line above me will be deleted!")
-
-    print("But the line above me won't!")
 ```
 
 This new feature will be applied to **all code blocks**: `def`, `class`, `if`, `for`,
 `while`, `with`, `case` and `match`.
+
+### Improved parentheses management
```
ichard26, line 73: Also probably worth mentioning that long return annotations will now be moved to their own line surrounded by parens (black used to just them alone violating the line length limit ;-;)

---

```python
def foo() -> (int):
    ...
```
will be changed to:
```python
def foo() -> int:
    ...
```
```python
def foo() -> intsdfsafafafdfdsasdfsfsdfasdfafdsafdfdsfasdskdsdsfdsafdsafsdfdasfffsfdsfdsafafhdskfhdsfjdslkfdlfsdkjhsdfjkdshfkljds:
    return 2
```
will be changed to:
```python
def foo() -> (
    intsdfsafafafdfdsasdfsfsdfasdfafdsafdfdsfasdskdsdsfdsafdsafsdfdasfffsfdsfdsafafhdskfhdsfjdslkfdlfsdkjhsdfjkdshfkljds
):
    return 2
```

---

I'm a fan of examples, so I also suggested the return annotation one.

docs/the_black_code_style/future_style.md
```diff
@@ -52,25 +52,37 @@ tracked in [this issue](https://github.com/psf/black/issues/2188).
...
-_Black_ will remove trailing newlines after code block openings. That means that the
-following code:
+_Black_ will remove trailing newlines after code block openings. For example:
 
 ```python
 def my_func():
 
     print("The line above me will be deleted!")
-
-    print("But the line above me won't!")
```
ichard26, line 63: I think this still has value since it's not immediately clear what "code block opening" _means_.

docs/the_black_code_style/future_style.md
```diff
@@ -52,25 +52,37 @@ tracked in [this issue](https://github.com/psf/black/issues/2188).
...
 ```
 
 This new feature will be applied to **all code blocks**: `def`, `class`, `if`, `for`,
 `while`, `with`, `case` and `match`.
+
+### Improved parentheses management
+
+_Black_ will format parentheses around return annotations, `await` expressions and
+`with` statements similarly to other sets of parentheses. For example:
```
ichard26, line 76: If we want to be specific, only return annotations have general better management. For `await` and `with` statements , Black only gained the ability to remove redundant parens. A super long context manager [is still formatted badly even though section hints at it being better now](https://github.com/psf/black/pull/2926#issuecomment-1079773063) for example.

ichard26 (commented):
Thanks for tackling this! I think you may have gone overboard with the terseness though :)

docs/the_black_code_style/future_style.md
felix-hilden: Nice suggestion, will do

docs/the_black_code_style/future_style.md
felix-hilden: Good clarification, I agree that we want to be specific here

docs/the_black_code_style/future_style.md
felix-hilden: Good observation. I'll change the wording instead, so that it's more explicit

ichard26 (approved):
It doesn't flow very nicely, but I can't think of a way to improve it so it's fine. I'd appreciate a re-review from Jelle though.

felix-hilden:
This is a pseudo-approval for Rich's changes!

PR #3152, opened 2022-07-03
Visually standardise badges

diksown:
### Description

The *Actions Status* badge (the one that displays Test: passing / Test: failing) had a different left side color from all the other badges. With this commit, a badge from the shields.io website is used, standardising the badge styles.

Comparing the change:

image

### Checklist - did you ...

This is a small visual change on the README, so no further work is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 <h2 align="center">The Uncompromising Code Formatter</h2>
 
 <p align="center">
-<a href="https://github.com/psf/black/actions"><img alt="Actions Status" src="https://github.com/psf/black/workflows/Test/badge.svg"></a>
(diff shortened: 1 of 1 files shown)


cooperlees (commented):
Thanks. I kind of prefer us to use GitHub's official badge. I have more faith in their ability to query their infra's runs more efficiently than a 3rd party. I won't block, but wonder what other maintainers think.

ichard26:
I agree, also this is IMO a bit pointless so there's no good reason to clutter up the file history. Thanks for the PR and the time you put into it, but I vote to reject this PR.

diksown:
Thanks for the feedback about the PR. I'm closing it, given the reasons presented.

Issue #2734, opened 2021-12-31
ESP: a standalone comment causes strings to be merged into one far too long (and requiring two passes to do so)


ichard26:
**To Reproduce**

For example, take this code:

```python
def a():
    CompiledSQL(
        "addresses.id AS addresses_id, "
        "addresses.email_address AS addresses_email_address "
        # note the comma-separated FROM clause
        "IN (__[POSTCOMPILE_primary_keys]) "
        "AND addresses.id = anon_1.id ORDER BY addresses.id",
        ["hmm"]
    )
```

And run it with these arguments:

```sh
$ black file.py --experimental-string-processing
```

The resulting error is:

```
error: cannot format test.py: INTERNAL ERROR: Black produced different code on the second pass of the formatter.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_wqp6guhd.log
Oh no! 💥 💔 💥
1 file failed to reformat.
```

```diff
Mode(target_versions={<TargetVersion.PY36: 6>, <TargetVersion.PY38: 8>, <TargetVersion.PY37: 7>}, line_length=88, string_normalization=True, is_pyi=False, is_ipynb=False, magic_trailing_comma=True, experimental_string_processing=True)
--- source
+++ first pass
@@ -1,9 +1,9 @@
 def a():
     CompiledSQL(
         "addresses.id AS addresses_id, "
         "addresses.email_address AS addresses_email_address "
         # note the comma-separated FROM clause
-        "IN (__[POSTCOMPILE_primary_keys]) "
-        "AND addresses.id = anon_1.id ORDER BY addresses.id",
-        ["hmm"]
+        "IN (__[POSTCOMPILE_primary_keys]) AND addresses.id = anon_1.id ORDER BY"
+        " addresses.id",
+        ["hmm"],
     )
--- first pass
+++ second pass
@@ -1,9 +1,8 @@
 def a():
     CompiledSQL(
         "addresses.id AS addresses_id, "
         "addresses.email_address AS addresses_email_address "
         # note the comma-separated FROM clause
-        "IN (__[POSTCOMPILE_primary_keys]) AND addresses.id = anon_1.id ORDER BY"
-        " addresses.id",
+        "IN (__[POSTCOMPILE_primary_keys]) AND addresses.id = anon_1.id ORDER BY addresses.id",
         ["hmm"],
     )
```

**Expected behavior**

It not crashing :slightly_smiling_face: 

**Environment**

[...]

PR #3162, opened 2022-07-13
Add parens around implicit string concatenations where increases readability

### LINKED ISSUE #3159: Improved string processing: add parentheses around implicit string concatenations to increase readability in certain contexts
**Describe the style change**

When splitting long string literals in _Black_'s `--preview` _Improved string processing_ style, add parentheses around implicit string concatenations to increase readability in certain contexts.

"certain contexts" include:

- Function parameters
- Sequence elements

**Examples of unformatted code**

```python
                    flash(
                        'None of the email addresses or domains you entered are valid',
                        'error',
                    )
```
```python
some_list = [
    ' lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et dolore magna aliqua Ut enim ad minim',
    ' veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo',
]
```

```python
                    flash(
                        'None of the email addresses or domains you entered'
                        ' are valid',
                        'error',
                    )
```
```python
some_list = [
    " lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor"
    " incididunt ut labore et dolore magna aliqua Ut enim ad minim",
    " veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo",
]
```

**Desired style**

```python
                    flash(
                        (
                            'None of the email addresses or domains you entered'
                            ' are valid'
                        ),
                        'error',
                    )
```
```python
some_list = [
    (
        " lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor"
        " incididunt ut labore et dolore magna aliqua Ut enim ad minim"
    ),
    " veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo",
]
```

**Additional context**

This issue was raised originally in https://github.com/psf/black/issues/2188#issuecomment-1009041704. Per @JelleZijlstra, I'm starting a new issue separately here to track the work.

The "certain contexts" list isn't exhaustive. I plan to read the code and play around with an implementation, then report back here.

yilei:
### Description

This PR resolves #3159. It adds parentheses around implicit string concatenations when it's inside a list, set, or tuple.

### Implementation notes

Looking at the order of the transformers [here](https://github.com/psf/black/blob/05b63c4bccbfb292a92d3ec962ce9b8fa4ebcfd5/src/black/linegen.py#L474-L480), we need to "wrap in parens" _before_ `string_split` runs. So my solution is to introduce a "collaboration" between `StringSplitter` and `StringParenWrapper` where the splitter "skips" the split until the wrapper adds the parens (and then the line after the paren is split by `StringSplitter`) in another pass.

I have also considered an alternative approach, where I tried to add a different "string paren wrapper" class, and it runs before `string_split`. Then I found out it requires a different `do_transform` implementation than `StringParenWrapper.do_transform`, since the later assumes it runs _after_ the `delimiter_split` transform. So I stopped researching that route.

### Checklist - did you ...

### DIFF
--- a/tests/data/preview/long_strings.py
+++ b/tests/data/preview/long_strings.py
@@ -289,10 +289,8 @@ x += (
 y = "Short string"
 
 print(
-    (
-        "This is a really long string inside of a print statement with extra arguments"
-        " attached at the end of it."
-    ),
+    "This is a really long string inside of a print statement with extra arguments"
+    " attached at the end of it.",
     x,
     y,
     z,
@@ -455,15 +453,13 @@ bad_split3 = (
 )
 
 bad_split_func1(
-    (
-        "But what should happen when code has already "
-        "been formatted but in the wrong way? Like "
-        "with a space at the end instead of the "
-        "beginning. Or what about when it is split too "
-        "soon? In the case of a split that is too "
-        "short, black will try to honer the custom "
-        "split."
-    ),
+    "But what should happen when code has already "
+    "been formatted but in the wrong way? Like "
+    "with a space at the end instead of the "
+    "beginning. Or what about when it is split too "
+    "soon? In the case of a split that is too "
+    "short, black will try to honer the custom "
+    "split.",
     xxx,
     yyy,
     zzz,
@@ -566,11 +562,9 @@ comment_string = (  # This comment gets thrown to the top.
 )
 
 arg_comment_string = print(
-    (  # This comment gets thrown to the top.
-        "Long lines with inline comments which are apart of (and not the only member"
-        " of) an argument list should have their comments appended to the reformatted"
-        " string's enclosing left parentheses."
-    ),
+    "Long lines with inline comments which are apart of (and not the only member of) an"
+    " argument list should have their comments appended to the reformatted string's"
+    " enclosing left parentheses.",  # This comment gets thrown to the top.
     "Arg #2",
     "Arg #3",
     "Arg #4",
@@ -630,31 +624,23 @@ return (
 )
 
 func_with_bad_comma(
-    (
-        "This is a really long string argument to a function that has a trailing comma"
-        " which should NOT be there."
-    ),
+    "This is a really long string argument to a function that has a trailing comma"
+    " which should NOT be there.",
 )
 
 func_with_bad_comma(
-    (  # comment after comma
-        "This is a really long string argument to a function that has a trailing comma"
-        " which should NOT be there."
-    ),
+    "This is a really long string argument to a function that has a trailing comma"
+    " which should NOT be there.",  # comment after comma
 )
 
 func_with_bad_comma(
-    (
-        "This is a really long string argument to a function that has a trailing comma"
-        " which should NOT be there."
-    ),
+    "This is a really long string argument to a function that has a trailing comma"
+    " which should NOT be there.",
 )
 
 func_with_bad_comma(
-    (  # comment after comma
-        "This is a really long string argument to a function that has a trailing comma"
-        " which should NOT be there."
-    ),
+    "This is a really long string argument to a function that has a trailing comma"
+    " which should NOT be there.",  # comment after comma
 )
 
 func_with_bad_parens_that_wont_fit_in_one_line(
--- a/tests/data/preview/long_strings.py
+++ b/tests/data/preview/long_strings.py
@@ -18,6 +18,14 @@ D3 = {x: "This is a really long string that can't possibly be expected to fit al
 
 D4 = {"A long and ridiculous {}".format(string_key): "This is a really really really long string that has to go i,side of a dictionary. It is soooo bad.", some_func("calling", "some", "stuff"): "This is a really really really long string that has to go inside of a dictionary. It is {soooo} bad (#{x}).".format(sooo="soooo", x=2), "A %s %s" % ("formatted", "string"): "This is a really really really long string that has to go inside of a dictionary. It is %s bad (#%d)." % ("soooo", 2)}
 
+L1 = ["The is a short string", "This is a really long string that can't possibly be expected to fit all together on one line. Also it is inside a list literal, so it's expected to be wrapped in parens when spliting to avoid implicit str concatenation.", short_call("arg", {"key": "value"}), "This is another really really (not really) long string that also can't be expected to fit on one line and is, like the other string, inside a list literal.", ("parens should be stripped for short string in list")]
+
+L2 = ["This is a really long string that can't be expected to fit in one line and is the only child of a list literal."]
+
+S1 = {"The is a short string", "This is a really long string that can't possibly be expected to fit all together on one line. Also it is inside a set literal, so it's expected to be wrapped in parens when spliting to avoid implicit str concatenation.", short_call("arg", {"key": "value"}), "This is another really really (not really) long string that also can't be expected to fit on one line and is, like the other string, inside a set literal.", ("parens should be stripped for short string in set")}
+
+S2 = {"This is a really long string that can't be expected to fit in one line and is the only child of a set literal."}
+
 func_with_keywords(my_arg, my_kwarg="Long keyword strings also need to be wrapped, but they will probably need to be handled a little bit differently.")
 
 bad_split1 = (
@@ -109,7 +117,7 @@ fstring_with_no_fexprs = f"Some regular string that needs to get split certainly
 
 comment_string = "Long lines with inline comments should have their comments appended to the reformatted string's enclosing right parentheses."  # This comment gets thrown to the top.
 
-arg_comment_string = print("Long lines with inline comments which are apart of (and not the only member of) an argument list should have their comments appended to the reformatted string's enclosing left parentheses.",  # This comment stays on the bottom.
+arg_comment_string = print("Long lines with inline comments which are apart of (and not the only member of) an argument list should have their comments appended to the reformatted string's enclosing left parentheses.",  # This comment gets thrown to the top.
     "Arg #2", "Arg #3", "Arg #4", "Arg #5")
 
 pragma_comment_string1 = "Lines which end with an inline pragma comment of the form `# <pragma>: <...>` should be left alone."  # noqa: E501
@@ -277,8 +285,10 @@ x += (
 y = "Short string"
 
 print(
-    "This is a really long string inside of a print statement with extra arguments"
-    " attached at the end of it.",
+    (
+        "This is a really long string inside of a print statement with extra arguments"
+        " attached at the end of it."
+    ),
     x,
     y,
     z,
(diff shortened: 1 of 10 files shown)


yilei:
I scanned through the diff-shades results, they all look expected to this PR's change.

jack1142:
> I scanned through the diff-shades results, they all look expected to this PR's change.

You have to put it behind the `--preview` flag.

yilei:
@jack1142 The changed code is already behind the `--preview` flag (`if Preview.string_processing in mode:`), the [diff from `assert-no-changes`](https://github.com/psf/black/runs/7312264292?check_suite_focus=true#step:10:1) is from https://github.com/ichard26/blackbench, which [enables it](https://github.com/ichard26/blackbench/blob/1e4e36b8de4ba351a95f4948866620bada64cf53/pyproject.toml#L38) via the (deprecated) `--experimental-string-processing` flag.

I assume this check doesn't force it to be false, or am I missing something?

ichard26:
Gah, I missed that >.<

Instead of fixing diff-shades, I'll just push a quick update to blackbench to switch to the `preview` option 'cause I maintain all of the black devtooling :P

yilei:
Thank you @ichard26!

Adding the new `--preview` diff report link for convenience: https://github.com/psf/black/runs/7313506857?check_suite_focus=true#step:10:1

tests/data/preview/long_strings.py
```diff
@@ -512,23 +568,31 @@ def foo():
...
+    (
+        "This is a really long string argument to a function that has a trailing comma"
+        " which should NOT be there."
+    ),
 )
 
 func_with_bad_comma(
-    "This is a really long string argument to a function that has a trailing comma"
-    " which should NOT be there.",  # comment after comma
+    (  # comment after comma
```
JelleZijlstra, line 592: hm, it seems like this comment gets moved to the beginning? That's unfortunate.

yilei: yeah, i'm actually not sure where the comment should go. there is another existing example:

https://github.com/psf/black/blob/1b6de7b0a33b568f71ff86e0e5fef6d4c479c2b7/tests/data/preview/long_strings.py#L91

https://github.com/psf/black/blob/1b6de7b0a33b568f71ff86e0e5fef6d4c479c2b7/tests/data/preview/long_strings.py#L447-L450

that throws the comment to the opening paren, which is the same behavior here.

looks like it was introduced in https://github.com/psf/black/pull/1132, but it's unclear whether this behavior is intentionally designed or not.

thoughts?

JelleZijlstra: I see, thanks. Maybe we can fix this, but seems like the issue is independent from this PR.

JelleZijlstra (approved):

ichard26 (changes requested):
I agree that this improves formatting. The code looks mostly good, but noticed few issues: 

- Tuple literals are not considered, that seems unconsistent
- Wrapping broken up string literals that live alone in a sequence literal that simply has a trailing comma seems overkill?

```diff
--- test.py	2022-08-03 17:43:27.228805 +0000
+++ test.py	2022-08-03 17:43:28.373631 +0000
@@ -1,8 +1,10 @@
 L2 = [
-    "This is a really long string that can't be expected to fit in one line and is the "
-    "only child of a list literal.",
+    (
+        "This is a really long string that can't be expected to fit in one line and is"
+        " the only child of a list literal."
+    ),
 ]
```

```diff
--- a/django:django/db/backends/mysql/features.py
+++ b/django:django/db/backends/mysql/features.py
@@ -122,12 +132,14 @@
             skips.update(
                 {
                     "https://jira.mariadb.org/browse/MDEV-19598": {
-                        "schema.tests.SchemaTests."
-                        "test_alter_not_unique_field_to_primary_key",
+                        (
+                            "schema.tests.SchemaTests."
+                            "test_alter_not_unique_field_to_primary_key"
+                        ),
                     },
                 }
             )
```
- It's wrapping broken up strings literals incorrectly when there's a comment in the middle, breaking valid code

```python
CompiledSQL(
    "SELECT addresses.user_id AS addresses_user_id, "
    "addresses.email_address AS addresses_email_address "
    # note the comma-separated FROM clause
    "FROM addresses, (SELECT addresses_1.id AS id FROM "
    "AND addresses.id = anon_1.id ORDER BY addresses.id",
    arg_func(),
)
```

```diff
--- test.py	2022-08-03 18:54:22.054128 +0000
+++ test.py	2022-08-03 18:57:45.657305 +0000
@@ -1,8 +1,9 @@
 CompiledSQL(
-    "SELECT addresses.user_id AS addresses_user_id, "
-    "addresses.email_address AS addresses_email_address "
+    (
+        "SELECT addresses.user_id AS addresses_user_id, "
+        "addresses.email_address AS addresses_email_address "
+    )
     # note the comma-separated FROM clause
-    "FROM addresses, (SELECT addresses_1.id AS id FROM "
-    "AND addresses.id = anon_1.id ORDER BY addresses.id",
+    "FROM addresses, (SELECT addresses_1.id AS id FROM AND addresses.id = anon_1.id ORDER BY addresses.id",
     arg_func(),
 )
```
(note that due to the forced second pass in `format_str`, the error output from Black is less than helpful, it claims the input is invalid, but really the input to the second pass was invalid because the first pass made it invalid ...)

JelleZijlstra:
> * Wrapping broken up string literals that live alone in a sequence literal that simply has a trailing comma seems overkill?

I think the parentheses are actually helpful here, because otherwise the code looks like a two-element list.

yilei:
quick status update: i'm currently on vacation. will address the comments in about two weeks. thanks for pointing out the issues!

ichard26:
enjoy your vacation! :beach_umbrella:

yilei:
looking at a fix for the last "comment in the middle of strings" issue, to actually wrap the entire string + comment in parens is really hard thing to do given the complexity of `STANDALONE_COMMENT`.

however, i think one reasonable option is to skip wrapping in parens for this case. to do that, the code change required seems to also fix the mentioned https://github.com/psf/black/issues/2734. so, i'll send a separate PR to fix https://github.com/psf/black/issues/2734 first.

yilei:
@ichard26 I've addressed your comments, please take another look:

> Tuple literals are not considered, that seems unconsistent

Tuples are now considered too

> Wrapping broken up string literals that live alone in a sequence literal that simply has a trailing comma seems overkill?

Working as intended per https://github.com/psf/black/pull/3162#issuecomment-1204392015

> It's wrapping broken up strings literals incorrectly when there's a comment in the middle, breaking valid code

This PR doesn't wrap those cases in parens, and the original incorrect results were fixed by https://github.com/psf/black/pull/3227

tests/data/preview/long_strings.py
```diff
@@ -345,6 +359,71 @@ def foo():
...
+        " expected to fit on one line and is, like the other string, inside a list"
+        " literal."
+    ),
+    "parens should be stripped for short string in list",
+]
+
+L2 = [
+    "This is a really long string that can't be expected to fit in one line and is the"
+    " only child of a list literal."
+]
```
ichard26, line 381: I would've thought this would be wrapped too, but it isn't because there isn't a trailing comma. Not sure if that's intentional or simply an oversight.

ichard26 (commented):
The amount of changes diff-shades reports for this PR scares me, especially since I'm not entirely sure if it's really an improvement in situations like these:
```diff
--- a/warehouse:warehouse/packaging/services.py
+++ b/warehouse:warehouse/packaging/services.py
@@ -33,13 +33,15 @@
         # This class should not be used in production, it's trivial for it to
         # be used to read arbitrary files from the disk. It is intended ONLY
         # for local development with trusted users. To make this clear, we'll
         # raise a warning.
         warnings.warn(
-            "LocalFileStorage is intended only for use in development, you "
-            "should not use it in production due to the lack of safe guards "
-            "for safely locating files on disk.",
+            (
+                "LocalFileStorage is intended only for use in development, you "
+                "should not use it in production due to the lack of safe guards "
+                "for safely locating files on disk."
+            ),
             InsecureStorageWarning,
         )
```
To be fair, the extra parenthesises and indention are mostly fine, but seem redundant. Probably doesn't make much sense to add even more complexity to the "certain contexts" part of this formatting rule though to keep stuff like this unchanged. Checking whether the string's previous (previous, cause the comma) or next (next) sibling is also a string sounds incredibly fragile.

tests/data/preview/long_strings.py
yilei: yes this is intentional as implemented in this PR.

yilei:
Agreed that this is causing scary amount of diffs.

Going back to the original intention of this PR, it tries to avoid bugs for implicit string concatenations that only commonly happen in lists/tuples/sets. As it's quite easy to miss commas:

```
some_list = [
    " lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor"
    " incididunt ut labore et dolore magna aliqua Ut enim ad minim",
    " veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo",
]
```

In function calls, this is less of an issue because missing commas should usually result in an immediate runtime error because of argument mis-match.

So, another options is to limit this PR to only list/tuple/set literals and exclude function calls? What do you think, @ichard26 @JelleZijlstra ?

JelleZijlstra:
> So, another option is to limit this PR to only list/tuple/set literals and exclude function calls? What do you think, @ichard26 @JelleZijlstra ?

Yes, I think that's a good idea. Maybe also dicts? Something like
```python
{
    "x": "some"
    "long string",
}
```
is more readable as
```python
{
    "x": (
        "some"
        "long string"
    ),
}
```

JelleZijlstra:
> So, another option is to limit this PR to only list/tuple/set literals and exclude function calls? What do you think, @ichard26 @JelleZijlstra ?

Yes, I think that's a good idea. Maybe also dicts? Something like
```python
{
    "x": "some"
    "long string",
}
```
is more readable as
```python
{
    "x": (
        "some"
        "long string"
    ),
}
```

yilei:
> Yes, I think that's a good idea

Great, I pushed new commits now this PR no longer wraps function call args.

>  Maybe also dicts?

You meant the wrapped example is more readable right? This is the value in dicts, which is already done by the current version of `black --preview`: https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADvAHJdAD2IimZxl1N_Wm_vzTQ-P7e9kejh_LvEiUE6tLUAZ1xVz5lZ6m-wo9D4ahShh_wm_7SxyQe1Rl-WZPSnzPI_DUNfuxeWB-Y41DUdrRGpmMMlfFExIzlSN1HqyMjBQ4vDQ-0a1ylSwHDuAyj3rp34nKMCAAAAAB6xJMz7xPYdAAGOAfABAABV9CtdscRn-wIAAAAABFla

ichard26 (approved):
Looks way better!

Still iffy about stuff like this:
```python
django_test_skips = {
    "opclasses are PostgreSQL only.": {
        (
            "indexes.tests.SchemaIndexesNotPostgreSQLTests."
            "test_create_index_ignores_opclasses"
        ),
    },
}
```
but as long as you can remove the trailing comma and Black collapses the wrapping parentheses, it's not a big problem. 

Thank you so much for working and sticking with this through the multiple rounds of review! Cheers to another merged PR! :tada:

yilei:
Thank _you_ for the thorough review and accepting my PR, I really appreciate it!

PR #3167, opened 2022-07-14
update on warning message

Panther-12:
### Description
I added a warning message just below the first paragraph noting that `blackd` should not run as a publicly accessible server.

### Checklist - did you ...

- [yes ] Add a CHANGELOG entry if necessary?
- [yes ] Add / update tests if necessary?
- [yes ] Add new / update outdated documentation?

### DIFF
--- a/docs/usage_and_configuration/black_as_a_server.md
+++ b/docs/usage_and_configuration/black_as_a_server.md
@@ -4,6 +4,11 @@
 protocol. The main benefit of using it is to avoid the cost of starting up a new _Black_
 process every time you want to blacken a file.
 
+```{warning}
+`blackd` should not be run as a publically accessible server as there are no security
+precautions in place to prevent abuse. **It is intended for local use only**.
+```
+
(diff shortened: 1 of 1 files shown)


ichard26:
A few more things:

The PR / commit title could be better, I'd suggest "Add warning not run blackd publicly in docs" or "Warn to not run blackd publicly in docs". A clearer title just makes things easier, especially when you come back to a PR / commit after _a while_ and forgot most if not all the context behind it. 

You can use special syntax in your PR description to link an issue. Example: #3167. If there's a relevant issue, you should always reference it to give reviewers more context to why the PR exists and what it changes. Also, if you add a special keyword (see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue for more) you can have the issue automatically closed when the PR is merged!

Panther-12:
Thanks for the insights. I'll be sure to keep that in mind.

PR #3168, opened 2022-07-14
Actually disable docstring prefix normalization with -S + fix instability

ichard26:
### Description
    
The former was a regression I introduced a long time ago. To avoid changing the stable style too much, the regression is only fixed if `--preview` is enabled.
   
*annoying enough, as we currently always enforce a second format pass if changes were made, there's no good way to prove the existence of the docstring quote normalization instability issue. For posterity, here's one failing example:

```diff
--- source
+++ first pass
@@ -1,7 +1,7 @@
 def some_function(self):
-    ''''<text here>
+    """ '<text here>
 
     <text here, since without another non-empty line black is stable>
 
-    '''
+    """
     pass
--- first pass
+++ second pass
@@ -1,7 +1,7 @@
 def some_function(self):
-    """ '<text here>
+    """'<text here>
 
     <text here, since without another non-empty line black is stable>
 
     """
     pass
```

### Checklist - did you ...

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -295,8 +295,8 @@ class LineGenerator(Visitor[Line]):
             # indentation of those changes the AST representation of the code.
             if Preview.normalize_docstring_quotes_and_prefixes_properly in self.mode:
                 # There was a bug where --skip-string-normalization wouldn't stop us
-                # from normalizing docstring prefixes. To maintain stability, we can only
-                # address this buggy behaviour while the preview style is enabled.
+                # from normalizing docstring prefixes. To maintain stability, we can
+                # only address this buggy behaviour while the preview style is enabled.
                 if self.mode.string_normalization:
                     docstring = normalize_string_prefix(leaf.value)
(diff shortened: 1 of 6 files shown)


src/black/linegen.py
```diff
@@ -293,7 +293,24 @@ def visit_STRING(self, leaf: Leaf) -> Iterator[Line]:
         if is_docstring(leaf) and "\\\n" not in leaf.value:
             # We're ignoring docstrings with backslash newline escapes because changing
             # indentation of those changes the AST representation of the code.
-            docstring = normalize_string_prefix(leaf.value)
+            if Preview.normalize_docstring_quotes_and_prefixes_properly in self.mode:
+                # There was a bug where --skip-string-normalization wouldn't stop us
+                # from normalize docstring prefixes. To maintain stability, we can only
```
JelleZijlstra, line 298:
```suggestion
                # from normalizing docstring prefixes. To maintain stability, we can only
```

JelleZijlstra (approved):

PR #3170, opened 2022-07-14
Fix typo in black docs for --extend-exclude

onescriptkid:
### Description

The black config docs provide an example for configuring `--extend-exclude` in the `pyproject.toml` as shown below.

```
extend-exclude = '''
# A regex preceded with ^/ will apply only to files and directories
# in the root of the project.
(
  ^/foo.py    # exclude a file named foo.py in the root of the project
  | *_pb2.py  # exclude autogenerated Protocol Buffer files anywhere in the project
)
'''
```

Using the configuration *verbatim* results in the following error

```
black --check --verbose .

Usage: black [OPTIONS] SRC ...
Try 'black -h' for help.

Error: Invalid value for '--extend-exclude': Not a valid regular expression: nothing to repeat at position 179 (line 5, column 5)
```

Replacing `| *_pb2.py` with `| *._pb2.py` fixes the regex complaint.

Checkout the [example](https://github.com/onescriptkid/test_black) to reproduce step-by-step.
You should see a similar result.

```
black --check --verbose .
...
subdir/should_be_skipped_pb2.py ignored: matches the --extend-exclude regular expression
should_be_skipped_pb2.py ignored: matches the --extend-exclude regular expression
would reformat should_be_reformatted.py
would reformat subdir/should_be_reformatted.py

Oh no! 💥 💔 💥
2 files would be reformatted.
```

### Checklist - did you ...

_A CHANGELOG entry would be overkill, but I'm not opposed._
_Nope, it's a change to only a markdown file._
_Yup, that's the goal of this PR._

### Specs

My development environment is
- _Ubuntu 20.04_
- _python 3.8.10_
- _black 22.6.0_


ichard26:
Thank you so much for your contribution! This project is only possible by contributions like these :black_heart:. You're awesome, @onescriptkid. I don't think I would've ever noticed this haha :p 

If you got any feedback to share on your experience contributing here, let us know via #2238!

PR #3175, opened 2022-07-16
filterwarnings=["errors", ... followup

graingert:
### Description

document filtered warnings that cannot be removed easily and move invalid escape sequence DeprecationWarning to those tests that need it

### Checklist - did you ...

### DIFF
--- a/tests/test_format.py
+++ b/tests/test_format.py
@@ -28,6 +28,7 @@ def check_file(
     assert_format(source, expected, mode, fast=False)
 
 
+@pytest.mark.filterwarnings("ignore:invalid escape sequence.*:DeprecationWarning")
 @pytest.mark.parametrize("filename", all_data_cases("simple_cases"))
 def test_simple_format(filename: str) -> None:
     check_file("simple_cases", filename, DEFAULT_MODE)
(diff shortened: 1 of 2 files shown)


tests/test_format.py
```diff
@@ -28,6 +28,7 @@ def check_file(
     assert_format(source, expected, mode, fast=False)
 
 
+@pytest.mark.filterwarnings("ignore:invalid escape sequence.*:DeprecationWarning")
```
graingert, line 31: this is caused by this example test case:

https://github.com/psf/black/blob/e026701b9a894ef19ad59eb6e394e5f3f7c18fdd/tests/data/simple_cases/docstring.py#L163

at some point a future python version will cause this test to fail

tests/test_format.py
```diff
@@ -132,6 +133,7 @@ def test_python_2_hint() -> None:
     exc_info.match(black.parsing.PY2_HINT)
 
 
+@pytest.mark.filterwarnings("ignore:invalid escape sequence.*:DeprecationWarning")
```
graingert, line 136: this is caused by this example test case:

https://github.com/psf/black/blob/e026701b9a894ef19ad59eb6e394e5f3f7c18fdd/tests/data/miscellaneous/docstring_no_string_normalization.py#L107

at some point a future python version will cause this test to fail

tests/test_format.py
graingert: or maybe it will just be a warning forever ;) ? https://github.com/python/cpython/issues/77093

PR #2941, opened 2022-03-18
Add ISort

Shivansh-007:
### Description

Closes #2930 

The formatting changes are part of commit 37356f6940773f0903486da9629b68bd78d47209 and edf37bf708006f12fc7d3cab43c21f19548b3035, the rest are config/chore changes.

### Checklist - did you ...
- [ ] Add a CHANGELOG entry if necessary?
- [X] Add / update tests if necessary? -> n/a
- [X] Add new / update outdated documentation? -> n/a

PR #3188, opened 2022-07-27
Add isort to linting toolchain

### LINKED ISSUE #2930: Addition of `isort` to our dev toolchain
Currently, the imports exist in any random order as added by the developers (see below), which sometimes makes it difficult to look through them in search of something. Furthermore, it is not very pretty or easy to read.

https://github.com/psf/black/blob/086ae68076de570b0cb1881a3c3b9da592b46ee0/src/black/__init__.py#L52-L59
https://github.com/psf/black/blob/086ae68076de570b0cb1881a3c3b9da592b46ee0/src/black/__init__.py#L6-L11

This is where `isort` comes in, `isort`, unlike `black` is a fantastic hyper configurable tool that will magically sort your imports so you don't have to care about them. It also allows you to format them while agreeing with the `black` style guide using [Black Profile](https://pycqa.github.io/isort/docs/configuration/profiles.html)

An experimental PR containing changes brought by isort could be made, and we could then decide how to proceed after reviewing the diff.

ichard26:
### Description

Resolves #2930 and closes #2941.

### Checklist - did you ...

### DIFF
--- a/src/black/nodes.py
+++ b/src/black/nodes.py
@@ -3,16 +3,7 @@ blib2to3 Node/Leaf transformation-related utility functions.
 """
 
 import sys
-from typing import (
-    Generic,
-    Iterator,
-    List,
-    Optional,
-    Set,
-    Tuple,
-    TypeVar,
-    Union,
-)
+from typing import Generic, Iterator, List, Optional, Set, Tuple, TypeVar, Union
 
 if sys.version_info >= (3, 8):
     from typing import Final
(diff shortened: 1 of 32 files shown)


cooperlees (approved):
Thanks. Sounds good to me.

src/black/nodes.py
```diff
@@ -25,14 +16,13 @@
...
-from blib2to3.pytree import Node, Leaf, type_repr, NL
-from blib2to3 import pygram
-from blib2to3.pgen2 import token
-
 from black.cache import CACHE_DIR
 from black.strings import has_triple_quotes
+from blib2to3 import pygram
+from blib2to3.pgen2 import token
 
+# lib2to3 fork
```
wookie184, line 24: Did this comment cause the next line to be in it's own section? It seems like it would make more sense above all blib2to3 imports.

Note there are a couple of other occurrences of this comment in different files that may need looking at. They could possibly all just be removed though tbh.

isidentical (approved):
Nice!

src/black/nodes.py
ichard26: I agree, I've removed them.

wookie184 (approved):

ichard26:
Thanks for the review folks! :black_heart:

PR #3017, opened 2022-04-15
 Consolidate test and release workflows + add concurrency limit to test + fuzz

ichard26:
### Description

This makes changes across workflow domains easier (like adding concurrency limits) and overall improves consistency. I plan to submit another PR integrating the mypyc workflow from ichard26/black-mypyc-wheels after this is merged. Here's an example run: https://github.com/ichard26/black/actions/runs/2144614522 (including the mypyc patch)

This is probably best landed with a rebase merge fyi.

### Checklist - did you ...

- [x] Add a CHANGELOG entry if necessary? -> n/a
- [x] Add / update tests if necessary? -> n/a
- [x] Add new / update outdated documentation?

PR #3189, opened 2022-07-27
Consolidate test CI and add concurrency limits

ichard26:
### Description

This comes from https://github.com/psf/black/pull/3017 which in hindsight was making too many changes at once. This should be easier to review, hopefully :)

This is not required for the upcoming 22.7.0 release, this is just the easiest part of #3017 to file separately. There's no rush to merge this.

### Checklist - did you ...


.github/workflows/lint.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set up Python
+      - name: Set up latest Python
```
cooperlees, line 19:
```suggestion
      - name: Set up latest actions/setup-python Python
```

.github/workflows/test.yml
```diff
@@ -66,17 +67,40 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - name: Coveralls finished
+      - name: Send finished signal to Coveralls
         uses: AndreMiras/coveralls-python-action@v20201129
         with:
           parallel-finished: true
           debug: true
+
+  uvloop:
```
cooperlees, line 82: So does this run in parallel out of the one .yml file? TIL. Nice.

cooperlees (approved):
All LGTM and I see no reason to merge.

.github/workflows/test.yml
ichard26: Yup, runs in parallel with the main test jobs just like how it did before when it was a separate workflow

.github/workflows/lint.yml
ichard26: Don't think this is necessary since it's IMO obvious we're going to be using actions/setup-python to set up Python. I know you want it to be explicit the latest Python is coming from that action, but it seems overly explicit.

cooperlees: It just hurts my OCD as there could be a later stable version of python released, but the action could not be using it ... That is all.

Issue #3185, opened 2022-07-25
Vim plugin: prefix errors with "Black"


tobast:
**Is your feature request related to a problem? Please describe.**

When Black is run automatically as a Vim plugin, eg. upon save, and an error occurs, it is not clear that the error comes from Black.

For instance, when saving a Python file using the `match ...:` new syntax with an old version of Black, vim logs

```
Cannot parse: 282:14:         match expr:
```

(this can be reproduced with `:Black target_version=py38` with an up-to-date Black)

When Black is run automatically, it is completely unclear that the error comes from Black, making it hard to understand that Black should be updated.

**Describe the solution you'd like**

When Black logs an error under Vim, prefix its output by eg. `Black:`.

**Additional context**

Versions:
```
Vim: 9.0 (2022 Jun 28, compiled Jun 28 2022 16:22:51), archlinux
Black: 22.6.0 from pip
```

ichard26:
Sounds reasonable to me. I'll note that no one active on the core team can write vimscript so we'd depend on an external contribution for this. Thanks for the suggestion!

tobast:
I'm not fluent in vimscript either, but this turned out to by Python in the end :) Here's a very simple PR.

ichard26:
Whoops, we forgot to link GH-3194 to this issue. Closing as it's been resolved :)

PR #3194, opened 2022-07-28
Vim: prefix messages with "Black:"

tobast:
### Description

As mentioned in #3185, when using Black as a Vim plugin, especially automatically on save, the plugin's messages can be confusing, as nothing indicates that they come from Black.

This PR simply prefixes the messages with "Black: ".

### Checklist - did you ...

- [X] Add / update tests if necessary?  [no changes needed]
- [X] Add new / update outdated documentation?  [no changes needed]

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,6 +52,7 @@
 
 - Change from deprecated `asyncio.get_event_loop()` to create our event loop which
   removes DeprecationWarning (#3164)
+- Prefix Black messages in Vim with `Black: ` (#3194)
 
 ### Packaging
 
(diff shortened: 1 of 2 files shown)


ichard26 (approved):
Do we want to prefix the installation messages with `Black: ` too, or is that overkill? Either way, thanks for the PR!

tobast:
I would guess that it would be overkill, as usually the user explicitly runs the installation commands, and they know that logged messages come from Black. I might be wrong, however.

ichard26:
I agree. Congrats on your first contribution here, thanks again!

PR #3197, opened 2022-07-30
Port & upstream mypyc wheel build workflow 

ichard26:
### Description

This was the last part of #3017. I say was because I rewrote most of the build setup in an effort to modernize it. 

- Test run: https://github.com/ichard26/black/actions/runs/2785946747
- Test run on TestPyPI: https://test.pypi.org/project/black-mypyc/2022.8.2a2/

I'll update the release documentation in a separate PR with the changes to both the release CD and release flow. 

### Checklist - did you ...

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -1,4 +1,4 @@
-name: pypi_upload
+name: Publish to PyPI
 
 on:
   release:
@@ -8,14 +8,14 @@ permissions:
   contents: read
 
 jobs:
-  build:
-    name: PyPI Upload
+  main:
+    name: sdist + pure wheel
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set up Python
+      - name: Set up latest Python
         uses: actions/setup-python@v4
         with:
           python-version: "*"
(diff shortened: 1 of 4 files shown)


.github/workflows/pypi_upload.yml
```diff
@@ -26,11 +26,93 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-20.04
+            name: linux-x86_64
+            # Black needs clang to compile successfully on Linux. -g0 disables debug symbols
+            # which cuts down the wheel size a fair bit!
+            extra_environment: "CC=clang CFLAGS='-g0 -v'"
+            extra_before_build: "&& yum install -y clang"
```
JelleZijlstra, line 47: I'm surprised this works, I don't think `yum` is an Ubuntu thing.

.github/workflows/pypi_upload.yml
```diff
@@ -26,11 +26,93 @@ jobs:
...
+          python -m pip install --upgrade --disable-pip-version-check pip
+          python -m pip install --upgrade twine
+
+      - name: Build wheels via cibuildwheel
+        uses: pypa/cibuildwheel@v2.3.0
+        env:
+          # So these are the environments we are targeting:
+          # - CPython 3.6 and higher (no PyPy)
+          # - 64-bit only
+          # - ARCH: amd64, x86_64, universal2, and arm64
```
JelleZijlstra, line 94: Aren't amd64 and x86_64 the same thing?

ichard26: Yes, I should use a slash instead.

.github/workflows/pypi_upload.yml
ichard26: Anyone serious about building wheels for Linux will use the [manylinux](https://github.com/pypa/manylinux/) docker images to help maximize compatibility. cibuildwheel uses these images which are based off Red Hat IIRC.

ichard26:
For the record, I've verified the wheels built by this new setup behave identically to the pure wheel (via diff-shades) on Ubuntu 20.04.04 for all Python versions we support. @zsol demonstrated that the arm64 wheels don't crash at least which is good. Windows is still relatively untested but I'm not particularly concerned. If I have the time, I'll do a diff-shades run on Windows too.

JelleZijlstra (approved):

Issue #3176, opened 2022-07-17
Upcoming aiohttp 4.0 is breaking our test suite with a new deprecation warning


ichard26:
See https://github.com/ichard26/black-deps-ci/runs/7374216061?check_suite_focus=true. We can either bump the lower bound requirement for aiohttp, or we can reuse the approach taken in https://github.com/psf/black/pull/2974/ for a similar issue. I'll note that aiohttp 4.0 is still under development so breakage will come later, but eventually :)

cc @zsol we should probably configure pytest to _not_ error out on warnings when running under mypy_primer since it'd be pointless busywork and noise for them.

graingert:
> cc @zsol we should probably configure pytest to not error out on warnings when running under mypy_primer since it'd be pointless busywork and noise for them.

Does mypy_primer run the black test suite?

ichard26:
@graingert There's a PR to make it do so: https://github.com/hauntsaninja/mypy_primer/pull/41

shivamdurgbuns:
So @ichard26 what is required to fix this, I can work on it.

graingert:
@shivamdurgbuns simplist fix is to append the relevant ignore to filterwarnings https://github.com/psf/black/blob/main/pyproject.toml#L51

shivamdurgbuns:
@graingert Understood! Will do accordingly.

PR #3198, opened 2022-07-30
Added depreciated warning in filter warning

shivamdurgbuns:
Signed-off-by: Shivam Durgbuns <shivamdurgbuns@gmail.com>
fixes: https://github.com/psf/black/issues/3176

### Checklist - did you ...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,8 +102,6 @@ markers = [
 xfail_strict = true
 filterwarnings = [
     "error",
-    # this ignore can be removed when @middleware is removed from the codebase.
-    '''ignore:Middleware decorator is deprecated since 4\.0 and its behaviour is default, you can simply remove this decorator:DeprecationWarning''',
     # this is mitigated by a try/catch in https://github.com/psf/black/pull/2974/
     # this ignore can be removed when support for aiohttp 3.7 is dropped.
     '''ignore:Decorator `@unittest_run_loop` is no longer needed in aiohttp 3\.8\+:DeprecationWarning''',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,8 +102,7 @@ markers = [
 xfail_strict = true
 filterwarnings = [
     "error",
-    # this is mitigated because Middleware decorator is deprecated in aiohttp since 4.10
-    # this ignore can be removed when Middleware decorator is removed.
+    # this ignore can be removed when @middleware is removed from the codebase.
     '''ignore:Middleware decorator is deprecated since 4\.0 and its behaviour is default, you can simply remove this decorator:DeprecationWarning''',
     # this is mitigated by a try/catch in https://github.com/psf/black/pull/2974/
     # this ignore can be removed when support for aiohttp 3.7 is dropped.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,9 @@ markers = [
 xfail_strict = true
 filterwarnings = [
     "error",
+    # this is mitigated because Middleware decorator is deprecated in aiohttp since 4.10
+    # this ignore can be removed when Middleware decorator is removed.
+    '''ignore:Middleware decorator is deprecated since 4\.0 and its behaviour is default, you can simply remove this decorator:DeprecationWarning''',
     # this is mitigated by a try/catch in https://github.com/psf/black/pull/2974/
     # this ignore can be removed when support for aiohttp 3.7 is dropped.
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -50,6 +50,9 @@ markers = [
 xfail_strict = true
 filterwarnings = [
     "error",
+    # this is mitigated because Middleware decorator is deprecated in aiohttp since 4.10
+    # this ignore can be removed when Middleware decorator is removed.
```
ichard26, line 54:
```suggestion
    # this ignore can be removed when @middleware is removed from the codebase.
```

ichard26 (commented):
There is no mitigation here, it's simply ignoring the issue. We might want to also try except this decorator though similar to https://github.com/psf/black/pull/2974 so when aiohttp does remove this decorator, nothing breaks :)

Thoughts @graingert ?

graingert:
> There is no mitigation here, it's simply ignoring the issue. We might want to also try except this decorator though similar to #2974 so when aiohttp does remove this decorator, nothing breaks :)
> 
> Thoughts @graingert ?

Yeah you need to mitigate this with a try/catch ImportError

pyproject.toml
```diff
@@ -50,6 +50,8 @@ markers = [
 xfail_strict = true
 filterwarnings = [
     "error",
+    # this ignore can be removed when @middleware is removed from the codebase.
```
graingert, line 53: My intention was that you'd apply the mitigation rather than change the comment ;)

shivamdurgbuns: @graingert I am not getting it. You want me to raise Import Error exception on catching Depreciation Warning?

JelleZijlstra:
Sounds like we should not do this and instead actually fix the warning.

pyproject.toml
graingert: I want you to catch the ImportError not raise it

shivamdurgbuns:
> Sounds like we should not do this and instead actually fix the warning.

Has the fix been made? Or should I add an exception for this?

pyproject.toml
shivamdurgbuns: @graingert can you review now?

ichard26:
Sorry @shivamdurgbuns, but looks like we haven't been clear enough on what needs to be done. We want to mitigate the deprecation of `@middleware` in aiohttp 4.x by adding a try/except in `src/blackd/middlewares.py` that simply does `middleware = lambda x: x` if `aiohttp.web_middlewares.middleware` doesn't exist. After that then it's safe to ignore the deprecation warning in `pyproject.toml`.

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 813e86b..849891f 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,6 +105,9 @@ filterwarnings = [
     # this is mitigated by a try/catch in https://github.com/psf/black/pull/2974/
     # this ignore can be removed when support for aiohttp 3.7 is dropped.
     '''ignore:Decorator `@unittest_run_loop` is no longer needed in aiohttp 3\.8\+:DeprecationWarning''',
+    # this is mitigated by a try/catch in https://github.com/psf/black/pull/3198/
+    # this ignore can be removed when support for aiohttp 3.x is dropped.
+    '''ignore:Middleware decorator is deprecated since 4\.0 and its behaviour is default, you can simply remove this decorator:DeprecationWarning''',
     # this is mitigated by https://github.com/python/cpython/issues/79071 in python 3.8+
     # this ignore can be removed when support for 3.7 is dropped.
     '''ignore:Bare functions are deprecated, use async ones:DeprecationWarning''',
diff --git a/src/blackd/middlewares.py b/src/blackd/middlewares.py
index 7abde52..2be42b0 100644
--- a/src/blackd/middlewares.py
+++ b/src/blackd/middlewares.py
@@ -1,9 +1,19 @@
-from typing import Awaitable, Callable, Iterable
+from typing import TYPE_CHECKING, Awaitable, Callable, Iterable, TypeVar
 
-from aiohttp.web_middlewares import middleware
 from aiohttp.web_request import Request
 from aiohttp.web_response import StreamResponse
 
+try:
+    from aiohttp.web_middlewares import middleware
+except ImportError:
+    # @middleware is deprecated and its behaviour is the default since aiohttp 4.0 so
+    # if it doesn't exist anymore, just define a no-op decorator for compatibility.
+    middleware = lambda x: x
+
+if TYPE_CHECKING:
+    F = TypeVar("F", bound=Callable[..., Any])
+    middleware: Callable[[F], F] = lambda x: x
+
 Handler = Callable[[Request], Awaitable[StreamResponse]]
```
Here are roughly the changes we'd like, although this is mostly untested.

PR #3202, opened 2022-08-01
makes install available for all users in docker image

### LINKED ISSUE #2975: Docker images use a user install of Black so don't work when running as another user
**Describe the bug**

Since [bd96139](https://github.com/psf/black/commit/bd961304b63ad8178610df7242edf180d825e336), Docker images use a user install of Black. This means they can't be used when running a container as a user other than root. 

This is a problem for me because we run Black inside of Jenkins and Jenkins runs _all_ containers with the same UID as the host, meaning that Black cannot run because it isn't on the PATH and in any case `/root/.local/bin/black` is inaccessible. It also seems like it could be problematic in any situation where the Black Docker container needs to write a file, because files it creates must be owned by root.

**To Reproduce**

Previously this would work:
```
$ docker run --rm -it -u $(id -u):$(id -g) -v $PWD:/workspace -w /workspace pyfound/black:21.9b0 black --check .
All done! ✨ 🍰 ✨
4 files would be left unchanged.
```
However on versions 21.10b0 and up, including the latest 22.3.0 I get the following output:
```
$ docker run --rm -it -u $(id -u):$(id -g) -v $PWD:/workspace -w /workspace pyfound/black:21.10b0 black --check .
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "black": executable file not found in $PATH: unknown.
$ docker run --rm -it -u $(id -u):$(id -g) -v $PWD:/workspace -w /workspace pyfound/black:22.3.0 black --check .
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "black": executable file not found in $PATH: unknown.
```

**Expected behavior**

I should be able to run the Black Docker container as any user, not just root.

**Environment**

- Black's version: 21.10b0, 22.3.0
- OS and Python version: [Linux/Python 3.10.4]

n-borges:
### Description

Changes the default docker image installation to be made in a virtualenv. This way we get to keep the handy multistage build without excluding non-root users.
fixes #2975 
### Checklist - did you ...

- [ x] Add a CHANGELOG entry if necessary?
- [ x] Add / update tests if necessary?
- [ x] Add new / update outdated documentation?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,16 +2,19 @@ FROM python:3-slim AS builder
 
 RUN mkdir /src
 COPY . /src/
+ENV VIRTUAL_ENV=/opt/venv
+RUN python -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 RUN pip install --no-cache-dir --upgrade pip setuptools wheel \
     # Install build tools to compile dependencies that don't have prebuilt wheels
     && apt update && apt install -y git build-essential \
     && cd /src \
-    && pip install --user --no-cache-dir .[colorama,d]
+    && pip install --no-cache-dir .[colorama,d]
 
 FROM python:3-slim
 
 # copy only Python packages to limit the image size
-COPY --from=builder /root/.local /root/.local
-ENV PATH=/root/.local/bin:$PATH
+COPY --from=builder /opt/venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
 
 CMD ["black"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,7 @@ RUN mkdir /src
 COPY . /src/
 ENV VIRTUAL_ENV=/opt/venv
 RUN python -m venv $VIRTUAL_ENV
-ENV PATH="$VIRTUAL_ENV/bin:$PATH"
-RUN pip install --no-cache-dir --upgrade pip setuptools wheel \
+RUN . /opt/venv/bin/activate && pip install --no-cache-dir --upgrade pip setuptools wheel \
     # Install build tools to compile dependencies that don't have prebuilt wheels
     && apt update && apt install -y git build-essential \
     && cd /src \
@@ -17,4 +16,4 @@ FROM python:3-slim
 COPY --from=builder /opt/venv /opt/venv
 ENV PATH="/opt/venv/bin:$PATH"
 
-CMD ["black"]
+CMD ["/opt/venv/bin/black"]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,7 +47,7 @@
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
 - Vim plugin: prefix messages with `Black: ` so it's clear they come from Black (#3194)
-- Docker: changes the installation path to be available to non-root users (#3202)
+- Docker: changed to a /opt/venv installation + added to PATH to be available to non-root users (#3202)
 
 ### Output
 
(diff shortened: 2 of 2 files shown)


ichard26:
hey thanks for picking this up so quickly, I really appreciate it! :black_heart: 

Anyway, I think this PR deserves a changelog entry under the `Integrations` subheader. If you could do that, that'd be great. Don't forget to include the PR number at the end of the entry.

n-borges:
> hey thanks for picking this up so quickly, I really appreciate it! black_heart
> 
> Anyway, I think this PR deserves a changelog entry under the `Integrations` subheader. If you could do that, that'd be great. Don't forget to include the PR number at the end of the entry.

Oh sorry about that, i overlooked that section in the changelog. I updated my PR accordingly!

Dockerfile
```diff
@@ -2,16 +2,19 @@ FROM python:3-slim AS builder
 
 RUN mkdir /src
 COPY . /src/
+ENV VIRTUAL_ENV=/opt/venv
+RUN python -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```
cooperlees, line 7: Don't know if this will work - Ignore if it does not ...

```suggestion
RUN source /opt/venv/bin/activate
```
Lets use the inbuilt `activate` script venv includes ... This way if any fixes are made in future python's venv module we get them for free.

Dockerfile
```diff
@@ -2,16 +2,19 @@ FROM python:3-slim AS builder
...
-    && pip install --user --no-cache-dir .[colorama,d]
+    && pip install --no-cache-dir .[colorama,d]
 
 FROM python:3-slim
 
 # copy only Python packages to limit the image size
-COPY --from=builder /root/.local /root/.local
-ENV PATH=/root/.local/bin:$PATH
+COPY --from=builder /opt/venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
```
cooperlees, line 18: Will this make the PATH persistent like this?

Dockerfile
```diff
@@ -2,16 +2,19 @@ FROM python:3-slim AS builder
...
 
 CMD ["black"]
```
cooperlees, line 20:
```suggestion
CMD ["/opt/venv/bin/black"]
```
Lets do this so it's more evident to people looking that we're in a venv and this is the black binary you should expect to use ...

cooperlees (changes requested):
Thanks. This makes sense. I just think we can tweak a few things.

Dockerfile
n-borges: Yes, it will prepend the PATH for all users. Here's a couple examples: 

```zsh
# test output as host user
❯ docker run --user $(id -u):$(id -g) nborges/black:pr which black
/opt/venv/bin/black

# test output as root
❯ docker run --user 0:0 nborges/black:pr which black
/opt/venv/bin/black
```

Dockerfile
n-borges: Thanks for the suggestion, i agree. Being that RUN instructions are independent from each other we are kind of forced to do this:
`RUN . /opt/venv/bin/activate && pip install ...`
It works indeed. I'm adding this, let me know what you think about it!

Dockerfile
n-borges: Got it, adding this as well

cooperlees (approved):
Thanks for the testing. This all looks good to go for me. Shouldn't break anyone unless they were using full path to the old black - Not a lot we can do there ...

n-borges:
Ok, thanks for the review!

cooperlees:
> Ok, thanks for the review!

Thanks for the fix and taking feedback so well + working together to get this in.

PR #3205, opened 2022-08-02
Remove invalid syntax in test

TomFryers:
### Description

`uR` is not a legal string prefix, so this test breaks (`AssertionError: cannot use --safe with this file; failed to parse source file AST: invalid syntax`) if changed to one in which the file is changed (e.g. in #2885). I've changed the last test to have `u` alone, and added an `R` to the test above instead.

### Checklist - did you ...


TomFryers:
I don't think a changelog entry is essential here!

ichard26 (approved):
ah of course I was the one who introduced this faulty test case haha, thank you for catching and fixing this!

Issue #3207, opened 2022-08-03
Misdetection of project root when `--stdin-filename` targets a filepath outside the current working directory


ichard26:
**To Reproduce**

```console
~/programming/oss/black on HEAD (b776bf9) [$?] via Python v3.8.5 (venv) 
❯ echo "print('Hello, world!')" > ../outside.py

~/programming/oss/black on HEAD (b776bf9) [$?] via Python v3.8.5 (venv) 
❯ cat ../outside.py | black --stdin-filename "../outside.py" -

No Python files are present to be formatted. Nothing to do 😴

~/programming/oss/black on HEAD (b776bf9) [$?] via Python v3.8.5 (venv) took 711ms 
❯ black ../outside.py 
reformatted ../outside.py

All done! ✨ 🍰 ✨
1 file reformatted.
```

**Expected behavior**

That `cat ../outside.py | black --stdin-filename "../outside.py" -` would also reformat `../outside.py`.

**Environment**

- Black's version:  b776bf92adb7f47cd92f550e33c2db445d226f78
- OS and Python version:  CPython 3.8.5 on Ubuntu 20.04.04 LTS

**Additional context**

I said I would look into this issue in https://github.com/microsoft/vscode-black-formatter/issues/52 where it was originally discovered but I never did. I'd appreciate it if someone else could look into this for me :)

Bobronium:
Thank you!

PR #3208, opened 2022-08-04
Fix misdetection of project root when `--stdin-filename` filepath is outside the current working directory

n-borges:
### Description
Fixes behaviour that prevented formatting for files outside the current working directory (#3207). 

```console
# before the fix
❯ cat ../script.py | python -m black --stdin-filename '../script.py' -

# after the fix 
❯ cat ../test.py | python -m black --stdin-filename '../test.py' -   
print("hello world")
All done! ✨ 🍰 ✨
1 file left unchanged.
```

Let me know what you think! Not sure about how elegant the solution is, but I would be happy to refactor it based on your suggestions.

### Checklist - did you ...

### DIFF
--- a/src/black/files.py
+++ b/src/black/files.py
@@ -161,7 +161,10 @@ def normalize_path_maybe_ignore(
         abspath = path if path.is_absolute() else Path.cwd() / path
         normalized_path = abspath.resolve()
         try:
-            root_relative_path = normalized_path.relative_to(root).as_posix()
+            if not str(normalized_path).startswith(str(root)):
+                root_relative_path = os.path.join(Path.cwd(), path)
+            else:
+                root_relative_path = normalized_path.relative_to(root).as_posix()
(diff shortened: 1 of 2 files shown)


felix-hilden:
Oh, this seems to overlap with #3216 most unfortunately. Although it is quite a simple solution. Not sure which would be better, but I'd like to see a more concrete test related to the original issue if you're up for the race :P

ichard26:
Hey, so https://github.com/psf/black/pull/3216 was filed as an alternative fix for the same issue. I chose it over yours as it's probably more maintainable in the longterm by maintaining the invariant the project root contains the files to be formatted. Thanks for submitting a PR anyway, perhaps they got interested in fixing this issue too by your PR :)

Have a great day/night!

n-borges:
Great, happy this issue has been solved!

Issue #3204, opened 2022-08-02
Black -C doesn't remove trailing commas in indices


TomFryers:
**Describe the bug**
Trailing commas inside indexing square brackets are not removed when running Black with `-C`. This also affects generic types with multiple parameters.

**To Reproduce**

For example, take this code:

```python
x = [0, 1,]
x = z[0, 1,]
```

and run with `-C`. We get

```python
x = [0, 1]
x = z[0, 1,]
```

**Expected behavior**

The final trailing comma should disappear.

```python
x = z[0, 1]
```
**Environment**

- Black's version: 22.6.0. Also tested on main on the playground, and with/without `--preview`.
- OS and Python version: Linux/Python 3.10.5

hauntsaninja:
Duplicate of https://github.com/psf/black/issues/2296

ichard26:
Good catch, thank you @hauntsaninja!

PR #3209, opened 2022-08-05
Strip trailing commas in subscripts with -C

### LINKED ISSUE #2296: black doesn't remove ugly trailing commas in subscripts
**Describe the bug**

We recently added `--skip-magic-trailing-comma` to black. When running with this option, we'd like it to always remove ugly unnecessary trailing commas that plagued some earlier versions of black, but it doesn't do this in the case of subscripts.

**To Reproduce**

```
(venv) ~/tmp/venv λ cat test.pyi
Tuple[A, B,]

def f(a, b,): ...
(venv) ~/tmp/venv λ black -l 130 test.pyi --skip-magic-trailing-comma --check --diff
--- test.pyi	2021-06-01 07:57:26.995533 +0000
+++ test.pyi	2021-06-01 07:58:31.094021 +0000
@@ -1,3 +1,3 @@
 Tuple[A, B,]
 
-def f(a, b,): ...
+def f(a, b): ...
would reformat test.pyi
Oh no! 💥 💔 💥
1 file would be reformatted.
```

**Expected behavior**

If the subscript expression has more than one element, we should remove the extra trailing comma.

hauntsaninja:
Fixes #2296, #3204

### DIFF
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -273,8 +273,8 @@ class Line:
         - it's not a single-element subscript
         Additionally, if ensure_removable:
         - it's not from square bracket indexing
-            - specifically, single-element square bracket indexing with
-              Preview.skip_magic_trailing_comma_in_subscript
+          (specifically, single-element square bracket indexing with
+          Preview.skip_magic_trailing_comma_in_subscript)
         """
         if not (
             closing.type in CLOSING_BRACKETS
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -308,7 +308,7 @@ class Line:
             if comma.parent is None:
                 return False
             if Preview.skip_magic_trailing_comma_in_subscript in self.mode:
-                return bool(
+                return (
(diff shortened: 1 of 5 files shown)


ichard26 (approved):
STGM. Thank you for fixing this so quickly!

src/black/lines.py
```diff
@@ -301,8 +303,22 @@ def has_magic_trailing_comma(
 
             if not ensure_removable:
                 return True
+
             comma = self.leaves[-1]
-            return bool(comma.parent and comma.parent.type == syms.listmaker)
+            if comma.parent is None:
+                return False
+            if Preview.skip_magic_trailing_comma_in_subscript in self.mode:
+                return bool(
```
JelleZijlstra, line 311:
```suggestion
                return (
```
Nit, but all the elements in the or are already bools.

JelleZijlstra (approved):
Looks good, one small nit

PR #3211, opened 2022-08-05
Reduce single file format overhead by lazily importing modules and loading .gitignore / --exclude

### LINKED ISSUE #2564: Black slow startup to format a single file

ichard26:
### Description

```
`black.reformat_many` depends on a lot of slow-to-import modules. When
formatting simply a single file, the time paid to import those modules
is totally wasted. So I moved `black.reformat_many` and its helpers
to `black.concurrency` which is now *only* imported if there's more
than one file to reformat. This way, running Black over a single file
is snappier

Here are the numbers before and after this patch running `python -m
black --version`:

- interpreted: 411 ms +- 9 ms -> 342 ms +- 7 ms: 1.20x faster
- compiled: 365 ms +- 15 ms -> 304 ms +- 7 ms: 1.20x faster
```

```
Loading .gitignore and compiling the exclude regex can take more than
15ms. We shouldn't and don't need to pay this cost if we're simply
formatting files given on the command line directly.

I would've loved to lazily import pathspec, but the patch won't be clean
until the file collection and discovery logic is refactored first.

Co-authored-by: Fabio Zadrozny <fabiofz@gmail.com>
```

Fixes #2564 and closes #2656.

### Checklist - did you ...

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1,6 +1,5 @@
 import io
 import json
-import os
 import platform
 import re
 import sys
@@ -27,11 +26,6 @@ from typing import (
     Tuple,
     Union,
 )
-
-if sys.version_info >= (3, 8):
-    from typing import Final
-else:
-    from typing_extensions import Final
 
 import click
 from click.core import ParameterSource
@@ -92,7 +86,6 @@ from blib2to3.pgen2 import token
 from blib2to3.pytree import Leaf, Node
 
 COMPILED = Path(__file__).suffix in (".pyd", ".so")
-DEFAULT_WORKERS: Final = os.cpu_count()
 
 # types
 FileContent = str
@@ -371,9 +364,8 @@ def validate_regex(
     "-W",
     "--workers",
     type=click.IntRange(min=1),
-    default=DEFAULT_WORKERS,
-    show_default=True,
-    help="Number of parallel workers",
+    default=None,
+    help="Number of parallel workers [default: number of CPUs in the system]",
 )
 @click.option(
     "-q",
@@ -448,7 +440,7 @@ def main(  # noqa: C901
     extend_exclude: Optional[Pattern[str]],
     force_exclude: Optional[Pattern[str]],
     stdin_filename: Optional[str],
-    workers: int,
+    workers: Optional[int],
     src: Tuple[str, ...],
     config: Optional[str],
 ) -> None:
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1,10 +1,8 @@
-import asyncio
 import io
 import json
 import os
 import platform
 import re
-import signal
 import sys
 import tokenize
 import traceback
@@ -13,10 +11,8 @@ from dataclasses import replace
 from datetime import datetime
 from enum import Enum
 from json.decoder import JSONDecodeError
-from multiprocessing import Manager, freeze_support
 from pathlib import Path
 from typing import (
-    TYPE_CHECKING,
     Any,
     Dict,
     Generator,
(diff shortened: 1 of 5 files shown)


ichard26:
For what it is worth, I'd prefer a rebase merge for this PR given the commits are mostly independent (other than the changelog entry).

ichard26:
Coolio, diff-shades is failing and I don't know why at all ...

ichard26:
Coolio, diff-shades is failing and I don't know why at all ...

Doyhenc99 (changes requested):
Sadang bib

src/black/__init__.py
```diff
@@ -90,10 +90,10 @@
 from blib2to3.pgen2 import token
 from blib2to3.pytree import Leaf, Node
 
-if TYPE_CHECKING:
-    from concurrent.futures import Executor
-
 COMPILED = Path(__file__).suffix in (".pyd", ".so")
+DEFAULT_WORKERS: Final = os.cpu_count()
```
JelleZijlstra, line 94: Hm this does a syscall at startup, we may want to put it into a cached-forever function instead.

src/black/__init__.py
```diff
@@ -1495,8 +1369,14 @@ def patch_click() -> None:
 
 
 def patched_main() -> None:
-    maybe_install_uvloop()
-    freeze_support()
+    global __BLACK_MAIN_CALLED__
+    __BLACK_MAIN_CALLED__ = True
```
JelleZijlstra, line 1373: Instead of this global, what if we return early in `maybe_install_uvloop` if uvloop is already in sys.modules?

Or we could just call it unconditionally. It's only called once per formatting run and it's pretty fast:

```
In [14]: %timeit black.concurrency.maybe_install_uvloop()
1.67 µs ± 30.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

src/black/__init__.py
JelleZijlstra: Though then again

```
In [16]: %timeit os.cpu_count()
1.05 µs ± 10.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

ichard26:
PTAL. Please note I'd prefer a **rebase merge** here if you approve and chose to merge right after.

src/black/concurrency.py
```diff
@@ -55,3 +71,129 @@ def shutdown(loop: asyncio.AbstractEventLoop) -> None:
...
+def reformat_many(
+    sources: Set[Path],
+    fast: bool,
+    write_back: WriteBack,
+    mode: Mode,
+    report: Report,
+    workers: Optional[int],
+) -> None:
+    """Reformat multiple files using a ProcessPoolExecutor."""
+    maybe_install_uvloop()
```
ichard26, line 88: I chose to keep the helper function and instead of always attempting an import of uvloop in the module scope just as insurance if we ever accidentally break the invariant `black.concurrency` should only be imported via `black.main` if there's more than one file to format.

ichard26:
This works on Windows with PyInstaller (just tested locally) so I'm merging.

PR #3215, opened 2022-08-10
Use debug f-strings for feature detection

### LINKED ISSUE #2907: When detecting Python versions, `f"{x=}"` should imply 3.8+
**Describe the bug**

The `=` specifier was added to f-strings in Python 3.8, but Black does not yet distinguish it from any other f-string, and therefore detects the version as 3.6+.  This leads to missed formatting opportunities, and sometimes problems in downstream projects since this is invalid syntax under earlier versions.  (e.g. https://github.com/Zac-HD/shed/issues/31, prompting this issue)

**To Reproduce**

```python
from black import detect_target_versions, parsing

code = parsing.lib2to3_parse("""
f'{x=}'
""")
print(detect_target_versions(code))
```
```
$ python repro.py
{<TargetVersion.PY36: 6>, <TargetVersion.PY37: 7>, <TargetVersion.PY38: 8>, <TargetVersion.PY39: 9>, <TargetVersion.PY310: 10>}
```

hauntsaninja:
Fixes #2907

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -87,6 +87,7 @@ from black.output import color_diff, diff, dump_to_file, err, ipynb_diff, out
 from black.parsing import InvalidInput  # noqa F401
 from black.parsing import lib2to3_parse, parse_ast, stringify_ast
 from black.report import Changed, NothingChanged, Report
+from black.trans import iter_fexpr_spans
 from blib2to3.pgen2 import token
 from blib2to3.pytree import Leaf, Node
 
@@ -1240,6 +1241,7 @@ def get_features_used(  # noqa: C901
 
     Currently looking for:
     - f-strings;
+    - self-documenting expressions in f-strings (f"{x=}");
     - underscores in numeric literals;
     - trailing commas after * or ** in function signatures and calls;
     - positional only arguments in function signatures and lambdas;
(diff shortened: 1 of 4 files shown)


tests/test_black.py
```diff
@@ -310,6 +310,24 @@ def test_detect_pos_only_arguments(self) -> None:
         versions = black.detect_target_versions(root)
         self.assertIn(black.TargetVersion.PY38, versions)
 
+    def test_detect_debug_f_strings(self) -> None:
```
isidentical, line 313: Can you also include a couple of normal cases like the ones in here (not all of them, a couple should be enough)

https://github.com/python/cpython/blob/71c3d649b5a0324c6eb01f9ad025c1e102b82bba/Lib/test/test_fstring.py#L1262-L1291

src/black/__init__.py
```diff
@@ -1261,6 +1263,11 @@ def get_features_used(  # noqa: C901
             value_head = n.value[:2]
             if value_head in {'f"', 'F"', "f'", "F'", "rf", "fr", "RF", "FR"}:
                 features.add(Feature.F_STRINGS)
+                if Feature.DEBUG_F_STRINGS not in features:
+                    for span_beg, span_end in iter_fexpr_spans(n.value):
```
isidentical, line 1267: Is this an expensive function, or is its speed negligible? If it is a bit slow, we could also add a more naive check above ("=" in n.value) to only include f-strings with equal signs (which should reduce the number of calls to this function by a lot). If it is not too expensive, I think it is alright to keep it as is.

isidentical (approved):
Looks amazing!

src/black/__init__.py
hauntsaninja: It is not too expensive, it's a single loop over the value. mypyc should also compile it well.

isidentical: Cool! Then no need.

ichard26 (approved):
Man you work fast :eyes: Thank you for all of your contributions to Black lately, I really appreciate it :)

PR #3216, opened 2022-08-10
Fix misdetection of project root with `--stdin-filename`

### LINKED ISSUE #3207: Misdetection of project root when `--stdin-filename` targets a filepath outside the current working directory

hauntsaninja:
There are a number of places this behaviour could be patched, for
instance, it's quite tempting to patch it in `get_sources`. However
I believe we generally have the invariant that project root contains all
files we want to format, in which case it seems prudent to keep that
invariant.

Fixes #3207

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -468,7 +468,13 @@ def main(  # noqa: C901
         out(main.get_usage(ctx) + "\n\nOne of 'SRC' or 'code' is required.")
         ctx.exit(1)
 
-    root, method = find_project_root(src) if code is None else (None, None)
+    if stdin_filename is not None:
+        src_with_stdin_filename = tuple(stdin_filename if s == "-" else s for s in src)
+    else:
+        src_with_stdin_filename = src
+    root, method = (
+        find_project_root(src_with_stdin_filename) if code is None else (None, None)
+    )
     ctx.obj["root"] = root
 
     if verbose:
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -469,12 +469,8 @@ def main(  # noqa: C901
         out(main.get_usage(ctx) + "\n\nOne of 'SRC' or 'code' is required.")
         ctx.exit(1)
 
-    if stdin_filename is not None:
-        src_with_stdin_filename = tuple(stdin_filename if s == "-" else s for s in src)
-    else:
-        src_with_stdin_filename = src
     root, method = (
-        find_project_root(src_with_stdin_filename) if code is None else (None, None)
+        find_project_root(src, stdin_filename) if code is None else (None, None)
     )
     ctx.obj["root"] = root
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -485,7 +485,9 @@ def main(  # noqa: C901
             )
 
             normalized = [
-                (normalize_path_maybe_ignore(Path(source), root), source)
+                (source, source)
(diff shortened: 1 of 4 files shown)


hauntsaninja:
I wasn't quite sure where to put a test. The existing `--stdin-filename` tests are mostly unit tests of `get_sources`. The existing `find_project_root` tests don't take in input.
One potential option would be to make `stdin_filename` a parameter of `find_project_root` and then move this logic into `find_project_root`? I could also make some complicated test that exercised all of main.

hauntsaninja:
Updated changelog.

ichard26 mentioned that the "Sources to be formatted" log became incorrect with this change.
Technically, this is a separate bug that can be reproduced on main with e.g. `~/dev/black/docs main λ cat conf.py | black --stdin-filename "outside.py" -vvv --check -`. Anyway, I pushed a quick fix for this, but in the future we'd ideally want "Sources to be formatted" logic to be more integrated with `get_sources`.

I still owe tests on this PR.

JelleZijlstra (approved):

src/black/__init__.py
```diff
@@ -469,7 +469,13 @@ def main(  # noqa: C901
...
         ctx.exit(1)
 
-    root, method = find_project_root(src) if code is None else (None, None)
+    if stdin_filename is not None:
+        src_with_stdin_filename = tuple(stdin_filename if s == "-" else s for s in src)
+    else:
+        src_with_stdin_filename = src
+    root, method = (
+        find_project_root(src_with_stdin_filename) if code is None else (None, None)
+    )
```
felix-hilden, line 478:
```suggestion
    if stdin_filename is not None:
        src = tuple(stdin_filename if s == "-" else s for s in src)

    root, method = find_project_root(src) if code is None else (None, None)
```

Perhaps we could make the above work below as well.

Also, I was unaware how passing multiple stdin placeholders works, but for `- - -` etc stdin is only read once (or does closing it prevent further reads, guess so), and with other files `-` is read as a filename. So it is safe to assume that there will only be one to convert. Or maybe we'll have to even guard against weird use, not sure what would be best.

Yeah, tests would be nice 👍

ichard26 (approved):
Awesome work, thank you! I pushed a change to your test since it wasn't actually testing the buggy behaviour due to how the `find_project_root` function works.

PR #3217, opened 2022-08-11
feat(#3195): add preview option support for blackd

### LINKED ISSUE #3195: Preview option support for blackd
It seems that **blackd** does not support [the preview option](https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html?highlight=preview#preview-style) now.

I suggest implementing an option for that like it is done for other features https://black.readthedocs.io/en/stable/usage_and_configuration/black_as_a_server.html#protocol

Something like `X-PREVIEW` would be fine.

I faced this while using **[intellij-blackconnect](https://github.com/lensvol/intellij-blackconnect/)** and there is an issue about that too https://github.com/lensvol/intellij-blackconnect/issues/37

Mogost:
### Description

Resolves #3195

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,7 +29,7 @@
 
 <!-- Changes to blackd -->
 
-- `blackd` now supports preview style via `X-PREVIEW` header (#3195)
+- `blackd` now supports preview style via `X-Preview` header (#3217)
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,8 @@
 
 <!-- Changes to blackd -->
 
+- `blackd` now supports preview style via `X-PREVIEW` header (#3195)
+
 ### Configuration
 
 <!-- Changes to how Black can be configured -->
--- a/docs/usage_and_configuration/black_as_a_server.md
+++ b/docs/usage_and_configuration/black_as_a_server.md
@@ -54,10 +54,10 @@ The headers controlling how source code is formatted are:
   command line flag. If present and its value is not the empty string, no string
   normalization will be performed.
 - `X-Skip-Magic-Trailing-Comma`: corresponds to the `--skip-magic-trailing-comma`
-  command line flag. If present and its value is not the empty string, trailing commas
+  command line flag. If present and its value is not an empty string, trailing commas
   will not be used as a reason to split lines.
-- `X-Preview`: corresponds to the `--preview`command line flag. If present and its value
-  is not the empty string, experimental and potentially disruptive style changes will be
+- `X-Preview`: corresponds to the `--preview` command line flag. If present and its value
+  is not an empty string, experimental and potentially disruptive style changes will be
   used.
 - `X-Fast-Or-Safe`: if set to `fast`, `blackd` will act as _Black_ does when passed the
   `--fast` command line flag.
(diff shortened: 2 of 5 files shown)


src/blackd/__init__.py
```diff
@@ -32,6 +32,7 @@
 PYTHON_VARIANT_HEADER = "X-Python-Variant"
 SKIP_STRING_NORMALIZATION_HEADER = "X-Skip-String-Normalization"
 SKIP_MAGIC_TRAILING_COMMA = "X-Skip-Magic-Trailing-Comma"
+PREVIEW = "X-PREVIEW"
```
ichard26, line 35:
```suggestion
PREVIEW = "X-Preview"
```
I know HTTP headers are case-insensitive, but we might as make it consistent with the other constants :)

docs/usage_and_configuration/black_as_a_server.md
```diff
@@ -56,6 +56,9 @@ The headers controlling how source code is formatted are:
 - `X-Skip-Magic-Trailing-Comma`: corresponds to the `--skip-magic-trailing-comma`
   command line flag. If present and its value is not the empty string, trailing commas
   will not be used as a reason to split lines.
+- `X-Preview`: corresponds to the `--preview`command line flag. If present and its value
+  is not the empty string, experimental and potentially disruptive style changes will be
+  used.
```
ichard26, line 61:
```suggestion
- `X-Skip-Magic-Trailing-Comma`: corresponds to the `--skip-magic-trailing-comma`
  command line flag. If present and its value is not an empty string, trailing commas
  will not be used as a reason to split lines.
- `X-Preview`: corresponds to the `--preview` command line flag. If present and its value
  is not an empty string, experimental and potentially disruptive style changes will be
  used.
```

CHANGES.md
```diff
@@ -29,6 +29,8 @@
 
 <!-- Changes to blackd -->
 
+- `blackd` now supports preview style via `X-PREVIEW` header (#3195)
```
ichard26, line 32:
```suggestion
- `blackd` now supports preview style via `X-Preview` header (#3217)
```
Changelog entries are linked to their PR, not issue.

ichard26 (commented):
Hey thanks for filing a PR, I really appreciate it :black_heart: 

The PR looks pretty good, got a few minor suggestions. I would've also suggested adding a testcase that verifies the preview style was actually enabled by using input that's only changed under the preview style, **however**, the test would have to be changed in 2023 when we promote many changes in preview to stable. Sooo it's probably fine to not have a test, but I just wanted to make sure we thought this through.

Mogost:
Thank you for your suggestions @ichard26. Just applied them.

> I would've also suggested adding a testcase that verifies the preview style was actually enabled by using input that's only changed under the preview style, however, the test would have to be changed in 2023 when we promote many changes in preview to stable. Sooo it's probably fine to not have a test, but I just wanted to make sure we thought this through.

Yes, that's the reason I didn't create this kind of test. Moreover, I noticed that the other headers have no tests at all. So I thought that a test that just sends this header would be enough.

ichard26 (approved):
I'll merge this in around 24 hours if no one yelps any objections. @Mogost feel free to add yourself to the AUTHORS list, either way thank you very much! Any feedback on your experience contributing here is welcome here: #2238.

Mogost:
Thank you @ichard26 !
I will take a look at the mentioned issue.

JelleZijlstra (approved):

PR #3220, opened 2022-08-13
Use --no-implicit-optional for type checking

hauntsaninja:
This makes type checking PEP 484 compliant (as of 2018).
mypy will change its defaults soon.

See:
https://github.com/python/mypy/issues/9091
https://github.com/python/mypy/pull/13401

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -22,6 +22,7 @@ warn_no_return=True
 warn_redundant_casts=True
 warn_unused_ignores=True
 disallow_any_generics=True
+no_implicit_optional=True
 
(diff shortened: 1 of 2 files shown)


hauntsaninja:
In general, given that black uses mypyc, feels like you should probably start from `strict` and then re-enable any looseness you need.

hauntsaninja:
Also skip news please! :-)

ichard26 (commented):
> In general, given that black uses mypyc, feels like you should probably start from strict and then re-enable any looseness you need. But that can be for later.

I agree :+1:

ichard26 (approved):
Misclicked :upside_down_face:

PR #3222, opened 2022-08-14
Use strict mypy checking

hauntsaninja:
Note that we still have:
```
src/blib2to3/pytree.py
13:# mypy: allow-untyped-defs, allow-incomplete-defs

src/blib2to3/pgen2/tokenize.py
4:# mypy: allow-untyped-defs, allow-untyped-calls

src/blib2to3/pgen2/conv.py
4:# mypy: ignore-errors
```

I took a solid look at removing those from pytree, but it seems like some of the existing type annotations are lies (e.g. things annotated as `_Results` can have list values).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,8 @@ repos:
           - types-typed-ast >= 1.4.1
           - click >= 8.1.0
           - platformdirs >= 2.1.0
+          - pytest
+          - hypothesis
 
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v2.7.1
--- a/mypy.ini
+++ b/mypy.ini
@@ -38,3 +38,6 @@ ignore_missing_imports = True
 
 [mypy-tokenize_rt.*]
 ignore_missing_imports = True
+
+[mypy-uvloop.*]
+ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -23,3 +23,18 @@ warn_unreachable=True
 # The following is because of `patch_click()`. Remove when
 # we drop Python 3.6 support.
 warn_unused_ignores=False
+
+[mypy-blib2to3.driver.*]
+ignore_missing_imports = True
+
+[mypy-IPython.*]
+ignore_missing_imports = True
+
+[mypy-colorama.*]
+ignore_missing_imports = True
+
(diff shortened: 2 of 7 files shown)


hauntsaninja:
Oh also I left `no_implicit_reexport = False` because a lot of our tests rely on this. black itself doesn't have a public API, so fixing this is low stakes.

ichard26:
We compile blib2to3 with mypyc, I guess the codepaths with the faulty annotations are just never hit in practice...?

hauntsaninja:
Yeah... I guess mypyc might not check the values of dicts? I'll get more familiar with the code and try again later

hauntsaninja:
Okay, I thought the `Error: Invalid value for 'analysis': File 'assert-no-changes-pr-3222-7e456de3e0.json' does not exist.` thing was spurious, but since it happened again I found the actual errors above. I was a little confused, but it turns out pre-commit secretly and evil-ly inserts `ignore_missing_imports` into mypy config. Anyway, everything is green now.

ichard26:
> Okay, I thought the Error: Invalid value for 'analysis': File 'assert-no-changes-pr-3222-7e456de3e0.json' does not exist. thing was spurious, but since it happened again I found the actual errors above. 

Hmm, I should document the fact that "analysis does not exist" errors may simply be the byproduct of other errors, ... or just change the workflow to not error out a second time. Sorry for the confusion!

ichard26:
About the compiler error, I opened https://github.com/mypyc/mypyc/issues/948 to track that. I pushed a fix that simply removes the type declaration in `NegatedPattern`. We don't use this part of `blib2to3` so simplicity is all that matters here.

ichard26:
Ugh, diff-shades is failing because mypy can't find `_black_version.py` since the local copy of Black isn't installed yet. `preview-changes` ran without issue because the baseline revision wasn't cached so it'd install local Black as part of the baseline analysis, letting the target revision (i.e. this PR) build just fine. The `assert-no-changes` job was able to use a cached baseline analysis hence the build error.

ichard26 (approved):
STGM, thank you for all of your work helping to migrate OSS projects to mypy post --strict-optional being on by default! Will ask @JelleZijlstra for a lookover though as I'm no pro at typing.

JelleZijlstra (approved):

Issue #2503, opened 2021-09-21
No branch stable to install Vim plugin with vim-plug


lcheylus:
According to the [docs](https://black.readthedocs.io/en/stable/integrations/editors.html#vim), to install Vim plugin with vim-plug, we must use branch `stable`. But there is no (more ?) branch `stable` only a tag.

`.vimrc` file :
`Plug 'psf/black', { 'branch': 'stable' }`

When I try to install the plugin with `:PlugInstall` command, I have this error : `Invalid branch HEAD (expected: stable).`

If I modify my Vim configuration to use tag instead of branch, it works as expected (Vim plugin installed and black tool installed in virtualenv) :
`Plug 'psf/black', { 'tag': 'stable' }`

Please recreate `stable` branch or modify doc to install Vim plugin : use tag instead of branch with `vim-plug`.

- Version: Black, version 21.9b0 on Python 3.9.7 (default, Sep  3 2021, 06:18:44)
- OS and Python version: Debian testing (bookworm) amd64, Python 3.9.7

pfheatwole:
Be aware that vim-plug does not force override local tags with modified remote tags. Because the remote `psf/black` repo deletes and recreates their `stable` tags, local copies will not stay in sync, and `:PlugUpdate` will complain about "pending changes".

If you want to track the non-stationary `stable` tag instead tracking the `main` branch then you'll need to do something like:

```
$ cd "~/.vim/bundle/black"
$ git fetch --tags --force
```
Modify the directory to match `PLUGIN_DIRECTORY` in the `call plug#begin(PLUGIN_DIRECTORY)` line near the top of your `~/.vimrc` or `~/.config/nvim/init.vim`.

Fogapod:
I have a different error trying to use vim-plug, both using `tag` and `branch`: `Plug 'psf/black', { 'tag': 'stable' }`

firts time it works, but next updates return error: `merge: origin/stable - not something we can merge`

docs say i should run this command, but it doesn't work:
```sh
$ git checkout origin/stable -b stable
fatal: 'origin/stable' is not a commit and a branch 'stable' cannot be created from it
```
repo information (cloned by vim-plug):
```sh
~/.local/share/nvim/plugged/black (tags/21.12b0) $ g status
HEAD detached at stable
nothing to commit, working tree clean
```

KevinWhalen:
This seems potentially related. I I thought I had updated in the past without problems.

    :PlugUpdate

`merge: origin/stable - not something we can merge`

[...]

PR #3223, opened 2022-08-15
Update stable branch after publishing to PyPI

ichard26:
### Description

We've decided to a) convert stable back into a branch and b) to update it immediately as part of the release process. We may as well automate it.

Test run: https://github.com/ichard26/black/runs/7831041412?check_suite_focus=true (note that in that run the job runs after the sdist/pure wheel job passes, while in this PR, it runs after all the PyPI jobs finish, I changed it in the test run because I didn't want to wait several minutes for the mypyc jobs to finish)

**This is a draft PR because the stable ref is currently a tag in this repository so things will break**. I'll have to convert the ref manually before landing this.

### Checklist - did you ...

- [x] Add new / update outdated documentation? -> will be done in a separate PR

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -74,3 +74,22 @@ jobs:
         env:
           TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
         run: pipx run twine upload --verbose -u '__token__' wheelhouse/*.whl
+
+  update-stable-branch:
+    name: Update stable branch
+    needs: [main, mypyc]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout stable branch
+        uses: actions/checkout@v3
+        with:
+          ref: stable
+          fetch-depth: 0
+
+      - name: Update stable branch to release tag & push
+        run: |
+          git reset --hard ${{ github.event.release.tag_name }}
+          git push --force
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -92,4 +92,4 @@ jobs:
       - name: Update stable branch to release tag & push
         run: |
           git reset --hard ${{ github.event.release.tag_name }}
-          git push --force
+          git push


JelleZijlstra (approved):

.github/workflows/pypi_upload.yml
```diff
@@ -74,3 +74,22 @@ jobs:
...
+      - name: Checkout stable branch
+        uses: actions/checkout@v3
+        with:
+          ref: stable
+          fetch-depth: 0
+
+      - name: Update stable branch to release tag & push
+        run: |
+          git reset --hard ${{ github.event.release.tag_name }}
+          git push --force
```
JelleZijlstra, line 95: We should be able to avoid the `--force`, right? Each release should be a direct descendant of the previous one.

cooperlees:
Where can I read again where a branch helps over moving a tag? I still don't get the advantage of the branch overhead when I tag should be able to do what we want.

felix-hilden (approved):
Content looks good, so given a consensus on branch > tag we're good to go (I don't see why not if Vim improves, RTD is already meh)

ichard26:
@cooperlees 
> I still don't get the advantage of the branch overhead **when a tag should be able to do what we want**. 

Can? Yes. Well? Definitely not. (also, what overhead does a branch have that a tag does not?)

[Git tags are not the right tool, at all](https://git-scm.com/docs/git-tag#_on_re_tagging). They come with the expectation that they will never change. Things will not work as expected if they do change, doubly so if they change regularly. Once you pull `stable` from the remote and it's copied in your local repository, no matter how many times you run `git pull` you'll never see it get updated automatically. Your only recourse is to delete the tag via `git tag -d stable` before pulling.

This gets annoying really quickly since `stable` is supposed to be the solution for folks "[who want to move along as Black developers deem the newest version reliable.](https://black.readthedocs.io/en/stable/contributing/release_process.html#moving-the-stable-tag)" [See this comment for how this impacts users using our Vim plugin](https://github.com/psf/black/issues/2503#issuecomment-1196357379). It also affects us developers[^1]. If you have `stable` locally, once we cut a new release and update the `stable` tag, a simple `git pull` / `git fetch` will not pull down the updated `stable` tag. Unless you remember to delete `stable` before pulling, `stable` will become stale and useless.

You can argue this is a good thing ("people should explicitly opt into updating stable"), but IMO it does not match user expectations nor developer expectations[^2]. Especially since not all our integrations that use `stable` are bound by this security measure, for example our GitHub Action (since it does a clean fetch of the repository every time it's used). I believe consistency would be good here.

Finally, ever since we switched to a tag, we've been facing issues with ReadTheDocs not picking up updates to `stable` unless we force a rebuild. The initial rebuild on the `stable` update just pulls the commit the tag previously pointed to. I'm not sure if switching back to a branch will fix this, but I'd wager it will.  
  
[^1]: In fairness, most folks working on Black probably don't use the `stable` ref anyway, especially us maintainers who'd know what is the latest version by heart, but it'd still be nice to make it usable for local dev though.  

[^2]: Also what benefit does a `stable` ref have over explicit version tags like `22.6.0`? If you're going to opt into some odd pin mechanism, might as well use explicit version tags for clarity and consistency.

PR #3226, opened 2022-08-15
Removed need for subprocess call in `action.yml`

ionite34:
### Description

Updated `actions.yml` to use the alternative `$GITHUB_ACTION_PATH` variable as documented [here](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsrun) instead of the original `github.action_path` which caused issues with bash on the Windows runners. This removes the need for a python subprocess to call the `main.py` script.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,6 +55,8 @@
 - Vim plugin: prefix messages with `Black: ` so it's clear they come from Black (#3194)
 - Docker: changed to a /opt/venv installation + added to PATH to be available to
   non-root users (#3202)
+- GitHub Action: Updated usage of $GITHUB_ACTION_PATH to remove the nested python 
+  subprocess call for actions/main.py (#3226)
 
 ### Output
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -55,6 +55,8 @@
 - Vim plugin: prefix messages with `Black: ` so it's clear they come from Black (#3194)
 - Docker: changed to a /opt/venv installation + added to PATH to be available to
   non-root users (#3202)
+- GitHub Action: Updated usage of $GITHUB_ACTION_PATH to remove the nested python
```
ichard26, line 58: I don't think a changelog entry is needed for this PR as this part should be totally transparent to users. I'll add a `skip news` label in the meanwhile.

ichard26 (approved):
I really appreciate you filing this PR @ionite34. I always love to remove hacky workarounds :)

I can confirm this works: https://github.com/ichard26/testblackaction/runs/7963173649?check_suite_focus=true (don't be worried by the red, the failure is expected as I committed some code that gets reformatted :slightly_smiling_face: -- the action startup clearly works fine)

Feel free to add yourself to the AUTHORS.md file, otherwise this mostly SGTM other than the changelog entry.

ichard26:
Thanks for the PR @ionite34! :wolf:

PR #3227, opened 2022-08-16
Fix a string merging/split issue caused by standalone comments.

### LINKED ISSUE #2734: ESP: a standalone comment causes strings to be merged into one far too long (and requiring two passes to do so)

yilei:
### Description

This PR fixes the issue where a standalone comment causes strings to be merged into one far too long (and requiring two passes to do so). (Fixes #2734.)

### Checklist - did you ...

### DIFF
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -553,6 +553,9 @@ class StringMerger(StringTransformer, CustomSplitMapMixin):
 
             next_str_idx += 1
 
+        # Take a note on the index of the non-STRING leaf.
+        non_string_idx = next_str_idx
+
         S_leaf = Leaf(token.STRING, S)
         if self.normalize_strings:
             S_leaf.value = normalize_string_quotes(S_leaf.value)
@@ -572,7 +575,19 @@ class StringMerger(StringTransformer, CustomSplitMapMixin):
         string_leaf = Leaf(token.STRING, S_leaf.value.replace(BREAK_MARK, ""))
 
         if atom_node is not None:
-            replace_child(atom_node, string_leaf)
+            # If not all children of the atom node are merged...
+            if non_string_idx - string_idx < len(atom_node.children):
+                # We need to replace the old STRING leaves with the new string leaf.
+                first_child_idx = None
+                for idx in range(string_idx, non_string_idx):
+                    child_idx = LL[idx].remove()
+                    if first_child_idx is None:
+                        first_child_idx = child_idx
+                if first_child_idx is not None:
+                    atom_node.insert_child(first_child_idx, string_leaf)
+            else:
+                # Else replace the atom node with the new string leaf.
+                replace_child(atom_node, string_leaf)
 
         # Build the final line ('new_line') that this method will later return.
         new_line = line.clone()
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -579,11 +579,9 @@ class StringMerger(StringTransformer, CustomSplitMapMixin):
             # when there is a standalone comment in the middle) ...
             if non_string_idx - string_idx < len(atom_node.children):
                 # We need to replace the old STRING leaves with the new string leaf.
-                first_child_idx: Optional[int] = None
-                for idx in range(string_idx, non_string_idx):
-                    child_idx = LL[idx].remove()
-                    if first_child_idx is None:
-                        first_child_idx = child_idx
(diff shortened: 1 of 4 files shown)


src/black/trans.py
```diff
@@ -572,7 +575,20 @@ def make_naked(string: str, string_prefix: str) -> str:
...
+            # when there is a standalone comment in the middle) ...
+            if non_string_idx - string_idx < len(atom_node.children):
+                # We need to replace the old STRING leaves with the new string leaf.
+                first_child_idx: Optional[int] = None
+                for idx in range(string_idx, non_string_idx):
+                    child_idx = LL[idx].remove()
+                    if first_child_idx is None:
+                        first_child_idx = child_idx
+                if first_child_idx is not None:
+                    atom_node.insert_child(first_child_idx, string_leaf)
```
ichard26, line 588:
```suggestion
            if non_string_idx - string_idx < len(atom_node.children):
                # We need to replace the old STRING leaves with the new string leaf.
                first_child_idx = LL[string_idx].remove()
                for idx in range(string_idx + 1, non_string_idx):
                    LL[idx].remove()
                atom_node.insert_child(first_child_idx, string_leaf)
```
I'm assuming that `LL[string_idx]` always points to a valid string Leaf here, but this should be a lot clearer.

ichard26 (approved):
Awesome work. Can't say I understand _everything_ here, but either way this logic should've existed in the first place. Thank you so much!

src/black/trans.py
yilei: great suggestion!

the `if first_child_idx is not None:` check is still needed for type checking, as `.remove()` returns an `Optional[int]`, though at runtime it shouldn't be `None`.

i updated with the latest commit.

PR #3232, opened 2022-08-19
Check if variable is already a bool before converting it

tshirtman:
### Description

I had issue running Black in vim after recent updates (notably using vim > 9, linked against python 3.10.4), getting error:

TypeError: bool object expected; got int

after some tinkering i found that the values defined (or not) in pyproject.toml were the culprit. But even adding
```
string_normalization = false                                    
fast = true
quiet = false
```
in the black block of pyproject.toml, didn't help, as it was trying to convert a bool to a bool, as if it was a string,

AttributeError: 'bool' object has no attribute 'lower'

but even adding quotes around the values still got me the original "bool object expected; got int", error.

So i replaced the usage of `strtobool` with something more robust, using `ConfigParser.BOOLEAN_STATES.get` and falling back on False.

### Checklist - did you ...

Not sure any of these are necessary, for the vim plugin.

### DIFF
--- a/autoload/black.vim
+++ b/autoload/black.vim
@@ -3,7 +3,7 @@ import collections
 import os
 import sys
 import vim
-from distutils.util import strtobool
+from configparser import ConfigParser
 
 
 class Flag(collections.namedtuple("FlagBase", "name, cast")):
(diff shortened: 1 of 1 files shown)


tshirtman:
Hm, seems i was not up to date, closing and reassessing.

ichard26:
No worries, glad you could sort out your issue in the end! :)

PR #3243, opened 2022-08-29
Strip even more redundant for parentheses

### LINKED ISSUE #3080: unnecessary parentheses in unpacking aren't removed
**Describe the style change**

Replace `(foo)` with `foo` when unpacking, where `foo` is an expression.

```python
points = [(1, 2), (3, 4)]
for (x), (y) in points:
    print(x, y)
```

**Desired style**

```python
points = [(1, 2), (3, 4)]
for x, y in points:
    print(x, y)
```

ichard26:
### Description

Fixes #3080.

What remains is removing unnecessary brackets because yes, you can do this:

```python
for [x, y] in points:
    print("this is valid code!!!")
```

I'm leaving this as a future task because it turns out simply converting these square brackets into parentheses is not an AST-safe transformation so letting the pre-existing logic handle the rest doesn't work. And `maybe_make_parens_invisible_in_atom` is currently not designed to handle square brackets and initial attempts to add a boolean `treat_square_brackets_as_parens` parameter were unsuccessful.

### Checklist - did you ...

### DIFF
--- a/tests/data/preview/remove_for_brackets.py
+++ b/tests/data/preview/remove_for_brackets.py
@@ -1,3 +1,6 @@
+for (((x))) in points:
+    pass
+
 # Only remove tuple brackets after `for`
 for (k, v) in d.items():
     print(k, v)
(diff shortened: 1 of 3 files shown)


tests/data/preview/remove_for_brackets.py
```diff
@@ -1,3 +1,6 @@
+for (((x))) in points:
```
JelleZijlstra, line 1: Can you add some tests where the assignment target is too long to fit in one line? Like `for (many, variables, that, go, on, over, eighty, eight, characters) in some_list:`.

JelleZijlstra (commented):
Thanks! I wouldn't worry about square brackets; that's a different and more rarely used syntax that we can handle later if we choose to make more AST-unsafe changes.

tests/data/preview/remove_for_brackets.py
ichard26: Done, 51e5be8dafcc97ec5c6f15ec51bcb8aa02b48acb.

tests/data/preview/remove_for_brackets.py
```diff
@@ -29,6 +81,21 @@
         module._verify_python3_env = lambda: None
 
 # Brackets remain for long for loop lines
+for (
+    one_super_long_name_as_the_for_target_list_because_why_not_you_dont_control_me
+) in points:
+    pass
+
+for one_super_long_name_as_the_for_target_list_because_why_not_you_dont_control_me_i_have_total_freedom in (points):
```
ichard26, line 89: The odd parentheses is a pre-existing bug in main. Something in `right_hand_split` or something similar is deciding to make the invisible parentheses around `points` visible even though the split failed. I'm happy to try to fix this before merging, but I probably won't get to it until September.

ichard26:
diff-shades is timing out, it could be simply GHA flakiness, but there could be a genuine infinite loop (or massive slowdown) somewhere. Since I'm in no rush to land this I'll mark this as draft.

cobaltt7:

PR #3246, opened 2022-08-31
add preview flag in vim plugin

SyPeter:
In order to use the --preview mode in the vim plugin this PR adds the flag to the needed vim files (autoload / plugin).

### DIFF
--- a/autoload/black.vim
+++ b/autoload/black.vim
@@ -30,6 +30,7 @@ FLAGS = [
   Flag(name="skip_string_normalization", cast=strtobool),
   Flag(name="quiet", cast=strtobool),
   Flag(name="skip_magic_trailing_comma", cast=strtobool),
+  Flag(name="preview", cast=strtobool),
 ]
 
 
@@ -145,6 +146,7 @@ def Black(**kwargs):
     string_normalization=not configs["skip_string_normalization"],
     is_pyi=vim.current.buffer.name.endswith('.pyi'),
     magic_trailing_comma=not configs["skip_magic_trailing_comma"],
+    preview=not configs["preview"],
     **black_kwargs,
   )
   quiet = configs["quiet"]
--- a/autoload/black.vim
+++ b/autoload/black.vim
@@ -146,7 +146,7 @@ def Black(**kwargs):
     string_normalization=not configs["skip_string_normalization"],
     is_pyi=vim.current.buffer.name.endswith('.pyi'),
     magic_trailing_comma=not configs["skip_magic_trailing_comma"],
-    preview=not configs["preview"],
+    preview=configs["preview"],
     **black_kwargs,
   )
   quiet = configs["quiet"]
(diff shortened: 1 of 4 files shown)


autoload/black.vim
```diff
@@ -145,6 +146,7 @@ def Black(**kwargs):
     string_normalization=not configs["skip_string_normalization"],
     is_pyi=vim.current.buffer.name.endswith('.pyi'),
     magic_trailing_comma=not configs["skip_magic_trailing_comma"],
+    preview=not configs["preview"],
```
ichard26, line 149: This doesn't seem right, `preview` isn't an inverted option like `skip-string-normalization`. 
```suggestion
    preview=configs["preview"],
```

ichard26 (changes requested):
Hi, thank you for the pull request! We depend on community contributions for the Vim plugin so it's really appreciated :black_heart: 

Anyway could you add an entry in the changelog under the integrations header? Don't forget to include your PR's number at the end. I also noted a potential error, but I may be wrong since I'm not that familiar with the Vim plugin.

autoload/black.vim
SyPeter: You are absolutely right! The `not` is wrong here.

ichard26 (approved):
Let's ship it! :ship:

ichard26:
Congrats on your first contribution here! If you have any feedback to share, let us know. Have a great day/night!

mvilanova:
Started getting the error in the screenshot when using the black plugin for Vim after using the changes introduced in this PR. Let me know if there's anything I can help with to troubleshoot and fix the issue.

Screen Shot 2022-09-04 at 12 03 01 PM

mvilanova:
I was able to fix the issue. Brew upgraded Vim to the the latest version, which depends on `python@3.10`. Deleting the `~/.vim/black` directory and explicitly calling Black from within Vim recreated the directory and solved the issue.

ichard26:
The root problem here that a clean reinstall fixed is that the Black installed in your Vim setup is too old and doesn't support the `--preview` flag. It's been out for eight months (since 22.1.0).

mvilanova:
@ichard26 what's the best way to keep it up to date? I run `:PluginUpdate` within Vim on a regular basis, but I think that only updates the code in `~/.vim/bundle/black`?

ichard26:
[According to our docs](https://black.readthedocs.io/en/stable/integrations/editors.html#vim) `:BlackUpgrade` is the command you're looking for.

GuiMarthe:
Hey folks! Thanks for this fix :D 

How can I know which python/black the vim `Black` plugin is pointing to? I have too many versions locally :sweat_smile:

ichard26:
@GuiMarthe [according to the docs](https://black.readthedocs.io/en/stable/integrations/editors.html#vim) there's no easy way of getting the Python executable filepath, but you can check the Black version via `:BlackVersion`. IIRC the plugin creates its own virtual environment to put a copy of Black in though,, so I'm not sure how useful `:BlackPythonInfo` or whatever would be. I don't know the plugin well though so perhaps I'm missing something :)

PR #3252, opened 2022-09-01
Added an option for number of spaces before inline comments

excitoon:
### Description

Well this is most obvious thing that shall be tuned after the length of line.

### Checklist - did you ...

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -36,6 +36,7 @@ from _black_version import version as __version__
 from black.cache import Cache, get_cache_info, read_cache, write_cache
 from black.comments import normalize_fmt_off
 from black.const import (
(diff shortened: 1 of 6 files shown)


JelleZijlstra:
Thanks for your work, but Black generally tries to minimize the number of configuration switches. There's been very little demand for configurability in this area, so I see no need to add a new option.

excitoon:
@JelleZijlstra Be open. "Members of the community are open to collaboration, whether it's on PEPs, patches, problems, or otherwise."

excitoon:
Fixed it in https://github.com/excitoon/lack/commit/a2d6493dcf2cc061159489ed9d4344b3a03de66d . New name was inspired by lack of openness in PSF.

PR #3254, opened 2022-09-03
Return pip install black[jupyter] hint with syntax to match platform

Mitchellpkt:
### Description

This change updates the message returned when Jupyter dependencies are not installed, so that it contains syntax corresponding to the platform.

Previously, attempting to apply black formatting to an ipynb notebook without black[jupyter] installed always returned the hint to run `pip install black[jupyter]`. This command does not work on Ubuntu with python3.

With this change, in the same scenario, the hint is `pip install 'black[jupyter]'` if `"linux"` is in the platform name, and the old hint `pip install black[jupyter]` otherwise.

The only function changed in the main codebase was `black.handle_ipynb_magics.jupyter_dependencies_are_installed`

I updated `tests/test_no_ipynb.py` to mock linux and non-linux versions of the existing tests.

Because this PR does not impact formatter rules or behavior, I did not update the CHANGELOG or documentation. However if I skipped any steps or missed anything please let me know, and I will take care of it :)

### Checklist - did you ...

- [X] Add a CHANGELOG entry if necessary? -> Not sure if applicable
- [X] Add / update tests if necessary? -> Yes (`tests/test_no_ipynb.py`)
- [X] Add new / update outdated documentation? -> N/A

### DIFF
--- a/src/black/handle_ipynb_magics.py
+++ b/src/black/handle_ipynb_magics.py
@@ -3,6 +3,7 @@
 import ast
 import collections
 import dataclasses
+import platform
 import secrets
 import sys
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


zsol:
Nit: this is not platform-dependent, but shell-dependent behavior.

Having said that: Couldn't we just change the prompt to be `pip install "black[jupyter]"` instead for all platforms?

Mitchellpkt:
Ah yes, technically shell not platform, my bad.

`pip install 'black[jupyter]'` worked for me on Linux and Mac, but not Windows.

Windows test:
```
> pip install 'black[jupyter]'
ERROR: Invalid requirement: "'black[jupyter]'"
```

jack1142:
You can use double quotes which should work on all shells.

Mitchellpkt:
That seems much more straightforward! I can make a fresh PR later without the message switching :+1:

Issue #3264, opened 2022-09-08
How do I match GitHub action with stability policy?


lofidevops:
**Is this related to a problem? Please describe.**

The pages on [stability policy](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy) and [GitHub Actions](https://black.readthedocs.io/en/stable/integrations/github_actions.html) are very clear. However, I don't think it's possible to enforce stability with GitHub actions.

**Describe the solution you'd like**

I would like to add `uses: psf/black@2022` in my GitHub action, and have this be equivalent to `black ~= 22.0` (or `black < 23`) in the developer toolchain. I believe this would require adding a `2022` tag, updating it on every release, then adding a `2023` tag for the first `23.x` release.

**Describe alternatives you've considered**

 * Pin specific version in GitHub action, leave toolchain at `~= 22.0`. When developer toolchain is updated to `~= 23.0`, re-pin in GitHub action. Downside: chance of discrepancy between 22.x (GitHub) and 22.y (toolchain).
 * Similar options (e.g. switch toolchain to `--preview` in November/December) require even more coordination.
 * Do not pin GitHub action or toolchain. Downside: defeats the point of stability policy, which I assume is to prevent scrambling in early 2023.
 * Pin specific version of GitHub Action and toolchain. Dowside: must update for every single release.

**Additional context**

This is a professional team, so the goal is to eliminate non-critical technical alerts. It could be that the final alternative is the only viable solution. Or there's a simple solution I have missed :D

jack1142:
You can just set version argument in the action to `<23`. The stability policy only applies to the code style so what matters is Black version, not the action version.

lofidevops:
@jack1142 Ok, so this example adapted from the docs would work. Thanks!

```
- uses: psf/black@stable
  with:
    options: "--check --verbose"
    src: "./src"
    version: "<23"
```

ichard26:
I don't believe that will work right now since if [the version option is provided `==` is simply prefixed](https://github.com/psf/black/blob/main/action/main.py#L19), breaking version ranges. I'd be happy to change the action to be smarter (defaulting to `==` but instead adds nothing if there's a version specifier type marker).

cooperlees:
> @jack1142 Ok, so this example adapted from the docs would work. Thanks!
> 
> ```
> - uses: psf/black@stable
>   with:
>     options: "--check --verbose"
>     src: "./src"
>     version: "<23"
> ```

[...]

PR #3265, opened 2022-09-08
Support version specifiers in GH action

jack1142:
### Description

Addresses the lack of support for version specifiers in GH action that was mentioned in #3264.

### Checklist - did you ...

    - I think there aren't any

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,9 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- Update GitHub action to support use of version specifiers (i.e. `<23`) for Black
+  version (#3265)
+
 ### Documentation
 
 <!-- Major changes to documentation and policies. Small docs changes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,7 +42,7 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
-- Update GitHub action to support use of version specifiers (i.e. `<23`) for Black
+- Update GitHub Action to support use of version specifiers (e.g. `<23`) for Black
   version (#3265)
 
 ### Documentation
--- a/action/main.py
+++ b/action/main.py
@@ -14,9 +14,10 @@ VERSION = os.getenv("INPUT_VERSION", default="")
 
 run([sys.executable, "-m", "venv", str(ENV_PATH)], check=True)
 
-req = "black[colorama]"
-if VERSION:
-    req += f"=={VERSION}"
+version_specifier = VERSION
+if VERSION and VERSION[0] in "0123456789":
+    version_specifier = f"=={VERSION}"
+req = f"black[colorama]{version_specifier}"
 pip_proc = run(
     [str(ENV_BIN / "python"), "-m", "pip", "install", req],
     stdout=PIPE,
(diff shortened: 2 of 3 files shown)


action/main.py
```diff
@@ -14,9 +14,10 @@
 
 run([sys.executable, "-m", "venv", str(ENV_PATH)], check=True)
 
-req = "black[colorama]"
-if VERSION:
-    req += f"=={VERSION}"
+version_specifier = VERSION
+if VERSION and VERSION[0] in "0123456789":
```
JelleZijlstra, line 18:
```suggestion
if VERSION and VERSION[0].isdigit():
```

jack1142, line 18: I could do `VERSION[0].isnumeric()` here though that's rather wide due to it supporting all Unicode characters that use `Numeric_Type` class. Personally, I tend to avoid isdecimal/isdigit/isnumeric methods because they evaluate to True for more than I actually want but I can change this.

jack1142: Ah damn, I didn't notice this comment before I made mine :smile: I can change it, though I would like to note that isdecimal is less wide check than isdigit so it might be better choice if we're switching it.

felix-hilden: I guess it's fine either way, since we only care about recognising version specifier characters, which won't be decimal or digits. But no biggie.

jack1142:
I added the documentation so it should be good to go now. Just let me know (or update the PR yourself) whether I should use `isdecimal()` or leave it as is.

action/main.py
ichard26: Do packaging tools normalize non-ascii digits? If so, I'd prefer `isdecimal()` otherwise I'd prefer being stricter.

CHANGES.md
```diff
@@ -42,6 +42,9 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- Update GitHub action to support use of version specifiers (i.e. `<23`) for Black
+  version (#3265)
```
ichard26, line 46:
```suggestion
- Update GitHub Action to support use of version specifiers (e.g. `<23`) for Black
  version (#3265)
```

ichard26 (approved):
LGTM (although I didn't test this, I hope there isn't a bug somewhere). Thank you very much!

action/main.py
jack1142: `pip` doesn't handle Unicode numbers:
```
❯ pip install 'black==२२.१'
ERROR: Could not find a version that satisfies the requirement black==२२.१ (from versions: 18.3a0, 18.3a1, 18.3a2, 18.3a3, 18.3a4, 18.4a0, 18.4a1, 18.4a2, 18.4a3, 18.4a4, 18.5b0, 18.5b1, 18.6b0, 18.6b1, 18.6b2, 18.6b3, 18.6b4, 18.9b0, 19.3b0, 19.10b0, 20.8b0, 20.8b1, 21.4b0, 21.4b1, 21.4b2, 21.5b0, 21.5b1, 21.5b2, 21.6b0, 21.7b0, 21.8b0, 21.9b0, 21.10b0, 21.11b0, 21.11b1, 21.12b0, 22.1.0, 22.3.0, 22.6.0, 22.8.0)
ERROR: No matching distribution found for black==२२.१
```

So in the end, it shouldn't matter whether we use `isdecimal()` or the `in` check as it will fail anyway.

PR #3266, opened 2022-09-08
[FIX] migrate-black.py: don't fail on binary files

hbrunn:
without this, the script fails if any commit contains non-text files

[docs](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---binary)

### DIFF
--- a/scripts/migrate-black.py
+++ b/scripts/migrate-black.py
@@ -49,6 +49,7 @@ def blackify(base_branch: str, black_command: str, logger: logging.Logger) -> in
             [
                 "git",
                 "diff",
+                "--binary",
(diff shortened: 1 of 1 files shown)


hbrunn:
i think the skip-news label is appropriate here

cooperlees (approved):
Thanks. why not. TIL too ...

Will leave it for a little bit to give some other people the change to review / question.

hbrunn:
this came up as the only hiccup with a migration of a mid size project to black, the devs were otherwise happy

ichard26 (approved):
I appreciate the follow-up!

PR #3270, opened 2022-09-14
Improve order of paragraphs on line splitting

TomFryers:
### Description

These two paragraphs were tucked away at the end of the section, after the diversion on backslashes. I nearly missed the first paragraph and opened a nonsense issue, and I think the second belongs higher up with it too.

### Checklist - did you ...

A changelog entry is unnecessary, I believe.
N/A
Yes

### DIFF
--- a/docs/the_black_code_style/current_style.md
+++ b/docs/the_black_code_style/current_style.md
@@ -85,6 +85,19 @@ def very_important_function(
         ...
 ```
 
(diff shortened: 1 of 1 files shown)


felix-hilden (approved):
Seems good to me 👍

ichard26 (approved):
Thank you, this makes much more sense!

PR #3274, opened 2022-09-17
Make context manager examples in future style docs consistent

Blandes22:
### Description

Noticed a typo in line 18, all calls to make_context_manager() had the number outside of parentheses.  
Changed `make_context_manager1()` to  `make_context_manager(1) `  for all four calls.

This should get a label to silence the CHANGELOG entry check.

### DIFF
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -15,7 +15,7 @@ managers.
 We don't want formatting like:
 
 ```py3
-with make_context_manager(1) as cm1, make_context_manager(2) as cm2, make_context_manager(3) as cm3, make_context_manager(4) as cm4:
+with make_context_manager1() as cm1, make_context_manager2() as cm2, make_context_manager3() as cm3, make_context_manager4() as cm4:
     ...  # nothing to split on - line too long
 ```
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
The current code seems correct to me. The intention is that we're making four different context managers with generic names, not that we're passing a number to the same context manager.

ichard26:
We should probably fix the examples that come after then as the numbers are used as arguments.

Blandes22:
Updated the PR to invert the changes I had one, returning the code block at line 18 to its original format and updating the code blocks starting at lines 24 and 41 to match the format of the previous context managers.

ichard26 (approved):
Awesome, thank you! It's embarrassing that I missed this when I reviewed the PRs that added these examples 😅

ichard26:
Sorry, was this intentional @Blandes22?

Blandes22:
It was not intentional, I was just trying to clean up my home page. Did I mess everything up?

ichard26:
You can recover your deleted fork of Black which will allow you to reopen this PR, but if you would prefer, I can refile this PR (keeping the author information attributed to you!) and recover your work that way.

Blandes22:
I will go ahead and recover the fork I deleted and reopen this pr. Sorry about that, that was my mistake.

ichard26:
Don't worry about it, it happens. All that matters is that you can recover from your mistakes -- which you can in this case! :)

Blandes22:
I think it should all be okay now, if there is something I missed please let me know.

PR #3275, opened 2022-09-17
Always call freeze_support() if sys.frozen is True

### LINKED ISSUE #3250: Fix start up code to also call freeze_support() on macOS when frozen so native macOS executables work
I broke the native macOS executables. [The build for 22.8.0 failed](https://github.com/psf/black/runs/8124297895?check_suite_focus=true) since the executable failed the basic sanity checks pre-upload
```
Usage: black_macos [OPTIONS] SRC ...
Try 'black_macos -h' for help.

Error: No such option: --multiprocessing-fork
Usage: black_macos [OPTIONS] SRC ...
Try 'black_macos -h' for help.

Usage: black_macos [OPTIONS] SRC ...
Try 'black_macos -h' for help.

Error: No such option: --multiprocessing-fork
Error: No such option: --multiprocessing-fork
error: cannot format src/black/__main__.py: A process in the process pool was terminated abruptly while the future was running or pending.
error: cannot format src/black/files.py: A process in the process pool was terminated abruptly while the future was running or pending.
```

When I first submitted https://github.com/psf/black/pull/3211 I checked how the stdlib guards the `freeze_support()` call and copied it:

- https://github.com/python/cpython/blob/2ecc195498f3b1256fabc2b66e0d8f6d671fa1d7/Lib/multiprocessing/context.py#L144-L150

So as far as I understand from reading the code, [macOS should not need `freeze_support()` to work with PyInstaller](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count:~:text=Calling%20freeze_support()%20has%20no%20effect%20when%20invoked%20on%20any%20operating%20system%20other%20than%20Windows), but apparently it does. Anyway let's fix this by including macOS in the OS check I guess.

https://github.com/psf/black/blob/2018e667a6a36ee3fbfa8041cd36512f92f60d49/src/black/__init__.py#L1377-L1384

ichard26:
### Description

This shouldn't be necessary, but in practice macOS also requires freeze_support() so let's play it safe and always call it when sys.frozen is truthy. Only a tiny sliver of users will be using these PyInstaller executables and the start-up performance is already terrible as the executable unzips itself on the fly anyway.

Fixes #3250.

### Checklist - did you ...

### Review notes

See https://github.com/ichard26/black/actions/runs/3074931212 which proves this works.


JelleZijlstra:
Are you sure this fixes the issue? It really does seem the case that freeze_support() only does anything on Windows: https://github.com/python/cpython/blob/246a044641388dfd5023cb4b8dbe5519cb41d943/Lib/multiprocessing/context.py#L144

ichard26:
I honestly don't know. I ran a test run with this PR and macOS passed: https://github.com/ichard26/black/actions/runs/3074931212 (see the issue for a link to the failing run for 22.8.0) but I still don't understand why it's broken in the first place.

JelleZijlstra:
I guess it's harmless to merge this, but maybe the previous failure was just random flakiness?

ichard26:
[Just triggered a rerun of the 22.8.0 macOS job to see if flakiness is the issue](https://github.com/psf/black/actions/runs/2967555863/jobs/5041109540).

JelleZijlstra (approved):

PR #3293, opened 2022-09-28
add \.ipynb_checkpoints to DEFAULT_EXCLUDES

raybellwaves:
### Description

I believe most users would not want to run black[jupyter] on files in .ipynb_checkpoints/.

This PR adds .ipynb_checkpoints/ to DEFAULT_EXCLUDES in src/black/const.py

I did a quick issue search and couldn't see if a PR like this is outside of scope or has been rejected in the past (https://github.com/psf/black/issues?q=is%3Aissue+ipynb_checkpoints)

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,7 +27,7 @@
 
 <!-- Changes to how Black can be configured -->
 
-- Files in \.ipynb_checkpoints are excluded by default. (#3293)
+\.ipynb_checkpoints
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,7 +27,7 @@
 
 <!-- Changes to how Black can be configured -->
 
-\.ipynb_checkpoints
+- `.ipynb_checkpoints` directories are now excluded by default (#3293)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,8 @@
 
 <!-- Changes to how Black can be configured -->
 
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Seems reasonable, but could you link to some documentation on these files so we can verify that it makes sense to exclude them?

Also, please add a changelog entry as this is a user-facing change.

raybellwaves:
> Seems reasonable, but could you link to some documentation on these files so we can verify that it makes sense to exclude them?

What is the path .ipynb_checkpoints/ ?
---------------------------------------

The best information I found is in stack overflow - https://stackoverflow.com/a/46422176/6046019

"Jupyter creates a checkpoint file every single time you create an .ipynb file, and then it updates the checkpoint file every single time you manually save your progress for the initial .ipynb."

"the checkpoint file is located within a hidden folder named .ipynb_checkpoints"

> Also, please add a changelog entry as this is a user-facing change.

Added in the Configuration section of the CHANGES.md

JelleZijlstra (approved):
Thanks!

CHANGES.md
```diff
@@ -27,6 +27,8 @@
 
 <!-- Changes to how Black can be configured -->
 
+- Files in \.ipynb_checkpoints are excluded by default. (#3293)
```
isidentical, line 30:
```suggestion
\.ipynb_checkpoints
```

ichard26 (approved):
Thank you for the PR! I know almost nothing about Jupyter Notebooks, you learn something new everyday :)

PR #3257, opened 2022-09-05
(🎁) Preserve crlf line endings in blackd

KotlinIsland:
### Description

Added CRLF line ending preservation to blackd, perfect for gorgeous looks 🚀

### Checklist - did you ... (know?)

- [x] Add a CHANGELOG entry if necessary?
- [x] Add / update tests if necessary?
- [x] Add new / update outdated documentation?

-  resolves #3255

PR #3299, opened 2022-09-29
Add option to skip the first line of source code

### LINKED ISSUE #3214: add -x : skipping first line like cpython does
https://docs.python.org/dev/using/cmdline.html#cmdoption-x
allows for non standard shebang on dos and web source files.
Black should allow the exact same option allowing to skip the first line and leave it untouched.

web use case : https://pygame-web.github.io/wiki/pygame-script/

https://pygame-web.github.io/showroom/pygame-scripts/org.pygame.touchpong.html
```
wget https://pygame-web.github.io/showroom/pygame-scripts/org.pygame.touchpong.html
# assuming pygame is installed
python3 -x org.pygame.touchpong.html
```

aaossa:
### Description

Adds a new CLI option to skip the first line of source code while formatting. To enable use the `-x`/`--skip-source-first-line` flag. The skipped line will be retained in the output. The option is also available for blackd.

Closes #3214 

### Checklist - did you ...

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -811,8 +811,8 @@ def format_file_in_place(
         raise ValueError(
             f"File '{src}' cannot be parsed as valid Jupyter notebook."
         ) from None
-    src_contents = header.decode() + src_contents
-    dst_contents = header.decode() + dst_contents
+    src_contents = header.decode(encoding) + src_contents
+    dst_contents = header.decode(encoding) + dst_contents
 
     if write_back == WriteBack.YES:
         with open(src, "w", encoding=encoding, newline=newline) as f:
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -798,7 +798,10 @@ def format_file_in_place(
         mode = replace(mode, is_ipynb=True)
 
     then = datetime.utcfromtimestamp(src.stat().st_mtime)
+    header = b""
     with open(src, "rb") as buf:
+        if mode.skip_source_first_line:
+            header = buf.readline()
         src_contents, encoding, newline = decode_bytes(buf.read())
     try:
         dst_contents = format_file_contents(src_contents, fast=fast, mode=mode)
@@ -808,6 +811,8 @@ def format_file_in_place(
         raise ValueError(
             f"File '{src}' cannot be parsed as valid Jupyter notebook."
         ) from None
+    src_contents = header.decode() + src_contents
+    dst_contents = header.decode() + dst_contents
 
     if write_back == WriteBack.YES:
         with open(src, "w", encoding=encoding, newline=newline) as f:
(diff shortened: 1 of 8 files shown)


tests/test_black.py
```diff
@@ -341,6 +341,20 @@ def test_string_quotes(self) -> None:
         black.assert_equivalent(source, not_normalized)
         black.assert_stable(source, not_normalized, mode=mode)
 
+    def test_skip_source_first_line(self) -> None:
```
JelleZijlstra, line 344: Add a test case where there is some valid code after the first line.

Also, what if the file doesn't use `\n` to separate lines?

aaossa: I'll add a valid code after the first line in the test file:

```python
This is not valid Python syntax
y = "This is valid syntax"
```

About the second concern, it is indeed valid. I solved it by using the `newline` variable that comes from `decode_bytes` since it's already available. But now several Blackd tests fail because the variable is not available when Blackd invokes `black.format_file_contents`.

https://github.com/psf/black/blob/07b68e2425d2caadc2edf8fe021f60a32c6f44a7/src/blackd/__init__.py#L132-L134

This seems related to a WIP at #3257 . Maybe a `newline` variable could be defined in Blackd before calling `black.format_file_contents`, using the same check that's implemented in the cited PR, something like:

```python
if req_str[req_str.find("\n") - 1] == "\r":
    newline = "\r\n"
else:
    newline = "\r\n"
```

That would work for me, because then in this PR I could invoke `black.format_file_contents` as:

```python
formatted_str = await loop.run_in_executor(
    executor, partial(black.format_file_contents, req_str, fast=fast, mode=mode, newline=newline)
)
```

Should I wait for #3257 to be merged and then apply this change? A possible patch is using `os.linesep`, but that completely ignores the current approach (Black takes the first newline in the file as a reference)

JelleZijlstra: I will hopefully be able to land #3257 soon.

aaossa: Now that #3257 has been merged, I'll rebase my PR and apply the changes listed above 👌 (Creating the `newline` variable first, pass the argument to `black.format_file_contents`, and use the variable in the code introduced by #3257 )

aaossa:
Just though of something... What if the first line ends with a different character than the second line? For example:

```python
Line skipped using -x but ends with \n
"First source code line, ending with \r\n"
```

At the moment, the `newline` argument for `format_file_contents` is defined according to the first line (`decode_bytes`), and then the first line is skipped. In this example, `decode_bytes` will return `newline = "\n"` (first line) but the first not-skipped line will end in `\r\n`. This is what happens in the current implementation. If this is a problem, maybe `decode_bytes` should know if the first line is skipped and define `newline` according to the second line of the file (if present), but I'm not sure what's the expected behavior or even if its relevant enough now.

---

BTW, adding an argument seems a bit too much now, because my current implementation (WIP locally) only uses the `newline` argument when the first line is skipped, but the value is not passed further. Maybe applying the partition on `"\n"` is the best approach at the moment since a line ending in `\r\n` should be split too (`\r` will remain untouched as part of the header)

src/black/__init__.py
```diff
@@ -800,6 +811,8 @@ def format_file_in_place(
         raise ValueError(
             f"File '{src}' cannot be parsed as valid Jupyter notebook."
         ) from None
+    src_contents = header.decode() + src_contents
```
JelleZijlstra, line 814: Doesn't this need to use the encoding we got from `decode_bytes()`?

JelleZijlstra (approved):

PR #3302, opened 2022-09-30
Enforce empty lines before classes/functions with sticky leading comments.

### LINKED ISSUE #246: Black doesn't enforce two empty lines before functions on module level when they have a sticky leading comment
Operating system: macOS High Sierra
Python version: 3.6.5
Black version: 18.5b0
pycodestyle version: 2.4.0

Should Black insert a newline between these variables and the function?

### pycodestyle complains about test.py but Black doesn't change it

```console
$ cat test.py
```
```python
from __future__ import print_function, unicode_literals

all_data = []
all_versions = set()

# Comment
def thing():
    pass
```
```console
$ pycodestyle test.py
test.py:7:1: E302 expected 2 blank lines, found 1
$ black test.py
/private/tmp/test.py already well formatted, good job.
```

### Both pycodestyle and Black are fine with test2.py

```console
$ cat test2.py
```
```python
from __future__ import print_function, unicode_literals

all_data = []
all_versions = set()


# Comment
def thing():
    pass
```
```console
$ pycodestyle test2.py
$ black test2.py
/private/tmp/test2.py already well formatted, good job.
```

### Black reformats test3.py, with no comment

```console
$ cat test3.py
```
```python
from __future__ import print_function, unicode_literals

all_data = []
all_versions = set()

def thing():
    pass
```
```console
$ pycodestyle test3.py
test3.py:6:1: E302 expected 2 blank lines, found 1
$ black test3.py
reformatted /private/tmp/test3.py
$ pycodestyle test3.py
$ cat test3.py
```
```python
from __future__ import print_function, unicode_literals

all_data = []
all_versions = set()


def thing():
    pass
```

### LINKED ISSUE #3300: Black is inconsistent with empty lines before an inner function after code block open, with or without leading comments
**To Reproduce**

Unformatted code:

```python
def main():
    if a:

        # Comment
        def b():
            pass

    if b:
        def c():
            pass
```

Running `black --preview`:

```python
def main():
    if a:
        # Comment
        def b():
            pass

    if b:

        def c():
            pass
```

**Expected behavior**

Empty line should be consistent added (or removed) between the code block open and inner function:

```python
def main():
    if a:

        # Comment
        def b():
            pass

    if b:

        def c():
            pass
```

Or,

```python
def main():
    if a:
        # Comment
        def b():
            pass

    if b:
        def c():
            pass
```

**Additional context**

I did a bisect and this was caused by https://github.com/psf/black/pull/3035.

Note that if the inner function doesn't have a leading comment, it won't remove the empty line:

```python
def main():
    if a:

        def b():
            pass

        def c():
            pass
```

Thus I believe this is an undesired behavior change in #3035?

yilei:
### Description

Enforce empty lines before classes/functions with leading comments:

- Fixes #246

`EmptyLineTracker` works on lines that haven't been split yet. When sticky leading comment is used before class/def lines, the tracker must look behind the possible comment lines. So I introduced a new `LinesBlock` class to keep track of those states.

### Checklist - did you ...

### DIFF
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -506,10 +506,9 @@ class EmptyLineTracker:
 
         # Maintain the semantic_leading_comment state.
         if current_line.is_comment:
-            if (
-                self.semantic_leading_comment is None
-                or self.previous_line is None
-                or (before and not self.previous_line.is_decorator)
+            if self.previous_line is None or (
+                not self.previous_line.is_decorator
+                and (self.semantic_leading_comment is None or before)
             ):
                 self.semantic_leading_comment = block
         elif not current_line.is_decorator:
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -513,6 +513,8 @@ class EmptyLineTracker:
         if current_line.is_comment:
             if self.previous_line is None or (
                 not self.previous_line.is_decorator
+                # `or before` means this comment already has an empty line before
+                and (not self.previous_line.is_comment or before)
                 and (self.semantic_leading_comment is None or before)
             ):
                 self.semantic_leading_comment = block
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -458,6 +458,7 @@ class LinesBlock:
 
     mode: Mode
     previous_block: Optional["LinesBlock"]
+    is_class: bool  # Whether the original line is a class def.
     before: int = 0
     content_lines: List[str] = field(default_factory=list)
     after: int = 0
@@ -500,7 +501,11 @@ class EmptyLineTracker:
             if self.previous_line is None
             else before - previous_after
         )
-        block = LinesBlock(mode=self.mode, previous_block=self.previous_block)
+        block = LinesBlock(
+            mode=self.mode,
+            previous_block=self.previous_block,
+            is_class=current_line.is_class,
+        )
         block.before = before
         block.after = after
 
@@ -623,6 +628,7 @@ class EmptyLineTracker:
                 in current_line.mode
                 and slc is not None
                 and slc.previous_block is not None
+                and not slc.previous_block.is_class
                 and slc.before <= 1
             ):
                 comment_to_add_newlines = slc
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -458,7 +458,7 @@ class LinesBlock:
 
     mode: Mode
     previous_block: Optional["LinesBlock"]
-    is_class: bool  # Whether the original line is a class def.
+    original_line: Line
     before: int = 0
     content_lines: List[str] = field(default_factory=list)
     after: int = 0
@@ -504,7 +504,7 @@ class EmptyLineTracker:
         block = LinesBlock(
             mode=self.mode,
             previous_block=self.previous_block,
-            is_class=current_line.is_class,
+            original_line=current_line,
             before=before,
             after=after,
         )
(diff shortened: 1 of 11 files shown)


yilei:
I pushed a few new commits to address a few unexpected diffs from diff-shades.

I read through the latest diff-shades results again, and they look all expected.

src/black/lines.py
```diff
@@ -458,33 +480,55 @@ class EmptyLineTracker:
...
+            else before - previous_after
         )
-        self.previous_after = after
+        block = LinesBlock(
+            mode=self.mode,
+            previous_block=self.previous_block,
+            is_class=current_line.is_class,
+        )
+        block.before = before
+        block.after = after
```
JelleZijlstra, line 510:
```suggestion
            is_class=current_line.is_class,
            before=before,
            after=after,
        )
```

Better to avoid modifying the object after creation.

JelleZijlstra (approved):

tests/data/preview/comments9.py
```diff
@@ -0,0 +1,240 @@
...
+
+# leading 4 that already has an empty line
+def decorated_with_split_leading_comments():
+    pass
+
+
+def main():
+    if a:
+
+        # Leading comment before inline function
```
felix-hilden, line 204: I'd expect this to not contain the newline

tests/data/preview/comments9.py
```diff
@@ -0,0 +1,240 @@
+# Test for https://github.com/psf/black/issues/246.
+
+some = statement
+# This comment should be split from the statement above by two lines.
+def function():
```
felix-hilden, line 5: Could we also test two lines of comments? I'd expect them both to be moved.

felix-hilden (changes requested):
Thanks a ton! Some comments below, and one discussion point: would it make sense to consider the comments to be in the block? Instead of having an outer "semantic comment" I'm not 100 % sure, but that could simplify things so that we can use ordinary newline rules for the whole block. Maybe? 🤔

tests/data/preview/comments9.py
yilei: Done

tests/data/preview/comments9.py
yilei: Adding an empty line is consistent with inline functions without leading comments, just like module-level functions with leading comments:

```python
def function():
    if a:

        def inline():
            pass

        def another_inline():
            pass
```

yilei:
> would it make sense to consider the comments to be in the block? Instead of having an outer "semantic comment" I'm not 100 % sure, but that could simplify things so that we can use ordinary newline rules for the whole block.

This is appealing. But I tried and couldn't yet find a way to actually simplify this by considering the comments to be in the block.

To decide whether a comment belongs to this block, we have to examine what the block is, as well as [the newlines in the first leaf prefix](https://github.com/psf/black/blob/b60b85b234d6a575f636d0a125478115f993c90c/src/black/lines.py#L488-L493). This logic is already part of EmptyLineTracker. So this would require a big surgery to redesign EmptyLineTracker and likely also the LineGenerator.

All the empty line logic is still contained inside the EmptyLineTracker, so I feel this PR's implementation isn't too bad.

tests/data/preview/comments9.py
yilei: i.e. this is fixing https://github.com/psf/black/issues/3300

felix-hilden: Commented on the issue, let's continue there.

yilei:
PR updated to keep the current behavior for inner functions after code block open. This now only fixes #246.

tests/data/preview/comments9.py
yilei: This PR now keeps the behavior for inner function after block open.

felix-hilden (approved):

JelleZijlstra (approved):

Issue #3292, opened 2022-09-27
String processing: wrap concatenated strings in parens in all cases


yilei:
Exending the dicussions in #3260 and #2553, I'd like to once more try to bring up the topic of how black should format concatenated strings.

After a lot of reading and experimenting, I'm now in favor of always wrapping concatenated strings in parens, i.e. extending the cases from #3159 to all. Examples:

```python
# Unformatted:
function_call(
    " lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et dolore magna aliqua Ut enim ad minim",
    " veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo",
)
some_list = [
    " lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et dolore magna aliqua Ut enim ad minim",
    " veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo",
]

# v22.8.0 --preview:
function_call(
    " lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor"
    " incididunt ut labore et dolore magna aliqua Ut enim ad minim",
    " veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo",
)
some_list = [
    (
        " lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor"
        " incididunt ut labore et dolore magna aliqua Ut enim ad minim"
    ),
    " veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo",
]

# Proposed:
function_call(
    (
        " lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor"
        " incididunt ut labore et dolore magna aliqua Ut enim ad minim"
    ),
    " veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo",
)
some_list = [
    (
        " lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor"
        " incididunt ut labore et dolore magna aliqua Ut enim ad minim"
    ),
    " veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo",
]
```

This in turn makes #3260 and #2553 redundant/obsolete.

## Pros

1. Code is more readable since the extra indentation makes the scope more clear.
[...]

PR #3307, opened 2022-10-03
Wrap concatenated strings used as function args in parens.

yilei:
### Description

This implements #3292.

I have only heard one 👍 from @felix-hilden. But I think it's also easier to decide after seeing the formatting changes introduced by this PR, thus I'm sending it now so we can visualize the diffs.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,8 @@
 
 - Fix a crash when formatting some dicts with parenthesis-wrapped long string keys
   (#3262)
+- Implicitly concatenated strings used as function args are now wrapped inside
+  parentheses (#3307)
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,8 +14,6 @@
 
 <!-- Changes that affect Black's preview style -->
 
-- Fix a crash when formatting some dicts with parenthesis-wrapped long string keys
-  (#3262)
 - Enforce empty lines before classes and functions with sticky leading comments (#3302)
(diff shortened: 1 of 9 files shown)


CHANGES.md
```diff
@@ -14,7 +14,11 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Fix a crash when formatting some dicts with parenthesis-wrapped long string keys
+  (#3262)
```
```suggestion
```
Bad merge resolution on my part

src/black/trans.py
```diff
@@ -1062,6 +1062,7 @@ def _prefer_paren_wrap_match(LL: List[Leaf]) -> Optional[int]:
             syms.listmaker,
             syms.dictsetmaker,
             syms.testlist_gexp,
+            syms.arglist,
         ]
         # If the string is an immediate child of a list/set/tuple literal...
```
JelleZijlstra, line 1067: Should update this comment.

Also, since we're now adding parens in basically all contexts, could we get rid of `matching_nodes` entirely?

JelleZijlstra (commented):
Looking at the diff-shades output I'm still a bit nervous whether this is a net positive, since it does often feel like it's adding unnecessary lines. But I guess that's what we have a preview style for.

src/black/trans.py
yilei: Done, and yes we no longer need `matching_nodes`.

yilei:
Updated, please take another look. Thanks Jelle!

JelleZijlstra (approved):

Issue #3308, opened 2022-10-04
please support use of the system black installation


bugfood:
**Is your feature request related to a problem? Please describe.**

As a new user of black, I installed the `black` package from my distribution to try it out. After that, I wanted to integrate it into vim. I set up the plugin, but discovered that the plugin requires building a virtualenv within which to install black. Using a virtualenv is undesirable to me for the following reasons:
* I already have black installed; there should be no need to install it again.
* The virtualenv adds clutter to my `~/.vim` directory.
* I'd prefer to upgrade/manage black via my distribution's package manager.
* In order to use a virtualenv, I would need to install additional python packages for which I have no other immediate use.

**Describe the solution you'd like**

I'd prefer the vim plugin to use the system installation of black. Since having a virtualenv is apparently desirable to other people, this should be configurable.

**Describe alternatives you've considered**

I've considered using the virtualenv.

**Additional context**

Please review the PR which I will submit shortly.

Thanks,
Corey

krowbar:
Doesn't setting 
`g:black_virtualenv = "/usr/local/"`
somewhere in vim startup files (or specifically for python files) do that?
The path may vary depending on your distribution, it will search for the site-packages below that point (see _get_virtualenv_site_packages function)

bugfood:
> Doesn't setting `g:black_virtualenv = "/usr/local/"` somewhere in vim startup files (or specifically for python files) do that? The path may vary depending on your distribution, it will search for the site-packages below that point (see _get_virtualenv_site_packages function)

Hmm. I hadn't looked at that before now, but I don't think it's suitable. I'll elaborate below.

That function is:
```
def _get_virtualenv_site_packages(venv_path, pyver):
  if sys.platform[:3] == "win":
    return venv_path / 'Lib' / 'site-packages'
  return venv_path / 'lib' / f'python{pyver[0]}.{pyver[1]}' / 'site-packages'

```

...and it's called as (abridged):
```
  virtualenv_path = Path(vim.eval("g:black_virtualenv")).expanduser()
  virtualenv_site_packages = str(_get_virtualenv_site_packages(virtualenv_path, pyver))
  if not virtualenv_path.is_dir():
    # (set up black virtualenv)...
```

So it looks like `g:black_virtualenv` needs to refer to a directory that has subdirectories `lib/python3.10/site-packages`, at least for my python version.

This doesn't match anything in the Debian package structure.
```
$ dpkg -L black
/.
/usr
/usr/bin
/usr/bin/black
[...]

PR #3309, opened 2022-10-04
Vim plugin: allow using system black rather than virtualenv

bugfood:
Provide a configuration parameter to the Vim plugin which will allow the plugin to skip setting up a virtualenv. This is useful when there is a system installation of black (e.g. from a Linux distribution) which the user prefers to use.

Using a virtualenv remains the default.

### Description

I have several reasons why using a virtualenv is undesirable to me. I have listed them in the following feature request.
Fixes: https://github.com/psf/black/issues/3308

### Checklist - did you ...

- [Y] Add an entry in `CHANGES.md` if necessary?
- [N/A?] Add / update tests if necessary?
- [Y] Add new / update outdated documentation?

Thanks,
Corey

### DIFF
--- a/autoload/black.vim
+++ b/autoload/black.vim
@@ -56,6 +56,16 @@ def _get_virtualenv_site_packages(venv_path, pyver):
   return venv_path / 'lib' / f'python{pyver[0]}.{pyver[1]}' / 'site-packages'
 
 def _initialize_black_env(upgrade=False):
+  if vim.eval("g:black_use_virtualenv ? 'true' : 'false'") == "false":
+    if upgrade:
+      print("Upgrade disabled due to g:black_use_virtualenv being disabled.")
+      print("Either use your system package manager (or pip) to upgrade black separately,")
+      print("or modify your vimrc to have 'let g:black_use_virtualenv = 1'.")
+      return False
+    else:
+      # Nothing needed to be done.
+      return True
+
   pyver = sys.version_info[:3]
   if pyver < (3, 7):
     print("Sorry, Black requires Python 3.7+ to run.")
--- a/docs/integrations/editors.md
+++ b/docs/integrations/editors.md
@@ -104,7 +104,7 @@ Commands and shortcuts:
   - you can optionally pass `target_version=<version>` with the same values as in the
     command line.
 - `:BlackUpgrade` to upgrade _Black_ inside the virtualenv;
-- `:BlackVersion` to get the current version of _Black_ inside the virtualenv.
+- `:BlackVersion` to get the current version of _Black_ in use.
 
 Configuration:
 
@@ -159,6 +159,18 @@ restarting Vim.
 If you need to do anything special to make your virtualenv work and install _Black_ (for
 example you want to run a version from main), create a virtualenv manually and point
 `g:black_virtualenv` to it. The plugin will use it.
+
+If you would prefer to use the system installation of _Black_ rather than a virtualenv,
+then add this to your vimrc:
+
+```
+let g:black_use_virtualenv = 0
+```
+
+Note that the `:BlackUpgrade` command is only usable and useful with a virtualenv, so
+when the virtualenv is not in use, `:BlackUpgrade` is disabled. If you need to upgrade
+the system installation of _Black_, then use your system package manager or pip--
+whatever tool you used to install _Black_ originally.
 
 To run _Black_ on save, add the following lines to `.vimrc` or `init.vim`:
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,9 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- Vim plugin: Optionally allow using the system installation of Black via
+  `let g:black_use_virtualenv = 0`(#3309)
+
 ### Documentation
 
 <!-- Major changes to documentation and policies. Small docs changes
--- a/plugin/black.vim
+++ b/plugin/black.vim
@@ -63,6 +63,9 @@ endif
 if !exists("g:black_target_version")
   let g:black_target_version = ""
 endif
+if !exists("g:black_use_virtualenv")
+  let g:black_use_virtualenv = 1
+endif
 if !exists("g:black_preview")
   let g:black_preview = 0
 endif


docs/integrations/editors.md
```diff
@@ -160,6 +160,16 @@ If you need to do anything special to make your virtualenv work and install _Bla
...
 
+If you would prefer to use the system installation of _Black_ rather than a virtualenv,
+then add this to your vimrc:
+
+```
+let g:black_use_system = 1
+```
+
+Note that when the system installation is in use, the `:BlackUpgrade` command is not
+usable.
```
felix-hilden, line 171: Is it hard to make it work?

docs/integrations/editors.md
```diff
@@ -160,6 +160,16 @@ If you need to do anything special to make your virtualenv work and install _Bla
 example you want to run a version from main), create a virtualenv manually and point
 `g:black_virtualenv` to it. The plugin will use it.
 
+If you would prefer to use the system installation of _Black_ rather than a virtualenv,
+then add this to your vimrc:
+
+```
+let g:black_use_system = 1
```
felix-hilden, line 167:
```suggestion
let g:black_system_install = 1
```
Or something, "use system" isn't immediately that obvious

felix-hilden (commented):
Thanks for the PR, some comments below!

docs/integrations/editors.md
bugfood: It would be impractical (or impossible) for the vim plugin to upgrade the system installation of `black. Maybe that's not what you're asking about, though.

Are you asking whether the vim plugin could still upgrade the virtualenv installation even when `g:black_use_system` is true? I think yes--it would be easy to redo the logic (untested):
```
  if vim.eval("g:black_use_system ? 'true' : 'false'") == "true" and not upgrade:
     # Nothing needed to be done.
      return True
```

...but I don't think that would be very good from a user standpoint. If a user has set `g:black_use_system`, then they have indicated that that do not want to use a virtualenv, so installing a virtualenv (or upgrading an existing one) would be pointless--the virtualenv won't be used anyway.

At worst, if a user has `g:black_use_system` enabled by error and _does_ want to use a virtualenv, then when installing/upgrading the virtualenv has no effect on the `black` version in use, that would make it appear as if the virtualenv installation/upgrade has done something wrong.


This is all from my perspective, at least--I just think it would be least surprising to have `:BlackUpgrade` visibly fail rather than do something that is unwanted or have the results be masked. If I'm wrong on that, let me know and I can certainly change the PR.

docs/integrations/editors.md
bugfood: Yes, naming is hard. Would it be better to have this be `g:black_use_virtualenv` and set the default to `1` (so the actual default behavior remains unchanged)?

docs/integrations/editors.md
felix-hilden: Yeah I meant upgrading the system installation. Fair enough!

docs/integrations/editors.md
felix-hilden: Sounds good 👍

bugfood: Ok, I updated the PR and re-tested it locally. This still seems to work for me, with the default behavior being the same.

I also added fleshed out the error message and the documentation to (hopefully) be a bit more helpful.

bugfood: ...and rebased now to fix merge conflict.

autoload/black.vim
```diff
@@ -56,6 +56,16 @@ def _get_virtualenv_site_packages(venv_path, pyver):
   return venv_path / 'lib' / f'python{pyver[0]}.{pyver[1]}' / 'site-packages'
 
 def _initialize_black_env(upgrade=False):
+  if vim.eval("g:black_use_virtualenv ? 'true' : 'false'") == "false":
```
felix-hilden, line 59:
```suggestion
  if not vim.eval("g:black_use_virtualenv"):
```
Is this idiomatic? Can we not check the original value like this or something? Sorry, not super familiar 🤔

bugfood: I'm not familiar either, and I'm open to better ways. I thought pretty carefully about this and tried to write it in a way that has no possibility of unexpected behavior.

Vim's truthy/falsey handling is rather different from python's.
https://vi.stackexchange.com/questions/36270/what-values-are-truthy-and-falsey
https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/

For the values set in the vimrc file, since the vimrc file is in vimscript, we want to be able to respect vim's handing of truthiness. The eventual handling within python code is abstracted away from the user (the user would have no way of knowing this without examining the plugin source code).

`vim.eval()` operates on vimscript, so vim's truthiness comes into play. This is what I wanted, but unfortunately, `vim.eval()`, as documented, doesn't return ever a boolean. From `:help python-eval` in vim:
```
vim.eval(str)                                           python-eval
        Evaluates the expression str using the vim internal expression
        evaluator (see expression).  Returns the expression result as:
        - a string if the Vim expression evaluates to a string or number
        - a list if the Vim expression evaluates to a Vim list
        - a dictionary if the Vim expression evaluates to a Vim dictionary
```

To get around that limitation, I used the ternary operator.

1. The ternary operator in vimscript converts arbitrary truthy/falsey values to known strings.
2. `vim.eval()` returns one of these two known strings.
3. Python does string comparison to make its decision.

felix-hilden: Alright, thank you for the detailed explanation!

felix-hilden (approved):

autoload/black.vim
bugfood: You're welcome. I could have used any string values, but `"true"` and `"false"` looked the most clear to me.

I couldn't have done this instead:
```
if vim.eval("g:black_use_virtualenv ? '1' : '0'"):
```
...because python considers the string `"0"` to be true.


This does work:
```
if vim.eval("g:black_use_virtualenv ? '' : '1'"):
```
...but I think the original version is more clear to the reader.

bugfood:
@felix-hilden (or anyone) is there anything I need to do to get this PR merged? So far, I don't think there are any objections.

Thanks,
Corey

cooperlees (approved):
Looks fine to me, felix approved, but I have very little vim plugin skill ... #blindMerge

bugfood:
Thank you.

naught101:
To use this, add this to your `.vimrc`:
```vim
let g:black_use_virtualenv = 0
```

PR #3322, opened 2022-10-10
added python version to parse error message

### LINKED ISSUE #3294: Clearer errorr message for "Cannot parse" when targetting multiple version

I was working on a project that is only intended to run on Python 3.10. However, the `black` configuration incorrectly listed multiple target versions because I copied the file from a different project.

When I tried to format a file with a match statement, the error was 

> error: cannot format foo.py: Cannot parse: 2:6: match bar:

I was initially confused as I knew that `black` had added Python 3.10 support. Then I noticed my mistake. By saying that the code should target a multiple Python versions, using a match statement would not be valid.

If `black` encounters a parsing error and it was configured to target multiple Python versions, I think it would be helpful to mention that the syntax "might be valid, but not across all target versions". This would be a hint to the user to think about if they should be using that syntax or that `black` is misconfigured (as it was in my case).

It would require more work, but the parser could back track and attempt each of the parser for each of the listed target versions. That would allow `black` provide an even more specific error message. Something like "this is valid for Python 3.10, but targeting Python 3.10 and 3.9", while also being able to identify an invalid syntax.

NNRepos:
### Description
improve parsing error message to make sure users know which version it was run on (fixes #3294).

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -148,6 +148,7 @@
 
 ### Output
 
+- Add python version to parse error message
 - Change from deprecated `asyncio.get_event_loop()` to create our event loop which
   removes DeprecationWarning (#3164)
 - Remove logging from internal `blib2to3` library since it regularly emits error logs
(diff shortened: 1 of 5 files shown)


NNRepos:
sorry about this, renamed my branch.

PR #3323, opened 2022-10-10
added python version to parse error message

### LINKED ISSUE #3294: Clearer errorr message for "Cannot parse" when targetting multiple version

nnrepos:
### Description
improve parsing error message to make sure users know which version it was run on (fixes #3294).

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -148,6 +148,7 @@
 
 ### Output
 
+- Add python version to parse error message
 - Change from deprecated `asyncio.get_event_loop()` to create our event loop which
   removes DeprecationWarning (#3164)
 - Remove logging from internal `blib2to3` library since it regularly emits error logs
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -133,7 +133,7 @@ plus a short summary.
 
 ```console
 $ black src/
-error: cannot format src/black_primer/cli.py: Cannot parse (python 3.10): 5:6: mport asyncio
+error: cannot format src/black_primer/cli.py: Cannot parse (python 3.10+): 5:6: mport asyncio
 reformatted src/black_primer/lib.py
 reformatted src/blackd/__init__.py
 reformatted src/black/__init__.py
@@ -151,7 +151,7 @@ Using configuration from /tmp/pyproject.toml.
 src/blib2to3 ignored: matches the --extend-exclude regular expression
 src/_black_version.py wasn't modified on disk since last run.
 src/black/__main__.py wasn't modified on disk since last run.
-error: cannot format src/black_primer/cli.py: Cannot parse (python 3.10): 5:6: mport asyncio
+error: cannot format src/black_primer/cli.py: Cannot parse (python 3.10+): 5:6: mport asyncio
 reformatted src/black_primer/lib.py
 reformatted src/blackd/__init__.py
 reformatted src/black/__init__.py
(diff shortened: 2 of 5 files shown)


CHANGES.md
```diff
@@ -148,6 +148,7 @@
 
 ### Output
 
+- Add python version to parse error message
```
felix-hilden, line 151:
```suggestion
- Add python version to parse error message (#3323)
```
Please include the PR number!

felix-hilden (approved):
Thanks for submitting this, look good apart from a changelog detail 👍

JelleZijlstra (commented):
I feel like this will create more user confusion:

```
% black --target-version py36  -c 'a++'
a++
error: cannot format <string>: Cannot parse (python 3.0): 1:3: a++
% black --target-version py39  -c 'a++'
a++
error: cannot format <string>: Cannot parse (python 3.7): 1:3: a++
```
Users are going to wonder why it says "3.0" or "3.7" when they never asked for that version.

nnrepos:
> I feel like this will create more user confusion:
> 
> ```
> % black --target-version py36  -c 'a++'
> a++
> error: cannot format <string>: Cannot parse (python 3.0): 1:3: a++
> % black --target-version py39  -c 'a++'
> a++
> error: cannot format <string>: Cannot parse (python 3.7): 1:3: a++
> ```
> 
> Users are going to wonder why it says "3.0" or "3.7" when they never asked for that version.

ok, can you help me understand why we parse `python 3.0`, when a user asks to parse `py36`?

JelleZijlstra:
> ok, can you help me understand why we parse `python 3.0`, when a user asks to parse `py36`?

We don't have separate grammars for every version. For py36 we use a grammar with version set to (3, 0).

nnrepos:
> > ok, can you help me understand why we parse `python 3.0`, when a user asks to parse `py36`?
> 
> We don't have separate grammars for every version. For py36 we use a grammar with version set to (3, 0).

ok, so do we have an access to the version that the user provided somewhere?

JelleZijlstra:
It's the target version, which gets passed as a command-line parameter and thence as an argument to `main()`.

nnrepos:
hey @JelleZijlstra @felix-hilden, i've made some changes, please let me know if this is helpful 😄

JelleZijlstra:
I'm still not enthusiastic about this:
- Hardcoding the list of versions is a little hacky and hard to maintain
- I'm afraid the error message may still be confusing for users, but don't have a better suggestion

nnrepos:
> I'm still not enthusiastic about this:
> 
> * Hardcoding the list of versions is a little hacky and hard to maintain
> * I'm afraid the error message may still be confusing for users, but don't have a better suggestion

* i've used the same hard-coding that is used by the parser, since there weren't any constants/functions which can be used to do that. that can be refactored of course.
* i think users are smart enough to understand what a range of numbers is, we're all programmers here...

feel free to close this if you're not interested

PR #3325, opened 2022-10-10
Clarify check argument is needed for github action workflow

jlplenio:
### Description

I used the _Black_ [github action workflow](https://black.readthedocs.io/en/stable/integrations/github_actions.html) in one of my projects and passed arguments via `options`. Because I had initially omitted `'--check'`, my workflow did not fail, although _Black_ found files that needed formatting. I would like to suggest this one sentence that makes that even clearer. It would have saved me some debugging time, as I am quite new to GitHub action workflows.

Clarifies that the check argument is required so that the workflow fails when Black finds files that need to be formatted based on the correct exit codes.

I have successfully run the [docs build test](https://black.readthedocs.io/en/latest/contributing/the_basics.html#docs-testing) locally.

### Checklist - did you ...

### DIFF
--- a/docs/integrations/github_actions.md
+++ b/docs/integrations/github_actions.md
@@ -45,8 +45,8 @@ extra. Installing the extra and including Jupyter Notebook files can be configur
 
 You can also configure the arguments passed to _Black_ via `options` (defaults to
 `'--check --diff'`) and `src` (default is `'.'`). Please note that the
-[`'--check'` argument](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#exit-code)
-is required so that the workflow fails if _Black_ finds files that need to be formatted.
+[`--check` flag](labels/exit-code) is required so that the workflow fails if _Black_
+finds files that need to be formatted.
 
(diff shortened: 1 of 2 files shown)


cooperlees (approved):
I agree and approve this extra addition. I feel it's too late now, but we should have probably defaulted with `--check`.

Should we think about a Major version release changing this default in the future? What are the downsides people see other than helping people find unformatted files?

cooperlees (changes requested):
Let's just please our prettier formatter on this markdown file please.

jlplenio:
I was unable to reproduce the prettier error on my WSL. He somehow does not care about line length.
I used the suggested edit from the [failed pre-commit workflow](https://github.com/psf/black/actions/runs/3222431515/jobs/5272797762).

ichard26 (approved):
I'm okay with adding `--check` to the argument list unconditionally starting with 23.0.0.

felix-hilden (approved):
LGTM and a forced `--check` sounds good @ichard26 👍

PR #3332, opened 2022-10-16
Add compatible `indentexpr` to Vim plugin

### LINKED ISSUE #3151: vim plugin: provide black-compatible `indentexpr`
The way that vim automatically indents python code[^1] differs from the black format.  I'm aware of two differences: (i) lines within parentheses/braces are indented twice rather than once and (ii) closing parentheses/braces are not dedented.  These differences can be seen clearly by looking at a function argument list that spans multiple lines.  Here is such a list as it is automatically formatted by vim (i.e. typed without hitting tab or backspace):
```
# $ vim -u /etc/vimrc +'filetype indent on' +'set ft=python'
def foo(
        a,
        b,
        c,
        ):
    pass
```
Here is the same argument list as it would be formatted by black (ignoring that it would be condensed onto a single line):
```
def foo(
    a,
    b,
    c,
):
    pass
```
Multiline list/dict/set comprehensions have the same behavior.

I personally find it annoying that the code I type does not immediately conform to the black style.  As such, I think it'd be nice if the vim plugin provided by this repository included an `indentexpr` that implemented the black indentation rules.

Is there any interest in having something like this?  I have some experience writing vim plugins, so I might be able to contribute something, but probably not in the near future.

Alternatively, is there an easy way to get vim to do what I want without having to write a bunch of vimscript code?  A brief search didn't turn up any existing plugins or relevant `vimrc` settings, but this feels like it should be a common desire and I'm half-convinced that I'm missing something simple.

[^1]: As specified by `/usr/share/vim/vim82/indent/python.vim`.

aus-hawk:
### Description

Fixes #3151

Before a patch that was made a few months ago in Vim ([9.0.259](https://github.com/vim/vim/commit/fd999452adaf529a30d78844b5fbee355943da29#diff-082e8db8a25bfef894eb32a5af1a1ad965b90c4a164fc33276e69f3522401cb8)), and less than a month ago in Neovim ([0.8.0](https://github.com/neovim/neovim/commit/d3cd79709b6491a5806c67ea52c19f244164954e#diff-082e8db8a25bfef894eb32a5af1a1ad965b90c4a164fc33276e69f3522401cb8)), the built in Python indent script poorly indented multiple lines within parenthesis by default and didn't even have an option to make closing parenthesis that start a line line up with the start of the statement, instead always being the same indent as the line before it. It is also still inconsistent with how `with` statements are proposed to be formatted in the future.

This adds to the Vim plugin its own indent script that can be run when the `g:black_indent` flag is enabled. It uses Vim's own Python indent except for cases where it is inconsistent, and sets the configs itself to be as consistent with Black as possible.

The flag is disabled by default since I think it's reasonable to assume that many Vim users probably already have a plugin to do this. I know I did.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,9 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- Vim plugin: can now indent while editing according to the _Black_ code style, which
+  can be enabled with the `g:black_indent` flag (#3332)
+
 ### Documentation
 
 <!-- Major changes to documentation and policies. Small docs changes
--- a/docs/integrations/editors.md
+++ b/docs/integrations/editors.md
@@ -137,8 +137,8 @@ $ git checkout origin/stable -b stable
 
 or you can copy the plugin files from
 [plugin/black.vim](https://github.com/psf/black/blob/stable/plugin/black.vim),
-[autoload/black.vim](https://github.com/psf/black/blob/stable/autoload/black.vim),
-and [indent/python.vim](https://github.com/psf/black/blob/stable/indent/python.vim).
+[autoload/black.vim](https://github.com/psf/black/blob/stable/autoload/black.vim), and
+[indent/python.vim](https://github.com/psf/black/blob/stable/indent/python.vim).
 
 ```
 mkdir -p ~/.vim/pack/python/start/black/plugin
--- a/docs/integrations/editors.md
+++ b/docs/integrations/editors.md
@@ -114,6 +114,7 @@ Configuration:
 - `g:black_virtualenv` (defaults to `~/.vim/black` or `~/.local/share/nvim/black`)
 - `g:black_quiet` (defaults to `0`)
 - `g:black_preview` (defaults to `0`)
+- `g:black_indent` (defaults to `0`)
 
 To install with [vim-plug](https://github.com/junegunn/vim-plug):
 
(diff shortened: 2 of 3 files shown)


aus-hawk:
Makes sense, I'll admit to definitely *not* being an expert in Vim plugins. I've been testing it on Vim 9.0.354-1 and 8.2.2434-3, and Neovim 0.8.0-3 and 0.4.4-1, just to see if I miss anything version related. I also ran:

```sh
find -regex ".*\\.py" -exec vim -c "normal =G" -c wq "{}" ";"
```

in `src` to see how it indents which is an okay way to check for holes, which the code still has (indentation breaks if a comment comes after an open parenthesis for example), so this is a lot more of a work in progress than I initially thought making the PR.

Ultimately, I don't think it's a good idea to keep trying to do this by patching the inconsistent parts of the built in `indentexpr` because there are bound to be some non-obvious ones I don't account for like the comments issue I mentioned earlier. It's probably better to just forget about trying to fix what's broken and make a standalone `indentexpr` function. It'll be a bit more effort but nothing that Vim breaks would affect it.

If you think that's a good idea, I just want to verify: should it be as opinionated as Black itself? By that I mean should it be written under the assumption that the user is already trying to stick with the style and isn't trying to do things like:

```py
def test(aaaaaaaaaaaa,
         bbbbbbbbbbbb,
         cccccccccccc):
    ...
```

Assuming they won't even try to do things like this should make it simpler to write by just not considering the cases like this, so if they try something like this it might indent completely wrong, but as long as they follow the style it would work exactly as expected. How I see it, if they don't want to use the Black format while writing code they shouldn't be using the Black plugin as the indent script and just keep it disabled.

aus-hawk:
I've updated the script to use its own standalone `indentexpr` function that assumes the user is going to already be writing in the Black style. This turned out to be a very good idea since:

1. The original script was a bit more buggy than I originally thought, and hacked together to rely on Vim staying consistent. Rewriting to be self sufficient means only consulting the single script for any issues ran into in the future.
2. The new script is consistent. Running the `find` command from the last comment in `src` with the old script used to reindent a few files, while the new script changes none of them.
1. The new script is *significantly* faster. I'm talking almost an order of magnitude. Running that `find` command in a virtual machine on `src` used to take ~2m20s. The same command with the new script takes less than 15 seconds.
2. I haven't tested it but I'm almost certain the new script works on Vim versions at least as old as 7 like the rest of the Vim plugin, and probably much older. The old script I wasn't so sure of.

I'm still of the opinion that not even worrying about trying to indent code that's inconsistent with the style is beneficial since it keeps the code simpler and if they don't want to write code according to Black they can just keep the flag disabled and use whatever script they typically use. If you disagree then I can try to flesh it out a bit more, but if not I think the script as it is now is perfectly usable.

fredizzimo:
Maybe it would make more sense to extract this as a standalone vim plugin? It would be easier to install, and the black maintainers don't need to deal with any issues that crops up.

BTW, I found this existing plugin https://github.com/jeetsukumaran/vim-python-indent-black, but I don't know how it compares to this implementation.

aus-hawk:
I think making a standalone repo is the way to go. Most (Neo)vim package managers tend to install the entire repository, which means installing Black twice for anyone that wants to use it as a Vim plugin (I don't). A lone plugin like the one you posted makes more sense if someone just wants the indentation.

As Felix said, no one officially maintains the Vim plugin (I sure wasn't planning on it), so it would be rarely-useful code that would inevitably go out of date and no one would understand. All that said, it's been open for over a year. There's clearly not enough of a demand for that to be worth it. I'll close this.

PR #3336, opened 2022-10-17
Apply .gitignore correctly in every source entry

### LINKED ISSUE #3291: .gitignore is ignored by black if more than one source location is specified
**Describe the bug**

When adding more than one source entry, black ignores the .gitignore entries.

**To Reproduce**

content of .gitignore

```
generated/*
```

Running

```
black test src
```
produces

```
$ black --check tests src
would reformat generated/jtd/__init__.py

Oh no! 💥 💔 💥
1 file would be reformatted, 29 files would be left unchanged.
```

Running
```
black src
```

produces

```
$ black --check src
All done! ✨ 🍰 ✨
```

**Expected behavior**

```
$ black --check test src
All done! ✨ 🍰 ✨
```

**Environment**

- Black's version: 22.8.0
- OS and Python version: MacOS m1

**Additional context**

works with Black version: 22.6.0

aaossa:
### Description

Fix #3291 : incorrectly ignoring .gitignore presence when more than one source directory is specified. Commit message of the commit that fixed the bug:

When passing multiple src directories, the root gitignore was only applied to the first processed source. The reason is that, in the first source, exclude is `None`, but then the value gets overridden by `re_compile_maybe_verbose(DEFAULT_EXCLUDES)`, so in the next iteration where the source is a directory, the condition is not met and sets the value of `gitignore` to `None`.

To fix this problem, we set the new value of `exclude` is a new variable (`_exclude`). This prevents entering the wrong condition on the second iteration and also keeps the origina value if the condition is not met.

Also, the value of both `_exclude` and `gitignore` is calculated only once by using `None` as default value. This help us to reduce the number of computations (only calculate the value once, if needed), and also retains the original value of `gitignore`, which is then merged with the value of `p_gitignore`. That's why we also define `_gitignore` to contain the result of the merge (add) operation.

### Checklist - did you ...

- [x] ~Add new / update outdated documentation?~


src/black/__init__.py
```diff
@@ -625,6 +625,9 @@ def get_sources(
     sources: Set[Path] = set()
     root = ctx.obj["root"]
 
+    _exclude = exclude
```
aaossa, line 628: Not sure how to fix the mypy check:

```
src/black/__init__.py:681:21: error: Argument 4 to "gen_python_files" has incompatible type "Optional[Pattern[str]]"; expected "Pattern[str]"  [arg-type]
```

The complain comes because `exclude` has type `Optional[Pattern[str]]`, so `_exclude` should have the same type. But `gen_python_files` always expects a `Pattern[str]`.

In the case that `exclude` is `None` (and therefore `_exclude` too), the value of `_exclude` then changes to `Pattern[str]`, so the value passed to `gen_python_files` will never be `None`. Any ideas?

aaossa: Maybe use[ `typing.cast`](https://docs.python.org/3/library/typing.html#typing.cast) when passing the value to `gen_python_files`? `cast(Pattern[str], _exclude)`. Feels bad but should make mypy happy

aaossa: Decided to add `cast(Pattern[str], _exclude)` when passing the value to `gen_python_files`

aaossa:
I'm not sure about why the checks failed 🤔  These seems to be not related with my changes, maybe this requires running the checks again?

* `Error: Invalid value for 'analysis': File 'preview-changes-pr-3336-f55a81e9e0.json' does not exist.`
* `Error: Invalid value for 'analysis': File 'assert-no-changes-pr-3336-f55a81e9e0.json' does not exist.`

src/black/__init__.py
```diff
@@ -625,6 +626,9 @@ def get_sources(
...
+    _gitignore = gitignore = None
```
aaossa, line 630: > `src/black/__init__.py:630: error: Local variable "gitignore" has inferred type None; add an annotation`

Solving by adding type annotation: `#type: Optional[PathSpec]`

src/black/__init__.py
```diff
@@ -659,25 +664,27 @@ def get_sources(
...
+                    _gitignore = gitignore
+                else:
+                    _gitignore = gitignore + p_gitignore
             sources.update(
                 gen_python_files(
                     p.iterdir(),
                     ctx.obj["root"],
                     include,
-                    exclude,
+                    cast(Pattern[str], _exclude),
```
JelleZijlstra, line 683: Instead of casting, we should assert that _exclude is not None. cast is unsafe.

aaossa: I'm not sure how to assert that. The value will never be `None`, but is a limitation (?) of mypy https://github.com/psf/black/pull/3336/files#r997119695

But, this issue should go away if we just initialize `_exclude` as `re_compile_maybe_verbose(DEFAULT_EXCLUDES)`, as you suggested. Then, casting nor asserting should be necessary (mypy will correctly identify `_exclude` as `Pattern[str]`)

JelleZijlstra (approved):

src/black/__init__.py
```diff
@@ -625,6 +626,11 @@ def get_sources(
     sources: Set[Path] = set()
     root = ctx.obj["root"]
 
+    exclude_is_None = exclude is None
+    exclude = re_compile_maybe_verbose(DEFAULT_EXCLUDES) if exclude is None else exclude
+    gitignore = None  # type: Optional[PathSpec]
+    root_gitignore = get_gitignore(root)
```
ichard26, line 632:
```suggestion
    using_default_exclude = exclude is None
    exclude = re_compile_maybe_verbose(DEFAULT_EXCLUDES) if exclude is None else exclude
    gitignore: Optional[PathSpec] = None
    root_gitignore = get_gitignore(root)
```

src/black/__init__.py
```diff
@@ -658,16 +664,14 @@ def get_sources(
 
             sources.add(p)
         elif p.is_dir():
-            if exclude is None:
-                exclude = re_compile_maybe_verbose(DEFAULT_EXCLUDES)
-                gitignore = get_gitignore(root)
+            if exclude_is_None:
```
ichard26, line 667:
```suggestion
            if using_default_exclude:
```

ichard26 (commented):
Hey I just wanted to say thanks for the PR. I really appreciate it! Also thank you @JelleZijlstra for the review! 🧡 

The PR overall looked good, but two things stood out to me on a quick scan.

aaossa:
> Hey I just wanted to say thanks for the PR. I really appreciate it! Also thank you @JelleZijlstra for the review! 🧡
> 
> The PR overall looked good, but two things stood out to me on a quick scan.

I'll tackle your suggestions while solving conflicts on #3338 . One of the conflicts is around the same area, so solving them there is "free"

src/black/__init__.py
aaossa: Change included in #3338

PR #3338, opened 2022-10-17
Compare each .gitignore found with an appropiate relative path

### LINKED ISSUE #3315: Incorrectly ignored files when `subfolder/.gitignore` contains */*
**Describe the bug**

When a subfolder contains a .gitignore file with the special rule `*/*` (ignore subdirs contents) Black incorrectly ignores any file at the same level of the .gitignore.

**To Reproduce**

A minimal git repository:
```
.
└── repository/
    ├── .git/
    └── subdir/
        ├── .gitignore  # Contains */*
        └── main.py
```

Any change in the contents of main.py will be "seen" by Git but not Black, resulting in `"No Python files are present to be formatted. Nothing to do 😴"` in the following cases:

* `black .`  (from repository/)
* `black subdir/` (from repository/)
* `black .` (from repository/subdir/)

It works as expected when `SRC` is the path to the file:

* `black subdir/main.py` (from repository/)
* `black main.py` (from repository/subdir/)

When using `black . --verbose`, the output confirms that the file is ignored because of the .gitignore file:

```
> black . --verbose
Identified `/repository` as project root containing a .git directory.
Sources to be formatted: "."
.git ignored: matches the --exclude regular expression
subdir/.gitignore ignored: matches the .gitignore file content
subdir/main.py ignored: matches the .gitignore file content
```

Notice that files in other folders of the repository are targeted as expected. Only the folder that contains the .gitignore file has its contents incorrectly ignored.

Also, the following repository structure works as expected for both Git and Black in any case, which confirms that the problem comes when the .gitignore is in a subdirectory with Python files:

```
.
└── repository/
    ├── .git/
    ├── .gitignore  # Contains */*
    └── main.py
```

**Expected behavior**

Black should be able to detect and format the contents of subdir/ that are visible for Git (like subdir/main.py).
 
**Environment**

- Black's version: main
- OS and Python version: Linux/Python 3.8.10

**TLDR**

These steps help to visualize the problem:

1. Setup the repository
  ```bash
  mkdir demo && cd demo
  git init
  echo "var=[1,2,3]" > root.py
  mkdir subdir
  echo "var=[1,2,3]" > subdir/subdir.py
  echo "*/*" > subdir/.gitignore
  git add -A
  git commit -m "First commit"
  ```

2.  Confirm that `subdir/subdir.py` is visible to Git:
```bash
echo "var=[3,4,5]" > subdir/subdir.py
git status
```

3. Black reaches `root.py` but not `subdir/subdir.py`
```bash
black . --verbose
```

aaossa:
### Description

Fix #3315 : Fix incorrectly applied .gitignore rules by considering the .gitignore location and the relative path to the target file. Commit message of the commit that fixed the bug:

When a .gitignore file contains the special rule to ignore every subfolder content (`*/*`) and the file is located in a subfolder
relative to where the command is executed (root), the rule is incorrectly applied and ignores every file at the same level of the .gitignore file.

The reason for this is that the `gitignore` variable accumulates the rules found in each .gitignore while traversing files and directories recursively. This makes sense and, in general, works as expected. The problem is that the gitignore rules are applied using as the root of the target directory as a reference. This is the cause of the bug.

The implemented solution keeps track of every .gitignore file found while traversing the targets and the location of each .gitignore. Then, when matching files to the .gitignore rules, compare each set of rules with the appropiate relative path to the candidate target file.

To make this possible, we changed the single `gitignore` object with a dictionary of similar objects, where the corresponding key is the path to the folder that contains that .gitignore file. This required changing the signature of the `get_sources` function. Also, we introduce a `is_ignored` function that compares a file with every set of rules. Finally, some tests required an update to pass the gitignore object in the new format.

### Checklist - did you ...

- [x] ~Add new / update outdated documentation?~


src/black/files.py
```diff
@@ -211,15 +211,27 @@ def gen_python_files(
...
     `report` is where output about exclusions goes.
     """
+
+    def is_ignored(
+        gitignore_dict: Dict[Path, PathSpec], child: Path, report: Report
+    ) -> bool:
+        for _dir, _gitignore in gitignore_dict.items():
+            relative_path = normalize_path_maybe_ignore(child, _dir, report)
+            if relative_path is None:
+                continue
```
aaossa, line 221: Seems like `return False` is a better alternative since a symlink that goes back inside root should be ignored (at least seems to be ignored now)

aaossa:
Converting to draft because I'm getting strange results locally

src/black/files.py
aaossa: Yeah, seems like the `continue` produces additional unnecessary output. I'll replace it with `break` so that the function returns `False`

src/black/files.py
```diff
@@ -244,6 +256,11 @@ def gen_python_files(
         if child.is_dir():
             # If gitignore is None, gitignore usage is disabled, while a Falsey
             # gitignore is when the directory doesn't have a .gitignore file.
+            if gitignore is None:
+                _gitignore = None
+            else:
+                _gitignore = gitignore.copy()
+                _gitignore[child] = get_gitignore(child)
```
aaossa, line 263: I'll also use the path from root (`root / child`) to the corresponding .gitignore. This solves producing additional and incorrect output when running black on a directory 👌 Change also applied on `src/black/__init__.py` (key `p` in the `_gitignore` dictionary was changed to `root / p`)

tests/test_black.py
```diff
@@ -2056,6 +2056,24 @@ def test_invalid_nested_gitignore(self) -> None:
         gitignore = path / "a" / ".gitignore"
         assert f"Could not parse {gitignore}" in result.stderr_bytes.decode()
 
+    def test_gitignore_that_ignores_subfolders(self) -> None:
```
aaossa, line 2059: Added a test case where the target folder was different to the root.

aaossa:
This is PR is back up again and ready for review 👍

src/black/__init__.py
```diff
@@ -660,14 +662,14 @@ def get_sources(
         elif p.is_dir():
             if exclude is None:
                 exclude = re_compile_maybe_verbose(DEFAULT_EXCLUDES)
-                gitignore = get_gitignore(root)
+                root_gitignore = get_gitignore(root)
                 p_gitignore = get_gitignore(p)
                 # No need to use p's gitignore if it is identical to root's gitignore
                 # (i.e. root and p point to the same directory).
```
JelleZijlstra, line 668: What if there are more than two levels? For example, we're formatting `a/b/c/` and `a/.gitignore/`, `a/b/.gitignore`, and `a/b/c/.gitignore` all exist.

src/black/files.py
```diff
@@ -211,15 +211,27 @@ def gen_python_files(
 
     `report` is where output about exclusions goes.
     """
+
+    def is_ignored(
```
JelleZijlstra, line 215: Doesn't seem like this needs to be a nested function.

src/black/files.py
```diff
@@ -198,7 +198,7 @@ def gen_python_files(
     extend_exclude: Optional[Pattern[str]],
     force_exclude: Optional[Pattern[str]],
     report: Report,
-    gitignore: Optional[PathSpec],
+    gitignore: Optional[Dict[Path, PathSpec]],
```
JelleZijlstra, line 201: Would suggest renaming this to `gitignore_dict` for clarity. Also, no reason to keep this Optional, we can just use an empty dict if there's no gitignores.

src/black/files.py
```diff
@@ -244,6 +256,11 @@ def gen_python_files(
         if child.is_dir():
             # If gitignore is None, gitignore usage is disabled, while a Falsey
             # gitignore is when the directory doesn't have a .gitignore file.
+            if gitignore is None:
+                _gitignore = None
+            else:
+                _gitignore = gitignore.copy()
+                _gitignore[root / child] = get_gitignore(child)
```
JelleZijlstra, line 263: Can do `new_gitignore = {**gitignore, root / child: get_gitignore(child)}`.

I'd avoid using leading underscores for locals; better to give an informative name.

src/black/__init__.py
aaossa: Good question. Right now, if the command is `black a/b/c/` then `a/b/.gitignore` is completely skipped while `a/.gitignore` and `a/b/c/.gitignore` are captured by `root_gitignore` and `p_gitignore`, respectively. What's the expected behavior?

src/black/files.py
aaossa: `gitignore` can be `None` if `exclude` is not `None`. In that case, no gitignore is considered at any level. I though that this is the expected behavior while reading the code. If `get_sources` receives a value for `exclude`, what's the expected behavior?

src/black/files.py
aaossa: Using a nested function allows me to short-circuit if a match is found and break the outer loop (`for child in paths:`) with its result. Using `continue` in an inner loop just jumps to the next iteration in that loop. Using `break` might help, but how to know if there was actually a match or not (breaking because `relative_path is None` or because there was a match)? I think that using a nested function makes it easier to read/understand.

src/black/__init__.py
aaossa: Should this be a separate issue/PR? If its an undefined behavior, then it should be documented properly I think

src/black/__init__.py
```diff
@@ -660,14 +662,14 @@ def get_sources(
...
-                if gitignore != p_gitignore:
-                    gitignore += p_gitignore
-            else:
-                gitignore = None
+                if root_gitignore == p_gitignore:
```
aaossa, line 669: Check if this is still necessary. If two set of rules are equal, it's still possible that they produce different results because they're in different locations. Might be better to compare the paths

src/black/__init__.py
aaossa: > What if there are more than two levels? For example, we're formatting `a/b/c/` and `a/.gitignore/`, `a/b/.gitignore`, and `a/b/c/.gitignore` all exist.

The current behavior seems to be that `black a/b/c` only detects `a/b/c/.gitignore` and completely ignores other gitignore files. This PR implements the same behavior

src/black/files.py
```diff
@@ -211,15 +211,27 @@ def gen_python_files(
...
+
+    def is_ignored(
+        gitignore_dict: Dict[Path, PathSpec], child: Path, report: Report
+    ) -> bool:
+        for _dir, _gitignore in gitignore_dict.items():
+            relative_path = normalize_path_maybe_ignore(child, _dir, report)
+            if relative_path is None:
+                break
+            if _gitignore is not None and _gitignore.match_file(relative_path):
+                report.path_ignored(child, "matches the .gitignore file content")
```
aaossa, line 223: Maybe include the (full) path to the .gitignore file that matched?

src/black/files.py
JelleZijlstra: I'm suggesting to make it a separate top-level function, wouldn't that be possible?

aaossa: Oh sure, I'll update the PR with that change

JelleZijlstra (approved):

PR #3339, opened 2022-10-17
[2097] Fix is_type_comment

itxasos23:
### Description

Issue https://github.com/psf/black/issues/2097 raised the issue that `is_type_comment` only works for comments starting with `# type:`, and more spaces/tabs between `#` and `type`, or `type`and `:` are valid as per typed-ast/ast, but not black.

This PR (for now!) only tries to make is_type_comment be consistent with typed-ast, not to remove the extra spaces/tabs if they are present.

I'd like some guidance as to:
- Using **regex**. Is there a way to perform this check fast, while not changing the ast parsing?
- **Testing**. I want to unit test the method to ensure it works as expected, but I don't see tests for similar functions in the tests directory. Is this by design?
- Is this **user facing?**
- Do you think it makes sense to parse these type_comments into its **component tokens**: `#`, `type`, `:` and the rest? I believe that would help with fixing the formatting.

### Checklist - did you ...


itxasos23:
I'd argue this PR doesn't merit a changelog entry by now (maybe it does if we decide to reformat `# type:*` comments?)

tests/test_typed_comments.py
```diff
@@ -0,0 +1,13 @@
...
+from blib2to3.pgen2 import token
+from blib2to3.pytree import Leaf
+
+
+@pytest.mark.parametrize(
+    "comment_str",
+    ["# type: int", "#  type: int", "#   type: int", "# type : int", "# type  : int"],
+)
+def test_typed_comments(comment_str: str) -> None:
+    assert is_type_comment(Leaf(token.COMMENT, comment_str))
```
clavedeluna, line 13: Would be good to have tests for `test_not_typed_comment`

clavedeluna: > Would be good to have tests for `test_not_typed_comment`

for example
` # this is a comment about a :param called type`

src/black/nodes.py
```diff
@@ -778,12 +779,21 @@ def is_import(leaf: Leaf) -> bool:
...
     """Return True if the given leaf is a special comment.
     Only returns true for type comments for now."""
     t = leaf.type
     v = leaf.value
-    return t in {token.COMMENT, STANDALONE_COMMENT} and v.startswith("# type:" + suffix)
+    if t not in {token.COMMENT, STANDALONE_COMMENT}:
+        return False
+
+    if not v.startswith("#"):
+        return False
```
clavedeluna, line 794: can combine if statements of ln790 and ln793 into one if statement since they both return the same thing

tests/test_typed_comments.py
itxasos23: Thanks for the suggestion! Added!

src/black/nodes.py
itxasos23: You're right! I started thinking about guardian clauses and got carried away :) 
Fixed!

clavedeluna (approved):
LGTM! Hopefully a maintainer will answer the questions on your PR description :)

ichard26:
Hey!

I just wanted to say thank you for submitting a PR. I know us maintainers haven't been very active reviewing PRs lately and that's disheartening. I speak for the rest of the maintainership team that we really appreciate all of your work.

I promise I've seen this come into my inbox and care about it. I've just been really busy lately and finding time (and the energy) to do reviews has been tough. I'll try my best to find time this week.

Thanks again.

cooperlees (commented):
I know there's probably been discussion, but I don't like we're introducing a regex to parse for broken syntactical usages of a deprecated (or are there valid uses I don't know of for type comments) feature of Python ...

src/black/nodes.py
```diff
@@ -778,12 +779,18 @@ def is_import(leaf: Leaf) -> bool:
     )
 
 
+is_type_comment_regex = re.compile(r"^#[\s\t]+type[\s\t]*:")
```
JelleZijlstra, line 782: Isn't `\t` included in `\s`?

JelleZijlstra (commented):
How does this affect formatting in practice? It's not clear to me whether this will need to go in the preview style.

Also, you need a changelog entry.

PR #3362, opened 2022-10-28
Format type hints after comments

### LINKED ISSUE #3350: Type hint wrapping after comment
**Describe the style change**

For some reason, black weirdly wraps a type hint _only_ if it is after a comment.

```python
def get_requires_for_build_sdist(
    config_settings: dict[str, str | list[str]] | None = None
) -> list[str]:
    return ["pathspec", "pyproject_metadata"]

def get_requires_for_build_sdist(
    # pylint: disable-next=unused-argument
    config_settings: dict[str, str | list[str]]
    | None = None
) -> list[str]:
    return ["pathspec", "pyproject_metadata"]
```

**Desired style**

```python
def get_requires_for_build_sdist(
    config_settings: dict[str, str | list[str]] | None = None
) -> list[str]:
    return ["pathspec", "pyproject_metadata"]

def get_requires_for_build_sdist(
    # pylint: disable-next=unused-argument
    config_settings: dict[str, str | list[str]] | None = None
) -> list[str]:
    return ["pathspec", "pyproject_metadata"]
```

**Additional context**

Tested with both regular black and `--preview`, and with target version.

charlie572:
### Description

This fixes #3350.

The problem is that when Black encounters standalone comments on one line inside brackets, it splits the line using delimiter_split. This will split on the comment, but it will also split the line on the highest priority delimiter, which isn't always necessary.

I've added a comma_split function, which is the same as delimiter_split, but `delimiter_priority = COMMA_PRIORITY`. Black will now try comment_split before it tries delimiter_split. This means it will first split on standalone comments and commas, and only split on other delimiters if the lines are still too long after that.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Stop type hints after comments from wrapping unnecessarily. (#3362)
+
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
--- a/docs/contributing/reference/reference_functions.rst
+++ b/docs/contributing/reference/reference_functions.rst
@@ -83,6 +83,8 @@ Split functions
 
 .. autofunction:: black.linegen.delimiter_split
 
+.. autofunction:: black.linegen.comma_split
+
 .. autofunction:: black.linegen.left_hand_split
 
 .. autofunction:: black.linegen.right_hand_split
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -538,7 +538,12 @@ def transform_line(
                 ]
         else:
             if line.inside_brackets:
-                transformers = [delimiter_split, standalone_comment_split, rhs]
+                transformers = [
+                    comma_split,
+                    delimiter_split,
+                    standalone_comment_split,
+                    rhs,
+                ]
             else:
                 transformers = [rhs]
     # It's always safe to attempt hugging of power operations and pretty much every line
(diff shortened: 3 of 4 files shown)


charlie572:
Sorry, I should have run diff-shades before submitting a pull request. I'll keep working on this, and see if I can fix this solution or think of a better one.

JelleZijlstra:
It's fine to rely on CI to run diff-shades for you. I haven't looked closely here but probably you just need to gate your changes to the preview style.

charlie572:
Should I rebase main onto here to resolve the merge conflict, or will the maintainer do that? The only merge conflict is one line in CHANGES.md.

JelleZijlstra:
Easiest to fix simple merge conflicts by going through GitHub's UI, I just did it for you.

charlie572:
Okay, thanks.

aqeelat:
@charlie572 I'm interested in this PR. Can you please rebase?

charlie572:
Great. I can do this tomorrow. Do I just need to rebase main onto this branch to fix merge conflicts?

hauntsaninja:
Looking at the diff shades report https://github.com/psf/black/actions/runs/5431147643/jobs/9880366015?pr=3362 it looks this causes the "magic" trailing comma to not get respected inside collection literals if there's a comment

charlie572:
I'm confused about the magic_trailing_comma attribute of the Line class. Why does it point at the closing bracket after the magic trailing comma, and not at the magic trailing comma token? This means that my code can't check for the magic trailing comma, since it is looking at a Line object containing just the comma-separated members, not the surrounding brackets.

I can think of two ways to make it handle magic trailing commas properly:
a. add another property that points at the magic trailing comma token
b. search the tokens for a magic trailing comma inside the transform_line function (around line 600).

Option a makes the most sense to me. Is that okay?

charlie572:
I implemented option a. But now the attribute names are weird. Maybe magic_trailing_comma should point at the comma, and bracket_after_magic_trailing_comma can point at the following bracket? Or maybe we can remove the latter attribute?

aqeelat:
@charlie572 I understand nothing from the code in this PR (I'm not familiar with the internals of Black) but I want to thank you for your effort.

charlie572:
Thank you.

Issue #3355, opened 2022-10-24
Lint `src/blib2to3`


clavedeluna:
**Is your feature request related to a problem? Please describe.**

As I've been working on some black issues and overall getting a hang of the repo, I got really frustrated when debugging methods which name an arg or a variable as the built-in `type`, for example:

`def printtoken(type, token, xxx_todo_changeme, xxx_todo_changeme1, line):  # for testing`
`def addtoken(self, type: int, value: Text, context: Context) -> bool:`

You can find many of these with `pylint src --disable=all --enable=redefined-builtin`

**Describe the solution you'd like**

There are quite a bit of linting issues in the repo in addition to this one:
- unused imports still at top of module
- unused variables
- etc

**Describe alternatives you've considered**

Can we add a pylint step to github actions so developers start using it locally, too? I think an iterative approach would be reasonable, so start with something like `pylint src --disable=all --enable=redefined-builtin` and slowly start enabling more checks as they're fixed in a new PR.

**Additional context**

This would help maintain code as well as debug. If you're using `pdb`. and you want to check the `type` of an object, if there's already a variable called `type` it doesn't work ....

JelleZijlstra:
We already have flake8 set up for linting. I haven't had positive experiences with pylint: it has too many checks that aren't useful and it's slow. I'd be open to adding more flake8 plugins or enable additional rules though.

clavedeluna:
@JelleZijlstra I'll focus on flake8 then. Im I mistaken or is the current pre-commit configuration for flake8 not running in some sub directories inside `src`? For example, running locally:

```
 $ flake8 src/
src/blib2to3/pygram.py:12:1: F401 '.pgen2.token' imported but unused
src/blib2to3/pytree.py:760:13: E731 do not assign a lambda expression, use a def
src/blib2to3/pgen2/token.py:81:21: E721 do not compare types, use 'isinstance()'
src/blib2to3/pgen2/pgen.py:20:1: F811 redefinition of unused 'grammar' from line 5
src/blib2to3/pgen2/pgen.py:85:13: E265 block comment should start with '# '
src/blib2to3/pgen2/pgen.py:195:13: F841 local variable 'oldlen' is assigned to but never used
src/blib2to3/pgen2/pgen.py:197:13: F841 local variable 'newlen' is assigned to but never used
src/blib2to3/pgen2/pgen.py:373:13: E722 do not use bare 'except'
src/blib2to3/pgen2/parse.py:12:1: F401 'copy' imported but unused
[...]

PR #3363, opened 2022-10-29
Lint blib2to3

clavedeluna:
### Description

The goal of this PR is to get files under blib2to3 to be linted during pre-commit. Unfortunately, many of these files have difficult to manage code so to make this PR manageable to review, I've decided to fix some of the flake8 errors while adding `noqa` to others. 

As an example of why I added `noqa` to some, there are some flake8 linting errors for `src/blib2to3/pgen2/tokenize.py: generate_tokens`, which indicate that a rehaul of this function is really necessary. 

### Checklist - did you ...

- [ ] Add an entry in `CHANGES.md` if necessary? **is this necessary for internal change?**

References #3355

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 # Note: don't use this config for your own repositories. Instead, see
 # "Version control integration" in docs/integrations/source_version_control.md
-exclude: ^(src/blib2to3/|profiling/|tests/data/)
+exclude: ^(profiling/|tests/data/)
 repos:
   - repo: local
     hooks:
@@ -36,6 +36,7 @@ repos:
           - flake8-bugbear
           - flake8-comprehensions
           - flake8-simplify
+          - flake8-builtins
 
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v0.971
--- a/src/blib2to3/README
+++ b/src/blib2to3/README
@@ -1,18 +1,19 @@
-A subset of lib2to3 taken from Python 3.7.0b2.
-Commit hash: 9c17e3a1987004b8bcfbe423953aad84493a7984
+A subset of lib2to3 taken from Python 3.7.0b2. Commit hash:
+9c17e3a1987004b8bcfbe423953aad84493a7984
 
 Reasons for forking:
+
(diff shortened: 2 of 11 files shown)


JelleZijlstra:
Not a full review (and I know this is still a draft), but please try to keep unnecessary changes to a minimum to reduce churn.

clavedeluna:
> Not a full review (and I know this is still a draft), but please try to keep unnecessary changes to a minimum to reduce churn.

For sure, small PRs are always my goal. So far, everything is needed to make things pass. Is there something you think is so far out of bounds?

clavedeluna:
I have to say that adding linting to blib2to3 has raised some pretty egregious code. I think it's worth splitting some of the changes to another PR. So this PR would be adding linting to that dir and maybe fixing just some minor flake8 failures, but anything that requires serious fixing I can add noqa to that line and open separate PRs to remove the noqas so we can discuss the code changes.

ichard26:
The PR seems fine, but I'm not a big fan of the `token` -> `token_mod` changes, it doesn't seem worth it.

clavedeluna:
> The PR seems fine, but I'm not a big fan of the `token` -> `token_mod` changes, it doesn't seem worth it.

Let me give you an example that hopefully convinces you.  The original import is 
`from blib2to3.pgen2 import token`

any code such as this
```
def printtoken(type, token, xxx_todo_changeme, xxx_todo_changeme1, line):  # for testing
    (srow, scol) = xxx_todo_changeme
    (erow, ecol) = xxx_todo_changeme1
    print(
        "%d,%d-%d,%d:\t%s\t%s"
        % (srow, scol, erow, ecol, token.tok_name[type], repr(token))
    )
```

has a `token` argument which is clearly overriding the import `token` name. In fact, when `token.tok_name` is called, attempting to index the `tok_name` dict in token.py, it thinks we're indexing the `token` argument. This is pretty egregious code (no offense to original authors....)

So modifying the import is absolutely necessary, though other options could be renaming the variable names instead of the import, changing `token.py` module name to something else, etc. Happy to hear preferred alternatives, but for now an import alias was the easiest to do.

hauntsaninja:
Thanks, but this PR is quite stale and has several merge conflicts. Here's a replacement #3794

PR #3368, opened 2022-10-31
Prefer splitting right hand side of assignment statements.

### LINKED ISSUE #1498: Prefer expanding RHS of item assignment over LHS
Hi guys, first of all thanks for this amazing tool! Glad to see it taking the Python world by storm.

This is a fairly minor style change request.

The formatting I take issue with occurs during item assignment. When the assignment exceeds the line length and the RHS has no "required" parentheses to break on, it chooses to break on the square brackets on the LHS of the assignment:

```py
# CURRENT BEHAVIOUR
# in:

some_dictionary["some_key"] = some_long_expression_causing_long_line

# out:

some_dictionary[
    "some_key"
] = some_long_expression_causing_long_line
```

Personally I think this is really ugly - looking at this expression, it is not immediately obvious that an **assignment** is happening, since the `=` is moved to the 3rd line. Of all the non-blacked code I've ever read, not one line has ever been formatted like that, which tells me it's not totally human-friendly.

My desired result for the above input would be like the following:

```py
# DESIRED BEHAVIOUR
some_dictionary["some_key"] = (
    some_long_expression_causing_long_line
)
```

Obviously for the cases where my desired behaviour would still result in the 1st line being longer than the desired line length, the current behaviour should be applied.

My current workaround for this styling caveat is to use `dict.update()` instead (though obviously this isn't a general solution since `__setitem__` is not exclusive to MutableMappings):

```py
some_dictionary.update(
    some_key=some_long_expression_causing_long_line,
)
```

yilei:
### Description

When an assignment statement's left hand side fits on a single line, prefer splitting the right hand side's outer optional parens.

Fixes #1498.

### Checklist - did you ...


yilei:
Looking at the [diff-shades results](https://github.com/psf/black/actions/runs/3365294050/jobs/5580624247#step:10:27), I need to fix the following isuses:

1. Trailing comma in a RHS tuple isn't handled:

```diff
-        (
-            field,
-            targets,
-            alias,
-            joins,
-            path,
-            opts,
-            transform_function,
-        ) = self._setup_joins(pieces, opts, alias)
+        (field, targets, alias, joins, path, opts, transform_function,) = (
+            self._setup_joins(pieces, opts, alias)
+        )
```

2. There are some unrelated changes where the RHS of an assignment statement has un-nested pairs of inner parens, example:

```diff
             clone.external_aliases[alias] = (
-                isinstance(table, Join)
-                and table.join_field.related_model._meta.db_table != alias
-            ) or (
-                isinstance(table, BaseTable) and table.table_name != table.table_alias
+                (
+                    isinstance(table, Join)
+                    and table.join_field.related_model._meta.db_table != alias
+                )
+                or (
+                    isinstance(table, BaseTable)
+                    and table.table_name != table.table_alias
+                )
             )
```

3. This is bad:

```diff
         y_hat_discrete[
             np.arange(y_hat_probas.shape[0]), y_hat_probas.argmax(axis=1)
-        ] = 1
+        ] = (1)
```

JelleZijlstra:
I was going to point out 1. but you beat me :)

I don't think 2. is a problem: the parentheses around each `and` are preserved, it just adds new parentheses around the whole RHS.

JelleZijlstra:
I noticed this one:

```
-    df["category"] = Series(
-        np.array(list("abcdefghij")).take(np.random.randint(0, 10, size=n))
-    ).astype("category")
+    df["category"] = (
+        Series(
+            np.array(list("abcdefghij")).take(np.random.randint(0, 10, size=n))
+        ).astype("category")
+    )
```
which I realize is another example of what you meant by 2.

JelleZijlstra:
Also wanted to mention that most of the changes in diff-shades are clear improvements. This is going to be a positive change.

yilei:
Adding a note, looks like the handling of trailing comma on the left side is related to an existing issue:

```python
(
    a,
    b,
    c,
    d,
) = func1(arg1)

(a, b, c, d,) = func1(
    arg1
) and func2(arg2)
```

[playground link](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ACwAGtdAD2IimZxl1N_WhQ0lZ8ePG18SrWnqEXm6MEpBB5Xxo7HNC1DU--HVO-aJQWwg_6UCkhbjSObDxvgZLassYf5fne9JNRJzx1OVHomO9iFUp1rOQhdodRYLgcwob9p60Ag9Sxa4HwqP5v4KEoAAAA4Z-SjB7OHFQABhwGxAQAAxZqMWbHEZ_sCAAAAAARZWg==)

yilei:
The three issues mentioned above are now fixed, this is ready to review again.

yilei:
Kindly ping? @JelleZijlstra

src/black/linegen.py
```diff
@@ -631,6 +644,22 @@ def right_hand_split(
...
+    )
+
+
+def _first_right_hand_split(
+    line: Line,
+    omit: Collection[LeafID] = (),
+) -> _RHSResult:
+    """Split the line into head, body, tail starting with the last bracket pair.
+
+    Note: this function should not have side effects. It's replied upon by
```
JelleZijlstra, line 659:
```suggestion
    Note: this function should not have side effects. It's relied upon by
```

JelleZijlstra (approved):
Looked over diff-shades again and everything looks good.

PR #3374, opened 2022-11-03
Fix printing of emoji on Windows when stdout is redirected

### LINKED ISSUE #3156: No Emojis on PowerShell
**Describe the bug**

I've read [Issue #1031](https://github.com/psf/black/issues/1031), but my terminal supports emojis, yet `black` is not printing them for some reason:

image

**To Reproduce**

Simply run `pre-commit run --all` with the following configuration settings:

#### `pyproject.toml` Configuration
```toml
[tool.black]
line-length = 90
target-version = ["py38"]
include = '.*\.pyi?$'
exclude = '\.eggs|\.git|\.mypy_cache|\.tox|\.venv|build|dist'

[tool.pycln]
all = true
include = '.*\.pyi?$'
extend_exclude = 'stubs/vtkmodules-stubs/all.pyi|stubs/vtkmodules-stubs/web/camera.pyi'
```

#### `.pre-commit-config.yaml`

```yaml
minimum_pre_commit_version: "2.19.0"
fail_fast: true
repos:
  - repo: https://github.com/python-jsonschema/check-jsonschema
    rev: 0.16.2
    hooks:
      - id: check-gitlab-ci
        name: (check-jsonschema) Check `.gitlab-ci.yml` formatted correctly

  - repo: local
    hooks:
      - id: pyupgrade
        name: (pyupgrade) Format source to latest python syntax
        language: system
        files: .*\.pyi?$
        entry: pyupgrade
        args: [
            "--py310-plus",
        ]

  - repo: local
    hooks:
      - id: pycln
        name: (pycln) Remove unused imports
        language: system
        files: .*\.pyi?$
        entry: pycln
        args: [
          "--config=pyproject.toml"
        ]

  - repo: local
    hooks:
      - id: black
        name: (black) Format source code
        language: system
        files: .*\.pyi?$
        entry: black
```

**Expected behavior**

Emoji's print

**Environment**

- Black's version: `22.6.0`
- Pre-Commit's version: `2.19.0`
- Pycln's version: `2.0.1`
- OS and Python version: `Windows 10 x64-bit`
- PowerShell: `7.2.5`
- VSCode: `1.68.1`

ashb:
This is most apparent when black is run via pre-commit which does
`subprocess.Popen(stdout=PIPE)`.

One option around this would be to instruct users to set `PYTHONUTF8=1`,
but that is not a very scalable approach, as basically everything supports
UTF-8 these days.

This fix was inspired by pycln and https://github.com/hadialqattan/pycln/issues/54

Fixes #3156

### Checklist - did you ...


ashb:
Sorry, I freshly cloned this on my windows machine and forgot to do a `pre-commit install`

JelleZijlstra (commented):
Thanks for your contribution!

Monkeypatching `sys.stdout` feels quite risky though. A few questions:
- What will happen on Windows terminals that use an encoding incompatible with UTF-8? I'm worried we'll get mojibake in that case.
- Could this be fixed in `click` instead? We use `click.echo` to output all emoji, and `click` is supposed to take care of output encoding for us.
- The `pycln` issue reported getting `UnicodeEncodeError`, while the OP of the Black issue saw literal `\u1234` output instead. Do you know why that is?

ashb:
> Thanks for your contribution!
> 
> Monkeypatching `sys.stdout` feels quite risky though. A few questions:

> - What will happen on Windows terminals that use an encoding incompatible with UTF-8? I'm worried we'll get mojibake in that case.

This isn't possible on ~"modern" Windows terminals (which I means Windows 10+, even the built in CMD.exe understands utf-8)

Also from python 3.7+ the output is utf8 when stdout is connected to a TTY, so this disparity only occurs when not in a TTY.

So in short I think the risk of mojibake is tiny. And whatever risk there is already applies without this change and you'd get an encoding error on trying to convert the extended unicode emoji character into cp1252.

One option I could do here is to limit this to Windows 10+. They should remove the risk entirely.

> - Could this be fixed in `click` instead? We use `click.echo` to output all emoji, and `click` is supposed to take care of output encoding for us.

It can, but for some reason it felt like a more disruptive change to make there rather then just to a single project (black) that we know we already expect to be able to print utf-8/high unicode.

> - The `pycln` issue reported getting `UnicodeEncodeError`, while the OP of the Black issue saw literal `\u1234` output instead. Do you know why that is?

I suspect the Unicode escape is something to do with how the output is captured by pre-commit. I can dig into that more if you'd like confirmation

ambv:
This is funny to even mention but since Black works on Python 3.7+, it will run on Windows 7, Windows 8, 8.1, 10, and 11. Maybe even Vista 😂

Python 3.9 and newer don't support Windows 7 but still support 8+.

So, while decreasingly likely, it is certainly possible for a user of an older Windows to run Black. I would add the version check to avoid regressions. Skipping an emoji and showing `\Uxxx` is less disruptive than mojibake.

ashb:
So adding a `int(platform..release()) > 10` sounds like a safer way of doing this then?

```python console
>>> platform.uname()
uname_result(system='Windows', node='sinope', release='10', version='10.0.22621', machine='AMD64')
>>> platform.release()
'10'
```

ashb:
Ah I've done a bit of digging and technically it's not even all versions of windows 10, but anyone since the Creators update, build 1809, so I'll make the check even more detailed.

(And I've checked that this will still work on Github Actions too. On Server 2022 it returns `10.0.20348` for the version.

src/black/__init__.py
```diff
@@ -1393,6 +1393,22 @@ def patch_click() -> None:
...
 
 def patched_main() -> None:
+    #: Fixes errors with emoji in  Windows terminals when output is redirected
+    # (i.e. pre-commit): https://github.com/psf/black/issues/3156
+    # To be as safe as possible, we only enable this for "Windows 10 October
+    # 2018 Update (version 1809)" and up
+    if (
+        "pytest" not in sys.modules
+        and platform.system() == "Windows"
+        and tuple(map(int, platform.version().split("."))) >= (10, 0, 1809)
```
ambv, line 1403: Can't decide if I love or hate that last line. Advantages:
- it's correct (as it uses `int` comparisons);
- it performs well (as the `and` short circuits for non-Windows users);
- fits in a single line.

The disadvantage is that it reads pretty clumsily ("make a tuple out of an int map of the string that `platform.version()` returns split by the period... and then compare it to (10, 0, 1809)"). Especially that int map bit.

But I guess the advantages outweigh my personal preference here.

src/black/__init__.py
```diff
@@ -1393,6 +1393,22 @@ def patch_click() -> None:
 
 
 def patched_main() -> None:
+    #: Fixes errors with emoji in  Windows terminals when output is redirected
+    # (i.e. pre-commit): https://github.com/psf/black/issues/3156
+    # To be as safe as possible, we only enable this for "Windows 10 October
+    # 2018 Update (version 1809)" and up
```
ambv, line 1399:
```suggestion
    # 2018 Update (version 1809)" and up.
```

CHANGES.md
```diff
@@ -42,6 +42,9 @@
 
 <!-- Changes to Black's terminal output and error messages -->
 
+- Force output messages to always be UTF-8 on Windows (so we can see the emoji), even if
+  terminal is redirected or captured (#3374)
```
ambv, line 46:
```suggestion
- Fix emoji rendering on Windows 10+ when standard output is redirected
  or captured by forcing UTF-8 encoding in this case (#3374)
```

src/black/__init__.py
ashb: Pretty much sums up my feelings when writing it!

CHANGES.md
```diff
@@ -51,6 +51,9 @@
 
 <!-- Changes to Black's terminal output and error messages -->
 
+- Fix emoji rendering on Windows 10+ when standard output is redirected
+  or captured by forcing UTF-8 encoding in this case (#3374)
```
ashb, line 55:
```suggestion
- Fix emoji rendering on Windows 10+ when standard output is redirected or captured by
  forcing UTF-8 encoding in this case (#3374)
```

src/black/__init__.py
```diff
@@ -1411,6 +1411,22 @@ def patch_click() -> None:
 
 
 def patched_main() -> None:
+    #: Fixes errors with emoji in  Windows terminals when output is redirected
```
ichard26, line 1414:
```suggestion
    #: Fixes errors with emoji in Windows terminals when output is redirected
```

ichard26 (approved):
Thank you! I appreciate all of your in-depth investigation, I don't understand what's wrong here myself, but this seems safe have read your comments. 

I'll reach out to the click folks tomorrow, but I agree it's very possible click won't fix this for us as it (seems to) requires monkeypatching the standard streams.

Sorry for taking so long to get back to you!

JelleZijlstra:
Coming back to this, I'm not enthusiastic about including this in Black:

- We're monkeypatching the standard library, always a risky thing to do, and we're doing it only in a particular set of obscure conditions (when pytest is absent and we're on a particular version of Windows). That feels like a future maintainability nightmare.
- Printing unicode isn't a core part of what Black is about. We're using Click to create Unicode-compatible output, so I feel like it should be Click's job to fix Unicode issues here. If Click doesn't think it's worth fixing, maybe that's a hint that we shouldn't either.
- The bug this fixes isn't particularly bad. Emojis get printed as "\u1234" sequences, which is ugly but doesn't make Black unusable.

ashb:
If we don't accept this (and I entirely understand why) is there perhaps a place in the docs that I could document the work around to users (setting `PYTHONUTF8=1`)?

JelleZijlstra:
Mentioning that workaround in the docs seems fine.

ashb:
One comment on the "We're monkeypatching the standard library, always a risky thing to do" -- if that is the only concern then I could replace the `sys.stdout = ...` with [contextlib.redirect_stdout](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout)

ashb:
Looks like this won't be accepted so I'll close the pr

PR #3383, opened 2022-11-14
Wordsmith current_style.md

sckarlin:
### Description

Minor tweak to the language in the description of the current style.

In a nutshell, I think the original author meant to say "real-world" rather than "realtime."

### Checklist - did you ...

- [x] Add an entry in `CHANGES.md` if necessary? -> not necessary
- [x] Add / update tests if necessary? -> not necessary

My understanding is that this PR should get the magical label to silence the CHANGELOG entry check.


cooperlees (approved):
Many thanks. Yup, this makes more sense.

If CI passes I'll merge.

PR #3400, opened 2022-12-01
Add lean flag to supress tempting cake emoji

janjur:
### Description

I wanted to start taking good care of quality of code in my project. In order to achieve this, obviously I had to implement black. When black corrected all my formatting mistakes and then I run it again (just in case) I saw following message (**TRIGGER WARNING**):  

`All done! ✨ 🍰 ✨`

And I was schocked. I did not even realise how hungry I was, and the cake looked to delicious. I should add that I'm trying to loose weight. This delightful piece of confectionery art woke my deepest and darkest instincts. I relapsed and went to the corner store to buy chocolate.  

I do not want anyone else to suffer as I did. So I added `--lean` flag which you can use to remove the cake from output.

### Checklist - did you ...

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -384,6 +384,14 @@ def validate_regex(
     ),
 )
 @click.option(
+    "-l",
+    "--lean",
+    is_flag=True,
+    help=(
+        "Do not use the cake emoji on successful run"
+    ),
+)
+@click.option(
     "-v",
     "--verbose",
     is_flag=True,
@@ -441,6 +449,7 @@ def main(  # noqa: C901
     experimental_string_processing: bool,
     preview: bool,
     quiet: bool,
+    lean: bool,
     verbose: bool,
     required_version: Optional[str],
     include: Pattern[str],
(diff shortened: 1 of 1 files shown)


cooperlees:
I am sorry about your reaction to our emoji output. That said, we're very hesitant to add CLI flags. It is very daunting when running `--help` with a new CLI tool and finding walls and walls of options. It has been a priority of black to not get to this state since day one.

Due to this I'd like you to investigate other options. I believe there has been prior discussions about removing emojis and I believe you should be able to add something to your environment to hide them.

JelleZijlstra:
There is no way to suppress the emoji and we've consistently rejected requests to add one. However, the `-q` option will suppress most output, including this message.

janjur:
Thank you for your rapid response.  I've add a little rant in the description as you stated you do not mind one.  There was only a grain of truth to it so please do not be sorry.
I totally understand and respect your opinion on keeping the interface minimalistic.
Best regards.

JelleZijlstra:
Thank you! 🥬

PR #3412, opened 2022-12-09
Fix CI with latest flake8-bugbear

JelleZijlstra:
### Checklist - did you ...


JelleZijlstra:
strict= is only in 3.10+. I'm going to disable the lint rule instead until we drop support for 3.9.

cooperlees (approved):
Ahhh yes - We should disable this. Forgot we had all optional enabled here.

Thanks!

PR #3422, opened 2022-12-10
Fix crash in return annotation with walrus

JelleZijlstra:
Fixes #3419

### Checklist - did you ...

- [ ] Add an entry in `CHANGES.md` if necessary?
- [ ] Add / update tests if necessary?
- [ ] Add new / update outdated documentation?

PR #3423, opened 2022-12-10
Check stability for both preview and non-preview styles

### LINKED ISSUE #3419: Preview style incorrectly removes parens from walrus in annotation
```
$ black -c 'def this_is_so_dumb() -> (please := no): ...'
def this_is_so_dumb() -> (please := no):
    ...
$ black --preview -c 'def this_is_so_dumb() -> (please := no): ...'
def this_is_so_dumb() -> (please := no): ...
error: cannot format <string>: Cannot parse: 1:32: def this_is_so_dumb() -> please := no:
$ black --version
black, 22.12.0 (compiled: yes)
Python (CPython) 3.9.14
```

Found this in CI for #3418, cf. #3407.

Seems like a result of #2990, cc @jpy-git.

Broader learning: for all stable style tests, we should run tests in preview style that assert that formatting at least succeeds.

### LINKED ISSUE #3420: Preview mode crash with `--fast` and walrus in except clause
```
$ black --fast -c '''
try: pass
except (a := 1) as (b := why_does_this_exist): pass
'''
try:
    pass
except (a := 1) as (b := why_does_this_exist):
    pass
$ black --preview --fast -c '''
try: pass
except (a := 1) as (b := why_does_this_exist): pass
'''

try: pass
except (a := 1) as (b := why_does_this_exist): pass
error: cannot format <string>: Cannot parse: 3:9: except a := 1 as (b := why_does_this_exist):
$ black --version
black, 22.12.0 (compiled: yes)
Python (CPython) 3.9.14
```

Similar to #3419, found this in CI for #3418. This needs `--fast` because CPython actually rejects walrus here (good call), but the grammar allows it.

This one probably comes from #2939, cc @jpy-git again.

JelleZijlstra:
For all format tests, check that both preview and non-preview succeed without crashes. This makes sure we catch bugs early where preview mode introduces a new crash.

Fixes #3419. Fixes #3420. Closes #3422 (this PR is a more general solution).


JelleZijlstra:
This idea can be generalized to other aspects of the mode: run the stability/equivalence tests only while setting some part of the mode to the opposite of what the test normally does. I tried locally and found that every new configuration adds about 4 s to the test runtime. The most interesting configuration was setting line-length to 1, which found #3424, #3425, #3427, and #3428. Setting `is_pyi` to true found #3427.

I'm going to add a case that does line length = 1, preview = False, because that configuration seems to be good at catching bugs.

tests/util.py
```diff
@@ -70,12 +71,46 @@ def assert_format(
...
+    fast: bool = False,
+    minimum_version: Optional[Tuple[int, int]] = None,
+) -> None:
     actual = black.format_str(source, mode=mode)
-    _assert_format_equal(expected, actual)
+    if expected is not None:
+        _assert_format_equal(expected, actual)
     # It's not useful to run safety checks if we're expecting no changes anyway. The
     # assertion right above will raise if reality does actually make changes. This just
     # avoids wasted CPU cycles.
```
ichard26, line 112:
```suggestion
```
Unfortunately this optimization is now gone and I can't find a way to bring it back :(

ichard26 (commented):
Is it clear when the secondary format runs are the ones erroring out? I just don't want this to confuse contributors unfamiliar with our testing setup. I'm not happy this quadruples the runtime of the `test_format.py` tests (12s -> 45s), but c'est la vie.

tests/util.py
JelleZijlstra: Not sure why? The optimization still applies.

ichard26: Ah, you are right. My bad.

JelleZijlstra:
> Is it clear when the secondary format runs are the ones erroring out? I just don't want this to confuse contributors unfamiliar with our testing setup.

I pushed a change that wraps failures from these checks to hopefully make it clear.

> I'm not happy that this quadruples the runtime of the `test_format.py` tests (12s -> 45s), but c'est la vie.

I don't think it's going to be quite so bad, where did you get that number from? The current version only adds two extra checks, and the test does more than just calling this function, so it shouldn't 3x the overall runtime.

tests/util.py
```diff
@@ -70,12 +75,58 @@ def assert_format(
...
+    try:
+        _assert_format_inner(
+            source,
+            None,
+            replace(mode, preview=False, line_length=1),
+            fast=fast,
+            minimum_version=minimum_version,
+        )
+    except Exception as e:
+        text = "non-preview" if mode.preview else "preview"
```
ichard26, line 109:
```suggestion
```

ichard26 (approved):
I should've clarified this was on my machine (which is probably underpowered at this point). On GHA CI, it's negligible. 

And yes when you take the runtime of the rest of the test suite, it's not that bad.

tests/util.py
JelleZijlstra: oops, thanks

PR #3426, opened 2022-12-10
Fix syntax error in match test

JelleZijlstra:
Found with a variation on #3423 that runs tests in pyi mode.


isidentical (approved):
Nice catch 💯

Issue #3331, opened 2022-10-16
Docstring reformatting causes line to be too long


adamjstewart:
**Describe the bug**

Black is overzealous about reformatting docstrings to a single line, and can cause the resulting line to be longer than the line length limit.

**To Reproduce**

For example, take this code:
```python
def f():
    """89 123456789 123456789 123456789 123456789 123456789 123456789 123456789 12345678
    """
```
The first line of this docstring is exactly 88 characters. After blackening this code, this results in:
```python
def f():
    """89 123456789 123456789 123456789 123456789 123456789 123456789 123456789 12345678"""
```
The line is now 91 characters, and flake8 complains.

**Expected behavior**

Honestly, I wouldn't expect black to reformat docstrings at all. But I definitely wouldn't expect it to reformat to the point where style-compliant code becomes non-style-compliant.

**Environment**

- Black's version: 22.10.0, main
- OS and Python version: macOS 12.6, Python 3.9.13

**Additional context**

This appears to be a regression of the same bug reported in #1632 and #2274.

felix-hilden:
With the "preview" toggle enabled, this is [fixed](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADVAF9dAD2IimZxl1N_WlbvK5V9Hvo0zZlhiNcX2rzeBITicv6kV76qn7XtBxdbDUKlpDsBRgxCmvXDH4y3cSCeHL0G8Ak1dw6rxU3pwHzEUI3v4fV9F1jWCB_Rc_g_r7L3HpzgAACmJaHJbC_t3wABe9YBAAAAQ8S5HLHEZ_sCAAAAAARZWg==). Also worth mentioning that we're discussing the exact opposite behavior in #3320.

adamjstewart:
Thanks, glad to know it's already been fixed!

Issue #2274, opened 2021-05-29
Docstring new line reformatting can violate line length constraint


DrGFreeman:
**Describe the bug**

Removal of new line in docstrings can violate the line length constraint if the last line is within two characters of the max line length value.

**To Reproduce** Steps to reproduce the behavior:

Run black on the following code with the default line-length constraint of 88 characters.

```python
def f():
    """89 123456789 123456789 123456789 123456789 123456789 123456789 123456789 12345678
    """
    pass
```

**Expected behavior**

No reformatting is performed.

**Actual behavior**

The code is reformatted as shown below, which violates the max line length constraint.

```python
def f():
    """89 123456789 123456789 123456789 123456789 123456789 123456789 123456789 12345678"""
    pass
```

In a CI setup with Flake8 running after black, Flake8 throws the following error:

```
test.py:2:89: E501 line too long (91 > 88 characters)
```

**Environment (please complete the following information):**

- All versions of black from 20.8b0 up to 21.5b1 as well as main (d3670d9c6568aef270ccda3006252ee3388cf587).
- Reproduced on Windows 10 & Ubuntu 18.04/20.04, Python 3.7 & 3.8

**Does this bug also happen on main?**

Yes

**Additional context**

black 19.10b0 does not have this behavior.
This was reported in https://github.com/psf/black/issues/1635#issuecomment-696230833.

ichard26:
I'll be closing this since this is a duplicate of GH-1632, thank you for the report though!

DrGFreeman:
Oh! I should have checked before. My bad. Sorry for wasting your time 😓

ichard26:
Don't worry about it, we do happen to have 250+ issues so it can be (very) annoying to discover whether your issue has been already reported or not. It's my job as a triager to catch the mistakes anyhow :)

PR #3430, opened 2022-12-11
Do not put the closing quotes in a docstring on a separate line

### LINKED ISSUE #3320: style regression: too long single line docstrings have their quotes moved to a new line.
**Describe the bug**

Single-line docstrings that are longer than the line length have the final three `"` moved to a new line.

**To Reproduce**

For example, take this code:
```py
from typing import Optional

# leaves this the same
def large_image_url(self) -> Optional[str]:
    """Optional[:class:`str`]: Returns a URL pointing to the large image asset of this activity, if applicable.
    """
    pass


# leaves this the same too
def large_image_url(self) -> Optional[str]:
    """
    Optional[:class:`str`]: Returns a URL pointing to the large image asset of this activity, if applicable.
    """
    pass


# depending on the length of the docstring, this one *may* be left as-is
def large_image_url(self) -> Optional[str]:
    """
    Optional[:class:`str`]: Returns a URL pointing to the large image asset of this activity, if applicable."""
    pass


# but this one is modified
def large_image_url(self) -> Optional[str]:
    """Optional[:class:`str`]: Returns a URL pointing to the large image asset of this activity, if applicable."""
    pass

```

```sh
$ black file.py --preview
```

We get this format:

```py
from typing import Optional

# leaves this the same
def large_image_url(self) -> Optional[str]:
    """Optional[:class:`str`]: Returns a URL pointing to the large image asset of this activity, if applicable.
    """
    pass


# leaves this the same too
def large_image_url(self) -> Optional[str]:
    """
    Optional[:class:`str`]: Returns a URL pointing to the large image asset of this activity, if applicable.
    """
    pass


# this one is also left as-is
def large_image_url(self) -> Optional[str]:
    """
    Optional[:class:`str`]: Returns a URL pointing to the large image asset of this activity, if applicable.
    """
    pass


# but this one is modified
def large_image_url(self) -> Optional[str]:
    """Optional[:class:`str`]: Returns a URL pointing to the large image asset of this activity, if applicable.
    """
    pass
```

**Expected behavior**

Single line docstrings are left alone. I understand what this change is supposed to do, and on already multiline docstrings this wouldn't be an issue (although it would be nice for the docstring formatter to be entirely consistent among all docstrings instead of having different behavior for the line length but i digress).

**Environment**

- Black's version:  22.10.0
- OS and Python version:  Python 3.11.0rc2, on linux, but i can recreate with python 3.8 too

**Additional Comments**

This seems like a regression from #3044

JelleZijlstra:
Fixes #3320. Followup from #3044.

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -389,19 +389,18 @@ class LineGenerator(Visitor[Line]):
                 # We need to find the length of the last line of the docstring
                 # to find if we can add the closing quotes to the line without
                 # exceeding the maximum line length.
-                # If docstring is one line, then we need to add the length
-                # of the indent, prefix, and starting quotes. Ending quotes are
-                # handled later.
+                # If docstring is one line, we don't put the closing quotes on a
+                # separate line because it looks ugly (#3320).
                 lines = docstring.splitlines()
                 last_line_length = len(lines[-1]) if docstring else 0
-
-                if len(lines) == 1:
-                    last_line_length += len(indent) + len(prefix) + quote_len
 
                 # If adding closing quotes would cause the last line to exceed
                 # the maximum line length then put a line break before the
                 # closing quotes
-                if last_line_length + quote_len > self.mode.line_length:
+                if (
+                    len(lines) > 1
+                    and last_line_length + quote_len > self.mode.line_length
+                ):
                     leaf.value = prefix + quote + docstring + "\n" + indent + quote
                 else:
                     leaf.value = prefix + quote + docstring + quote
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@
 <!-- Changes that affect Black's preview style -->
 
 - Fix a crash in preview style with assert + parenthesized string (#3415)
+- Do not put the closing quotes in a docstring on a separate line, even if the line is
+  too long (#3430)
 
 ### Configuration
 
--- a/tests/data/preview/docstring_preview.py
+++ b/tests/data/preview/docstring_preview.py
@@ -54,13 +54,11 @@ def single_quote_docstring_over_line_limit2():
 
 
 def docstring_almost_at_line_limit():
-    """long docstring.................................................................
-    """
+    """long docstring................................................................."""
 
 
 def docstring_almost_at_line_limit_with_prefix():
-    f"""long docstring................................................................
-    """
+    f"""long docstring................................................................"""
 
 
 def mulitline_docstring_almost_at_line_limit():


ichard26 (commented):
Code looks fine in general, but this code is hard to grok so let's see what diff-shades has to say first.

JelleZijlstra:
The diff-shades report is as expected, just removing newlines between one-line docstrings and the closing quotes. Example:

```
 def _static_hasattr(value: object, attr: str) -> bool:
-    """Returns whether this value has the given attribute, ignoring __getattr__ overrides.
-    """
+    """Returns whether this value has the given attribute, ignoring __getattr__ overrides."""
     try:
```

src/black/linegen.py
```diff
@@ -389,19 +389,18 @@ def visit_STRING(self, leaf: Leaf) -> Iterator[Line]:
...
-                    last_line_length += len(indent) + len(prefix) + quote_len
-
                 # If adding closing quotes would cause the last line to exceed
                 # the maximum line length then put a line break before the
                 # closing quotes
-                if last_line_length + quote_len > self.mode.line_length:
+                if (
+                    len(lines) > 1
+                    and last_line_length + quote_len > self.mode.line_length
+                ):
```
ichard26, line 403: Why does this not take indentation into account when determining whether the line length would be violated?

ichard26 (commented):
On second look, this code looks suspicious.

src/black/linegen.py
ichard26: Oh hmm, sometimes it does take into account the indentation. Looks like `lines` doesn't contain what I expected it to. Nothing here makes sense :)

ichard26 (approved):
Actually this seems fine on a third look.

ichard26:
Thanks!

jensguballa:
I have created a [comment](https://github.com/psf/black/pull/3044#issuecomment-1366026400) on #3044, but as this here is the issue that has been merged, I am providing my thought here again.

After going over the issue I am a little bit puzzled now, as I am not really sure what the intended behavior should be.

My assumption was that black should not apply reformattings which will introduce new line length violations. In my opinion this sounds reasonable, given that you might have a processing pipeline where e.g. flake8 is doing its job hereafter and would fail (like in my case). But it looks like the resolution for this issue is going in another direction.

My view: For the [example](https://github.com/psf/black/pull/3044#issuecomment-1364710654) I have provided black needs to take a decision: either reformatting to a one-line string and violating the maximum line length, or, leaving the doc string as it is and keeping the conformity.

My preference is clearly to keep the conformity, but if I understand the resolution here correctly, reformatting to a one-line string is given priority. I am not in favor of this as special crafting of the doc string by the developer just to work around this issue would go against the philosophy of black.

So would it be reasonable not to reformat the doc string for the example I provided?

ichard26:
The crux of the problem here is that Black shouldn't move quotes that already violate the line length limit onto a new line (it's ugly and isn't compliant with PEP 257), but _also_ shouldn't move quotes that were already on their own line. This is impossible to do w/ Black's philosophy of avoiding looking at the original formatting as much as possible.  

I agree Black shouldn't cause more violations of the line length limit. Technically this introduces one more situation where Black is dependant on pre-existing formatting, but I don't think docstrings can be formatted in an "one size for all" way IMHO.

jensguballa:
I do know nothing about black's internals, but for the example I provided would it be possible to implement something like this for one-line doc strings:

Original code (maximum line length not exceeded):
```
class _YamlBlockStyle(str):
    """Class used to indicate that a string shall be serialized using the block style.
    """
```

What black produces currently (one-line doc string, maximum line length exceeded):
```
class _YamlBlockStyle(str):
    """Class used to indicate that a string shall be serialized using the block style."""
```

What I propose black could produce (multi-line doc string, maximum line length not exceeded):
```
class _YamlBlockStyle(str):
    """
    Class used to indicate that a string shall be serialized using the block style.
    """
```

I think that output would comply to the maximum line length as well as to PEP 257. If the latter output would still exceed the maximum line length, then of course there is nothing black can do.

adamjstewart:
If I'm understanding correctly, this PR re-introduces the following issues: #1632, #2274, #3331?

It's great that we're now PEP 257 compliant, but now black reformats my code such that flake8 complains about line length. What's the suggested approach for users using both black and flake8, to add `# noqa: E501` to every docstring in this situation? I would argue that this is uglier than having the closing triple quotes on a separate line...

PR #3439, opened 2022-12-13
Fix a crash when a colon line is marked between `# fmt: off` and `# fmt: on`

### LINKED ISSUE #3436: Crash on conditional inside # fmt: off, depending on comment indentation
**Describe the bug**

If a condition statement is wrapped in `# fmt: off`/`# fmt: on`, Black crashes with a parse error on the subsequent line.
The crash only occurs if the `# fmt: off` comment is indented the same as the wrapped condition.

**To Reproduce**

The following minimal code breaks parsing with Black 22.10 and later:
```python
def do_something(x, y):
    if x:
        return x
    # fmt: off
    elif y:
    # fmt: on
        return y
```

```sh
$ black test.py
```

The resulting error is:

> error: cannot format test.py: Cannot parse: 7:4:     return y

**Expected behavior**

Black can parse this valid Python snippet regardless of how the `# fmt: off` comment is indented.
This is the case for Black 22.6 and older.

**Environment**

- Black's version: 22.10
- OS and Python version: Linux/Python 3.8

**Additional context**

Interestingly, Black doesn't crash if the fmt-off comment is indented one more level.
This is parsed fine:
```python
def do_something(x, y):
    if x:
        return x
        # fmt: off
    elif y:
    # fmt: on
        return y
```
However, the instruction is also ignored, ie. the `# fmt: off` command isn't recognised as such.

yilei:
### Description

Fixes #3436.

Note that the `# fmt: on` line is always normalized to the next statement so its indentation is moved in this case. This isn't ideal, but at least it shouldn't crash.

### Checklist - did you ...

### DIFF
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -232,7 +232,7 @@ def generate_ignored_nodes(
 
         # fix for fmt: on in children
         if children_contains_fmt_on(container, preview=preview):
-            for child in container.children:
+            for index, child in enumerate(container.children):
                 if isinstance(child, Leaf) and is_fmt_on(child, preview=preview):
                     if child.type in CLOSING_BRACKETS:
                         # This means `# fmt: on` is placed at a different bracket level
@@ -240,6 +240,14 @@ def generate_ignored_nodes(
                         # we include this closing bracket in the ignored nodes.
                         # The alternative is to fail the formatting.
                         yield child
+                    return
+                if (
+                    child.type == token.INDENT
+                    and index < len(container.children) - 1
+                    and children_contains_fmt_on(
+                        container.children[index + 1], preview=preview
+                    )
+                ):
                     return
                 if children_contains_fmt_on(child, preview=preview):
                     return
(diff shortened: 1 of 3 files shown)


src/black/comments.py
```diff
@@ -241,6 +241,14 @@ def generate_ignored_nodes(
...
                         yield child
                     return
+                if (
+                    child.type == token.INDENT
+                    and index < len(container.children) - 1
+                    and children_contains_fmt_on(
+                        container.children[index + 1], preview=preview
+                    )
+                ):
+                    return
```
ichard26, line 251: Comment please :)

ichard26 (approved):
I feel like we should write developer docs on how `fmt: off/on/skip` work. The logic as-is is already complicated as heck and the edge cases are incredibly confusing sometimes...

Also technically this is invalid use of `# fmt: off/on` but this is just a limitation of Black since there's no way to represent a `# fmt: on` at the right indent level right between indention levels.

ichard26:
Oh where are my manners?... Thank you very much for investigating and filing a PR, it's appreciated! :black_heart:

src/black/comments.py
yilei: Good request, done :)

JelleZijlstra (approved):
Thank you!

Issue #1197, opened 2019-12-09
Line length should consider Unicode character width


uranusjr:
**Is your feature request related to a problem? Please describe.** A clear and concise

In some cases one would want to put non-ASCII characters in source. This is most common
for strings, but in Python 3 you can use non-ASCII identifiers as well.

But consider the following example:

```python
parser.add_argument(
    "poem", help="Poem to process",
    default="ちはやふる神代もきかず竜田川からくれなゐに水くくるとは",
)
```

The `default` line is visually long, and I’d like to put it on its own line. But Black would insist formatting it into

```python
parser.add_argument(
    "poem", help="Poem to process", default="ちはやふる神代もきかず竜田川からくれなゐに水くくるとは"
)
```

since it thinks there is enough space (indeed, the formatted line is only 73 characters long).

**Describe the solution you'd like**

When calculating line lengths, Black should consult the [width attribute](http://www.unicode.org/reports/tr11/tr11-36.html) of each character, and count double width for characters with EAW *F* or *W*.

**Describe alternatives you've considered**

Having strictly single-width characters in source, and pull in double-width strings at runtime (via i18n or `\u` sequences. The translation system is usually an overkill cases (e.g. developing an in-house CLI where all users speak a common East Asian language), and replacing all strings with `\u` make the source illegible to humans.

Usgin `fmt: on/off` becomes tedious and hard to the eye quickly when you need a lot of string scattered around (e.g. for logging purposes).

**Additional context** Add any other context or screenshots about the feature request
here.

![Screenshot 2019-12-09 at 15 00 35](https://user-images.githubusercontent.com/605277/70414310-ae4d3980-1a94-11ea-9ad2-c5864807b0e5.png)

mk-pmb:
:+1: and I'd be willing to try and implement it.

mk-pmb:
Related @ maintainers: Would you also want [zero-width characters](https://www.ptiglobal.com/2018/04/26/the-beauty-of-unicode-zero-width-characters/) to not add to line length?

JelleZijlstra:
This makes sense to me in general, but I'm not familiar with the Unicode subtleties involved here.

I think the relevant question for computing line length is how good, Unicode-aware code editors display these characters. So if zero-width characters are actually normally displayed as zero-width, we should count them as zero-width.

I'm also concerned about whether this change could affect performance for the common case of having only normal-width characters.

mk-pmb:
> for the common case of having only normal-width characters.

[...]

PR #1879, opened 2020-12-22
Improve multiline string handling

aneeshusa:
### Description

Improve formatting of multiline strings, especially in function calls,
by updating black to look at the context around multiline strings
to decide if they should be inlined or split to a separate line.
Currently behind the --preview flag.

Performance tested the new functionality in https://github.com/psf/black/pull/1879#issuecomment-1292668826.
Fixes https://github.com/psf/black/issues/256.

### Checklist - did you ...

- [x] Add an entry in `CHANGES.md` if necessary?
- [x] Add / update tests if necessary?
- [x] Add new / update outdated documentation?

PR #3445, opened 2022-12-17
Let string splitters respect `East_Asian_Width` property

dahlia:
### Description

This patch changes the preview style so that string splitters respect Unicode [East Asian Width](https://www.unicode.org/reports/tr11/) property.  If you are not familiar to CJK languages it is not clear immediately.  Let me elaborate with some examples.

Traditionally, East Asian characters (including punctuations) have taken up space twice than European letters and stops when they are rendered in monospace typeset.  Compare the following characters:

```
abcdefg.
글、字。
```

The characters at the first line are half-width, and the second line are full-width.  (Also note that the last character with a small circle, the East Asian period, is also full-width.)  Therefore, if we want to prevent those full-width characters to exceed the maximum columns per line, we need to count their *width* rather than the number of characters.  Again, the following characters:

```
글、字。
```

These are just 4 characters, but their total width is 8.

Suppose we want to maintain up to 4 columns per line with the following text:

```
abcdefg.
글、字。
```

How should it be then?  We want it to look like:

```
abcd
efg.
글、
字。
```

However, Black currently turns it into like this:

```
abcd
efg.
글、字。
```

It's because Black currently counts the number of characters in the line instead of measuring their width.  So, how could we measure the width?  How can we tell if a character is full- or half-width?  What if half-width characters and full-width ones are mixed in a line?  That's why Unicode defined an attribute named `East_Asian_Width`.  Unicode grouped every single character according to their width in fixed-width typeset.

So, this patch tried to change how Black measure line widths using `East_Asian_Width` attribute, and I believe it works well with full-width characters now.  However, I am not confident if I implemented things in a proper way, so please let me know if I need to adjust!

Also, I believe it partially addresses #1197, but I touched only string splitters.  Other parts probably need to be fixed as well.

### Checklist - did you ...


src/black/strings.py
```diff
@@ -4,8 +4,9 @@
 
 import re
 import sys
+import unicodedata
 from functools import lru_cache
-from typing import List, Pattern
+from typing import List, Pattern, Text
```
JelleZijlstra, line 9: Text is only relevant for Python 2/3 compatibility. Black is Python 3-only, so just use `str`.

dahlia: Instead of importing `typing.Text`, I replaced `Text` with `str`!

item4:
How nice!

But I have a question. How about handling for non string literal such as variable name, function name, class name, comment?

Examples are below
```py
# comment

# 이 값을 건드리면 당신은 야근을 면치 못 할 것입니다. 이상하게 보여도 절대 수정하지 마세요 모두를 위해 그냥 둡시다. - 전임자 드림
ANSWER = 42
```
```py
# function name

def 영어주소2한글주소(country: str, province: str, city: str, street_address1: str, street_address2: str, extra_data: dict[str, any]) -> str:
    pass
```

dahlia:
@item4 Other than string literals are not covered by this patch.  I would appreciate if someone works on it further.

dahlia:
Here's also few more points we need to address in the future:

- There are zero-width characters like U+200B ZERO WIDTH SPACE.  They are designated `N` (narrow) according to `East_Asian_Width`, which would be counted as a single column.  We need to count them as zero column.
- Multiple code points can be combined together and look like a single character when they are rendered.  They are called combining characters or grapheme clusters.  (E.g., umlaut letters, emoji with skin color.)  Those characters should be counted together.

src/black/lines.py
```diff
@@ -718,7 +719,7 @@ def is_line_short_enough(line: Line, *, line_length: int, line_str: str = "") ->
     if not line_str:
         line_str = line_to_string(line)
     return (
-        len(line_str) <= line_length
+        str_width(line_str) <= line_length
```
JelleZijlstra, line 722: This change appears to affect the stable style. It must be made in the preview style only. (The trans.py code is already preview-only so no need to change anything there.)

src/black/trans.py
```diff
@@ -71,6 +73,8 @@ class CannotTransform(Exception):
 TResult = Result[T, CannotTransform]  # (T)ransform Result
 TMatchResult = TResult[Index]
 
+SPLIT_SAFE_CHARS = frozenset(["\u3001", "\u3002", "\uff0c"])  # East Asian stops
```
JelleZijlstra, line 76: I see the motivation for this change but it feels separate from the width change, so it should be called out separately in the changelog.

JelleZijlstra:
Also, I'm a little worried about performance here. `char_width()` is obviously going to be slower than `len()`, but will this noticeably affect Black's overall performance?

Possible optimizations:
- Use a fast path for ascii-only strings (`str.isascii` method), since many programs will contain mostly ASCII-only strings
- Cache the result of the computation per string; this is only helpful if we often re-compute the width for the same string

src/black/lines.py
dahlia: Would it be a good idea to let `is_line_short_enough()` to take `mode: Mode` instead of `line_length: int`?  Or would it be better to just add a new parameter `preview: bool` to the function?

dahlia: I just added a new keyword parameter `preview: bool`.  Please let me know if it's not appropriate.

src/black/trans.py
dahlia: Indeed.  I added one more changelog for this.

dahlia:
I also added a fast path for ASCII-only strings!

src/black/strings.py
```diff
@@ -236,3 +237,40 @@ def normalize_string_quotes(s: str) -> str:
...
+
+
+def str_width(line_str: str) -> int:
+    """Return the width of `line_str` as it would be displayed in a terminal
+    or editor (which respects Unicode East Asian Width).
+
+    You could utilize this function to determine, for example, if a string
+    is too wide to display in a terminal or editor.
+    """
+    if line_str.isalpha():
```
JelleZijlstra, line 259:
```suggestion
    if line_str.isascii():
```
`isascii` is O(1) in CPython because string objects store internally whether they're ASCII-only. `isalpha` has to actually loop over the string. It's also not obvious that `isalpha` is correct; I'd have to look up whether any East Asian characters are counted as "alpha" for Unicode purposes.

dahlia: Actually, it was not intended, but just a typo.  I amended the commit!

ZeroRin:
Consider using [wcwidth](https://github.com/jquast/wcwidth)? In my experience in other projects this is more accurate for characters other than east-asian, see [this discussion](https://github.com/masaccio/compact-json/issues/4#issuecomment-1357049123)
Also wcwidth uses a `lru_cache` which should be helpful for fast computation

dahlia:
> Consider using [wcwidth](https://github.com/jquast/wcwidth)? In my experience in other projects this is more accurate for control characters, non-asian characters, etc. (see [this](https://github.com/masaccio/compact-json/issues/4#issuecomment-1357049123)) Also wcwidth uses `lru_cache` which should be helpful for the performance

Sounds good to me.  @JelleZijlstra I wonder if it's appropriate to add a new dependency for this specific logic.

ZeroRin:
There also exists [Rich](https://github.com/Textualize/rich/blob/master/rich/cells.py) who generates a table with wcwidth and does not need it as dependency. Same strategy can be adopted here as well. I personally prefer using existing packages instead of copying the code though

JelleZijlstra:
Looks like `wcwidth` is a pure-Python package with no further dependencies, so it wouldn't be too bad to add.

However, I am a little concerned about making formatting output depend on a third-party package. Right now if you have the same version of Black installed as someone else, you'll get the same formatting. But with wcwidth, perhaps different versions of the library will count line length differently, which would cause different formatting. That could create a confusing experience for users. We could get around it by pinning wcwidth to a specific version, but that's not great either.

ichard26:
If we want to use wcwidth, I'd prefer using a mechanism similar to rich to avoid the issues Jelle brings up. Otherwise pinning is "fine," but that's not great either (may cause dependency conflicts for some users if they use another tool that uses wcwidth). I'm kinda surprised rich seems to use a binary search instead of a dictionary lookup, but maybe there's some complexity I'm overlooking?

ZeroRin:
> I'm kinda surprised rich seems to use a binary search instead of a dictionary lookup, but maybe there's some complexity I'm overlooking?
Actually wcwidth itself is already using bisearch. To my understanding the key reason is that the width information is stored as a interval table instead of a full dictionary. A full dictionary for every unicode character would have ~a million records i guess, but the interval table consists of only ~500 records, as a block of  characters tend to have same width.
The rich implementation differs from the original wcwidth in that it merges bisearch table for 0-width and 2-width into a single, and that it caches width for strings with less then 512 characters for reuse.

src/black/strings.py
ZeroRin: strictly speaking `isascii` does not ensure a string width being the same as its length since `\u0000`~`\u001f` are control characters with no width. But that should not be a problem for formatting code

dahlia:
If we decide to generate Black's own table to avoid a runtime dependency on wcwidth, should the generated table be version-controlled, or should it be generated at build time (through a [PEP-517]-style custom build backend or a release script in the CI configuration)?

[PEP-517]: https://peps.python.org/pep-0517/

ichard26:
While writing a plugin for Hatchling (our build backend) would be super cool, it's probably not the best idea given it still gives room for install-to-install variability as Jelle brought up. It would be rare given almost no one installs from sdist, but the possibility is still there. Extending our release CD would be nice, but we haven't tried adding any automated source changes before and in general it sounds like it could devolve into a debugging nightmare. Doing it the rich way (feel free to add the new logic in subpackage of the Black package) would probably be best. 

However, I'd like @JelleZijlstra's input.

JelleZijlstra:
Generating a file and checking it in seems like the best way. That way, we can make sure it stays the same for the whole year and keep the stable style stable.

dahlia:
Now `char_width()` depends on the generated `wcwidth()` table instead of `unicodedata.east_asian_width()`, and it's LRU-cached.  I also changed the fast path condition of `str_width()` to address [what @ZeroRin pointed out](https://github.com/psf/black/pull/3445#discussion_r1062985596).

@JelleZijlstra Could you take a look again?

src/black/linegen.py
```diff
@@ -508,7 +508,12 @@ def transform_line(
         and not line.should_split_rhs
         and not line.magic_trailing_comma
         and (
-            is_line_short_enough(line, line_length=mode.line_length, line_str=line_str)
+            is_line_short_enough(
+                line,
+                line_length=mode.line_length,
```
JelleZijlstra, line 513: You can use the `ll` local variable here.

src/black/strings.py
```diff
@@ -236,3 +238,57 @@ def normalize_string_quotes(s: str) -> str:
...
+
+
+def str_width(line_str: str) -> int:
+    """Return the width of `line_str` as it would be displayed in a terminal
+    or editor (which respects Unicode East Asian Width).
+
+    You could utilize this function to determine, for example, if a string
+    is too wide to display in a terminal or editor.
+    """
+    if COMMON_SINGLE_WIDTH_CHARS.match(line_str):
```
JelleZijlstra, line 277: The reason I suggested `.isascii()` specifically is that thanks to a peculiarity of how strings are implemented, it can just look at an internal flag on the string object and so it takes O(1) time. Therefore, it's significantly faster than other similar-seeming methods like `isalpha()`, which in turn are quite a bit faster than running the regex engine.

To demonstrate:
```
In [3]: COMMON_SINGLE_WIDTH_CHARS: Final = re.compile(r"^[\x20-\x7e\xa0-\u02ff]*$")

In [4]: s = "a" * 40

In [5]: %timeit COMMON_SINGLE_WIDTH_CHARS.match(s)
229 ns ± 0.664 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [6]: %timeit s.isascii()
11.5 ns ± 0.0159 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

In [7]: %timeit s.isalpha()

79.2 ns ± 0.102 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

So `isascii` is about 20x faster than this regex.

If we use `isascii`, we'll no longer identify control chars as width 0. But I think that might be the right call anyway, because many editors will display control-chars (e.g. here is VSCode on a file with all the ASCII-range characters):

Screen Shot 2023-01-20 at 4 27 06 AM

So I think we should stick with `isascii`.

scripts/make_width_table.py
```diff
@@ -0,0 +1,50 @@
...
+    if start_codepoint >= 0:
+        yield (start_codepoint, end_codepoint, range_width)
+
+
+def main() -> None:
+    table_path = join(dirname(__file__), "..", "src", "black", "_width_table.py")
+    with open(table_path, "w") as f:
+        f.write(
+            f"""# Generated by {basename(__file__)}
+import sys
```
JelleZijlstra, line 33: The output should include the version of relevant tools that was used to generate this. I think that can include:

- wcwidth
- The Unicode version that wcwidth uses
- The Python version, if wcwidth relies on the Unicode DB that comes bundled with Python

scripts/make_width_table.py
```diff
@@ -0,0 +1,50 @@
...
```
JelleZijlstra, line 1: Please add a docstring explaining what this file does and when it should be run.

pyproject.toml
```diff
@@ -78,6 +78,7 @@ colorama = ["colorama>=0.4.3"]
 uvloop = ["uvloop>=0.15.2"]
 d = [
   "aiohttp>=3.7.4",
+  "wcwidth==0.2.5",
```
JelleZijlstra, line 81: Don't think we need this (this extra is for blackd).

dahlia:
@JelleZijlstra I addressed things you pointed out.  Could you take a look again?

scripts/make_width_table.py
```diff
@@ -0,0 +1,69 @@
...
+import wcwidth
+
+
+def make_width_table() -> Iterable[Tuple[int, int, int]]:
+    start_codepoint = -1
+    end_codepoint = -1
+    range_width = -2
+    for codepoint in range(0, sys.maxunicode + 1):
+        width = wcwidth.wcwidth(chr(codepoint))
+        if width == 1:
```
JelleZijlstra, line 31: Should we also ignore width = 0 here? That would be consistent with the `isascii` check.

I think for Black's purposes we can treat width-0 characters like width-1, in the expectation that editors will display the character in some way.

dahlia: Fixed!

JelleZijlstra (approved):

JelleZijlstra:
diff-shades is failing on pandas apparently because there are no files to analyze, cc @ichard26

dahlia:
I rebased commits on the latest main and resolved conflicts (mostly from #1879).  Please take a look again!

JelleZijlstra (approved):

dahlia:
The test suite apparently failed on the CI job.  Should I try to fix it?

JelleZijlstra:
@yilei kindly fixed it in #3615. I'll merge in main to hopefully get tests to pass.

ichard26 (approved):
I didn't check the width table code too closely since I presume that they were borrowed from rich. Everything else looks good though. Many thanks for the PR!

PR #3461, opened 2022-12-19
Do not move docker `latest_release` tag for Pre-Releases

### LINKED ISSUE #3453: Docker latest_release is pointing to an alpha version
The latest docker release on docker hub is pointing to an alpha version. I would have expected `latest_release` to be the latest release, and not the very latest on git. 

Is there a way to fix this to go back to 22.12.0?

cooperlees:
- When we do a pre-release lets not move the latest_release tag
  - This tag should only move on official real releases

Fixes #3453

*Would love any ideas on how to test this ...*

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -44,7 +44,9 @@ jobs:
           tags: pyfound/black:latest,pyfound/black:${{ env.GIT_TAG }}
 
       - name: Build and push latest_release tag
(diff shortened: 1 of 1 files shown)


ichard26 (approved):
I don't understand [the GitHub Webhooks documentation at all](https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=released#release), but this looks about right.

ichard26:
We also need to block the stable job from the PyPI workflow from running on a prerelease. That would be easy enough to test and I'll do that soon.

PR #3465, opened 2022-12-20
Add latest_prerelease Docker Hub tag for following the latest alpha release

matthewarmand:
### Description

Over here, https://github.com/psf/black/issues/3453#issuecomment-1357988758 while fixing an issue with the Docker image publishing workflow, it seemed there's a legitimate desire for a docker tag which follows alpha (prerelease) releases of Black. This PR creates a new tag to be pushed to Docker Hub going forward (`latest_prerelease` which will be updated with every alpha version release.

**I do have another change to request along with this...**

If someone comes across the official Black image on Docker Hub, the description is a shortened version of the README with a link to the full docs. There isn't an obvious pointer to the documentation of which Docker image tags are in use and for what purpose. I don't think duplicating the information there is necessarily the right way to go (unless it's sourced in this repository), but having another explicit link to that specific docs page would grease the wheels just a bit for folks discovering it on Docker Hub.

I don't see where in the source code that Description on DockerHub is maintained, and so I think it's probably been manually set by a maintainer. Linking to Black's Full Documentation still makes sense to me, but I think an additional line just above that would be helpful, something like the following:

> Official Docker Image Documentation: https://black.readthedocs.io/en/latest/usage_and_configuration/black_docker_image.html

### Checklist - did you ...

- [X] Add / update tests if necessary? (not necessary)

### DIFF
--- a/docs/usage_and_configuration/black_docker_image.md
+++ b/docs/usage_and_configuration/black_docker_image.md
@@ -13,7 +13,8 @@ _Black_ images with the following tags are available:
 - `latest_prerelease` - tag created when a new alpha (prerelease) version of _Black_ is
   released.\
   ℹ Recommended for users who want to preview or test alpha versions of _Black_. Note that
-  the most recent release may be newer than any prerelease.
+  the most recent release may be newer than any prerelease, because no prereleases
+  are created before most releases.
 - `latest` - tag used for the newest image of _Black_.\
   ℹ Recommended for users who always want to use the latest version of _Black_, even before
   it is released.
--- a/docs/usage_and_configuration/black_docker_image.md
+++ b/docs/usage_and_configuration/black_docker_image.md
@@ -13,8 +13,8 @@ _Black_ images with the following tags are available:
 - `latest_prerelease` - tag created when a new alpha (prerelease) version of _Black_ is
   released.\
   ℹ Recommended for users who want to preview or test alpha versions of _Black_. Note that
-  the most recent release may be newer than any prerelease, because no prereleases
-  are created before most releases.
+  the most recent release may be newer than any prerelease, because no prereleases are created
+  before most releases.
 - `latest` - tag used for the newest image of _Black_.\
   ℹ Recommended for users who always want to use the latest version of _Black_, even before
   it is released.
--- a/docs/usage_and_configuration/black_docker_image.md
+++ b/docs/usage_and_configuration/black_docker_image.md
@@ -10,7 +10,8 @@ _Black_ images with the following tags are available:
 - `latest_release` - tag created when a new version of _Black_ is released.\
   ℹ Recommended for users who want to use released versions of _Black_. It maps to [the latest release](https://github.com/psf/black/releases/latest)
   of _Black_.
-- `latest_prerelease` - tag created when a new alpha (prerelease) version of _Black_ is released.\
+- `latest_prerelease` - tag created when a new alpha (prerelease) version of _Black_ is
+  released.\
   ℹ Recommended for users who want to preview or test alpha versions of _Black_.
 - `latest` - tag used for the newest image of _Black_.\
   ℹ Recommended for users who always want to use the latest version of _Black_, even before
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Awesome. Thanks. LGTM - Will leave it a day or two to give other maintainers a chance to voice any thoughts etc.

We don't do pre-relaases that often.

docs/usage_and_configuration/black_docker_image.md
```diff
@@ -10,6 +10,9 @@ _Black_ images with the following tags are available:
 - `latest_release` - tag created when a new version of _Black_ is released.\
   ℹ Recommended for users who want to use released versions of _Black_. It maps to [the latest release](https://github.com/psf/black/releases/latest)
   of _Black_.
+- `latest_prerelease` - tag created when a new alpha (prerelease) version of _Black_ is
+  released.\
+  ℹ Recommended for users who want to preview or test alpha versions of _Black_.
```
JelleZijlstra, line 15: One concern here is that we don't usually use prereleases, we're just doing it right now as we're preparing for the 23.1 release. I'm expecting that for the rest of 2023 we won't have any prereleases, then we'll have 24.1a releases again in December 2023. So I think we should at least call out here that there may be a release that's more recent than the latest prerelease.

cooperlees: Good call. I did think of this but thought people could deal with it. But being more explicit is only going to help.

Maybe just call it prerelease and state it's only planned generally once a year for testing `--preview` moving into releases.

matthewarmand: That's good to know, and from a documentation standpoint I think it fits best in [the Release Process docs](https://black.readthedocs.io/en/latest/contributing/release_process.html). Currently there's no mention there of alpha versions or this yearly process, and if there's a thought that this will happen each December, it might be a good addition there.

Then, this bit here can be a bit more specific and possibly also point to the release process docs for more detail.

Thoughts?

JelleZijlstra: That makes sense! I kind of came up with this alpha process on the fly because I thought it'd be helpful. However, writing that new documentation would be a bigger change that's outside the scope of this PR, so for now we can just have a brief note here like "Note that the most recent release may be newer than any prerelease."

matthewarmand: Added, take a gander at that whenever you have a second

docs/usage_and_configuration/black_docker_image.md
```diff
@@ -10,6 +10,10 @@ _Black_ images with the following tags are available:
 - `latest_release` - tag created when a new version of _Black_ is released.\
   ℹ Recommended for users who want to use released versions of _Black_. It maps to [the latest release](https://github.com/psf/black/releases/latest)
   of _Black_.
+- `latest_prerelease` - tag created when a new alpha (prerelease) version of _Black_ is
+  released.\
+  ℹ Recommended for users who want to preview or test alpha versions of _Black_. Note that
+  the most recent release may be newer than any prerelease.
```
JelleZijlstra, line 16:
```suggestion
  the most recent release may be newer than any prerelease, because no prereleases
  are created before most releases.
```
I feel like this will make the statement easier to understand for most users.

matthewarmand: Lol I was wondering whether there'd be more changes to this. I agree that's clearer :+1:

JelleZijlstra:
Thanks @matthewarmand for your contribution!

matthewarmand:
No worries at all @JelleZijlstra, thanks for the feedback and the support.

Just to tie this off, there is still the suggestion of linking directly to the Docker Image docs from the Docker Hub overview; again I didn't see that source in this repo so if maintainers like that idea someone would probably need to go update that manually on https://hub.docker.com/r/pyfound/black.

Thanks again, I really appreciate your extremely timely support and attention.

PR #3467, opened 2022-12-20
Significantly speedup ESP on large expressions that contain many strings

### LINKED ISSUE #3340: maximum recursion depth exceeded while calling a Python object
**Describe the bug**

When running black using pre-commit against a generated Django migration file, the following error is observed:  
```
➜  backend git:(main) ✗ pre-commit run black --all-files             
- hook id: black
- exit code: 123

error: cannot format backend/src/black_test.py: maximum recursion depth exceeded while calling a Python object

Oh no! 💥 💔 💥
167 files left unchanged, 1 file failed to reformat.
```

**To Reproduce**

Here is a file that can reproduce the above issue: https://gist.github.com/trumpet2012/057e1e6bc381081e06a2eaf9e22db45a

Install pre-commit: https://pre-commit.com/index.html#installation

Our pre-commit config is:
```yaml
default_language_version:
  python: python3
exclude: (^|/)\.idea/
repos:
    rev: 22.10.0
    hooks:
      - id: black
        args:
          # enable preview mode to get the new string processing
          - --preview
```

Then running:

```sh
$ pre-commit run black --all-files
```

Results in the above error.

**Expected behavior**

Black should be able to correctly reformat the file.

**Environment**

- Black's version: 22.10.0
- OS and Python version: Mac OS Monterey, Python 3.9.14
- pre-commit version: 2.20.0

**Additional context**
It works when you run black without the `--preview` flag.

Interestingly, it also works running black directly (bypassing pre-commit) even with the `--preview` flag enabled.

Also when I was generating the example file above, I was going through and replacing our internal column names and help text with fake data and when I replaced the `help_text=` strings with just `Lorem ipsum dolores` it also worked correctly. Increasing the length of the help text strings then caused the issue to reappear.

yilei:
### Description

Previously, when ESP 1) merges strings in `StringMerger`; 2) strips parens in `StringParenStripper`, it finds and does one transform for one string group. Each transform will create a new `line` with 1) one group of strings merged; 2) one group of strings' parens stripped. This new `line` is then re-checked by those transformers. This is `O(n^2)` complexity.

Since these transformers won't cause line breaks, the same transforms can be done at one pass, and it only creates one new `line`. The new approach is `O(n)`.

Tested on the example from #3340 (not compiled, since I'm failing to use cibuildwheel to compile with mypyc on my machines):

1. Before this change, it took ~111 seconds to finish after I increased the recursionlimit to 1M
2. After this change, it took ~3.7s
3. In the stable style without ESP, this file takes ~2.5s. The increase is now sane.

Fixes #3340, since this no longer triggers recursion limit errors.

Hopefully this also solves #2314

### Checklist - did you ...

### DIFF
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -881,8 +881,7 @@ class StringParenStripper(StringTransformer):
     ) -> Iterator[TResult[Line]]:
         LL = line.leaves
 
-        # A list of tuples of (string_idx, rpar_idx)
-        string_and_rpar_indices: List[Tuple[int, int]] = []
+        string_and_rpar_indices: List[int] = []
         for string_idx in string_indices:
             string_parser = StringParser()
             rpar_idx = string_parser.parse(LL, string_idx)
@@ -895,7 +894,7 @@ class StringParenStripper(StringTransformer):
                     should_transform = False
                     break
             if should_transform:
-                string_and_rpar_indices.append((string_idx, rpar_idx))
+                string_and_rpar_indices.extend((string_idx, rpar_idx))
 
         if string_and_rpar_indices:
             yield Ok(self._transform_to_new_line(line, string_and_rpar_indices))
@@ -905,18 +904,24 @@ class StringParenStripper(StringTransformer):
             )
 
     def _transform_to_new_line(
-        self, line: Line, string_and_rpar_indices: List[Tuple[int, int]]
+        self, line: Line, string_and_rpar_indices: List[int]
     ) -> Line:
         LL = line.leaves
 
         new_line = line.clone()
         new_line.comments = line.comments.copy()
 
-        previous_rpar_idx = -1
-        for string_idx, rpar_idx in string_and_rpar_indices:
+        previous_idx = -1
+        # We need to sort the indices, since string_idx and its matching
+        # rpar_idx may not come in order, e.g. in
+        # `("outer" % ("inner".join(items)))`, the "inner" string's
+        # string_idx is smaller than "outer" string's rpar_idx.
+        for idx in sorted(string_and_rpar_indices):
+            leaf = LL[idx]
+            lpar_or_rpar_idx = idx - 1 if leaf.type == token.STRING else idx
             try:
                 append_leaves(
-                    new_line, line, LL[previous_rpar_idx + 1 : string_idx - 1]
+                    new_line, line, LL[previous_idx + 1 : lpar_or_rpar_idx]
                 )
             except BracketMatchError:
                 # HACK: I believe there is currently a bug somewhere in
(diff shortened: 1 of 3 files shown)


src/black/trans.py
```diff
@@ -198,14 +198,19 @@ def __init__(self, line_length: int, normalize_strings: bool) -> None:
...
-            * Ok(string_idx) such that `line.leaves[string_idx]` is our target
-            string, if a match was able to be made.
+            * Ok(string_indices) such that for each index, `line.leaves[index]`
+            is our target string if a match was able to be made. For
+            transformers that don't result in more lines (e.g. StringMerger,
+            StringParenStripper), multiple matches and transforms are done at
+            once to reduce the complexity.
                 OR
-            * Err(CannotTransform), if a match was not able to be made.
+            * Err(CannotTransform), if no match was not able to be made.
```
JelleZijlstra, line 207:
```suggestion
            * Err(CannotTransform), if no match could be made.
```
I think the old wording was correct, but this suggested wording is more clear.

JelleZijlstra (commented):
Thanks, this is an exciting improvement! I'll need to study the code some more and see how it affects perf on my codebase where I previously saw terrible performance.

src/black/trans.py
yilei: Updated.

yilei:
I ran this over our codebase, and fixed a bug.

[This](https://github.com/psf/black/blob/1e8217fd6284bdb020e7ca70964d677a3016f914/src/black/linegen.py#L544-L562) is the outer loop, previously `string_merge` and `string_paren_wrap` are generating one transformed new line per string group in the large expression. With this PR, they now generate a single new line for all the transformed string groups.

JelleZijlstra:
Thanks for the report!

Still seeing somewhat bad performance on an internal file with a large expression (nested calls, dictionaries, lots of strings). I think I reported #2314 based on a similar file. On this PR branch:

```
$ git checkout .; rm -rf ~/.cache/black/
Updated 57 paths from the index
$ time black --preview path/to/file.py
reformatted path/to/file.py

All done! ✨ 🍰 ✨
1 file reformatted.

real    1m33.226s
user    1m32.556s
sys     0m0.468s
$ git checkout .; rm -rf ~/.cache/black/
Updated 1 path from the index
$ time black path/to/file.py
All done! ✨ 🍰 ✨
1 file left unchanged.

real    0m19.135s
user    0m18.740s
sys     0m0.176s
$ black --version
black, 22.12.1.dev27+ga120a1d (compiled: no)
Python (CPython) 3.9.14
```

However, on 22.12.0 (compiled), the same file in preview mode takes:
```
real    8m7.575s
user    7m57.408s
sys     0m9.836s
```

JelleZijlstra:
So this PR clearly makes things better, but the performance penalty on ESP is still big enough (probably; I haven't tried with mypyc) that we should have a conversation about whether the tradeoff is acceptable. That can wait though.

src/black/trans.py
```diff
@@ -371,30 +376,50 @@ def do_match(self, line: Line) -> TMatchResult:
 
         is_valid_index = is_valid_index_factory(LL)
 
-        for i, leaf in enumerate(LL):
+        string_indices = []
+        idx = 0
+        while is_valid_index(idx):
```
JelleZijlstra, line 381: Probably best for another PR, but there may be an optimization opportunity here. `is_valid_index(idx)` is basically equivalent to `idx < len(LL)`, but it does so through some indirection and a nested function. According to Jukka mypyc isn't good at compiling nested functions, so this code may get a good speedup if we just use `idx < len(LL)`.

JelleZijlstra (approved):

src/black/trans.py
JelleZijlstra: I tried this (https://github.com/JelleZijlstra/black/commit/9b1f0b11ffd449d6cbf9f2326e316257a707a008) but didn't see a significant difference under ESP, though I didn't do rigorous benchmarking.

PR #3468, opened 2022-12-20
Vim plugin docs improvements

matthewarmand:
### Description

This is a few documentation changes for the Vim plugin. I included them in the same PR to avoid clutter and because they're all pretty minor and easily reviewed together. I also made a commit for each thematic change, so feel free to look at specific commits if you want to see the changes in isolation. I didn't think this was a substantial enough change to warrant a `CHANGES.md` entry, but am happy to add one if maintainers disagree.

There are 4 main changes in here:

#### Formatting and Organization

Commit: https://github.com/psf/black/pull/3468/commits/26c56f2a902dc59690a6526d9fd7e5d369f57b5b and https://github.com/psf/black/pull/3468/commits/197edbd58a55b8e9446a61bd8e1ff2d8b26b32bd

Currently there are no sub-headers under the Official Plugin section, which can make it a bit tough to read and especially to hyperlink to. Organized it into Installation, Usage, and Troubleshooting as felt appropriate.

#### Add missing configuration options

Commit: https://github.com/psf/black/pull/3468/commits/9929893e57154e4b4d4707feeba7376039286056

A few configuration options exist but weren't mentioned with their defaults at the top of the Vim plugin documentation section, so I added those.

#### Add plugin documentation for Arch Linux

Commit: https://github.com/psf/black/pull/3468/commits/bb0f44602edb0215159cb232ea4fede4f84d14ce

Refer to this issue on the Arch Linux `python-black` package: https://bugs.archlinux.org/task/73024

Since earlier this calendar year, the Black package distributed on Arch Linux ([`python-black`](https://archlinux.org/packages/community/any/python-black/)) also ships the Vim plugin, so I added a section to the Installation section mentioning that Arch Linux users need no further steps to use the plugin than installing Black.

#### Fix the vim-plug installation recommendation

Commit: https://github.com/psf/black/pull/3468/commits/1a73b724677ebf54894a238172b6074d950ea8cc

Refer to this discussion on the `vim-plug` repo: https://github.com/junegunn/vim-plug/pull/720#issuecomment-1105829356

Using the `stable` tag as a `branch` in `vim-plug` is a bit of an antipattern that doesn't work correctly, as it doesn't always re-resolve the SHA when the `stable` tag is moved. According to the maintainer, using the `tag` key with a shell wildcard is the better way to go. I selected `'*.*.*'` as the wildcard since it matches [Black's documented versioning scheme](https://black.readthedocs.io/en/latest/contributing/release_process.html\#cutting-a-release) and will not match on the current standard for Alpha versions (which only have one dot). You can see how this works in the image below; it matches on `22.12.0` (the latest stable release) even though the latest version overall at time of writing is `23.1a1`.

image

### Checklist - did you ...

- [X] Add an entry in `CHANGES.md` if necessary? (Seems a small enough change to not need an entry, do maintainers agree?)
- [X] Add / update tests if necessary? (just docs, not necessary)

### DIFF
--- a/docs/integrations/editors.md
+++ b/docs/integrations/editors.md
@@ -111,7 +111,10 @@ Configuration:
 - `g:black_fast` (defaults to `0`)
 - `g:black_linelength` (defaults to `88`)
 - `g:black_skip_string_normalization` (defaults to `0`)
+- `g:black_skip_magic_trailing_comma` (defaults to `0`)
 - `g:black_virtualenv` (defaults to `~/.vim/black` or `~/.local/share/nvim/black`)
+- `g:black_use_virtualenv` (defaults to `1`)
+- `g:black_target_version` (defaults to `""`)
 - `g:black_quiet` (defaults to `0`)
 - `g:black_preview` (defaults to `0`)
 
--- a/docs/integrations/editors.md
+++ b/docs/integrations/editors.md
@@ -164,19 +164,6 @@ curl https://raw.githubusercontent.com/psf/black/stable/autoload/black.vim -o ~/
 Let me know if this requires any changes to work with Vim 8's builtin `packadd`, or
 Pathogen, and so on.
 
-##### With ALE
-
-1. Install [`ale`](https://github.com/dense-analysis/ale)
-
-1. Install `black`
-
-1. Add this to your vimrc:
-
-   ```vim
-   let g:ale_fixers = {}
-   let g:ale_fixers.python = ['black']
-   ```
-
 #### Usage
 
 This plugin **requires Vim 7.0+ built with Python 3.7+ support**. It needs Python 3.7 to
@@ -271,6 +258,19 @@ If you later want to update _Black_, you should do it like this:
 ```console
 $ pip install -U black --no-binary typed-ast
 ```
+
+### With ALE
+
+1. Install [`ale`](https://github.com/dense-analysis/ale)
+
+1. Install `black`
+
+1. Add this to your vimrc:
+
+   ```vim
+   let g:ale_fixers = {}
+   let g:ale_fixers.python = ['black']
+   ```
 
 ## Gedit
 
--- a/docs/integrations/editors.md
+++ b/docs/integrations/editors.md
@@ -120,6 +120,10 @@ Configuration:
 
 #### Installation
 
+This plugin **requires Vim 7.0+ built with Python 3.7+ support**. It needs Python 3.7 to
+be able to run _Black_ inside the Vim process which is much faster than calling an
+external command.
+
 ##### `vim-plug`
 
 To install with [vim-plug](https://github.com/junegunn/vim-plug):
@@ -137,14 +141,14 @@ which matches the given pattern.
 
 The following matches all stable versions (see the
 [Release Process](../contributing/release_process.md) section for documentation of
-version scheme used by Black),
+version scheme used by Black):
 
 ```
 Plug 'psf/black', { 'tag': '*.*.*' }
 ```
 
 and the following demonstrates pinning to a specific year's stable style (2022 in this
-case).
+case):
 
 ```
 Plug 'psf/black', { 'tag': '22.*.*' }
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):

JelleZijlstra:
I think it may be worth calling out in CHANGES.md as we're changing the recommended way to install the vim plugin.

JelleZijlstra:
Also there's a CI failure, you'll need to use prettier to reformat the .md file.

JelleZijlstra:
Oh and the comment you linked about the versioning scheme refers to the stable *tag*. But a while ago we changed to have a stable *branch*, because using a tag that keeps changing is indeed weird. Does that change the situation?

matthewarmand:
> Oh and the comment you linked about the versioning scheme refers to the stable tag. But a while ago we changed to have a stable branch, because using a tag that keeps changing is indeed weird. Does that change the situation?

I'd missed that shift, I think it probably does change things. IIRC specifying a branch works better with `PlugUpdate`, so that would actually work fine. I guess at that point it's user preference whether they're fine with that branch or they prefer a branch pattern. Up to you really then whether we change this, add `tag` as an alternative, or just leave it as is, since I think functionally the current recommendation is probably fine.

> I think it may be worth calling out in CHANGES.md as we're changing the recommended way to install the vim plugin.

I'm assuming this conditional is solely based on the `vim-plug` section, rather than the Arch one, correct? In other words, if it's decided above to leave the `vim-plug` docs alone, we can ignore the change log?

> Also there's a CI failure, you'll need to use prettier to reformat the .md file.

All glory to the ~~hypnotoad~~ linter

JelleZijlstra:
> > Oh and the comment you linked about the versioning scheme refers to the stable tag. But a while ago we changed to have a stable branch, because using a tag that keeps changing is indeed weird. Does that change the situation?
> 
> I'd missed that shift, I think it probably does change things. IIRC specifying a branch works better with `PlugUpdate`, so that would actually work fine. I guess at that point it's user preference whether they're fine with that branch or they prefer a branch pattern. Up to you really then whether we change this, add `tag` as an alternative, or just leave it as is, since I think functionally the current recommendation is probably fine.

I might like the `tag: *.*.*` pattern better because it provides an obvious way to instead pin a specific version if you're so inclined, or to pin to `tag: 22.*.*` if you want the 2022 stable style. I don't use the plugin though, so no strong opinion either way.

Also, users might have weird cached state from back when `stable` was a tag.

> 
> > I think it may be worth calling out in CHANGES.md as we're changing the recommended way to install the vim plugin.
> 
> I'm assuming this conditional is solely based on the `vim-plug` section, rather than the Arch one, correct? In other words, if it's decided above to leave the `vim-plug` docs alone, we can ignore the change log?

Agree.

matthewarmand:
>I might like the tag: *.*.* pattern better because it provides an obvious way to instead pin a specific version if you're so inclined, or to pin to tag: 22.*.* if you want the 2022 stable style. I don't use the plugin though, so no strong opinion either way.
>
>Also, users might have weird cached state from back when stable was a tag.

Yeah I agree, I kinda prefer the `tag` method anyway for a few reasons, and especially with the versatility of pinning to a year, that's a good callout. I'll expand the section just a bit to explain those options (and probably keep the old one) and then add a CHANGES entry as discussed.

docs/integrations/editors.md
```diff
@@ -111,16 +111,47 @@ Configuration:
...
 Plug 'psf/black', { 'branch': 'stable' }
 ```
 
+Another option which is a bit more explicit and offers more control is to use
+`vim-plug`'s `tag` option with a shell wildcard. This will resolve to the latest tag
+which matches the given pattern.
+
+The following matches all stable versions (see the
+[Release Process](../contributing/release_process.md) section for documentation of
+version scheme used by Black),
```
JelleZijlstra, line 140:
```suggestion
version scheme used by Black):
```

docs/integrations/editors.md
```diff
@@ -111,16 +111,47 @@ Configuration:
...
+
+```
+Plug 'psf/black', { 'tag': '*.*.*' }
+```
+
+and the following demonstrates pinning to a specific year's stable style (2022 in this
+case).
```
JelleZijlstra, line 147:
```suggestion
case):
```

docs/integrations/editors.md
```diff
@@ -148,6 +187,8 @@ curl https://raw.githubusercontent.com/psf/black/stable/autoload/black.vim -o ~/
 Let me know if this requires any changes to work with Vim 8's builtin `packadd`, or
 Pathogen, and so on.
 
+#### Usage
+
 This plugin **requires Vim 7.0+ built with Python 3.7+ support**. It needs Python 3.7 to
```
JelleZijlstra, line 192: This paragraph doesn't really fit under "Usage". Maybe put it at the top of the "Installation" section?

matthewarmand: Yep good call, moved.

JelleZijlstra (approved):

Issue #3486, opened 2023-01-09
Wrap multiple context managers in parens when targeting Python 3.10+


yilei:
So that we actually implement https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html#using-backslashes-for-with-statements:

> Although when the target version is Python 3.9 or higher, Black will use parentheses instead since they’re allowed in Python 3.9 and higher.

I'll work on a PR to implement this when targeting Python 3.9+ ~~3.10+ (since it's only officially allowed in 3.10 https://docs.python.org/3/whatsnew/3.10.html#parenthesized-context-managers, though accidentally enabled in Python 3.9 with the new parser).~~ (per discussion, enabling it for Python 3.9+ should be fine in practice)

Note: #2926 is related.

bstableynv5:
btw, Ruff, probably using Black style rules, is reformatting multi-context manager using \ to use () and we've seen python raise syntax errors because of it.

ArcGIS Pro v3.1.2 on python 3.9.16 (Python 3.9.16 [MSC v.1931 64 bit (AMD64)] :: Anaconda, Inc. on win32)
<img width="855" height="262" alt="Image" src="https://github.com/user-attachments/assets/c60f714d-761f-4c34-b712-a221027bce06" />

JelleZijlstra:
Bugs in Ruff are better reported to Ruff, not here. If you report a bug, please include a precise code sample with the unexpected behavior.

Also, Python 3.9 is now unsupported.

bstableynv5:
I didn't expect support. Just thought you might like to be aware.

PR #3488, opened 2023-01-09
Documentation: clarify the state of multiple context managers.

yilei:
The backslash & paren-wrapping formatting for multiple context managers aren't yet implemented.

Related to #3486.


ichard26 (approved):
Thanks!

PR #3489, opened 2023-01-10
Wrap multiple context managers in parentheses when targeting Python 3.9+

### LINKED ISSUE #3486: Wrap multiple context managers in parens when targeting Python 3.10+

yilei:
### Description

Fixes #3486.

This also fixes the target version detection logic for parenthesized context managers (3.9) and match statements (3.10). But as far as I can tell, this didn't have user visible impact (until this PR).

I added a dedicated test data folder `preview_context_managers`, because this relies on the version detection logic when `--target-version` is not explicitly specified.

### Checklist - did you ...


yilei:
I manually ran `diff-shades analyze base.json --work-dir projects-cache --force-preview-style -- --target-version=py39` forcing to target 3.9. The attached diffs look good.

[cm.diff.zip](https://github.com/psf/black/files/10387302/cm.diff.zip)

First example:

```diff

╭──────────────────────── Summary ─────────────────────────╮
│ 12 projects & 29 files changed / 608 changes [+347/-261] │
│                                                          │
│ ... out of 2 384 528 lines, 11 521 files & 23 projects   │
╰──────────────────────────────────────────────────────────╯

[aioexabgp - https://github.com/cooperlees/aioexabgp.git]
╰─> revision fc83c5c705e3c818934ce0cdd04d6169a6977996
--- a/aioexabgp:aioexabgp/tests/fibs_tests.py
+++ b/aioexabgp:aioexabgp/tests/fibs_tests.py
@@ -190,31 +190,37 @@
                 )
             )

     def test_del_all_routes(self) -> None:
         # Test delete all with a nexthop
-        with patch(
-            f"{BASE_MODULE}.LinuxFib.get_route_table",
-            fibs_tests_fixtures.mocked_get_route_table,
-        ), patch(
-            f"{BASE_MODULE}.LinuxFib.del_route", return_value=True
-        ) as mock_del_route:
+        with (
+            patch(
+                f"{BASE_MODULE}.LinuxFib.get_route_table",
+                fibs_tests_fixtures.mocked_get_route_table,
+            ),
+            patch(
+                f"{BASE_MODULE}.LinuxFib.del_route", return_value=True
+            ) as mock_del_route,
+        ):
             self.assertTrue(
                 self.loop.run_until_complete(
                     self.lfib.del_all_routes(ip_address("fd00::4"))
                 )
             )
             # 1 prefix/route from v4 and 1 from v6 table
             self.assertEqual(2, mock_del_route.call_count)

         # Test no next hope and ensure we delete them all
-        with patch(
-            f"{BASE_MODULE}.LinuxFib.get_route_table",
-            fibs_tests_fixtures.mocked_get_route_table,
-        ), patch(
-            f"{BASE_MODULE}.LinuxFib.del_route", return_value=True
-        ) as mock_del_route:
+        with (
+            patch(
+                f"{BASE_MODULE}.LinuxFib.get_route_table",
+                fibs_tests_fixtures.mocked_get_route_table,
+            ),
+            patch(
+                f"{BASE_MODULE}.LinuxFib.del_route", return_value=True
+            ) as mock_del_route,
+        ):
             self.assertTrue(
                 self.loop.run_until_complete(self.lfib.del_all_routes(None))
             )
             # 2 prefix/route from v4 and 1 from v6 table
             self.assertEqual(3, mock_del_route.call_count)
```

tests/data/preview_context_managers/auto_detect/features_3_8.py
```diff
@@ -0,0 +1,18 @@
+# This file doesn't used any Python 3.9+ only grammars.
```
JelleZijlstra, line 1: Can you test that something like `with (a): pass` doesn't get autodetected as 3.9+?

yilei: Done

JelleZijlstra (approved):

JelleZijlstra:
Thank you as always!

PR #3500, opened 2023-01-17
Use dashes for pycodestyle max line length config

meliache:
### Description

The option is `max-line-length` with dashes, not underscores. The config option name is given in the output of `pycodestyle -h`, which can also be checked on https://pep8.readthedocs.io/en/stable/intro.html#example-usage-and-output:
```
Configuration:
    The project options are read from the [pycodestyle] section of the
    tox.ini file or the setup.cfg file located in any parent folder of the
    path(s) being processed.  Allowed options are: exclude, filename,
    select, ignore, max-line-length, max-doc-length, hang-closing, count,
    format, quiet, show-pep8, show-source, statistics, verbose
```

### Checklist - did you ...

- ~[ X] Add an entry in `CHANGES.md` if necessary?~
- ~[ X] Add / update tests if necessary?~
- [ X] Add new / update outdated documentation?


JelleZijlstra:
I just tried this and both options work with pycodestyle (version 2.10.0 on Python 3.11). Still, it's better to be consistent with pycodestyle's own docs, so I'll accept the change.

However, CI is currently failing because of an issue on main that's already fixed. I would update the branch myself but it looks like you didn't give permission for that. Could you merge main into your branch to make CI pass?

meliache:
Merged the main branch via the "sync" button in github's UI, hopefully that works and tests will pass now.

(This is the first PR that I did without ever cloning the repo, but completely in the github UI. Which I would never do do for code changes due to the absence of linting etc, but for a minor documentation change it seems okay.)

JelleZijlstra:
Yes, editing from the UI is always a bit tricky.

Seems like that didn't work, there are now a lot of unrelated changes.

Normally I'd click the "Update branch" button next to the merge button, but it seems like you disallowed edits by maintainers.

meliache:
I found the issue, this is a PR to merge my branch into `stable` :man_facepalming: . This is also the reason why it was behind `main` and why merging main introduced more changes. I suggest to close this now and re-open a PR onto `main`, will do it from my editor/cli to be sure. Btw, the _allow edits from maintainers_ tickmark is set, but probably the _update_ function might not work properly when it's not a PR onto stable, don't know :shrug:

JelleZijlstra:
Oh yes, that would do it! Glad we added the lint check against PRs to `stable`. I *think* it should be possible to retarget a PR to a different branch, but yeah it might be simpler to just make a new PR.

meliache:
Re-opened as #3513, sorry for the noise.

Issue #3501, opened 2023-01-17
Please make emojis optional


fredrgrs:
It seems that black indiscriminently prints emojis.  Could you PLEASE give us an option to disable unicode output, just like we can disable color output?  Not everyone is able to or chooses to use unicode-capable terminals, and these multi-byte characters will screw up many terminals, as they can be interpreted as terminal control codes or expected to render as single characters per byte.

fredrgrs:
Thanks so much for jumping on this!

Jackenmen:
This has been repeatedly rejected, do you have any reason for wanting this which hasn't already been mentioned in these issues?

See:
#223
#1031
#1255
#1270
#3359

fredrgrs:
Wow, it's obvious from these other issues that this is a common complaint and that you're really willing to die on this hill.  I expect this to be closed as the others have.  It's a brave new world, I guess, and we dinosaurs can get stuffed.

JelleZijlstra:
I am rejecting this feature request since there's nothing different from the last few times it has come up.

cladmi:
A reason for asking this, the normal black output when pasted to `Atlassian/Stash` breaks on the database https://jira.atlassian.com/browse/BSERV-8490

Yeah really annoying for anybody pasting "proof" of output. Fixing Atlassian tools is a lifelong goal to reach anything.

swills1:
Some of us chain checks together. I have a Bash script that runs on Git's `pre-commit` hook. It runs black, isort, mypy, and flake8. The emojis make more info to trudge through and makes my output longer / inconsistent. It also hurts logging. I am getting around this currently by redirecting the output;

```
black --check $directory_to_check #> /dev/null 2>&1
exit_code=$?

if [ $exit_code -eq 0 ]; then
    echo "Pass: black passed"
else
    echo "Fail: black failed"
fi
```
If I see it failed, I can run black without the redirect and get the error info.

PR #3502, opened 2023-01-17
Add --emoji/--no-emoji option

Secrus:
### Description

Resolves: #3501 
This PR adds a CLI option to disable emojis in output messages.

### Checklist - did you ...

  **(Didn't find any tests for the output, but can add them if necessary)**
   **(Could use some directions as to where to put the update about new CLI option)**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -64,6 +64,7 @@
 
 - Verbose logging now shows the values of `pyproject.toml` configuration variables
   (#3392)
+- Added `--emoji/--no-emoji` option to disable emojis in output messages (#3502)
 
 ### _Blackd_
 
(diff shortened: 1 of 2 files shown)


.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 # B905 should be enabled when we drop support for 3.9
-ignore = E203, E266, E501, W503, B905
+ignore = E203, E266, E501, W503, B905, B028
```
Secrus, line 3: The addition to ignore of this new option was discussed on Discord.

JelleZijlstra: Would you mind doing this in a separate PR too? We definitely need this change but I'm skeptical about the `--emoji` option.

Secrus: Done in #3502. 

Feel free to unalive this PR if you don't feel like this belongs in the codebase. I just saw the issue and it got me thinking about how hard could this be. One thing led to another and the PR appeared :smile:

JelleZijlstra:
Thanks for your work, but we're going to reject this feature request.

PR #3503, opened 2023-01-18
Add flake8-bugbear B907 to ignore list

Secrus:
### Description

This adds newly introduced flake8-bugbear `B028` rule to ignore list
### Checklist - did you ...

**None of those are necessary here.**

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,6 @@
 [flake8]
 # B905 should be enabled when we drop support for 3.9
-ignore = E203, E266, E501, W503, B905
+ignore = E203, E266, E501, W503, B905, B907
 # line length is intentionally set to 80 here because black uses Bugbear
 # See https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length for more details
 max-line-length = 80


JelleZijlstra (approved):

cooperlees:
Due to user dislike - New flake8-bugbear released with this check moved to optional B907 ...

.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 # B905 should be enabled when we drop support for 3.9
-ignore = E203, E266, E501, W503, B905
+ignore = E203, E266, E501, W503, B905, B028
```
cooperlees, line 3: Maybe we should just move forward flake8-bugbear version to 23.1.17 + disable B907
```suggestion
ignore = E203, E266, E501, W503, B905, B907
```

cooperlees (commented):
Sorry for the churn.

.flake8
JelleZijlstra: flake8-bugbear isn't pinned for us, so I think we're automatically using the latest.

Secrus: Done

Issue #3507, opened 2023-01-20
Autoload vim script uses non-specific path to python binary


kephircheek:
**Describe the bug**
Following with the official [docs](https://black.readthedocs.io/en/stable/integrations/editors.html#vim)
> On first run, the plugin creates its own virtualenv using the right Python version and automatically installs Black.

but failed with 
```
Error detected while processing /Users/kephircheek/rebecca/vim/plugged/black/autoload/black.vim:
line  214:
Traceback (most recent call last):
  File "<string>", line 113, in <module>
  File "<string>", line 89, in _initialize_black_env
  File "/opt/homebrew/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/venv/__init__.py", line 468, in create
    builder.create(env_dir)
  File "/opt/homebrew/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/venv/__init__.py", line 74, in create
    self.setup_python(context)
  File "/opt/homebrew/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/venv/__init__.py", line 292, in setup_python
    copier(context.executable, path)
  File "/opt/homebrew/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/venv/__init__.py", line 235, in symlink_or_copy
    shutil.copyfile(src, dst)
  File "/opt/homebrew/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/shutil.py", line 256, in copyfile
    with open(src, 'rb') as fsrc:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/bin/python3'
Error detected while processing function black#Black:
```

**Explanations**
My default `python3` has version `3.10.9` but `python3.11` also installed.
Path  `'/opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/bin/` exists and fit look like
```bash
$ ls -1 /opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/bin/
2to3-3.11
idle3.11
pip3
pip3.11
pydoc3.11
python3.11
python3.11-config
```
[...]

PR #3508, opened 2023-01-20
Specify python binary path (#3507)

kephircheek:
### Description
See 'How to fix` in #3507 

### Checklist - did you ...

- [X] Add an entry in `CHANGES.md` if necessary


JelleZijlstra:
Would this break environments that have only python3 but not python3.x available? Not sure if those exist, but maybe it's safer to try python3.x first and only then fall back to python3.

kephircheek:
Ok, sure. What about such idea? [54881f2](https://github.com/psf/black/pull/3508/commits/54881f251fd96f6d01035fd376f5fe3222df3c5b)

JelleZijlstra (approved):

PR #3520, opened 2023-01-24
Update document now that paren wrapping CMs on Python 3.9+ is implemented

yilei:
Implemented in #3489.


JelleZijlstra:
Maybe add something like "in the preview style only"?

yilei:
Like this?

JelleZijlstra (approved):

PR #3531, opened 2023-01-30
Github action now reports state of execution #3531

rasjani:
### Description

Adds 2 new outputs to github actions which would allow pipeline maintainer to do conditional steps based on actions black did.  These variables include `is_formatted` which defaults to 0 if no changes where made and 1 when files where formatted. Another one is `changed_files`, which defaults to 0 but reports the number of files changed during the action execution.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,9 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- GitHub Action: adds 2 action outputs to report changes that can be used in other
+  steps. Outputs: `is-formatted` and `changed-files` (#3531)
+
 ### Documentation
 
 <!-- Major changes to documentation and policies. Small docs changes
--- a/action.yml
+++ b/action.yml
@@ -27,6 +27,13 @@ inputs:
     description: 'Python Version specifier (PEP440) - e.g. "21.5b1"'
     required: false
     default: ""
+outputs:
+  is-formatted:
+    description: "Files checked are formatted to the current black formatter style."
+    value: steps
+  changed-files:
+    description: "Number of files black changed"
+    value: steps
 branding:
   color: "black"
   icon: "check-circle"
--- a/action/main.py
+++ b/action/main.py
@@ -2,9 +2,11 @@ import os
 import shlex
 import sys
 from pathlib import Path
+from re import MULTILINE, search
 from subprocess import PIPE, STDOUT, run
 
 ACTION_PATH = Path(os.environ["GITHUB_ACTION_PATH"])
+GITHUB_OUTPUT = Path(os.environ["GITHUB_OUTPUT"])
 ENV_PATH = ACTION_PATH / ".black-env"
 ENV_BIN = ENV_PATH / ("Scripts" if sys.platform == "win32" else "bin")
 OPTIONS = os.getenv("INPUT_OPTIONS", default="")
@@ -12,6 +14,10 @@ SRC = os.getenv("INPUT_SRC", default="")
 JUPYTER = os.getenv("INPUT_JUPYTER") == "true"
 BLACK_ARGS = os.getenv("INPUT_BLACK_ARGS", default="")
 VERSION = os.getenv("INPUT_VERSION", default="")
+
+_is_formatted_re = r"\s?(?P<changed-files>[0-9]+)\sfiles?\sreformatted(\.|,)\s?"
+
+_outputs = {"is-formatted": "false", "changed-files": "0"}
 
 run([sys.executable, "-m", "venv", str(ENV_PATH)], check=True)
 
@@ -38,8 +44,25 @@ if pip_proc.returncode:
 base_cmd = [str(ENV_BIN / "black")]
 if BLACK_ARGS:
     # TODO: remove after a while since this is deprecated in favour of SRC + OPTIONS.
-    proc = run([*base_cmd, *shlex.split(BLACK_ARGS)])
+    proc = run([*base_cmd, *shlex.split(BLACK_ARGS)], stderr=PIPE)
 else:
-    proc = run([*base_cmd, *shlex.split(OPTIONS), *shlex.split(SRC)])
+    proc = run(
+        [*base_cmd, *shlex.split(OPTIONS), *shlex.split(SRC)],
+        stderr=PIPE
+    )
+# Re-emit stderr back to console so that action output is visible to pipeline
+# Do note, click will strip terminal control codes if the output is not TTY
+# and thus, this will not show colors anymore.
+print(proc.stderr, file=sys.stderr, flush=True)
+
+_output = proc.stderr.decode("utf-8")
+matches = search(_is_formatted_re, _output, MULTILINE)
+if matches:
+    _outputs["is-formatted"] = "true"
+    _outputs["changed-files"] = str(matches.group("changed-files"))
+
+with GITHUB_OUTPUT.open("a+", encoding="utf-8") as f:
+    for k, v in _outputs.items():
+        f.write(f"{k}={v}\n")
 
 sys.exit(proc.returncode)
--- a/docs/integrations/github_actions.md
+++ b/docs/integrations/github_actions.md
@@ -70,3 +70,43 @@ If you want to match versions covered by Black's
     src: "./src"
     version: "~= 22.0"
 ```
+
+## Outputs
+
+This action will output two variables for further processing of changes black has made
+against `src` set.
+
+### `is-formatted`
+
+Defaults to `"false"`, set to `"true"` if black changed any files.
+
+### `changed-files`
+
+Defaults to `"0"`, set to string representation of integer value of how many files black
+modified.
+
+### Usage
+
+One could use either of these output variables to further have conditional steps within
+the same pipeline, like creating a pull request after black has done changes to the code
+base.
+
+```yaml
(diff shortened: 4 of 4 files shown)


action.yml
```diff
@@ -27,6 +27,13 @@ inputs:
     description: 'Python Version specifier (PEP440) - e.g. "21.5b1"'
     required: false
     default: ""
+outputs:
+  is_formatted:
+    description: "Whether the files were formatted using the black formatter."
+    value: steps
+  changed_files:
+    description: "no. of files black changed"
```
cooperlees, line 35:
```suggestion
    description: "Number of files black changed."
```
Lets save the ambiguity for non native English speakers.

action.yml
```diff
@@ -27,6 +27,13 @@ inputs:
     description: 'Python Version specifier (PEP440) - e.g. "21.5b1"'
     required: false
     default: ""
+outputs:
+  is_formatted:
+    description: "Whether the files were formatted using the black formatter."
```
cooperlees, line 32:
```suggestion
    description: "Files checked are formatted to the current black formatter style."
```
Just trying to be clear - Happy for variations of suggestions.

action/main.py
```diff
@@ -38,8 +44,23 @@
 base_cmd = [str(ENV_BIN / "black")]
 if BLACK_ARGS:
     # TODO: remove after a while since this is deprecated in favour of SRC + OPTIONS.
-    proc = run([*base_cmd, *shlex.split(BLACK_ARGS)])
+    proc = run([*base_cmd, *shlex.split(BLACK_ARGS)], stdout=PIPE, stderr=STDOUT)
```
cooperlees, line 47:
```suggestion
    proc = run([*base_cmd, *shlex.split(BLACK_ARGS)], stderr=PIPE, encoding="utf8")
```
All the output you're after here goes to stderr. You can also set the encoding on the call.

docs/integrations/github_actions.md
```diff
@@ -70,3 +70,43 @@ If you want to match versions covered by Black's
...
+This action will output two variables for further processing of changes black has made
+against `src` set.
+
+### `is_formatted`
+
+Defaults to `"0"`, set to `"1"` if black changed any files.
+
+### `changed_files`
+
+Defaults to `"0"`, set to string representation of integer value of how maney files
```
cooperlees, line 85:
```suggestion
Defaults to `"0"`, set to string representation of integer value of how many files
```

action/main.py
```diff
@@ -38,8 +44,23 @@
...
 else:
-    proc = run([*base_cmd, *shlex.split(OPTIONS), *shlex.split(SRC)])
+    proc = run(
+        [*base_cmd, *shlex.split(OPTIONS), *shlex.split(SRC)],
+        stdout=PIPE,
+        stderr=STDOUT,
+    )
```
cooperlees, line 53:
```suggestion
    proc = run(
        [*base_cmd, *shlex.split(OPTIONS), *shlex.split(SRC)],
        stderr=PIPE,
        encoding="utf8",
    )
```

action/main.py
```diff
@@ -38,8 +44,23 @@
...
+
+_output = proc.stdout.decode("utf-8")
+
+matches = search(_is_formatted_re, _output, MULTILINE)
```
cooperlees, line 57:
```suggestion
matches = search(_is_formatted_re, proc.stderr, MULTILINE)
```

cooperlees (changes requested):
Thanks for the contribution. Looks good. I think there is a few things we can clean up here tho. Lets try these, if any of my suggestions don't work, I apologize, just say so.

CHANGES.md
```diff
@@ -84,6 +84,8 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- Github Actions: adds 2 variables as github action outputs to report changes that can
+  be used in other steps. (#3531)
```
ichard26, line 88:
```suggestion
- GitHub Action: adds 2 action outputs to report changes that can
  be used in other steps. (#3531)
```
Please also mention the output names in this changelog entry.

ichard26 (changes requested):
Thanks for submitting a PR but as it stands currently, this hurts UX in the general case. This captures the output of the Black invocation and never reemits it. That's already not great, but it gets worse. If Black errors out for some reason, the developer reading the logs will have no idea what's wrong since the logs will be almost entirely empty. There are two ways of fixing this:

- Reemit the captured output (always or on a non-zero return code; I'd prefer the former)  
- Tee or somehow duplicate the output stream so it can be captured _and_ printed to the console as usual
 
The second could preserve the colour output and allow for live feedback (unlike the first option), however, I have never done this before so no clue whether this is possible. ASNI codes make it sound really annoying.

I know this Action's output could be improved in general, but that's dependant on GitHub fixing their own infrastructure (which will never happen IME), but if it's possible let's not make it worse.

Also, the general convention for GitHub Actions are to use dashes instead of underscores.  [Here's an example.](https://github.com/marketplace/actions/close-stale-issues).

Finally, I'm not keen on using using numbers as boolean, we could use `true` / `false` instead. I can go either way on this point though.

I know this is a lot of words, but I do really mean it that I appreciate the effort you've put into this. I'm just one of those "let's right it the first time" folks (un)fortunately :)

CHANGES.md
rasjani: 0 problems, I'll make the requested changes ..

action/main.py
rasjani: Regarding what @ichard26 suggested below, if terminal session here is not utf-8 (which is common in windows?), and i'd re-emit the stderr back to console, wouldn't decoding to utf-8 and then re-encoding back to whatever sys.stdout.encoding was cause possible issues ? Not sure about that - asking for clarification

CHANGES.md
rasjani: Regarding re-emitting the stderr - I have a feeling that black uses some mechanism to check if its output is actual console or just a file buffer and it will strip the terminal escape sequences based on the target. This will happen with straight piping and also when piped thru  `tee`

action.yml
rasjani: Ill amend the patch with local changes

action.yml
rasjani: Ill amend the patch, changed made in new version..

CHANGES.md
rasjani: - is_formatted & changed_files renamed to use - instead of _
- outputs mentioned in changelog and docs updated to reflect the rename
- changed is-formatted to have str like bool.

action/main.py
```diff
@@ -38,8 +44,25 @@
...
 else:
-    proc = run([*base_cmd, *shlex.split(OPTIONS), *shlex.split(SRC)])
+    proc = run(
+        [*base_cmd, *shlex.split(OPTIONS), *shlex.split(SRC)],
+        stderr=PIPE
+    )
+# Re-emit stderr back to console so that action output is visible to pipeline
+# Do note, click will strip terminal control codes if the output is not TTY
+# and thus, this will not show colors anymore.
+print(proc.stderr, file=sys.stderr, flush=True)
```
rasjani, line 56: Im still wondering if i should proc.stderr should match sys.stdout.encoding ?

rasjani: and to give a bit of background, character encoding on github runners with windows as os, is reported to be `cp1252` while ubuntu and macos default to utf-8 .. enforcing utf-8 on subprocess.run() feels dirty if the output is going back to windows runner ..

cooperlees: I guess, using sys.stdout.encoding is prob the better thing to do 
- Is there a `sys.stderr.encoding`? - I just know all our output will always be utf8 compliant ...

cooperlees: Since I did miss this swallowing of the output, it's prob better to not encode @ subprocess (like you had it - sorry) and only decode for the regex and reprint the raw bytes. That might preserve the colors etc. too

cooperlees:
Another option would be to have black itself check if it's running in a GitHub action environment and if so dump a file or set environment variable(s) actions could use maybe?

rasjani:
> Another option would be to have black itself check if it's running in a GitHub action environment and if so dump a file or set environment variable(s) actions could use maybe?

Not against rewriting the patch for that if the approach is acceptable ?  Less issues to tackle for sure but personal take; approach doesn't feel like "keeping it simple".  

Anyhuu, I'll make another pr against the core black and if that approach is valid, this pr can be closed.

cooperlees:
> Anyhuu, I'll make another pr against the core black and if that approach is valid, this pr can be closed.

Lets give it a few hours for others to see if they like the approach better ... I don't want to eat your time ... I'm also for going this way of people are happy about it ... 

I just floated my idea as alarm bells always go off in my head when automation is using output designed for humans / wrapping tools if it can be avoided.

rasjani:
@cooperlees https://github.com/rasjani/black/commit/367beb839046471fd2a5c803ad916ede008e6295 shows alternative approach baked directly into black itself and not into action ..  

And don't worry about wasting my time - at the moment - im "in between" customers and doing foss contributions is preferred as positive way to spend time on the bench :)

cooperlees:
Your POC commit looks good to me. I feel lets do that and read from the dumped file (I think JSON is a better choice) leaving stderr alone :)

cobaltt7:

PR #3533, opened 2023-01-31
Fix unsafe cast in linegen.py

### LINKED ISSUE #3532: Preview crashes on await (yield)
**Describe the bug**
`black --preview` crashes on seeing `await (yield)`
This is certainly not good code or anything, but is code we have [in one of our test files](https://github.com/Zac-HD/flake8-trio/blob/ec04f75f6c58515f58316870272fab315b828a35/tests/eval_files/trio108.py#L31-L34) in flake8-trio. Non-preview black has no issues with it.

**To Reproduce**

For example, take this code:

```python
async def foo():
    await (yield)
```

```sh
$ black --preview myfile.py 
```

The resulting error is:

```
error: cannot format myfile.py: blib2to3.pytree.Node object expected; got blib2to3.pytree.Leaf

Oh no! 💥 💔 💥
1 file failed to reformat.
```

**Expected behavior**

Not crashing

**Environment**

```sh
$ black --version
black, 22.12.0 (compiled: yes)
Python (CPython) 3.11.0
$ python --version
Python 3.11.0
$ uname -sro
Linux 6.1.1-arch1-1 GNU/Linux
```

JelleZijlstra:
Fixes #3532

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,7 @@
 - Fix an invalid quote escaping bug in f-string expressions where it produced invalid
   code. Implicitly concatenated f-strings with different quotes can now be merged or
   quote-normalized by changing the quotes used in expressions. (#3509)
+- Fix crash on `await (yield)` when Black is compiled with mypyc (#3533)
 
 ### Configuration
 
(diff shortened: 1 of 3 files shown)


ichard26 (approved):
Thank you!

PR #3546, opened 2023-02-04
Fix this import, it's blib2to3.pgen2.driver, not blib2to3.driver.

yilei:
I'm not sure why mypy doesn't complain.

This was found while I was playing with Pytype on blib2to3.


ichard26 (approved):
Thanks!

JelleZijlstra:
Mypy doesn't catch it because we have `ignore_missing_imports = True` for blib2to3 in mypy.ini. It might be worth checking if we can remove some of that.

PR #3567, opened 2023-02-12
Update documentation regarding isort compatibility

### LINKED ISSUE #3565: Outdated documentation for isort compatibility
**Is this related to a problem? Please describe.**

I believe that the documentation for compatibility with isort may be outdated. The [Why those options above?](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#why-those-options-above) section states:

>_Black_ wraps imports that surpass `line-length` by moving identifiers into their own indented line. If that still doesn’t fit the bill, it will put all of them in separate lines and put a trailing comma.

In my experience (and supported by this [old conversation](https://github.com/PyCQA/isort/issues/694#issuecomment-473063700)), its seems that _Black_ no longer makes a first attempt to move identifiers to their own indented line, and it just defaults to newlines for each (with trailing commas). 

Assuming I'm not misunderstanding something, the documentation should probably be updated to reflect this chanage. I think the following wording might suffice:

> _Black_ wraps imports that surpass `line-length` by moving identifiers onto separate lines and by adding a trailing comma after each.

I'm happy to submit the MR with that change if this checks out. (PS: thanks for an awesome tool!)

mitchnegus:
### Description

This updates the documentation regarding isort compatibility, as described in #3565.

(Closes #3565)

### Checklist - did you ...

    <small> - I consider this a small docs change, so probably not worth an entry</small>

### DIFF
--- a/docs/guides/using_black_with_other_tools.md
+++ b/docs/guides/using_black_with_other_tools.md
@@ -51,9 +51,9 @@ line_length = 88
 
 #### Why those options above?
 
-_Black_ wraps imports that surpass `line-length` by moving identifiers into their own
-indented line. If that still doesn't fit the bill, it will put all of them in separate
-lines and put a trailing comma. A more detailed explanation of this behaviour can be
+_Black_ wraps imports that surpass `line-length` by moving identifiers onto separate
+lines and by adding a trailing comma after each. A more detailed explanation of this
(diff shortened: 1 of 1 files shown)


cooperlees (approved):
SGTM - Both make sense ...

PR #3578, opened 2023-02-22
Remove docker support

brodycritchlow:
### Description

This PR removes the `Dockerfile` as it is not required for the use case of this fork. The `Dockerfile` is used to define the environment and dependencies for a Docker container. However, as we do not plan to use Black in a Docker container, we can safely remove this file.

### Checklist

- [x] Add an entry in `CHANGES.md` to inform users about the removal of `Dockerfile`
- [ ] Add / update tests if necessary
- [ ] Add new / update outdated documentation

By removing the `Dockerfile`, we can slim down the repository and make it easier to maintain. This change does not affect the functionality of Black in any way.


brodycritchlow:
Woops

Issue #3585, opened 2023-02-27
[tests] deprecation of setuptools 67.3.0


stanislavlevin:
Tests fail against `setuptools` 67.3.0:
```
==================================== ERRORS ====================================
____________________ ERROR collecting tests/test_blackd.py _____________________
tests/test_blackd.py:11: in <module>
    from aiohttp import web
/usr/lib64/python3/site-packages/aiohttp/__init__.py:212: in <module>
    from .worker import GunicornUVLoopWebWorker, GunicornWebWorker
/usr/lib64/python3/site-packages/aiohttp/worker.py:11: in <module>
    from gunicorn.config import AccessLogFormat as GunicornAccessLogFormat
/usr/lib/python3/site-packages/gunicorn/config.py:20: in <module>
    from gunicorn import __version__, util
/usr/lib/python3/site-packages/gunicorn/util.py:25: in <module>
    import pkg_resources
/usr/lib64/python3/site-packages/pkg_resources/__init__.py:3258: in <module>
    def _initialize_master_working_set():
/usr/lib64/python3/site-packages/pkg_resources/__init__.py:3232: in _call_aside
    f(*args, **kwargs)
/usr/lib64/python3/site-packages/pkg_resources/__init__.py:3283: in _initialize_master_working_set
    tuple(dist.activate(replace=False) for dist in working_set)
/usr/lib64/python3/site-packages/pkg_resources/__init__.py:3283: in <genexpr>
    tuple(dist.activate(replace=False) for dist in working_set)
/usr/lib64/python3/site-packages/pkg_resources/__init__.py:2804: in activate
    declare_namespace(pkg)
/usr/lib64/python3/site-packages/pkg_resources/__init__.py:2298: in declare_namespace
    warnings.warn(msg, DeprecationWarning, stacklevel=2)
E   DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('zope')`.
E   Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
```

https://setuptools.pypa.io/en/stable/history.html#v67-3-0
> [#3434](https://github.com/pypa/setuptools/issues/3434): Added deprecation warning for pkg_resources.declare_namespace. Users that wish to implement namespace packages, are recommended to follow the practice described in [PEP 420](https://peps.python.org/pep-0420/) and omit the __init__.py file entirely.

[...]

PR #3586, opened 2023-02-27
tests: Ignore pkg_resources.declare_namespace warnings

stanislavlevin:
### Description
https://setuptools.pypa.io/en/stable/history.html#v67-3-0
> Added deprecation warning for pkg_resources.declare_namespace.
  Users that wish to implement namespace packages, are recommended
  to follow the practice described in PEP 420 and omit the
  __init__.py file entirely.

However, projects (for example, `zope`) may use `pkg_resources`'s namespace. This warning can be ignored.

Fixes: https://github.com/psf/black/issues/3585


JelleZijlstra:
Thanks, but I don't feel that this is the right place to fix the problem (see my comment on the issue).

PR #3594, opened 2023-03-04
Support files with type comment syntax errors

### LINKED ISSUE #3593: Let black run on files even if `ast.parse` with `type_comments=True` fails

Right now, if a file parsing because of invalid type comments (even if the code runs just fine), black won't format the file.

For example:

```python
import typing

def foo(
    # type: typing.List
    bar,
):
    print(bar)

foo(10)
```

While the code still runs:

```console
$ python3 asd.py            
10
```

Black no longer formats it:

```console
$ black asd.py        
error: cannot format asd.py: cannot use --safe with this file; failed to parse source file AST: invalid syntax (<unknown>, line 4)
This could be caused by running Black with an older Python version that does not support new syntax used in your source file.

Oh no! 💥 💔 💥
1 file failed to reformat.
```

This is because `ast.parse` on that code fails if you pass `type_comments=True`:

```pycon
>>> import ast
>>> ast.parse(open('asd.py').read(), type_comments=True)
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 4
    # type: typing.List
            ^^^^^^^^^^^
SyntaxError: invalid syntax
```

(Black does work if you do `black asd.py --fast`, as it no longer does the AST based check.)

If `ast.parse(..., type_comments=True)` fails for the before/after check, maybe doing it with `type_comments=False` is fine, just for that file?

None so far.

**Additional context**

I don't think this can be considered an upstream bug, as this is indeed invalid syntax for a type checker's perspective. But black shouldn't care if the types are correct, in any way shape or form.

tusharsadhwani:
### Description

Resolves #3593 

### Checklist - did you ...

### DIFF
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -148,7 +148,7 @@ def lib2to3_unparse(node: Node) -> str:
 
 
 def parse_single_version(
-    src: str, version: Tuple[int, int], type_comments: bool
+    src: str, version: Tuple[int, int], *, type_comments: bool
 ) -> Union[ast.AST, ast3.AST]:
     filename = "<unknown>"
     # typed-ast is needed because of feature version limitations in the builtin ast 3.8>
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -185,7 +185,6 @@ def parse_ast(src: str) -> Union[ast.AST, ast3.AST]:
             if not first_error:
                 first_error = str(e)
 
-
     # Try to parse without type comments
     for version in sorted(versions, reverse=True):
         try:
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -185,11 +185,13 @@ def parse_ast(src: str) -> Union[ast.AST, ast3.AST]:
             if not first_error:
                 first_error = str(e)
 
-            # Try to parse without type comments
-            try:
-                return parse_single_version(src, version, type_comments=False)
-            except SyntaxError:
-                pass
+
+    # Try to parse without type comments
+    for version in sorted(versions, reverse=True):
+        try:
+            return parse_single_version(src, version, type_comments=False)
+        except SyntaxError:
+            pass
 
     raise SyntaxError(first_error)
 
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -165,9 +165,12 @@ def parse_single_version(
         else:
             return ast3.parse(src, filename)
     else:
-        # Typed-ast is guaranteed to be used here and automatically tracks type
-        # comments separately.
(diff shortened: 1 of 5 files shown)


tests/test_black.py
```diff
@@ -1938,6 +1938,14 @@ def test_equivalency_ast_parse_failure_includes_error(self) -> None:
         err.match("invalid character")
         err.match(r"\(<unknown>, line 1\)")
 
+    def test_type_comment_syntax_error(self) -> None:
```
JelleZijlstra, line 1941: Would prefer a test in test_format.py where we don't have to write a new test function.

src/black/parsing.py
```diff
@@ -175,11 +180,17 @@ def parse_ast(src: str) -> Union[ast.AST, ast3.AST]:
     first_error = ""
     for version in sorted(versions, reverse=True):
         try:
-            return parse_single_version(src, version)
+            return parse_single_version(src, version, type_comments=True)
         except SyntaxError as e:
             if not first_error:
                 first_error = str(e)
 
+            # Try to parse without type comments
```
JelleZijlstra, line 188: Need to think more about whether we should have a separate loop instead. I *think* that should never affect behavior, just performance.

src/black/parsing.py
```diff
@@ -148,24 +148,29 @@ def lib2to3_unparse(node: Node) -> str:
 
 
 def parse_single_version(
-    src: str, version: Tuple[int, int]
+    src: str, version: Tuple[int, int], type_comments: bool
```
JelleZijlstra, line 151:
```suggestion
    src: str, version: Tuple[int, int], *, type_comments: bool
```
Let's make it keyword-only

src/black/parsing.py
tusharsadhwani: I'll move it to an outer for loop anyway. Just for guaranteed backwards compatibility.

JelleZijlstra (approved):

tusharsadhwani:
@ichard26 can you help with re-triggering the workflows?

JelleZijlstra:
@ichard26 do you still want to look at this or can I merge it?

ichard26 (approved):
I wanted to take a look because this PR falls back to using the built-in ast module even on Python 3.7 which we avoided doing previously, but that's a really minor edge case — so it's fine. All it might do is make some files with invalid type comments still unparsable.

ichard26:
Thanks for the PR @tusharsadhwani!

PR #3632, opened 2023-03-30
chore: move configurations to `pyproject.toml`

SauravMaheshkar:
### Description

This PR aims to minimize the number of configuration files for the project. Instead of a separate `.coveragerc` and `mypy.ini`, we now would have all the configurations within the `pyproject.toml` file leading to a overall minimal code structure.

### Checklist

- [x] Add an entry in `CHANGES.md` if necessary? (`skip news` ??)


pyproject.toml
```diff
@@ -215,3 +215,51 @@ filterwarnings = [
...
+module="tokenize_rt.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module="uvloop.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module="_black_version.*"
+ignore_missing_imports = true
```
Secrus, line 265: you can put all those under one subtable with modules as list. see: [example](https://github.com/python-poetry/poetry/blob/c8e963e668af2f71ddce2053e575ae0427b363e6/pyproject.toml#L179)

SauravMaheshkar: @Secrus thanks for the feedback. Addressed in 52dbc57f542e6aff419c4c22143fdbdfd040eaa8

SauravMaheshkar:
@JelleZijlstra Request for Review

cooperlees (changes requested):
Seems we need:
- toml support for coverage
- mypy is very unhappy now showing lots of type errors
  - Haven't looked - but are these valid or problem with getting it's settings from the toml?

PR #3640, opened 2023-04-07
Do not wrap implicitly concatenated strings used as func args in parens.

yilei:
### Description

This reverts the change in #3307, as the changes that would have been introduced in the 2024 style are too destructive based on feedback. See https://github.com/psf/black/issues/2188#issuecomment-1489053232.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Implicitly concatenated strings used as function args are no longer wrapped inside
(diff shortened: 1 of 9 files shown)


yilei:
@JelleZijlstra any opinion on this? I think the preview string processing overall is a very big change, so I think reverting #3307 would be beneficial to get it in the 2024 stable style.

JelleZijlstra (approved):
OK, let's change this back for now.

PR #3648, opened 2023-04-14
Fix two more mypyc issues with mypyc v1.2.0.

yilei:
1. `math.inf` is a `float`, but `bracket_depth` is `int`. Alternatively we could use a really large int for maximum.
1. The `Final` in `nodes.py` is the same as #3645

Btw, any instructions how to build the mypyc wheels locally in the same way as CI? I only manually compiled the files excluding the ones excluded in pyproject.toml.

### DIFF
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -790,7 +790,7 @@ def is_line_short_enough(  # noqa: C901
     # store the leaves that contain parts of the MLS
     multiline_string_contexts: List[LN] = []
 
-    max_level_to_update = math.inf  # track the depth of the MLS
+    max_level_to_update: Union[int, float] = math.inf  # track the depth of the MLS
     for i, leaf in enumerate(line.leaves):
         if max_level_to_update == math.inf:
             had_comma: Optional[int] = None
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Not sure how to do a mypyc build like in CI, sorry. But it's definitely good to move towards newer versions of mypyc; will merge once CI is clean.

For the `math.inf` change, I wonder if we'd get better performance if we stuck to a single type, so that mypyc could unbox the value.

yilei:
With this PR, I was able to build mypyc 1.2.0 compiled linux wheels.

taleinat:
For future reference, the steps to locally build with mypyc via cibuildwheel were based on the following sequence I'd used a month or two ago (on Ubuntu 20.04):

0. Have Docker installed, the Docker daemon running, and permission to run the docker cli.
1. Create a venv and run `pip install cibuildwheel`
2. `cibuildwheel --platform linux --help`
3. `cibuildwheel --platform linux --print-build-identifiers`
4. `cibuildwheel --platform linux --only cp310-manylinux_x86_64`
5. `pip install ./wheelhouse/black-22.12.1.dev37+g18fb884-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`

Issue #3641, opened 2023-04-09
Documenting black-jupyter pre-commit hook


jamesbraza:
**Is this related to a problem? Please describe.**

The [Version control integration](https://black.readthedocs.io/en/stable/integrations/source_version_control.html#version-control-integration) briefly mentions `black-jupyter`:

> The `black-jupyter` hook is only available from version 21.8b0 and onwards.

Also, searching the docs for it, barely anything comes up.  To figure out how to use, I came upon this: https://github.com/psf/black/blob/23.3.0/.pre-commit-hooks.yaml

And some results ([here](https://github.com/shanku007/pro-orch/blob/master/.pre-commit-config.yaml), [here](https://github.com/rocs-org/datascience-template/blob/main/.pre-commit-config.yaml)) from a GitHub advanced search.

**Describe the solution you'd like**

It would be useful for `black` to document this integration officially, and maybe address a few questions:
- Why does `black-jupyter` require `additional_dependencies: [".[jupyter]"]`?
    - Shouldn't `black-jupyter` self-install `jupyter`?
- Should it include `types_or: [python, pyi, jupyter]`, `entry`?
- Why is `require_serial: true` not in the suggested the previously documented Black config?

```yaml
---
default_language_version:
  python: python3

repos:
  - repo: https://github.com/psf/black
    rev: 23.3.0
    hooks:
      - id: black  # Does one still need black here?
      - id: black-jupyter
        entry: black
        minimum_pre_commit_version: 2.9.2
        require_serial: true
        types_or: [python, pyi, jupyter]
        additional_dependencies: [".[jupyter]"]
```

**Describe alternatives you've considered**

N/a

**Additional context**

N/a

MarcoGorelli:
Hey @jamesbraza 

> Why does black-jupyter require additional_dependencies: [".[jupyter]"]? Shouldn't black-jupyter self-install jupyter?

You're right, it does - you don't need to manually specify `additional_dependencies`

> Should it include types_or: [python, pyi, jupyter], entry?

The hook already contains this, you don't need to specify it

> Why is require_serial: true not in the suggested the previously documented Black config?

It's already in the hook, you don't need to specify it yourself:

https://github.com/psf/black/blob/a552f7096a9f6e016c9bb1df1e0a77a17caeec1c/.pre-commit-hooks.yaml#L16

>       - id: black  # Does one still need black here?

[...]

PR #3650, opened 2023-04-17
Documented black-jupyter hook

jamesbraza:
### Description

Closes https://github.com/psf/black/issues/3641.

### Checklist - did you ...

I locally built the docs and saw it looked fine.

### DIFF
--- a/docs/integrations/source_version_control.md
+++ b/docs/integrations/source_version_control.md
@@ -24,8 +24,9 @@ expect][pre-commit-mutable-rev].
 
 ## Jupyter Notebooks
 
-If you want to include support for Jupyter Notebooks, then simply replace the hook's
-`id: black` with `id: black-jupyter` in the `.pre-commit-config.yaml`:
+There is an alternate hook `black-jupyter` that expands the targets of `black` to
+include Jupyter Notebooks. To use this hook, simply replace the hook's `id: black` with
+`id: black-jupyter` in the `.pre-commit-config.yaml`:
 
 ```yaml
 repos:
@@ -33,7 +34,6 @@ repos:
     rev: 23.3.0
     hooks:
       - id: black-jupyter
-        description: Black, with Jupyter Notebook support
         # It is recommended to specify the latest version of Python
         # supported by your project here, or alternatively use
         # pre-commit's default_language_version, see
--- a/docs/integrations/source_version_control.md
+++ b/docs/integrations/source_version_control.md
@@ -42,7 +42,7 @@ repos:
         # supported by your project here, or alternatively use
         # pre-commit's default_language_version, see
         # https://pre-commit.com/#top_level-default_language_version
-        language_version: python3.9
+        language_version: python3.11
 ```
 
 [black-tags]: https://github.com/psf/black/tags
--- a/docs/integrations/source_version_control.md
+++ b/docs/integrations/source_version_control.md
@@ -22,11 +22,28 @@ Feel free to switch out the `rev` value to something else, like another
 branches or other mutable refs since the hook [won't auto update as you may
 expect][pre-commit-mutable-rev].
 
-If you want support for Jupyter Notebooks as well, then replace `id: black` with
-`id: black-jupyter`.
+## Jupyter Notebooks
+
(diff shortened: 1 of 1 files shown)


jamesbraza:
Cc @MarcoGorelli

jamesbraza:
Please add the skip news label to this PR, as I think it's not worth being in the news

docs/integrations/source_version_control.md
```diff
@@ -22,11 +22,27 @@ Feel free to switch out the `rev` value to something else, like another
...
-If you want support for Jupyter Notebooks as well, then replace `id: black` with
-`id: black-jupyter`.
+## Jupyter Notebooks
+
+If you want to include support for Jupyter Notebooks, then simply replace the hook's
+`id: black` with `id: black-jupyter` in the `.pre-commit-config.yaml`:
 
 ```{note}
-The `black-jupyter` hook is only available from version 21.8b0 and onwards.
+The `black-jupyter` hook became available in version 21.8b0.
```
JelleZijlstra, line 31: Let's put this note below the yaml sample.

docs/integrations/source_version_control.md
```diff
@@ -22,11 +22,27 @@ Feel free to switch out the `rev` value to something else, like another
...
+  - repo: https://github.com/psf/black
+    rev: 23.3.0
+    hooks:
+      - id: black-jupyter
+        description: Black, with Jupyter Notebook support
+        # It is recommended to specify the latest version of Python
+        # supported by your project here, or alternatively use
+        # pre-commit's default_language_version, see
+        # https://pre-commit.com/#top_level-default_language_version
+        language_version: python3.9
```
JelleZijlstra, line 45:
```suggestion
        language_version: python3.11
```
Not necessary to keep this up to date with every Python version, but better to set a good example and use the latest version.

jamesbraza: Yeah sure, thanks for the suggestion!

docs/integrations/source_version_control.md
jamesbraza: Yeah I was thinking about this too.  Moved it down 👍

JelleZijlstra (approved):
Looks good, thanks! @MarcoGorelli any feedback?

docs/integrations/source_version_control.md
```diff
@@ -14,19 +14,35 @@ repos:
...
+If you want to include support for Jupyter Notebooks, then simply replace the hook's
+`id: black` with `id: black-jupyter` in the `.pre-commit-config.yaml`:
+
+```yaml
+repos:
+  - repo: https://github.com/psf/black
+    rev: 23.3.0
+    hooks:
+      - id: black-jupyter
+        description: Black, with Jupyter Notebook support
```
ichard26, line 36: [The `description` key is only used in `.pre-commit-hooks.yaml`, not `.pre-commit-config.yaml`](https://pre-commit.com/#pre-commit-configyaml---hooks). In other words, it goes on the provider side, not the consumer side.

jamesbraza: Good point, I removed the `description` field.

One question I had in my head when I made the original issue was, does `black-jupyter` run _only_ on Jupyter Notebooks?  And if yes, should I still keep the `id: black` line in my `.pre-commit-config.yaml`?  The answer is no, `black-jupyter` runs on both normal Python files and Jupyter Notebooks.

I thought having the `description` entry would answer this question within the `pre-commit` file, instead of requiring devs to pull up the `black` docs to search.

I think once one already knows `black-jupyter` runs on normal Python files, this `description` field isn't necessary.  It's only useful when one doesn't have this knowledge.  Thus, I think you're right let's remove the `description`, and hopefully new users find it intuitive (and can read the updated docs otherwise).

---

_Aside_

I would most like to see the below, like one does for `flake8` configs.  I think stating `jupyter` under `additional_dependencies` is intuitive.  However, it doesn't quite work as it's missing `types_or` for `jupyter` file type:

```yaml
repos:
    rev: 23.3.0
    hooks:
      - id: black
        additional_dependencies:
          - jupyter  # Also run on Jupyter Notebooks
```

PR #3661, opened 2023-04-26
Fix "# pyright: ignore" comment mobility

### LINKED ISSUE #3946: Black breaks pyright compatibility
Hi,
I've switched to VS code and pyright recently, and I realized black is breaking a lot of `# pyright: ignore` comments.

Input (both are valid for pyright):
```python
asm_client: SecretsManagerClient = boto3.client("secretsmanager") # pyright: ignore[reportUnknownMemberType]
asm_client: SecretsManagerClient = boto3.client( # pyright: ignore[reportUnknownMemberType]
    "secretsmanager"
)  
```

Output:
```python
asm_client: SecretsManagerClient = boto3.client(
    "secretsmanager"
)  # pyright: ignore[reportUnknownMemberType]
asm_client: SecretsManagerClient = boto3.client(
    "secretsmanager"
)  # pyright: ignore[reportUnknownMemberType]
```

Is there a way to stick the pyright comment to the original line ?

I'm using:
```
black, 23.9.1 (compiled: yes)
Python (CPython) 3.11.6
```

Thanks

etripier:
### Description

Closes #3946

Hi! It looks like you have some special handling for `# type: ignore` comments to avoid moving said comments around due to positional meaning. I think this same special handling should also apply to [`# pyright: ignore` comments](https://microsoft.github.io/pyright/#/comments?id=line-level-diagnostic-suppression).

The use case for `# pyright: ignore` comments is being able to ignore Pyright-specific diagnostic rules that might not make sense in the context of more generic `# type: ignore` comments.

### Checklist - did you ...

### DIFF
--- a/src/black/nodes.py
+++ b/src/black/nodes.py
@@ -822,7 +822,7 @@ def is_type_comment(leaf: Leaf, suffix: str = "") -> bool:
     t = leaf.type
     v = leaf.value
     return t in {token.COMMENT, STANDALONE_COMMENT} and (
-        v.startswith(f"# pyright:{suffix}") or v.startswith(f"# type:{suffix}")
+        v.startswith((f"# pyright:{suffix}", f"# type:{suffix}"))
     )
 
 
--- a/src/black/nodes.py
+++ b/src/black/nodes.py
@@ -821,7 +821,9 @@ def is_type_comment(leaf: Leaf, suffix: str = "") -> bool:
     Only returns true for type comments for now."""
     t = leaf.type
     v = leaf.value
-    return t in {token.COMMENT, STANDALONE_COMMENT} and v.startswith("# type:" + suffix)
+    return t in {token.COMMENT, STANDALONE_COMMENT} and (
+        v.startswith(f"# pyright:{suffix}") or v.startswith(f"# type:{suffix}")
+    )
 
 
 def wrap_in_parentheses(parent: Node, child: LN, *, visible: bool = True) -> None:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Black now avoids breaking up lines containing a comment beginning with "# pyright:
+  ignore". (#3661)
+
 ### Configuration
 
 <!-- Changes to how Black can be configured -->
--- a/tests/data/preview/comments7.py
+++ b/tests/data/preview/comments7.py
@@ -187,6 +187,7 @@ def func():
         0.0789,
         a[-1],  # type: ignore
     )
+    c = call(0.0123, 0.0456, 0.0789, 0.0123, 0.0789, a[-1])  # pyright: ignore[reportGeneralTypeIssues]
     c = call(0.0123, 0.0456, 0.0789, 0.0123, 0.0789, a[-1])  # type: ignore
     c = call(
         0.0123,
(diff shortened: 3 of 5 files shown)


tests/data/miscellaneous/long_strings_flag_disabled.py
```diff
@@ -276,6 +276,11 @@ def foo():
 
 x = f"This is a {{really}} long string that needs to be split without a doubt (i.e. most definitely). In short, this {string} that can't possibly be {{expected}} to fit all together on one line. In {fact} it may even take up three or more lines... like four or five... but probably just four."
 
+long_unmergable_string_with_pragma = (
```
etripier, line 279: Minor note: `unmergeable`

JelleZijlstra:
Should we add this to `black.comments.contains_pragma_comment` instead? I think these pyright comments are more like `noqa` than like `# type:`, which is treated specially by the Python parser.

etripier:
Ah, my understanding is that the pragma stuff only applies to string concatenation and splitting. I'm really hoping to fix the annoyance of having

```py
blah_blah_blah = [ # pyright: ignore
    ...
]
```

shift into

```py
blah_blah_blah = [
    ...
] # pyright: ignore
```

The behavior in https://github.com/psf/black/blob/de65741b8d49d78fa2675ef79b799cd35e92e7c1/src/black/lines.py#L263 is what I'd love to have, ideally.

src/black/nodes.py
```diff
@@ -821,7 +821,9 @@ def is_type_comment(leaf: Leaf, suffix: str = "") -> bool:
     Only returns true for type comments for now."""
     t = leaf.type
     v = leaf.value
-    return t in {token.COMMENT, STANDALONE_COMMENT} and v.startswith("# type:" + suffix)
+    return t in {token.COMMENT, STANDALONE_COMMENT} and (
+        v.startswith(f"# pyright:{suffix}") or v.startswith(f"# type:{suffix}")
```
Gobot1234, line 825:
```suggestion
        v.startswith((f"# pyright:{suffix}", f"# type:{suffix}"))
```
Is marginally more performant probably

roshanjrajan-zip:
Hi! I wanted to see if there any concerns about gettin this merged other than fixing the conflicts?

If it isn't clear we get scenarios  where the code gets transformed from this
```py
some_long_variable_name = some_long_dict_name[0] # pyright: ignore[reportGeneralTypeIssues]
```
to this
```py
some_long_variable_name = some_long_dict_name[
	0
] # pyright: ignore[reportGeneralTypeIssues]
```

where after the transformation, the ignore is no longer being applied to the actual line causing pyright to error. This looks similiar to the `# type: ignore` fix and this would helpful to have! Let me know if you have any other questions!

roshanjrajan-zip:
@JelleZijlstra In this case, is it better to do change it in the `contains_pragma_comment` to get the intended behavior or does it make sense to keep it in `is_type_comment`?

JelleZijlstra:
I need to look more into this to figure out exactly how we use these, but a few thoughts:

- I think we should treat the pyright comments more like the pylint comments than like the `# type: ignore` comments. `# type:` comments are extra special because they are treated specially by the Python AST, so it makes sense to be even more conservative with them. 
- I'd prefer to find a solution that avoids Black having too much knowledge of magic comments used by specific tools, since that's bad for long-term maintainability.

cobaltt7:

PR #3674, opened 2023-05-03
Fix new mypy error in blib2to3

JelleZijlstra:
See python/mypy#15174

### Checklist - did you ...

### DIFF
--- a/src/blib2to3/pgen2/tokenize.py
+++ b/src/blib2to3/pgen2/tokenize.py
@@ -599,7 +599,7 @@ def generate_tokens(
                     if token[-1] == "\n":  # continued string
                         strstart = (lnum, start)
                         maybe_endprog = (
-                            endprogs[initial]
+                            endprogs.get(initial)
                             or endprogs.get(token[1])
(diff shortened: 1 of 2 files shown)


ichard26 (approved):

JelleZijlstra:
Seems like there's a new crash on pandas code. Trying to reproduce it locally but no luck so far.

JelleZijlstra:
Got it, reduced repro case:

```python
print(
    f"Tried directories {str(rootdirs)} \
    but none started with prefix {parentdir_prefix}"
)
```

JelleZijlstra:
I'm trying to understand now why mypy was fine with the old code. Ivan on the mypy PR said it was some special-casing for `dict.update`, but I can't find where that is, and on the playground I get an error as expected: https://mypy-play.net/?mypy=latest&python=3.7&flags=strict&gist=9acbef1e99c9355bb6ff87f5d318171b

JelleZijlstra:
Seems like it's because the `_combinations` function is unannotated.

PR #3686, opened 2023-05-12
add option: allow_block_newline

### LINKED ISSUE #3551: Allow a single newline after opening blocks, possibly via an option setting

Previously black allowed to have a newline after a code opening black. This can indeed improve readability.

Compare for instance this

image

with

image

There are two reasons, why a single newline at the beginning of a block can improve the readability:

- **Emphasis / separation**: If the following line is complex/long, giving it some space on top makes it stand out more, and easier to grasp visually. In the example, the first `some_long_function_and_condition(a, b, c) == expected` feels overshadowed by the `def some_long_function_name(with_many: List[int], params: Optional[T]):`
- **Symmetry reasons / consistent margins**: If the function contains multiple blocks which have a certain semantic symmetry, it makes sense to mirror that visually. In the example, all three blocks are identical, so why can't they have identical margins? Using consistent margins is a key feature to proper graphical design, and collapsing this one margin is something that typically wouldn't be approved by an experienced designer.

From a practical perspective: Going from black version 22 to 23 in our code base is causing a massive diff! The vast majority of the diff (perhaps ~95%) is coming from this very change, which isn't even listed in the [release notes of version 23](https://black.readthedocs.io/en/stable/change_log.html#id1). Basically all the changes would make the code more ugly due to either of the two points above.

From a design perspective, https://github.com/psf/black/pull/3035 was probably a mistake. Since it is probably too late to revert that change, I'd hope that this is one of the few cases where we can introduce an option that allows the old behavior. Something like `allow_newline_at_beginning_of_blocks`.

Perhaps even reverting https://github.com/psf/black/pull/3035 ?

**Additional context**

Anything else needed?

mathause:
### Checklist - did you ...

- [x] Potentially closes #3551

This is an attempt to close #3551 via an option. This is certainly a WIP but I'd like to get feedback before continuing.

1. Is the approach via an option what you want to go for? Or do you consider reverting the style change for 2024 (as @JelleZijlstra hinted in https://github.com/psf/black/issues/3551#issuecomment-1496192763)?
2. What about the name of the option? I went for `allow_block_newline` but am of course fine to change. (For similar options you rather went with `skip_*`, so `skip_removing_block_newline`?)

(Another possibility would be to do `block_newline = Literal["remove", "allow", "enforce"]`, just to throw that into the discussion.)

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -269,6 +269,11 @@ def validate_regex(
     help="Don't use trailing commas as a reason to split lines.",
 )
 @click.option(
+    "--allow-block-newline",
+    is_flag=True,
+    help="Allow a single opening newline after a block opens.",
+)
+@click.option(
     "--experimental-string-processing",
     is_flag=True,
     hidden=True,
@@ -439,6 +444,7 @@ def main(  # noqa: C901
     skip_source_first_line: bool,
     skip_string_normalization: bool,
     skip_magic_trailing_comma: bool,
+    allow_block_newline: bool,
     experimental_string_processing: bool,
     preview: bool,
     quiet: bool,
@@ -547,6 +553,7 @@ def main(  # noqa: C901
         skip_source_first_line=skip_source_first_line,
         string_normalization=not skip_string_normalization,
         magic_trailing_comma=not skip_magic_trailing_comma,
+        allow_block_newline=allow_block_newline,
         experimental_string_processing=experimental_string_processing,
         preview=preview,
         python_cell_magics=set(python_cell_magics),
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -636,7 +636,11 @@ class EmptyLineTracker:
         ):
             return before, 1
 
-        if self.previous_line and self.previous_line.opens_block:
+        if (
+            not self.mode.allow_block_newline
+            and self.previous_line
+            and self.previous_line.opens_block
+        ):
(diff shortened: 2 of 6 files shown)


JelleZijlstra:
I don't want a new option for this. We minimize the use of options so that Black's style is consistent. Whatever we end up with, it should be the same for all users.

return42:
> I don't want a new option for this.

@JelleZijlstra  OK / I understand ... but ...

> We minimize the use of options so that Black's style is consistent. Whatever we end up with, it should be the same for all users.

when [pin Black](https://github.com/psf/black/issues/3551#issuecomment-1496270142) is not a alternative and an option is not wanted .. what alternatives are you suggest?

JelleZijlstra:
> when [pin Black](https://github.com/psf/black/issues/3551#issuecomment-1496270142) is not a alternative and an option is not wanted .. what alternatives are you suggest?

Either you pin Black, and always get the same formatting, or you don't pin it, and you sometimes get changes (but only when a new year comes around or you use `--preview`). You can't have it both ways.

mathause:
> Either you pin Black, and always get the same formatting, or you don't pin it, and you sometimes get changes (but only when a new year comes around or you use `--preview`). You can't have it both ways.

I don't think the problem is that applying black leads to changes but with this _specific_ change that makes the code less readable (at least that is my issue).

PR #3691, opened 2023-05-17
sort DEFAULT_EXCLUDES and add .vscode, .pytest_cache and .ruff_cache

raybellwaves:
### Description

I would like to expand the excluded files/folder given in the default value (DEFAULT_EXCLUDES).

I would like to exclude the folders of .vscode, .pytest_cache and .ruff_cache. I tend to add these in the exclude arg of `[tool.black]` in my project.toml files. I can't see others wanting to run black on these unless i'm mistaken. I'm therefore wanted to upstream my excludes and save potential work for others who add this to configurations excludes.

To make it easier to add files/folders in the future I sorted the the inputs of DEFAULT_EXCLUDES.

### Checklist - did you ...

- [ ] Add / update tests if necessary? - NA
- [ ] Add new / update outdated documentation? - NA


src/black/const.py
```diff
@@ -1,4 +1,4 @@
 DEFAULT_LINE_LENGTH = 88
-DEFAULT_EXCLUDES = r"/(\.direnv|\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|venv|\.svn|\.ipynb_checkpoints|_build|buck-out|build|dist|__pypackages__)/"  # noqa: B950
+DEFAULT_EXCLUDES = r"/(\.direnv|\.eggs|\.git|\.hg|\.ipynb_checkpoints|\.mypy_cache|\.nox|\.pytest_cache|\.ruff_cache|\.tox|\.svn|\.venv|\.vscode|__pypackages__|_build|buck-out|_build|dist|venv)/"  # noqa: B950
```
JelleZijlstra, line 2: `_build` is duplicated

raybellwaves: thanks for the spot

JelleZijlstra (approved):

PR #3696, opened 2023-05-19
Avoid EncodingWarning. Fixes #3695.

### LINKED ISSUE #3695: EncodingWarning in pgen
Relating to #3658, if PYTHONWARNDEFAULTENCODING is set, this library raises a warning at:

https://github.com/psf/black/blob/2fd9d8b339e1e2e1b93956c6d68b2b358b3fc29d/src/blib2to3/pgen2/pgen.py#L42

jaraco:
Fixes #3695

### DIFF
--- a/src/blib2to3/pgen2/pgen.py
+++ b/src/blib2to3/pgen2/pgen.py
@@ -30,7 +30,6 @@ class PgenGrammar(grammar.Grammar):
 
 
 class ParserGenerator(object):
-
     filename: Path
     stream: IO[Text]
     generator: Iterator[GoodTokenInfo]
@@ -39,7 +38,7 @@ class ParserGenerator(object):
     def __init__(self, filename: Path, stream: Optional[IO[Text]] = None) -> None:
         close_stream = None
         if stream is None:
-            stream = open(filename)
+            stream = open(filename, encoding="utf-8")
             close_stream = stream.close
         self.filename = filename
         self.stream = stream


cooperlees (approved):
Makes sense to me, as we assume source we parse is UTF8 compatible ...

JelleZijlstra:
I don't think we can assume Python source is UTF-8. However, seems like this is parsing the grammar files for blib2to3, which should be in UTF-8. I want to check the code some more, though.

jaraco:
According to [PEP 597 (motivation)](https://peps.python.org/pep-0597/#using-the-default-encoding-is-a-common-mistake), UTF-8 is the default encoding on non-Windows systems, so by specifying it, you provide clarity and consistent behavior across platforms. You could specify `encoding="locale"`, but I'd only recommend that if compatibility is essential and the variance is desirable. In my experience, it's almost always the case that using UTF-8 as the specified encoding provides the best experience and rarely leads to surprise. You should ask yourselves, do you expect that users have specifically put non-ascii non-UTF-8 characters in the grammar files and expect those to work only in that locale. If not, you probably want UTF-8. If so, you may want to migrate users away from that assumption.

JelleZijlstra:
Users shouldn't put anything in the grammar files, we control all the grammar files and we don't support users providing their own. My hesitation is only about verifying that this codepath is purely about opening the internal grammar files, not user code.

JelleZijlstra:
Thanks @jaraco!

PR #3697, opened 2023-05-19
Build with mypyc 1.3

stianjensen:
### Description

Several new versions of mypyc has been released since the last upgrade, and they include some performance improvements which could make the compiled version of Black run faster.

https://mypy-lang.org/news.html

The latest version of hatch-mypyc allows being installed next the 1.x series of mypy.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,8 @@
 
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
+- Upgrade mypyc from 0.991 to 1.3 (#3697)
+
 ### Parser
 
 <!-- Changes to the parser or to version autodetection -->
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -119,8 +119,8 @@ sources = ["src"]
 [tool.hatch.build.targets.wheel.hooks.mypyc]
 enable-by-default = false
 dependencies = [
-  "hatch-mypyc>=0.13.0",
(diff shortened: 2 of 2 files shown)


JelleZijlstra:
Thanks! I think we should jump straight to 1.3 if possible.

stianjensen:
Let's try 1.3, then!

I'm not sure if any part of the mypyc release flow is really possible to test locally – from my understanding, that is only setup on CI?

JelleZijlstra:
I believe it's testable by running `cibuildwheel` directly. This was discussed on our Discord (https://discord.com/channels/267624335836053506/846434317021741086/1096514507486023700).

stianjensen:
> I believe it's testable by running cibuildwheel directly.

Great! Managed to follow those instructions and build at least a couple Linux wheels now and then test them inside a docker python image. I don't really have a rig for perf testing (only an old macbook, and building for mac required replacing my global python binary), but at the very least I managed to use my built version of black to reformat a tiny python file.

Let me know if there's anything else I can do – either upload some of my built wheels, or specific tests that would be good to perform with them.

JelleZijlstra:
Thanks, that sounds great! @ichard26 is our mypyc expert and he tagged himself, so I'll give him some time to add any feedback.

JelleZijlstra:
It's been a month, let's merge this.

PR #3712, opened 2023-05-30
blackd: show default values for options

joelsgp:
I personally find this very useful.

Reference: https://click.palletsprojects.com/en/8.1.x/api/#click.Option

### Checklist - did you ...


cooperlees (approved):
LGTM thanks, I've always wished one could turn this on all option somewhere ...

Or it was the default, to show defaults ... haha :)

PR #3723, opened 2023-06-07
Add CITATION.cff file

### LINKED ISSUE #3722: Add CITATION.cff file

I am unsure how to cite Black in my research articles.

It would be useful to have an explicit metadata component describing how to cite Black.

A CITATION.cff file could be added to the root of the repository. This would allow GitHub to automatically generate a citation in both APA and BibTeX format. Additionally, reference manager plugins such as Zotero will be able to automatically capture the citation information.

The citation information could be added directly to the documentation, or an alternative citation format could be used such as BibTeX. But .cff seems to be the method [recommended by GitHub](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files#about-citation-files).

**Additional context**

Similar citation files are used by popular libraries such as [pandas](https://github.com/pandas-dev/pandas) and [tensorflow](https://github.com/tensorflow/tensorflow), although some use BibTeX format instead such as [NumPy](https://github.com/numpy/numpy). However, using BibTeX means that GitHub cannot automatically generate the citations (in APA) but instead just point to the citation file.

schackartk:
### Description

This pull request adds a CITATION.cff file to the root of the repository. This resolves #3722 and follows [GitHub's recommendation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files#citation-formats).

It will achieve several things:

- Provide explicit information (metadata) on how to cite *Black*
- Allow GitHub to automatically generate APA and BibTeX formatted citations
- Allow reference managers to automatically access the citation metadata

Hopefully, this will make it easier to give due credit to *Black*!

### Further considerations

What metadata to include in this file is subjective, and up to the *Black* development team. Here are a few considerations. Most are tradeoffs between accuracy and increased maintenance:

- Should an exhaustive list of authors be added (and maintained), or Łukasz Langa only, or add "the Black development team" as another author, or just have the dev team listed as the author? The latter option seems the most common.
- Should the version number be added (and maintained)?

### Checklist - did you ...

I don't believe that anything needs to be added to `CHANGES.md`. Maybe it could be added to the Documentation section?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,7 +57,8 @@
 
 ### Documentation
 
-- Add a CITATION.cff file to the root of the repository, containing metadata on how to cite this software (#3723)
+- Add a CITATION.cff file to the root of the repository, containing metadata on how to
+  cite this software (#3723)
 
 <!-- Major changes to documentation and policies. Small docs changes
      don't need a changelog entry. -->
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,8 @@
 
 ### Documentation
 
+- Add a CITATION.cff file to the root of the repository, containing metadata on how to cite this software (#3723)
+
 <!-- Major changes to documentation and policies. Small docs changes
      don't need a changelog entry. -->
 
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -7,6 +7,7 @@ type: software
 authors:
   - family-names: Langa
     given-names: Łukasz
+  - family-name: contributors to Black
 repository-code: 'https://github.com/psf/black'
 url: 'https://black.readthedocs.io/en/stable/'
 abstract: >-
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,27 +1,22 @@
 cff-version: 1.2.0
-title: 'Black: The uncompromising Python code formatter'
+title: "Black: The uncompromising Python code formatter"
 message: >-
-  If you use this software, please cite it using the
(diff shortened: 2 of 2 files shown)


ichard26:
Thanks for the PR!

We can steal what's in the footer of the documentation: 

> Łukasz Langa and contributors to Black

And I'd discourage maintaining a copy of the version number in the citation file. I recognize it may be helpful for citers, but it would introduce more friction during releases (which I'd like to avoid).

A changelog entry isn't necessary here, but it would be nice either way. Putting it under `Documentation` would be the way to go.

JelleZijlstra:
I think I'd actually prefer a changelog entry, as presumably we want users to know about this. We should omit a changelog entry only if users don't need to know about the change.

schackartk:
I have updated the authorship information and the changelog. Hopefully the authorship information looks okay. If not, I can change it to a single author with the literal string "Łukasz Langa and contributors to Black" instead of divided into 2 author fields.

Regarding version number, I agree, it is much more friction. Also, while version number is important, most citation styles don't include version numbers anyway. So if a user wants to be explicit, they can include that information manually, usually directly in the text.

schackartk:
I see Linting is failing. The message looks like it is making the changes itself, is that correct? Or do I need to make the changes it suggests and commit them?

JelleZijlstra:
You need to commit the changes.

PR #3731, opened 2023-06-12
Max line length with bugbear

### LINKED ISSUE #3716: Shouldn't max-line-length be 80 when using bugbear?
**Is this related to a problem? Please describe.**

The [documentation](https://black.readthedocs.io/en/latest/the_black_code_style/current_style.html#line-length) suggests using `flake8-bugbear` instead of just bumping `max-line-length` to 88, as bugbear already handles a line max 10% longer than `max-line-length`. However, the example _with_ bugbear below it still uses 88.

What this effectively means is that flake8 (with bugbear) won't complain about lines up to 96 characters long.

I think the example value should be 80:

```ini
[flake8]
max-line-length = 80
...
select = C,E,F,W,B,B950
extend-ignore = E203, E501
```

Alveel:
### Description

Fixes #3716

The current phrasing of the Flake8 section of the [line length documentation](https://black.readthedocs.io/en/latest/the_black_code_style/current_style.html#line-length) is very open.

This PR makes it more concise and explicit, as well as change the `max-line-length` example when using [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear) to 80, as explained in #3716.

### Checklist - did you ...

### DIFF
--- a/docs/the_black_code_style/current_style.md
+++ b/docs/the_black_code_style/current_style.md
@@ -160,6 +160,8 @@ harder to work with line lengths exceeding 100 characters. It also adversely aff
 side-by-side diff review on typical screen resolutions. Long lines also make it harder
 to present code neatly in documentation or talk slides.
 
+#### Flake8
+
 If you use Flake8, you have a few options:
 
 1. Recommended is using [Bugbear](https://github.com/PyCQA/flake8-bugbear) and enabling its B950 check instead of using Flake8's E501, because it aligns with Black's 10% rule. Install Bugbear and use the following config:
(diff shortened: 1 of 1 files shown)


Alveel:
I don't think an entry in `CHANGES.md` and adding/updating tests is necessary for this change, because:

1. what the documentation says is essentially the same
2. only documentation is updated, no code changes are made

Alveel:
Is it okay to add a header for flake8 or should I remove that? I noticed it would be the only third-level header in the page 😇

cooperlees (approved):
Thanks!

PR #3743, opened 2023-06-22
Allow specifying `--workers` via environment variable

### LINKED ISSUE #3742: Allow specifying `--workers` via environment variable

I ran into https://github.com/psf/black/issues/2513 in my CI system

While I can hardcode `--workers` into my invocations of black (in this case, via pre-commit), it's a little unnatural, since the number of cores differs between CI workers and say locally. E.g., I can hardcode the minimum of them, but then we leave some perf on the table.

I'd like to specify it by environment variable, say `BLACK_NUM_WORKERS`. This isn't unprecedented, e.g. `BLACK_CACHE_DIR` exists. This does increase configuration complexity a little.

Just hardcoding `--workers`. It's not the biggest deal.

hauntsaninja:
Fixes #3742

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,8 @@
 ### Configuration
 
 <!-- Changes to how Black can be configured -->
+- The `--workers` argument to Black can now be specified via the `BLACK_NUM_WORKERS`
+  environment variable (#3743)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,7 @@
 ### Configuration
 
 <!-- Changes to how Black can be configured -->
+
 - The `--workers` argument to Black can now be specified via the `BLACK_NUM_WORKERS`
   environment variable (#3743)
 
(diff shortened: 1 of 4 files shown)


JelleZijlstra (commented):
Looks good, but this should also be mentioned in the documentation, maybe somewhere in https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html (which I think is rather different on main, I need to do a release sometime).

hauntsaninja:
It looks like it will automatically have something added over here from the `--help`: https://github.com/psf/black/blob/9b2b048692f0bd642f38230b4e61c778a4653f91/docs/usage_and_configuration/the_basics.md#L35
```
  -W, --workers INTEGER RANGE     Number of parallel workers [default:
                                  BLACK_NUM_WORKERS environment variable or
                                  number of CPUs in the system]  [x>=1]
```
Should I document this somewhere beyond that? I don't see any other documentation about `--workers`.

hauntsaninja:
I guess I could make a new section below that called "environment variable options" and mention both this and `BLACK_CACHE_DIR`?

JelleZijlstra:
> I guess I could make a new section below that called "environment variable options" and mention both this and `BLACK_CACHE_DIR`?

Yes, sounds good!

JelleZijlstra (approved):

PR #3745, opened 2023-06-23
Fix removed comments in stub files

### LINKED ISSUE #1020: INTERNAL ERROR: *.pyi file + function + comment + ellipsis
Seems like a minimal example requires:
* a `.pyi` file
* a comment after the function declaration
* an ellipsis in the function body 

`example.pyi`:
```python
def f():
    # hi
    ...
```

```
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G95

$ python --version
Python 3.7.4

$ python -m black --version
black.py, version 19.3b0

$ cat example.pyi
def f():
    # hi
    ...

$ python -m black example.pyi
error: cannot format example.pyi: INTERNAL ERROR: Black produced invalid code: expected an indented block (<unknown>, line 3). Please report a bug on https://github.com/ambv/black/issues.  This invalid output might be helpful: /var/folders/m_/ffctv9r113lb285667lc7fh800k0_0/T/blk_2fvut1n6.log
All done! 💥 💔 💥
1 file failed to reformat.

$ cat /var/folders/m_/ffctv9r113lb285667lc7fh800k0_0/T/blk_iqek_luy.log
  File "/Users/dustiningram/.pyenv/versions/3.7.4/lib/python3.7/site-packages/black.py", line 3388, in assert_equivalent
    dst_ast = ast.parse(dst)
  File "/Users/dustiningram/.pyenv/versions/3.7.4/lib/python3.7/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
def foobar():  # hi

...
```

```
$ python -m black --version
black.py, version 19.3b0+82.gae5588c

$ python -m black example.pyi
error: cannot format example.pyi: INTERNAL ERROR: Black produced invalid code: invalid syntax (<unknown>, line 2). Please report a bug on https://github.com/psf/black/issues.  This invalid output might be helpful: /var/folders/m_/ffctv9r113lb285667lc7fh800k0_0/T/blk_yzbb8_6s.log
Oh no! 💥 💔 💥
1 file failed to reformat.
```

### LINKED ISSUE #1239: Comments removed in pyi files
Hi! I accidentally ran black on my entire venv in a project, and one of the files it spat out "INTERNAL ERROR: Black produced code that is not equivalent to the source." on was typeshed/stdlib/3/configparser.pyi. I'm not sure what version of typeshed that is, but the issue doesn't repro with the current version on GitHub.

Regardless, here's the diff:

```
--- src
+++ dst
@@ -5135,8 +5135,6 @@
   type_ignores=
     TypeIgnore(
     )  # /TypeIgnore
     TypeIgnore(
     )  # /TypeIgnore
-    TypeIgnore(
-    )  # /TypeIgnore
 )  # /Module
```

JelleZijlstra:
Fixes #1239. Fixes #1020.

I believe we can fix this without going through the preview style as we're only changing code that was not yet formatted with black; https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy explicitly calls out that we can fix bugs where we remove comments.


tests/data/simple_cases/ignore_pyi.py
```diff
@@ -0,0 +1,41 @@
...
+# whitespace doesn't matter (note the next line has a trailing space and tab)
+class z: ...
+
+def g():
+    # hi
+    ...
+
+def h():
+    ...
+    # bye
```
JelleZijlstra, line 41: Arguably we could turn this into `def h(): ...  # bye`, but that's more complicated to implement and if the user put it on a separate line, we might as well keep it there.

hauntsaninja (approved):
Looks great!

hauntsaninja:
I guess it could make more sense for the testing to look at ".pyi" extension instead of "_pyi" stem, but whatever

JelleZijlstra:
I think I tried that briefly but there was a fairly deep assumption in the test framework that everything has a `.py` extension, so I used this approach instead. Agree that solution would be cleaner though.

ichard26:
Totally out of the loop, but would you think it would be better in the long run to migrate away from assuming everything is a `.py` file? If so, let's open an issue.

JelleZijlstra:
Yes, let's do that

PR #3748, opened 2023-06-24
Decrease cost of ipynb code path when unneeded

hauntsaninja:
IPython is a very expensive import, like, at least 300ms. I'd also venture that it's much more common than tokenize-rt, which is like 30ms. I work in a repo where I use black, have IPython installed and there happen to be a couple notebooks (that we don't want formatted). I know I can force exclude ipynb, but this change doesn't really have a cost.

### DIFF
--- a/src/black/handle_ipynb_magics.py
+++ b/src/black/handle_ipynb_magics.py
@@ -61,6 +61,7 @@ def jupyter_dependencies_are_installed(*, verbose: bool, quiet: bool) -> bool:
         # isort: off
         import tokenize_rt  # noqa:F401
         import IPython  # noqa:F401
+
         # isort: on
     except ModuleNotFoundError:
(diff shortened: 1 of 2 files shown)


src/black/handle_ipynb_magics.py
```diff
@@ -58,8 +58,11 @@ class Replacement:
 @lru_cache()
 def jupyter_dependencies_are_installed(*, verbose: bool, quiet: bool) -> bool:
     try:
-        import IPython  # noqa:F401
+        # isort: off
```
JelleZijlstra, line 61:
```suggestion
        # tokenize_rt is less commonly installed than IPython, and IPython is expensive to import,
        # so import tokenize_rt first.
        # isort: off
```
So future readers know why we're doing this

PR #3749, opened 2023-06-24
Integrate verbose logging with get_sources

hauntsaninja:
Currently the verbose logging for "Sources to be formatted" is a little suspect in that it is a completely different code path from `get_sources`.

This can result in bugs like https://github.com/psf/black/pull/3216#issuecomment-1213557359 and generally limits the value of these logs.

This does change the "when" of this log and splits it out across multiple lines, but the colours help separate it from the even more verbose logs.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -631,7 +631,7 @@ def get_sources(
             is_stdin = False
 
         if is_stdin or p.is_file():
-            normalized_path = normalize_path_maybe_ignore(p, ctx.obj["root"], report)
+            normalized_path: Optional[str] = normalize_path_maybe_ignore(p, ctx.obj["root"], report)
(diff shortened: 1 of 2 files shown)


hauntsaninja:
I'm also interested in this now because we spend a lot of time in `Path.resolve` via `normalize_path_maybe_ignore` and I want to see if I can reduce that cost.

src/black/__init__.py
```diff
@@ -653,7 +633,11 @@ def get_sources(
         if is_stdin or p.is_file():
             normalized_path = normalize_path_maybe_ignore(p, ctx.obj["root"], report)
```
JelleZijlstra, line 634: Seems like the mypyc build is broken here (see the diff-shades output in CI); it infers None for this variable.

PR #3750, opened 2023-06-24
Check self format for the whole repo

hauntsaninja:
`black .` is changing things in gallery and scripts for me


hauntsaninja:
This could probably use a skip news label

JelleZijlstra:
Thanks! We might want to delete `gallery/`.

PR #3751, opened 2023-06-25
Improve performance by skipping unnecessary normalisation

hauntsaninja:
This speeds up black by about 40% on my repo at work when the cache is full. This change is intended to exactly preserve behaviour (other than a slight improvement to verbose logging).


src/black/files.py
```diff
@@ -276,15 +276,24 @@ def normalize_path_maybe_ignore(
...
-def path_is_ignored(
-    path: Path, gitignore_dict: Dict[Path, PathSpec], report: Report
+def _path_is_ignored(
+    root_relative_path: str,
+    root: Path,
+    gitignore_dict: Dict[Path, PathSpec],
+    report: Report,
 ) -> bool:
+    path = root / root_relative_path
+    # Note that this logic is sensitive to the ordering of gitignore_dict. Callers must
```
JelleZijlstra, line 286: Is it? Looks like we check all entries either way.

hauntsaninja: I might be being really dumb, but I think it is, because of the break on L285 / L292. You can hit that in the case of symlinks.

I also think it's maybe a little questionable to be applying gitignore to the resolved file rather than the symlink when formatting the symlink (e.g. a gitignored symlink can now affect what gets formatted), but whatever.

JelleZijlstra: OK I stared at the code some more and this does make sense now. (I thought it mattered only for broken symlinks, but that's not the case.)

We pass a dict here that gets built separately for every directory, since we look for `.gitignore` in every directory, so given a directory structure `root/a/b/c.py`, the dict will contain `root/.gitignore`, `root/a/.gitignore`, `root/a/b/.gitignore` in that order. The `break` happens for the first path where the symlink points outside the directory, so it makes sense to ignore all subsequent gitignores as they definitely won't match either.

Agree that it probably makes more sense to apply gitignore *before* resolving symlinks, but that's something do in a separate PR.

JelleZijlstra (approved):

Issue #3708, opened 2023-05-26
Black Github action causes subsequent actions to fail on self-hosted runner


cjproud:
I'm running a couple of GHA jobs, one of them being the black formatter inside an Ubuntu container on a self-hosted runner. I've noticed that after running the black formatting action there are files leftover in the `_actions` directory that aren't owned by the runner's user but rather by `root`:

```
[...]

PR #3759, opened 2023-07-03
Remove ENV_PATH on Black action completion.

cjproud:
### Description

Relates to #3708 by deleting the `.black.env` folder specified by `ENV_PATH` on action completion.

I had a difficult time replicating the issue above @JelleZijlstra and I can't see that any changes were made to stable that would effect this behaviour. Probably worth deleting that `.black.env` dir anyway?

### Checklist - did you ...

### DIFF
--- a/action/main.py
+++ b/action/main.py
@@ -1,7 +1,7 @@
 import os
 import shlex
-import sys
 import shutil
+import sys
 from pathlib import Path
 from subprocess import PIPE, STDOUT, run
 
--- a/action/main.py
+++ b/action/main.py
@@ -74,6 +74,6 @@ else:
         stderr=STDOUT,
         encoding="utf-8",
     )
-shutil.rmtree(ENV_PATH)
+shutil.rmtree(ENV_PATH, ignore_errors=True)
 print(proc.stdout)
 sys.exit(proc.returncode)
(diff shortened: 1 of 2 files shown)


action/main.py
```diff
@@ -73,5 +74,6 @@
         stderr=STDOUT,
         encoding="utf-8",
     )
+shutil.rmtree(ENV_PATH)
```
ichard26, line 77: Is it worth passing `ignore_errors=True` to this call? An error is unlikely, but it also shouldn't fail the entire action IMO.

ichard26 (commented):
Yeah, I'm not sure how this would cause issues, but this is a good idea either way. Thanks for filing the PR!

Please add a changelog entry though.

ichard26:
Don't mind the PyPy failures, it seems like a recent PyYAML release broke things on PyPy... :/

action/main.py
cjproud: I agree, I think an error is unlikely but yeah shouldn't fail the entire action.

cjproud:
@ichard26 I've added a changelog entry and the `ignore_errors=True` flag to the `shutil.rmtree` function 👍

hauntsaninja (approved):
Thanks, looks like another user confirmed the fix as well!

Issue #3658, opened 2023-04-24
Enable PEP-597 `EncodingWarning` in CI


Zac-HD:
[PEP-597](https://peps.python.org/pep-0597/) added an option to warn when the encoding argument to `open()` is omitted and the default locale-specific encoding is used, because this can cause bugs when the locale is incorrectly assumed to be e.g. utf-8, which may not be the case on Windows, or various systems in e.g. Japan.

Setting `PYTHONWARNDEFAULTENCODING=1` in CI, and fixing any existing issues, will ensure that downstream projects can also adopt this check without awkward workarounds.

PR #3763, opened 2023-07-04
Enable `PYTHONWARNDEFAULTENCODING = 1` in CI

Zac-HD:
Fixes https://github.com/psf/black/issues/3658, and ensures that https://github.com/psf/black/issues/3695 won't regress later.

For clarity I switched most uses of `with open(...)` to use `pathlib`, including a few where the encoding was already OK - I find it much easier to read boilerplate-y tests when they're consistent, and I thought it was worth maintaining that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,6 +69,9 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- Black is now tested with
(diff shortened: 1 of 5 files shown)


hauntsaninja (approved):
Thanks for fixing these!

Zac-HD:
No problem!  I'm looking forward to https://github.com/HypothesisWorks/hypothesis/pull/3691 running clean on upstream releases 😁

PR #3764, opened 2023-07-05
Better error message for invalid exclude types

### LINKED ISSUE #3574: extend-exclude with an array is a footgun
I accidentally had the following configuration in a project:

```
[tool.black]
extend-exclude = [
    "backendapp/migrations"
]
```

If i then run `black . -vvv`, it tells me e.g.

```
manage.py ignored: matches the --extend-exclude regular expression
```

which is really confusing because `backendapp/migrations` is neither a regex nor should it match `manage.py`.

The solution is `extend-exclude = "backendapp/migrations"`, which i only realized when looking into other projects' pyproject.toml files.

I would be nice if black would fail or at least warn when the configuration uses the wrong datatype. Additionally, there aren't any examples of `extend-exclude` in pyproject.toml in the docs, the only usage in the repo is black's own pyproject.toml

n/a

**Additional context**

### LINKED ISSUE #3754: Doesn't honor extend-exclude for nested file
**Describe the bug**

Given the following configuration in `pyproject.toml`
```
[tool.black]
line-length = 120
extend-exclude = [
    '^(?:([^\/]+\/)+(?:test_.*\.py|.*_test\.py))$',
    '.github',
    '.mypy_cache',
    '.pytest_cache',
    'venv'
]
```

Black fails with the error:

```
Error: Invalid value for '--extend-exclude': Not a valid regular expression: unbalanced parenthesis at position 17
```

But this is a valid regex that's supposed to match any files starting or ending in `test` or files in a directory named `test`. For example, the following path matches the `test` directory:
myapp/folder/test/test_something.py

**To Reproduce**

I guess use the regex in a unit test that parses it?

**Expected behavior**

It should work. It would be even better if Black could use the `extend-exclude` list from flake8 or mypy, so that I don't have repeat it. The [documentation](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html) only shows the `max-line-length` property. 

**Environment**

- Black's version: 23.3.0
- OS: macOS Ventura 13.4.1, M1 chip
- Python version: (CPython) 3.11.3

hauntsaninja:
Currently if this is a list, this ends up using the repr of the list as a regex.

Fixes #3574, fixes #3754

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -33,7 +33,7 @@
 - Fix black not honouring `pyproject.toml` settings when running `--stdin-filename` and
   the `pyproject.toml` found isn't in the current working directory (#3719)
 - Black will now error if `exclude` and `extend-exclude` have invalid data types in
-  `pyproject.toml`, instead of silently doing the wrong thing.
+  `pyproject.toml`, instead of silently doing the wrong thing (#3764)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,8 @@
 - `.pytest_cache`, `.ruff_cache` and `.vscode` are now excluded by default (#3691)
 - Fix black not honouring `pyproject.toml` settings when running `--stdin-filename` and
   the `pyproject.toml` found isn't in the current working directory (#3719)
+- Black will now error if `exclude` and `extend-exclude` have invalid data types in
+  `pyproject.toml`, instead of silently doing the wrong thing.
 
 ### Packaging
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -159,9 +159,7 @@ def read_pyproject_toml(
 
     exclude = config.get("exclude")
     if exclude is not None and not isinstance(exclude, str):
-        raise click.BadOptionUsage(
-            "exclude", "Config key exclude must be a string"
-        )
+        raise click.BadOptionUsage("exclude", "Config key exclude must be a string")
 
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -32,6 +32,8 @@
 - `.pytest_cache`, `.ruff_cache` and `.vscode` are now excluded by default (#3691)
 - Fix black not honouring `pyproject.toml` settings when running `--stdin-filename` and
   the `pyproject.toml` found isn't in the current working directory (#3719)
+- Black will now error if `exclude` and `extend-exclude` have invalid data types in
+  `pyproject.toml`, instead of silently doing the wrong thing.
```
JelleZijlstra, line 36:
```suggestion
  `pyproject.toml`, instead of silently doing the wrong thing. (#3764)
```

hauntsaninja: Github is doing something really weird where the pushes show on the branch in my fork, but don't get reflected in the PR https://github.com/hauntsaninja/black/commit/9f55e735ddbe600565e033543208689787b4b7db

JelleZijlstra: Yeah seems like there is something weird going on. In any case, this PR looks good once GitHub cooperates.

hauntsaninja: For whatever reason, the force in force push seems to wake it up. So I just need to commit an empty commit, push it, delete the commit and force push.

JelleZijlstra: Sounds like a great workflow

JelleZijlstra (approved):

PR #3765, opened 2023-07-05
Drop support for Python 3.7

hauntsaninja:
#3766


cooperlees (approved):
Awesome. So much cleanup going to >= 3.8.

I also agree the other patches etc. That we can remove are better for separate PRs so we can ensure more closely they don't change any behavior.

JelleZijlstra (approved):

PR #3768, opened 2023-07-05
Remove click patch

hauntsaninja:
Apparently this was only needed on Python 3.6. We've now dropped support for 3.6 and 3.7. It's also not needed on new enough click.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,7 @@
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
 - Upgrade mypyc from 0.991 to 1.3 (#3697)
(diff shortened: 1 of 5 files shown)


cooperlees (approved):
Wow. This brings back memories from where the terrorized us at work everywhere too. What a time to be alive deleting legacy on black 🥰

cooperlees:
Let's add a change log so if people see issues with click things on newer python we know something is up here ...

hauntsaninja:
Hmm, that's weird, added a changelog and now tests fail. Looks like there's a new Click release...

hauntsaninja:
Okay, this is https://github.com/pallets/click/issues/2558

JelleZijlstra (approved):

PR #3770, opened 2023-07-08
Fix CI for Click typing issue

hauntsaninja:
https://github.com/pallets/click/issues/2558


hauntsaninja:
Oh duh, it's not working because the baseline is broken

hauntsaninja:
Okay, I've removed the specifier from the user facing places. I think this will need to be yolo merged

PR #3773, opened 2023-07-09
Fix crash on type comment with trailing space

### LINKED ISSUE #3496: Internal error on type hinted comment with trailing space
**Describe the bug**

Black is encountering an `INTERNAL ERROR` when a type hinted comment has a trailing space.

**To Reproduce**

For example, take this code:
With the trailing space after `# type: dict[str, str]  ` black encounters an internal error
```python
d = {}  # type: dict[str, str] 
```
remove the white space character like `# type: dict[str, str]` and the file is formatted.

```sh
$ black t.py 

```

The resulting error is:

> error: cannot format t.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_66snb7vb.log

**Expected behavior**

**Environment**

- Black's version: black==22.12.0
- OS and Python version: WSL Ubuntu 22.04 Python 3.10.6

**Additional context**
here's the log
```
--- src
+++ dst
@@ -8,11 +8,11 @@
             )  # /Store
           id=
             'd',  # str
         )  # /Name
       type_comment=
-        'dict[str, str] ',  # str
+        'dict[str, str]',  # str
       value=
         Dict(
           keys=
           values=
         )  # /Dict

```

hauntsaninja:
Fixes #3496

### DIFF
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -208,15 +208,18 @@ def stringify_ast(node: ast.AST, depth: int = 0) -> Iterator[str]:
 
         else:
             normalized: object
-            # Constant strings may be indented across newlines, if they are
-            # docstrings; fold spaces after newlines when comparing. Similarly,
-            # trailing and leading space may be removed.
             if (
                 isinstance(node, ast.Constant)
                 and field == "value"
                 and isinstance(value, str)
             ):
+                # Constant strings may be indented across newlines, if they are
+                # docstrings; fold spaces after newlines when comparing. Similarly,
+                # trailing and leading space may be removed.
                 normalized = _normalize("\n", value)
+            elif field == "type_comment" and isinstance(value, str):
+                # Trailing spaces in type comments are removed.
+                normalized = value.rstrip(" ")
             else:
                 normalized = value
             yield f"{'  ' * (depth+2)}{normalized!r},  # {value.__class__.__name__}"
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -218,8 +218,8 @@ def stringify_ast(node: ast.AST, depth: int = 0) -> Iterator[str]:
                 # trailing and leading space may be removed.
                 normalized = _normalize("\n", value)
             elif field == "type_comment" and isinstance(value, str):
-                # Trailing spaces in type comments are removed.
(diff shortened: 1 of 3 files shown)


src/black/parsing.py
```diff
@@ -208,15 +208,18 @@ def stringify_ast(node: ast.AST, depth: int = 0) -> Iterator[str]:
...
                 isinstance(node, ast.Constant)
                 and field == "value"
                 and isinstance(value, str)
             ):
+                # Constant strings may be indented across newlines, if they are
+                # docstrings; fold spaces after newlines when comparing. Similarly,
+                # trailing and leading space may be removed.
                 normalized = _normalize("\n", value)
+            elif field == "type_comment" and isinstance(value, str):
+                # Trailing spaces in type comments are removed.
```
JelleZijlstra, line 221: Interesting that we apparently do not remove trailing tabs, only spaces. That might be a bug?

In any case, this change looks good, thanks!

hauntsaninja: Oh, we do remove tabs. Sorry, I forgot those exist.

JelleZijlstra: To be clear we actually *don't* strip trailing tabs:

```
(py311) jelle@m2mb-jelle black % black -c "x = int # type: int\t"     
x = int  # type: int\t
(py311) jelle@m2mb-jelle black % black -c "x = int # type: int   "
x = int # type: int   
error: cannot format <string>: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /var/folders/hh/hdb5qd2x3v51v6l0gknpfjfr0000gp/T/blk_7xu6fnsp.log
```

JelleZijlstra: Oh wait, that's not actually a tab, is it? Just a \t in the comment.

JelleZijlstra (approved):

src/black/parsing.py
hauntsaninja: Yeah, that's not a tab. `black -c $'x = int # type: int\t'`

PR #3775, opened 2023-07-09
Upgrade to latest mypy

hauntsaninja:
Newer mypy is typically much faster, a lot has happened since 0.991


cooperlees (approved):
Thanks.

Let's activate the auto updating of the .pre-commit-config.yaml - I thought we had.

JelleZijlstra:
I enabled pre-commit.ci on this repo, hope that's enough.

PR #3776, opened 2023-07-10
Prepare release 23.7.0

JelleZijlstra:


JelleZijlstra:
Nice, we get automatic prettier fixes from pre-commit now

hauntsaninja (approved):
Looks good (but didn't check if there were any other changes to be made outside of diff)

cooperlees (approved):
LGTM

PR #3779, opened 2023-07-11
Fix typo in CITATION.cff

skykasko:
### Description

Add a space between "of" and "hand-formatting."


ichard26 (approved):
Thank you very much!

PR #3782, opened 2023-07-11
Avoid importing `IPython` if notebook cells do not contain magics

neutrinoceros:
### Description

follow up to #3748: completely avoid import overhead in `jupyter_dependencies_are_installed`. This function now takes <0.2ms regardless of the result, while the current implementation takes >200ms on first call when the result is `True`.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,8 @@
 
 <!-- Changes that improve Black's performance. -->
 
+- minimize cost of introspection to detect jupyter dependencies (#3782)
+
 ### Output
 
 <!-- Changes to Black's terminal output and error messages -->
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,7 +30,7 @@
 
 <!-- Changes that improve Black's performance. -->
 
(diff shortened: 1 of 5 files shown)


hauntsaninja (commented):
Hmm, I thought about this, but when this returns true we'll end up importing these anyway right? In which case this introduces overhead

hauntsaninja:
I guess we can save if there isn't a cell magic in the notebook

neutrinoceros:
TBH I didn't look at how the function was integrated in the rest of the control flow. From an external point of view I thought it might be preferable to decouple introspection and actual imports, especially if the added overhead is 3 orders of magnitude bellow the import cost, but your call ! :)

drunkwcodes (approved):

ichard26 (approved):
The idea sounds reasonable and the changes look good. Please address failing lint CI though.

hauntsaninja:
Yeah, sorry, lint is failing because of click stuff again that's unrelated to these changes

ichard26:
Thanks @neutrinoceros and @hauntsaninja!

PR #3783, opened 2023-07-11
Remove unneeded mypy dependencies

AlexWaygood:
Black no longer uses typed-ast or the dataclasses backport, so these should both be unnecessary now!


JelleZijlstra:
Thanks!

PR #3791, opened 2023-07-17
Continue to avoid Click typing issue

hauntsaninja:


ichard26:
Sorry for not being up to date on the situation, but do you think the typing issue will be addressed soon? It's a shame that we seemingly have to continue to exclude new click releases one by one :/

Anyway, feel free to merge once lint CI passes.

hauntsaninja:
Yes, I think 8.1.5 was meant to fix the issue but didn't fully fix things

PR #3796, opened 2023-07-17
More concise formatting for dummy implementations

### LINKED ISSUE #1797: Allow reduced number of blank lines for dummy implementations
Hi!

I would like to start a discussion about reducing blank lines in some special cases, I hope this is welcome :)

The [Blank Lines](https://www.python.org/dev/peps/pep-0008/#blank-lines) section in PEP8 states:

> Blank lines may be omitted between a bunch of related one-liners (e.g. a set of dummy implementations).

While this is in the same paragraph as a sentence about functions, I would like to propose a liberal but well defined interpretation of this. It boils down to allowing reducing blank lines for empty classes and functions. This will make code that uses a lot of type definitions or overloads more readable. In the case of overloads this especially makes sense as it will allow overload definitions to be visually grouped with the overloaded function.

My opinion is also that the requirement for classes to take at least two lines and require being surrounded by two blank lines makes these definition _feel_ heavy. Now talking about how code feels might seem misplaced and unscientific mumbo-jumbo, but I think it is important to allow class definitions to feel light when they're used in a Type-driven development style. Some of these usecases for class definitions that would benefit from _feeling_ lighter are custom exceptions, and intersections. Why should a type definition like `A = NewType("A", int)` only require one line while a type definition like `class Intersection(A, B): ...` require six? I argue that making these definitions feel heavy will make developers refrain from using more elegant solutions that utilizes the type system over solutions that feel light but are in reality less expressive and doesn't describe business domains as fully as they could.

1. Allow putting the entire definition on one line if the entire body consists of `pass` or `...` (ellipses).

```python
class CustomError(Exception): ...

def fn(a): ....
```

2. Allow removing blank lines between empty classes and between functions, but require *one*  blank line between an empty class and an empty function. Functions and classes with a docstring as the entire body will also be considered "empty", but not allowed to be defined on a single line.

```python
class CustomError(Exception):
    """Allow documented classes to feel cozy too, as we don't want to discourage docs."""
class OtherError(CustomError): ...

def dummy(a): ...
def other(b): ...
```

3. Non-dummy classes and functions should remain their current set of rules for surrounding blank lines. With the exception of overloads. The rule for overloads could be generalized as to allow functions with the same name to skip blank lines between them, but to require two blank lines surrounding them.

```python
class Error(Exception): ...


class Implementation(Exception):
    """I am not an empty class and require some space"""

    a = 1


def fn(): ...


@overload
def a(arg: int) -> int: ...
@overload
def a(arg: str) -> str: ...
@overload
def a(arg: object) -> NoReturn: ...
def a(arg: Union[int, str, object]) -> Union[int, str]:
    if not isinstance(arg, (int, str)):
        raise TypeError
    return arg
```

While I think this could become well defined, and should before implementation work starts, I'm sure I've missed some edge cases that needs defining. Let's hammer them out *if* this is welcomed as a good idea in the first place, and not shot down ;)

### Current behaviour

Example 2. is currently reformatted by black as (14 lines instead of 6):

```python
class CustomError(Exception):
    """Allow documented classes to feel cozy too, as we don't want to discourage docs."""


class OtherError(CustomError):
    ...


def dummy(a):
    ...


def other(b):
    ...
```

Example 3. is currently formatted like this with black (33 instead of 22). Notice how information is lost here, the original formatting communicates to a reader that the overloads are related to each other:

```python
class Error(Exception):
    ...


class Implementation(Exception):
    """I am not an empty class and require some space"""

    a = 1


def fn():
    ...


@overload
def a(arg: int) -> int:
    ...


@overload
def a(arg: str) -> str:
    ...


@overload
def a(arg: object) -> NoReturn:
    ...


def a(arg: Union[int, str, object]) -> Union[int, str]:
    if not isinstance(arg, (int, str)):
        raise TypeError
    return arg
```

hauntsaninja:
Fixes #1797, in particular, see https://github.com/psf/black/issues/1797#issuecomment-1528381732


hauntsaninja:
Thanks for improving the comment! This is my testing PR, was looking at https://github.com/psf/black/issues/1797#issuecomment-1528381732 and wanted to see what diff-shades has to say about potential changes :-)

ichard26:
Ah I didn't realize that, haha. I thought you were wondering what was up with the comment looking wonky :P

src/black/lines.py
```diff
@@ -165,6 +165,13 @@ def is_def(self) -> bool:
             and second_leaf.value == "def"
         )
 
+    @property
+    def is_stub_def(self) -> bool:
```
hauntsaninja, line 169: Mirrors `is_stub_class` from a couple lines above

src/black/lines.py
```diff
@@ -714,7 +725,14 @@ def _maybe_empty_lines_for_class_or_def(
             else:
                 newlines = 0
         else:
-            newlines = 1 if current_line.depth else 2
+            if (
+                Preview.dummy_implementations in self.mode
+                and self.previous_line.is_stub_def
+                and (current_line.is_stub_def or current_line.is_decorator)
+            ):
+                newlines = user_hint_before
```
hauntsaninja, line 733: This is perhaps controversial, but it allows users to group things as they want (e.g. in the test case, dummy and other are grouped separately) and solves some lookahead problems when dealing with decorators

src/black/lines.py
```diff
@@ -578,6 +585,8 @@ def _maybe_empty_lines(self, current_line: Line) -> Tuple[int, int]:
             first_leaf.prefix = ""
         else:
             before = 0
+
+        user_hint_before = before
```
hauntsaninja, line 589: (Note that this has already been min-ed with max_allowed)

src/black/lines.py
```diff
@@ -165,6 +165,13 @@ def is_def(self) -> bool:
             and second_leaf.value == "def"
         )
 
+    @property
+    def is_stub_def(self) -> bool:
+        """Is this line a function definition with a body consisting only of "..."?"""
+        return self.is_def and self.leaves[-3:] == [
```
JelleZijlstra, line 171: Doesn't this also match something like `def f(x): y = ...`?

tests/data/preview/dummy_implementations.py
```diff
@@ -0,0 +1,71 @@
+from typing import NoReturn, Protocol, Union, overload
+
+
+def dummy(a): ...
+def other(b): ...
+
+
+@overload
+def a(arg: int) -> int: ...
+@overload
```
JelleZijlstra, line 10: Add a test where there are blank lines between overloads in the input.

hauntsaninja:
I didn't like the way my previous implementation special cased decorators, so I've taken a more normalised approach in this latest commit.

src/black/lines.py
hauntsaninja: Yes, this is true (and true for the is `is_stub_class` heuristic above as well). I've made the heuristic slightly better by including the colon. I didn't see anything adverse in the diff shades output.

Also note some of the pyi code we now use in linegen.py will trigger in some cases involving other statements. Didn't see anything in diff shades and it's not unreasonable when it happens.

JelleZijlstra (approved):

JelleZijlstra:
This change isn't in any release, and is currently only in the preview mode. If you don't want such new changes, use the stable style. Also, if you are using Black to format your code, probably best to not also run the PyCharm formatter, as we don't generally attempt to keep Black compatible with it.

PR #3797, opened 2023-07-17
Simplify empty line tracker

hauntsaninja:


src/black/lines.py
```diff
@@ -49,7 +49,7 @@
 class Line:
     """Holds leaves and comments. Can be printed with `str(line)`."""
 
-    mode: Mode
+    mode: Mode = field(repr=False)
```
hauntsaninja, line 52: This makes debugging much less verbose; plenty of other opportunities to see what the mode is.

hauntsaninja:
I was struggling to understand what was going on here, until I realised the loop body was actually just mostly unrelated to the loop.

It's interesting to me how code evolves over time... the core concepts become slowly become harder to discern, and sometimes if that core concept contained an extra indentation, the effects ripple on till today: https://github.com/psf/black/blame/72de89f54e744a8b3bbc9d3ea71f78a67d8ba82e/src/black/lines.py#L449-L454

PR #3806, opened 2023-07-23
Fix unintentionally swapped words in index.md

pradeep90:
### Description

I think the intent was to say "large changes in formatting", because it doesn't make sense to say "large formatting in changes".

### Checklist - did you ...


JelleZijlstra (approved):
Good catch!

JelleZijlstra:
Looks like this PR is targeting `stable` instead of `main`, could you fix that? Might have to file a new PR.

pradeep90:
Put up https://github.com/psf/black/pull/3809 targeting `main` instead.

PR #3810, opened 2023-07-24
Refactoring the assert statement.

freddiewanah:
### Description

Hello, first-time submitting PR here.

This PR updates an assert statement that checks the bounds of a string-slicing operation. The updated assertion provides more accurate and informative error handling by specifically checking the relative values of the indices and the string length.

The original assertion was essentially checking if Python's string slicing was behaving as expected. However, it wasn't providing any guarantees or useful information about the bounds i and j themselves.

The updated assertion checks that the indices used for slicing are within the bounds of the string. It will throw an AssertionError if the indices are out of bounds or if i > j, providing a more specific and informative error.

### Checklist - did you ...

Happy to update more related issues in the test code if needed :)

### DIFF
--- a/tests/test_trans.py
+++ b/tests/test_trans.py
@@ -13,7 +13,7 @@ def test_fexpr_spans() -> None:
         # a glance than only spans
         assert len(spans) == len(expected_slices)
         for (i, j), slice in zip(spans, expected_slices):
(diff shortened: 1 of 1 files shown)


ichard26 (approved):
Thanks!

PR #3817, opened 2023-07-27
Fix typo in `target-version` param wrongly used in plural

zyv:
### Description

In the code and tests the parameter is called `target-version`, not `target-versionS`. I think it's not good to have this typo here as it present a copy & paste risk of confusion.

### Checklist - did you ...


cooperlees (approved):
Thanks for taking the time to fix this!

zyv:
Thank you for quick processing and maintaining the tool!

Issue #109, opened 2018-04-05
Suggestion: Cache already formatted files


ojii:
We have a decently sized Python code base and running black is "slow" (~10 seconds), which can be a bit annoying when run often. A solution for this would be to cache already formatted files (eg by storing a checksum of them in a file somewhere).

I've implemented a hacky solution for this and for my project, on a fully black-ed + cacheed run takes ~1s vs the ~10s for fully black-ed + uncached. The initial run (fully non-black + uncached) takes significantly longer, so clearly more work is needed should this feature be desired, but before I continue working on this I want to know if this is something that is desired and could eventually be merged in.

ambv:
If this is in the interest of speeding things up then yeah, you have my full attention.

Playing devil's advocate here for a while:
1. Maybe you could look into a solution like [pre-commit](https://pre-commit.com) which essentially gives you what you want: it will only reformat files that are staged and doesn't touch anything else?
2. Maybe you need integration with your text editor? That way you only format the files you're working on, not running it on everything every time.
3. Where would Black look for this cache you're speaking of? The point would be to make statting for such file as fast as possible.

Take your time to answer. Also, there are two big optimizations that I'm planning to do myself. First is cythonizing everything, which should already cut down startup time and maybe 10% runtime. Second is `blackd` with a simple Unix socket interface that is going to circumvent startup time entirely for text editor integration.

If any of those sounds interesting then I'd be thrilled to work together on those things.

ambv:
The reason I asked for cache location is that if subsequent runs are unable to find it (because they were running from a different directory) then it's not very useful. And if Black starts walking directory trees, that will also impacts its speed (hence no configuration for now for example).

If you haven't tried that yet, try pre-commit.com, which only acts on changed files, so you get the best of both worlds.

Working this into an in-memory cache in blackd seems like the best option for me.

ojii:
I've actually thought more about this, and the way I cache it now (checksum per file) is the wrong approach. Instead, there should be a "global" file (somewhere in `~/.cache/black` on linux for example) which simply stores checksums. If two files are the same, we don't care which path they're at, if either of them is properly formatted, so is the other. 

The naive approach with this would be to just add all checksums of properly formatted files to that cache, but this would result in the cache file infinitely growing. Instead, it should probably be a LRU cache of a fixed maximum size. 

An eventual `blackd` could then also pick up this cache file, since the current directory no longer matters.

ojii:
[...]

PR #3821, opened 2023-07-29
Improve caching by comparing file hashes as fallback for mtime and size

cdce8p:
### Description
Rewrite and improve caching implementation to use file hashes as fallback for mtime and size comparisons. Especially for CI systems, comparing just based on mtime makes caching practically useless. With each new run and git checkout it changes and the cache will miss even if the file didn't change.

This PR adds a fallback to compare file hashes if the mtime changed to resolve that. It's not as fast, but still much faster than formatting the file outright. This approach is used by other tools as well, like mypy.
https://github.com/python/mypy/blob/v1.4.1/mypy/fswatcher.py#L80-L88

For the initial caching implementation comparing hashes was dismissed because the benefit was seen as not worth it (at least at first) https://github.com/psf/black/issues/109#issuecomment-381535094. However, as mentioned above and later in the issue https://github.com/psf/black/issues/109#issuecomment-1289352249, it's necessary for CI systems.

A quick performance comparison for https://github.com/home-assistant/core, roughly 10.000 files, run with pre-commit and Github actions.
- `23.7.0`: ~3:30 - 4:00min
- With PR: ~10s

### Checklist - did you ...

### DIFF
--- a/src/black/cache.py
+++ b/src/black/cache.py
@@ -1,21 +1,28 @@
 """Caching of formatted files with feature-based invalidation."""
-
+import hashlib
 import os
 import pickle
+import sys
 import tempfile
+from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Dict, Iterable, Set, Tuple
+from typing import Dict, Iterable, NamedTuple, Set, Tuple
 
 from platformdirs import user_cache_dir
 
 from _black_version import version as __version__
 from black.mode import Mode
 
-# types
-Timestamp = float
-FileSize = int
-CacheInfo = Tuple[Timestamp, FileSize]
-Cache = Dict[str, CacheInfo]
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
+
+
+class FileData(NamedTuple):
+    st_mtime: float
+    st_size: int
+    hash: str
 
 
 def get_cache_dir() -> Path:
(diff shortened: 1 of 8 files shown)


src/black/cache.py
```diff
@@ -37,61 +44,103 @@ def get_cache_dir() -> Path:
...
+
+    def filtered_cached(self, sources: Iterable[Path]) -> Tuple[Set[Path], Set[Path]]:
+        """Split an iterable of paths in `sources` into two sets.
+
+        The first contains paths of files that modified on disk or are not in the
+        cache. The other contains paths to non-modified files.
+        """
+        changed: Set[Path] = set()
+        done: Set[Path] = set()
+        for src in sources:
```
hauntsaninja, line 114: Why doesn't this use `is_changed`?

src/black/cache.py
```diff
@@ -37,61 +44,103 @@ def get_cache_dir() -> Path:
...
+                new_hash = Cache.hash_digest(res_src)
+                if st.st_size != old.st_size or new_hash != old.hash:
+                    changed.add(src)
+                    continue
+            done.add(src)
+        return changed, done
 
-def filter_cached(cache: Cache, sources: Iterable[Path]) -> Tuple[Set[Path], Set[Path]]:
-    """Split an iterable of paths in `sources` into two sets.
+    def update(self, sources: Iterable[Path]) -> None:
```
hauntsaninja, line 130: nit: can we name this something that makes it clearer it doesn't write? I'm also not seeing where we benefit from the separation, so maybe we should just make this write?

hauntsaninja (commented):
Thanks, this is a nice feature!

src/black/cache.py
```diff
@@ -37,61 +44,103 @@ def get_cache_dir() -> Path:
...
+
+        return cls(mode, cache_file, file_data)
+
+    @staticmethod
+    def hash_digest(path: Path) -> str:
+        """Return hash digest for path."""
+
+        with open(path, "rb") as fp:
+            data = fp.read()
+        return hashlib.sha256(data).hexdigest()
```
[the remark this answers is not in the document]
cdce8p: `sha256` works well for mypy. I would stick with it in this case.
https://github.com/python/mypy/blob/v1.5.1/mypy/util.py#L501-L510

tests/test_black.py
```diff
@@ -1979,31 +1980,32 @@ def test_cache_multiple_files(self) -> None:
...
+            with patch.object(black.Cache, "read") as read_cache, patch.object(
+                black.Cache, "write"
             ) as write_cache:
                 cmd = [str(src), "--diff"]
                 if color:
                     cmd.append("--color")
                 invokeBlack(cmd)
                 cache_file = get_cache_file(mode)
                 assert cache_file.exists() is False
+                read_cache.assert_called_once()
```
hauntsaninja, line 2007: This is a change in behaviour from eagerly reading the cache. Doesn't seem like a big deal

tests/test_black.py
```diff
@@ -2057,21 +2059,65 @@ def test_filter_cached(self) -> None:
...
-                cache, {uncached, cached, cached_but_changed}
-            )
+            cache = black.Cache.read(DEFAULT_MODE)
+
+            orig_func = black.Cache.get_file_data
+
+            def wrapped_func(path: Path) -> FileData:
+                if path == cached:
+                    return orig_func(path)
+                return FileData(0.0, 0, "")
```
hauntsaninja, line 2069: nit:
```suggestion
                if path == cached_but_changed:
                    return FileData(0.0, 0, "")
                raise AssertionError
```

hauntsaninja (approved):
Thanks for the updates! Looks good, but I'd like for Jelle to take a look before merging

tests/test_black.py
cdce8p: Yeah, this was a result of initializing the cache with `cache = Cache.read(mode)`.  Technically it did one unnecessary read in some cases.

Changing it wasn't too difficult. I just pushed 85b4a91c1b9dc70cb53af2dc37c9eaf4b985e453 to delay the cache read.

hauntsaninja (commented):
Hmm, I'm not sure I like this impl of delayed cache read, it seems too easy to forget to call `cache.read()`. I think I'd prefer one of:
a) make `file_data` a property that reads the cache file if `self._file_data` is not set
b) initialise `file_data` to None, so we get errors if we try to use the Cache without `read`-ing first
c) revert to before the last commit

cdce8p:
> Hmm, I'm not sure I like this impl of delayed cache read, it seems too easy to forget to call `cache.read()`.

True, even though that is how the current implementation (on main) does it. Always initialize `cache = {}` and do `read_cache(mode)` if necessary. But I agree, it's easy to accidentally forget the read.

> I think I'd prefer one of:
> a) make `file_data` a property that reads the cache file if `self._file_data` is not set
> b) initialise `file_data` to None, so we get errors if we try to use the Cache without `read`-ing first
> c) revert to before the last commit

I don't like (a). Doing the `read` when accessing the `file_data` is not something one would expect. It should either be in the constructor or a separate method. That's one thing I liked about the initial implementation `Cache.read(mode)` as classmethod was quite explicit and you immediately knew what would happen.

I reverted the commit. Long term it might make sense to use the cache in `--diff` mode as well in which case it's always needed anyway.

hauntsaninja:
@JelleZijlstra this looks good to me, but I'd like to wait until you have a chance to take a look!

src/black/cache.py
```diff
@@ -37,61 +44,93 @@ def get_cache_dir() -> Path:
...
+                return cls(mode, cache_file)
+
+        return cls(mode, cache_file, file_data)
+
+    @staticmethod
+    def hash_digest(path: Path) -> str:
+        """Return hash digest for path."""
+
+        with open(path, "rb") as fp:
+            data = fp.read()
```
JelleZijlstra, line 81:
```suggestion
        data = path.read_bytes()
```

src/black/cache.py
```diff
@@ -37,61 +44,93 @@ def get_cache_dir() -> Path:
...
+        return hashlib.sha256(data).hexdigest()
+
+    @staticmethod
+    def get_file_data(path: Path) -> FileData:
```
JelleZijlstra, line 85: Why not a global function? staticmethods often feel a bit useless.

cdce8p: I like it here as it helps to group these methods nicely. Obviously personal preference. Though, if you want me to change it, I can do that too.

JelleZijlstra: Let's just leave it as is, thanks!

JelleZijlstra (approved):

hauntsaninja:
Hmm we'll see if https://github.com/psf/black/pull/3843 fixes CI

cdce8p:
Any indication when the next release will be? I wouldn't ask normally, but the change will improve CI times drastically for larger projects when combined with caching. Would love to start using it.

JelleZijlstra:
It's been about two months, so I'll start the release process over the next few days.

PR #3822, opened 2023-07-29
Add PEP701 support

### LINKED ISSUE #3746: Support PEP 701: Syntactic formalization of f-strings
https://peps.python.org/pep-0701/, which will be in Python 3.12, adds support for several f-string variations that our current parser won't handle:

```
>>> f"{f"{"f"}"}"
'f'
```

We should support this new syntax, somehow.

I don't know if this is doable with blib2to3.

tusharsadhwani:
### Description

Adds support for [PEP 701](https://peps.python.org/pep-0701/): Syntactic formalization of f-strings to black's tokenizer.

Resolves #3746 

Given this Python file:

```python
x = f"foo{2 + 2}bar"
```

#### Previous output:

```console
$ python src/blib2to3/pgen2/tokenize.py asd.py
1,0-1,1:        NAME    'x'
1,2-1,3:        OP      '='
1,4-1,20:       STRING  'f"foo{2 + 2}bar"'
1,20-1,21:      NEWLINE '\n'
2,0-2,0:        ENDMARKER       ''
```

#### Current output

```console
$ python src/blib2to3/pgen2/tokenize.py asd.py
1,0-1,1:        NAME    'x'
1,2-1,3:        OP      '='
1,4-1,2:        FSTRING_START   'f"'
1,2-1,5:        FSTRING_MIDDLE  'foo'
1,5-1,6:        LBRACE  '{'
1,6-1,7:        NUMBER  '2'
1,8-1,9:        OP      '+'
1,10-1,11:      NUMBER  '2'
1,12-1,13:      RBRACE  '}'
1,12-1,15:      FSTRING_MIDDLE  'bar'
1,15-1,20:      FSTRING_END     '"'
1,20-1,21:      NEWLINE '\n'
2,0-2,0:        ENDMARKER       ''
```

### Checklist - did you ...


tusharsadhwani:
@ambv This is the approach that I'm thinking of: defining a `tokenize_string` generator, which in turn will call `generate_tokens()` inside it to be able to tokenize Python source inside it.

Does this seem ok?

tusharsadhwani:
cc @JelleZijlstra as well, for the above comment.

src/blib2to3/pgen2/tokenize.py
```diff
@@ -694,8 +717,82 @@ def generate_tokens(
...
+                yield token_with_updated_pos
+
+    yield (FSTRING_END, endquote, (lnum, max), endpos, line)
+
+
 if __name__ == "__main__":  # testing
     if len(sys.argv) > 1:
         tokenize(open(sys.argv[1]).readline)
-    else:
-        tokenize(sys.stdin.readline)
```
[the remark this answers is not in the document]
tusharsadhwani: unrelated, will revert

tusharsadhwani:
Also, although I don't think this change will be backwards incompatible, should it be possible to keep the old behaviour intact? Should this be behind the experimental flag?

JelleZijlstra:
@tusharsadhwani It's fine if we parse this code in earlier Python versions too. Catching syntax errors is not a goal for Black.

tusharsadhwani:
@ambv wouldn't the stack frames made by calling `generate_tokens()` recursively do the same job of maintaining a stack?

You can think of every call to `tokenize_string()` to adding an fstring-mode to the stack, and the `generate_tokens()` call inside it to pushing a normal mode to the stack.

ambv:
No, the call stack isn't automatically compatible with the stack we need here.

You need to change how `endprog` in `generate_tokens` works for strings. This is where the new stack needs to be.

When you encounter `f"` then you switch to string literal tokenizing, looking for either `"` or `{` as `endprog`. When you encounter `{` you add to the stack and switch back to regular tokenizing... but since the stack is not empty, you watch out for a `}` to return back to string tokenizing.

That way you can encounter another `"`, switch to string tokenizing, and add a `"` as `endprog` to switch back to regular parsing. And when you reach *that* `"`, then you remove it from the stack, switch to regular parsing, but there's still `}` on the stack so you keep looking for that.

At no point are you iterating a counter like `curly_brace_level`. The `endprog` stack depth is the level.

Is that clearer now?

tusharsadhwani:
Yes, it's clearer now.

I'll implement the rest.

tusharsadhwani:
@ambv For code like this (which was invalid in Python3.11 but is valid in Python3.12):

```python
f' \' {f"'"} \' '
```

Black should try to normalize it to something like this:


```python
f" ' {f"'"} ' "
```

Right?

This I believe would require a well thought-out algorithm to figure out exactly which f-strings we normalize and which we don't. Because, depending on whether we normalize the inner one or not, the number of `\'` present in the complete output could change.

I'm not sure what the right approach is here, should I leave that work for later, and only worry about, say, the top level f-string normalization for now?

tusharsadhwani:
cc @JelleZijlstra

ambv:
Good point that we should do backslash normalization here as well for 3.12+ source code input. Sounds like a non-trivial design surface so it wouldn't be an issue if we don't ship this right away.

tusharsadhwani:
@ambv I'm trying to write logic to split/not split the line on a `LBRACE` inside an f-string. The tests with long lines containing f-strings fail right now, everything else passes,

But there is one edge case:

```python
f"{
  x
}"
```

is valid syntax, but

```python
f"{
  x:d
}"
```

is not valid syntax in 3.12. Is this intentionally so? And if yes, where do I write the logic to not do RHS split if there's a format specifier?

tusharsadhwani:
Secondly, what's your opinion on RHS splitting inside f-strings?

```diff
-msg = (
-    lambda x: (
-        f"this is a very very very long lambda value {x} that doesn't fit on a single"
-        " line"
-    )
-)
+msg = lambda x: f"this is a very very very long lambda value {
+    x
+} that doesn't fit on a single line"
```

Is this the expected behaviour now?

tusharsadhwani:
I've hit an edge case that I don't know where to fix:

```python
x = f"""foo {
    2 + 2
}bar"""
```

So when `--preview` flag is passed, it seems to be working fine:
```console
$ black /tmp/foo.py --preview --diff       
All done! ✨ 🍰 ✨

$ black /tmp/foo.py --preview --target-version=py312 --diff
--- /tmp/foo.py 2023-10-08 12:24:32.356259+00:00
+++ /tmp/foo.py 2023-10-08 12:27:38.957618+00:00
@@ -1,4 +1,2 @@
 # So this keeps appending shit below the brackets, but only when preview is false
-x = f"""foo {
-    2 + 2
-}bar"""
+x = f"""foo {2 + 2}bar"""
would reformat /tmp/foo.py

All done! ✨ 🍰 ✨
1 file would be reformatted.
```

But without preview, it fails because it adds these newlines (running with `--fast` to show diff):

```console
$ black /tmp/foo.py --fast --diff                          
--- /tmp/foo.py 2023-10-08 12:24:32.356259+00:00
+++ /tmp/foo.py 2023-10-08 12:29:03.398299+00:00
@@ -1,4 +1,6 @@
 # So this keeps appending shit below the brackets, but only when preview is false
 x = f"""foo {
     2 + 2
+
+
 }bar"""
would reformat /tmp/foo.py

All done! ✨ 🍰 ✨
1 file would be reformatted.
```

With `--target-version=py312` it still works:

```console
$ black /tmp/foo.py --target-version=py312 --diff          
--- /tmp/foo.py 2023-10-08 12:24:32.356259+00:00
+++ /tmp/foo.py 2023-10-08 12:28:48.807668+00:00
@@ -1,4 +1,2 @@
 # So this keeps appending shit below the brackets, but only when preview is false
-x = f"""foo {
-    2 + 2
-}bar"""
+x = f"""foo {2 + 2}bar"""
would reformat /tmp/foo.py

All done! ✨ 🍰 ✨
1 file would be reformatted.
```

fredgido:
in this branch also didn't like it did this

```
info_text = (
    f"{a_count} a,"
    f"{b_count} b,"
    f"{c_count} c,"
    f"{d_count} d,"
    f"{e_count} e,"
    f"{f_count} f,"
    f"{g_count} g,"
    f"{h_count} h,"
)
```
becames

```
info_text = f"{a_count} a, " f"{b_count} b, " f"{c_count} c," f"{d_count} d," f"{e_count} e," f"{f_count} f," f"{
    g_count
} g," f"{h_count} h,"
```

tusharsadhwani:
@fredgido noted, i'll handle that case.

tusharsadhwani:
@ambv Small stuff and de-duplicating tokenizer code is left, can you do an initial review?

I've added a test file called `pep_701.py`.

There are 7 failing tests in `test_format.py`, and all of them seem like they due to line wrapping inside long f strings.

I'm unable to wrap my head around the right hand split code, and I need help with figuring out the logic for these two things:

- Not formatting f-string braces if `Feature.FSTRING_PARSING` is not available.
- Not formatting f-string braces if the brace ends in a `=`, eg. `f'{x    =    }'` shouldn't be formatted otherwise runtime behaviour changes.

cc @JelleZijlstra

ambv:
AFK right now but I'll do a review tomorrow. Thanks for working on this!

tusharsadhwani:
Here's one way that I thought to prevent fstring formatting in python 3.11 and below:

```diff
diff --git a/src/black/__init__.py b/src/black/__init__.py
index 7cb1e3b..38a66f9 100644
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1094,13 +1094,16 @@ def _format_str_once(src_contents: str, *, mode: Mode) -> str:
         future_imports = get_future_imports(src_node)
         versions = detect_target_versions(src_node, future_imports=future_imports)
 
-    context_manager_features = {
+    line_generator_features = {
         feature
-        for feature in {Feature.PARENTHESIZED_CONTEXT_MANAGERS}
+        for feature in {
+            Feature.PARENTHESIZED_CONTEXT_MANAGERS,
+            Feature.FSTRING_PARSING,
+        }
         if supports_feature(versions, feature)
     }
     normalize_fmt_off(src_node)
-    lines = LineGenerator(mode=mode, features=context_manager_features)
+    lines = LineGenerator(mode=mode, features=line_generator_features)
     elt = EmptyLineTracker(mode=mode)
     split_line_features = {
         feature
diff --git a/src/black/linegen.py b/src/black/linegen.py
index 5f346a2..b666245 100644
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -501,6 +501,13 @@ class LineGenerator(Visitor[Line]):
         yield from self.visit_default(leaf)
 
     def visit_fstring(self, node: Node) -> Iterator[Line]:
+        if Feature.FSTRING_PARSING not in self.features:
+            string = str(node)
+            string_leaf = Leaf(token.STRING, string)
+            node.replace(string_leaf)
+            yield from self.visit_default(string_leaf)
+            return
+
         fstring_start = node.children[0]
         fstring_end = node.children[-1]
         assert isinstance(fstring_start, Leaf)
```

But this caused some tests to fail, mostly around empty lines not getting added/removed?

tusharsadhwani:
@ambv small nudge

src/blib2to3/pgen2/tokenize.py
```diff
@@ -418,6 +467,15 @@ def untokenize(iterable: Iterable[TokenInfo]) -> str:
     return ut.untokenize(iterable)
 
 
+def is_fstring_start(token: str) -> bool:
+    # TODO: builtins.any is shadowed :(
```
JelleZijlstra, line 471: Would suggest adding `import builtins` and using `builtins.any`.

JelleZijlstra (commented):
Thanks! I haven't reviewed this in detail, but it looks promising. Two points:

- This PR also makes us format code inside f-strings. That's a good thing, but we should do it only in the preview style, as otherwise we'd break our stability guarantee.
- I don't like that the style currently puts line breaks inside the f-string placeholders:

```
% black --line-length=10 -c 'f"aaaaa{xxxx}aaa"'
f"aaaaa{
    xxxx
}aaa"
```
That looks weird and it's not how a human would split the string.

tusharsadhwani:
@JelleZijlstra fair point about not splitting inside f strings! but I wasn't able to wrap my head around the right hand split logic enough to try out stuff with it.

Are there any conditions where you'd want to split fstrings on multiple lines? what if the line is very long?

JelleZijlstra:
Possibly if the code inside the placeholder is very complex? It should certainly be uncommon.

It's fine to leave this PR focused purely on parsing PEP 701 syntax and leave decisions on how exactly to format them to another PR, though I recognize it may not be easy to separate those things out.

tusharsadhwani:
@ambv I'll be at PyCon TH next month, maybe we could work on this then?

tusharsadhwani:
@ambv bumping this, we had to take the conversation forward.

src/black/strings.py
```diff
@@ -217,6 +216,7 @@ def normalize_string_quotes(s: str) -> str:
             s = f"{prefix}{orig_quote}{body}{orig_quote}"
         new_body = sub_twice(escaped_orig_quote, rf"\1\2{orig_quote}", new_body)
         new_body = sub_twice(unescaped_new_quote, rf"\1\\{new_quote}", new_body)
+    # TODO: can probably be removed
```
JelleZijlstra, line 219: Probably only needed in pre-3.12 mode.

src/blib2to3/pgen2/driver.py
```diff
@@ -167,7 +167,9 @@ def parse_tokens(self, tokens: Iterable[GoodTokenInfo], debug: bool = False) ->
             if type in {token.INDENT, token.DEDENT}:
                 prefix = _prefix
             lineno, column = end
-            if value.endswith("\n"):
+            # FSTRING_MIDDLE is the only character that can end with a newline, and
```
JelleZijlstra, line 170:
```suggestion
            # FSTRING_MIDDLE is the only token that can end with a newline, and
```

src/blib2to3/pgen2/tokenize.py
JelleZijlstra: ^

tests/util.py
```diff
@@ -237,7 +237,7 @@ def get_flags_parser() -> argparse.ArgumentParser:
         "--target-version",
         action="append",
         type=lambda val: TargetVersion[val.upper()],
-        default=(),
+        default=[],
```
JelleZijlstra, line 240: Seems unnecessary?

JelleZijlstra (commented):
I think this is getting quite close now. Thanks for your patience. @ambv evidently doesn't have time for a further review, which is fine.

I have a couple of minor pieces of feedback. I will also want to do some additional testing to ensure this works correctly, and then it should be ready to merge.

tests/util.py
tusharsadhwani: this was causing some mypy issue locally. I'll check once again.

JelleZijlstra:
I pushed a couple of small lint fixes. However, there is at least one crash that appears when formatting Black itself:

```
% black --unstable -c '''raise ValueError(
                "xxxxxxxxxxxIncorrect --line-ranges format, expect 'START-END', found"
                f" {lines_str!r}"
            )'''
raise ValueError(
                "xxxxxxxxxxxIncorrect --line-ranges format, expect START-END, found"
                f" {lines_str!r}"
            )
error: cannot format <string>: INTERNAL ERROR: Black produced invalid code: unterminated string literal (detected at line 2) (<unknown>, line 2). Please report a bug on https://github.com/psf/black/issues.  This invalid output might be helpful: /var/folders/hh/hdb5qd2x3v51v6l0gknpfjfr0000gp/T/blk_2on9v_hn.log
```

dhruvmanila:
Hi 👋, we (Ruff) have moved ahead with f-string formatting with minimal style changes: https://github.com/astral-sh/ruff/pull/9642

Relevant discussion: https://github.com/astral-sh/ruff/discussions/9785

The style changes are kept to the minimal. Regarding breaking expressions across multiple lines, we've decided to use the heuristic to break only if at least one of the expression is multi-line i.e., the control of having multi-line expression or not is in the hands of the programmer. You can find more details in the PR description. I hope the maintainers can chime in on the style discussion if possible :)

tusharsadhwani:
@JelleZijlstra I'm unable to reproduce that case on this branch as of now:

```console
$ black --unstable -c '''raise ValueError(
                "xxxxxxxxxxxIncorrect --line-ranges format, expect 'START-END', found"
                f" {lines_str!r}"
            )'''
raise ValueError(
    "xxxxxxxxxxxIncorrect --line-ranges format, expect START-END, found" f" {
        lines_str!r
    }"
)
```

tusharsadhwani:
The tests are passing.

The one case that's broken right now:

```python
f"{x:{y}d}"
```

This should have produced a `FSTRING_MIDDLE` for `d` at the end there, but right now it produces a `NAME`. It's due to a problem in how format specifier state is handled right now.

I'll work on that tomorrow, and hopefully that will fix the fuzzing.

tusharsadhwani:
@JelleZijlstra there's a couple other TODOs in `pep_701.py` that I'm working on, but they're fairly edge cases. Please test the code and see if you can find any other problems?

src/blib2to3/pgen2/driver.py
JelleZijlstra: ^ This suggestion still seems valid

JelleZijlstra (commented):
Thanks for all your work here!

I ran this PR branch on the mypy codebase with `-t py312` and noticed some issues:

```
         if missing_error_codes:
             raise ValueError(
-                f"Some error codes are not documented: {', '.join(sorted(missing_error_codes))}"
+                f"Some error codes are not documented: {", ".join(sorted(missing_error_codes))}"
             )
 
```

We shouldn't change the quote style here.

```
     assert match is not None, f"Couldn't identify origin's owner: {output!r}"
-    assert (
-        match.group("repo").removesuffix(".git") == "mypy"
-    ), f'Unexpected repo: {match.group("repo")!r}'
+    assert match.group("repo").removesuffix(".git") == "mypy", f"Unexpected repo: {
+        match.group("repo")!r
+    }"
     return match.group("owner")
```

This is I suppose what you mentioned about the styling changes only applying in py312+ mode. But I don't want Black's style output to change here when people upgrade to py312+, and I generally don't think we should put line breaks inside of f-strings in almost any circumstances.

tusharsadhwani:
@JelleZijlstra So for now I've commented out the part where we try to format f-string children. Basically the f-string formatting should be idential for previous and current black, but 312 features (such as multiline children, comments, and double quotes inside double quotes) should be parse-able. f-strings will be left untouched.

src/black/strings.py
tusharsadhwani: removing it seems to fail tests, so i'm keeping the code.

JelleZijlstra:
I found a crash:

```
f'\N{GREEK CAPITAL LETTER DELTA}'
```

```
error: cannot format /Users/jelle/py/black/tests/nuni.py: Cannot parse: 1:11: f'\N{GREEK CAPITAL LETTER DELTA}'
```

But this is legal syntax; it evaluates the same as if there was no f:

```
>>> '\N{GREEK CAPITAL LETTER DELTA}'
'Δ'
>>> f'\N{GREEK CAPITAL LETTER DELTA}'
'Δ'
```

It also fails on:

```
fr'\{{\}}'
```

```
error: cannot format /Users/jelle/py/black/tests/nuni.py: Cannot parse: 1:6: fr'\{{\}}'
```

Found by running Black with this PR applied on `Lib/test/test_fstring.py` in CPython. These parse fine with current Black, so it's important that we keep them supported.

tusharsadhwani:
@JelleZijlstra Awesome, thanks! I'll fix these in a couple hours.
In hindsight I could have copied my test cases from cpython, oh well.

JelleZijlstra:
Thanks! Unfortunately `fr'\{{\}}'` still crashes.

tusharsadhwani:
Yeah, haven't had sufficient time to fix that one, tomorrow hopefully!

tusharsadhwani:
`'foo \{2+2}'` is also broken, the string tokenizing regex needs a revisit.

```console
$ python -m blib2to3.pgen2.tokenize
f'foo \{2+2}'
1,0-1,2:        FSTRING_START   "f'"
1,2-1,12:       FSTRING_MIDDLE  'foo \\{2+2}'
1,12-1,13:      FSTRING_END     "'"
1,13-1,14:      NEWLINE '\n'
```

tusharsadhwani:
@JelleZijlstra everything in `test_fstring.py` parses now, and all the tests in that file pass before and after formatting. I think that should cover everything.

JelleZijlstra:
Thank you! Testing on our internal codebase found another crash:

```
f"""
    WITH {f'''
    {1}_cte AS ()'''}
"""
```
produces
```
error: cannot format /Users/jelle/py/black/nested.py: Cannot parse: 3:7:     {1}_cte AS ()'''}
```

Interestingly this happens only if there is a newline in the inner f-string.

This is Python 3.9 code, so valid before and after PEP 701.

JelleZijlstra:
Other than the crashes though, this looks good and I'm hopeful we can merge it soon.

We should add a preview style feature to enable formatting inside f-strings too, but that can wait for another PR.

tusharsadhwani:
Cool, let me take a look.

tusharsadhwani:
@JelleZijlstra edge case taken care of.

JelleZijlstra:
Thanks! I ran the branch on our company repo and on a venv full of interesting installed packages and found no more bugs. I'll read over the diff one more time and hopefully then we'll be ready.

tests/data/cases/pep_701.py
```diff
@@ -0,0 +1,216 @@
...
+    2+2:d
+}"
+f"This is a really long string, but just make sure that you reflow fstrings correctly {2+2:d}"
+
+f"{2+2=}"
+f"{2+2    =    }"
+f"{     2      +     2    =    }"
+
+# TODO:
+# f"""foo {
```
JelleZijlstra, line 87: Is it easy to fix this TODO in Black? I checked and this syntax works in 3.11 and with current Black, so it would be a regression if we start failing on it.

tests/data/cases/pep_701.py
```diff
@@ -0,0 +1,216 @@
...
+
+f"{x:{y}d}"
+
+x = f"a{2+2:=^{x}}b"
+x = f"a{2+2:=^{foo(x+y**2):something else}}b"
+x = f"a{2+2:=^{foo(x+y**2):something else}one more}b"
+f'{(abc:=10)}'
+
+f"This is a really long string, but just make sure that you reflow fstrings {
+    2+2:d
```
JelleZijlstra, line 78: This is a SyntaxError in 3.12.0 but not 3.12.1, I suppose due to python/cpython#112059. Ran into this because I ran the test suite on 3.12.0 and it failed.

No change requested, but it seems possible this will cause trouble again in the future.

tests/data/cases/pep_701.py
```diff
@@ -0,0 +1,216 @@
...
+got {escape}"
+
+x = f'\N{GREEK CAPITAL LETTER DELTA} \N{SNOWMAN} {x}'
+fr'\{{\}}'
+
+f"""
+    WITH {f'''
+    {1}_cte AS ()'''}
+"""
+
```
JelleZijlstra, line 108: Suggest adding this test case:

```
f"{
X
!r
}"
```

It works already, but good to cover it in the test suite.

tests/data/cases/pep_701.py
tusharsadhwani: This is fixed now.

tusharsadhwani:
#4321 somehow broke changes in this PR, even though visit_STRING and visit_NUMBER were not touched in this PR. Looking into it cc @hauntsaninja

tusharsadhwani:
Is it OK to just revert that one?

JelleZijlstra (approved):

JelleZijlstra:
Congratulations, and thank you!

Since this fixes such a commonly encountered issue, I'll make a release soon so people can start using it.

ichard26:
Congratulations y'all, good work! 🖤

Issue #3820, opened 2023-07-29
Black trims whitespace around walrus operators in slices


charliermarsh:
**Describe the bug**

Black trims whitespace around `:=` operators within slices. I'm not sure if this is a feature or a bug (if it's intended, feel free to close of course).

The behavior is similar to Black's treatment of `:` operators in slices as described [here](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#slices), but I can't tell if the same treatment is intentional for `:=` operators.

**To Reproduce**

For example, given this code:

```python
x[y := 1]
```

The formatted code is:

```python
x[y:=1]
```

Similar to treatment of the `:` operator, Black does _not_ reformat these:

```python
x[(y := 1)]
x[y := f()]
```

Here's a [playground link](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ACPAFhdAD2IimZxl1N_WmyxXI25ydt_l-wQhA0VY58ypdO-f0EIorSuhckQkLRD1Y6D1pZQmLM0fnhDVbMMexZjk98IbWC5ckOcJ6Vc8k6der8aXC3kIo2BGf_R6H4AViGa7H6QU0kAAXSQAQAAAH_bXDqxxGf7AgAAAAAEWVo=) to demonstrate the above cases.

JelleZijlstra:
That feels like a bug, we should add spaces here. PR welcome!

thomasjo:
> Hey @JelleZijlstra I would like to work on this issue.

@Utkarshn10 it's already being worked on: https://github.com/psf/black/pull/3823

Utkarshn10:
Thanks for letting me know :)

> > Hey @JelleZijlstra I would like to work on this issue.
> 
> @Utkarshn10 it's already being worked on: #3823

PR #3823, opened 2023-07-29
Avoid removing whitespace for walrus operators within subscripts

charliermarsh:
### Description

At present, Black removes spaces around walrus operators within subscripts. For example, this is stable style:

```python
x[y:=1]
```

While rare in practice, it appears that this behavior is unintentional, and falls out from Black's removal of spaces around colons in slices (e.g., `x[y:z]`).

This PR removes that behavior, instead requiring spaces around walrus operators in such cases.

Closes https://github.com/psf/black/issues/3820.

### Checklist - did you ...


charliermarsh:
A related question -- Black's stable style uses:

```python
a[:(x := 13)]
```

Perhaps this should instead be:

```python
a[: (x := 13)]
```

(But I don't feel at all strongly here -- more asking if the stable style is intentional.)

JelleZijlstra:
> A related question -- Black's stable style uses:
> 
> ```python
> a[:(x := 13)]
> ```
> 
> Perhaps this should instead be:
> 
> ```python
> a[: (x := 13)]
> ```
> 
> (But I don't feel at all strongly here -- more asking if the stable style is intentional.)

I think the latter would be more consistent, since we add spaces to e.g. `'x[:a+4]'` and the walrus should generally be treated like any other operator.

JelleZijlstra (changes requested):
This looks good, thanks! However, the change should be made only in the preview style: we'll promote it to stable in January.

JelleZijlstra (approved):

hauntsaninja (approved):
Thanks, this looks good! I made some minor changes to preview handling, tests and reformatted the source

charliermarsh:
Thanks!

JelleZijlstra:
Tests are failing on 3.8/3.9 because of a change noted in https://docs.python.org/3.10/whatsnew/3.10.html#other-language-changes: "Assignment expressions can now be used unparenthesized within set literals and set comprehensions, as well as in sequence indexes (but not slices)."

We should split out the offending tests into a separate test file that gets run only on 3.10+.

PR #3844, opened 2023-08-19
Remove tox pin

cdce8p:
### Description
A more permanent fix for tox issue. This reverts #3843

Starting with `4.9`, tox will throw an error for unspecified environment names. The run command was missing the python version. _Doing the fix inline might not be as nice as a separate variable but works for all three OS without issues._

https://tox.wiki/en/latest/changelog.html#v4-9-0-2023-08-16
https://github.com/tox-dev/tox/issues/2858


JelleZijlstra:
Thanks!

PR #3858, opened 2023-08-30
Move coverage configurations to `pyproject.toml`

jnhyperion:
File `.coveragerc` is no longer needed since library `coverage` already support configurations in `pyproject.toml`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,7 +62,7 @@ jobs:
         if:
           github.repository == 'psf/black' && matrix.os == 'ubuntu-latest' &&
           !startsWith(matrix.python-version, 'pypy')
-        uses: AndreMiras/coveralls-python-action@v20201129
+        uses: AndreMiras/coveralls-python-action@develop
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           parallel: true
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         if:
           github.repository == 'psf/black' && matrix.os == 'ubuntu-latest' &&
           !startsWith(matrix.python-version, 'pypy')
-        uses: AndreMiras/coveralls-python-action@v20201129
+        uses: AndreMiras/coveralls-python-action@develop
```
JelleZijlstra, line 65: We should pin to a specific version for security.

cooperlees: Yeah, I'd also prefer to depend on dependabot to update this ... Is there a feature you needed in develop tag / branch?

hauntsaninja: Tags are mutable, if we care about security don't we need a hash?

jnhyperion: Make sense, updated to commit id.

cooperlees (approved):
Sweet - Will have to update all my other projects ...

cooperlees:
Do we want a change log here? I don't feel we need one ...

ichard26:
Thanks for the PR @jnhyperion, congrats on your first PR here! :tada:

PR #3870, opened 2023-09-10
mypyc builds on PRs, skip mypyc wheels for 3.12

hauntsaninja:
My mess around PR. Skip compilation on 3.12, skip compilation of black.cache


hauntsaninja:
grant me the serenity to accept the things I cannot change,
the courage to change the things I can,
and the wisdom to know the difference

PR #3867, opened 2023-09-09
Run wheel build on PRs

JelleZijlstra:
So we don't just test this workflow on releases. Maybe every PR is too much, but I want it at least on PRs that prepare releases.

PR #3875, opened 2023-09-10
Re-export black.Mode

JelleZijlstra:
And test that certain common external use cases (taken from https://grep.app/search?q=black.Mode&case=true) keep working.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -67,7 +67,7 @@ from black.mode import (
     FUTURE_FLAG_TO_FEATURE,
     VERSION_TO_FEATURES,
     Feature,
-    Mode,
+    Mode as Mode,  # re-exported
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Motivation is that otherwise mypyc on 3.12 (for unclear reasons) doesn't allow you to use `black.Mode`, as discussed in https://github.com/psf/black/pull/3867#issuecomment-1712919666

hauntsaninja (approved):

cdce8p:
I'm still seeing the error for Python 3.12, at least with cibuildwheel, which still uses `3.12.0rc0`. Can't reproduce it locally with `3.12.0rc1` though 🤔

```
AttributeError: module 'black' has no attribute 'Mode'. Did you mean: 'mode'?
```

PR #3877, opened 2023-09-10
Pickle raw tuples in FileData cache

JelleZijlstra:


JelleZijlstra:
Discussion and perf numbers in https://github.com/psf/black/pull/3867#issuecomment-1712954290.

I also tried pickling and pickling raw tuples is ~2x as fast as pickling NamedTuples for me:

```
In [88]: %timeit pickle.dumps([(*FileData(i, "x", 1.0),) for i in range(100)])
63.7 µs ± 5.33 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [89]: %timeit pickle.dumps([FileData(i, "x", 1.0) for i in range(100)])
124 µs ± 3.71 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

These numbers are without mypyc compilation so they may not completely reflect reality, but if anything I'd expect mypyc to magnify the difference (because the raw tuple version does more work in user code that would be sped up by mypyc, and the NamedTuple version does more work inside `pickle` that mypyc can't help with).

cdce8p (approved):

hauntsaninja (approved):
(can also use pickle protocol 5 if we want to now)

PR #3881, opened 2023-09-11
mypyc build improvements

### LINKED ISSUE #3879: Run wheel build on fewer PRs?
As of #3870 we now run the full mypyc wheel build workflow on all PRs. This should prevent a repeat of the 23.9.0 release, which didn't come with mypyc wheels because after the release was cut we discovered that the mypyc build was broken (#3865).

It's very important to me that we can test the wheel build *before* the release, as is happening right now on #3878 for the upcoming release 23.9.1. That gives me confidence to cut a release knowing that the build will actually succeed.

Some of the issues with the current wheel build were caused by #3821. Ideally, we would have noticed these issues before that PR was merged, not when we tried to cut a release that included it.

However, the full wheel build is pretty slow and it may not be all that useful on most PRs, so running the full workflow may end up slowing things down too much. Here's a few non-exclusive ideas that could help:

- We could trigger the workflow only for PRs with a particular label. We would add that label to release PRs, and also to bigger changes like #3821 to confirm they don't break mypyc.
- Running only the Linux wheel build (which tends to be the fastest) is often good enough, as most mypyc issues don't tend to be OS-specific. So maybe for low-risk PRs we could just run the Linux workflow.
- We could run the workflow at scheduled times (e.g. once a week). Not sure how much that would help though, as the wheel build should ideally depend just on what's in the repo (e.g., we pin the mypy version).

Happy to hear any other ideas too.

hauntsaninja:
- Build in separate jobs. This makes it clearer if e.g. a single Python version is failing. It also potentially gets you more parallelism.
- Build everything on push to master.
- Only build Linux 3.8 and 3.11 wheels on PRs.

Fixes #3879

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -1,9 +1,12 @@
-name: Build wheels and publish to PyPI
+name: Build and publish
 
 on:
   release:
     types: [published]
   pull_request:
+  push:
+    branches:
+      - main
 
 permissions:
   contents: read
@@ -12,6 +15,7 @@ jobs:
   main:
     name: sdist + pure wheel
     runs-on: ubuntu-latest
+    if: github.event_name == 'release'
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


hauntsaninja:
Oh we also now don't build universal wheels on macOS. No one really uses these, we dropped them in mypy a while ago and never looked back.

JelleZijlstra (approved):

Issue #3887, opened 2023-09-13
E704 and concise formatting for dummy implementations


hauntsaninja:
This preview style change https://github.com/psf/black/pull/3796 violates E704 as implemented by flake8 / pycodestyle.

Note that the change itself is PEP 8 compliant. E704 is disabled by default in pycodestyle for this reason. ruff intentionally does not implement E704, instead explicitly deferring to Black on this question.

At a minimum we should update https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#flake8 , but figured I'd open an issue to discuss.

cooperlees:
I'm not a fan of this change, but can live with it. I do agree with the removing of the new lines after a class definition here, but I don't see the benefit with overloads stated in the issue within these Abstract Base classes etc.

This has made some projects I maintain require a [change to line number set in tests for flake8-bugbear unittests](https://github.com/PyCQA/flake8-bugbear/pull/409) and have set the formatting off for a chunk of [bandersnatch code](https://github.com/pypa/bandersnatch/blob/main/src/bandersnatch/storage.py#L35-L53) as I don't find it easier to read. Saving vertical space has never been a goal of black.

I tend to agree with flake8's E704 here. I like not to put multiple statements on a line. But if we stick with this we will be forcing people to have to add exempting E704 in their flake8 configs on projects. This will cause CI to miss other multiple statements per line on black formatted projects if this behavior is maintained.

hauntsaninja:
I'm a little confused by the last paragraph. E704 is something that is typically completely obviated by an autoformatter (which is e.g. why ruff doesn't implement the check). So the only "other multiple statements per line" I could see CI missing are due to disabling autoformatting (or due to bugs in Black). If someone wants multiple statements on a line between `# fmt: off` and `# fmt: on`, but still wants flake8 to warn them about that, that feels a little adverse.

I don't expect formatting changes to be uncontroversial. But I was hopeful, since in addition to my personal preference, https://github.com/psf/black/issues/1797 is fairly popular, PEP 8 itself uses this kind of formatting in a couple places, and pycodestyle also disables E704 by default.

I'll update the docs, so that main is at least self-consistent. Hopefully `--preview` users give us some more feedback and we get a better sense of whether this change should be modified or rolled back.

JelleZijlstra:
[...]

PR #3888, opened 2023-09-13
Document disabling E704

hauntsaninja:
Linking #3887

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,6 @@
 [flake8]
 # B905 should be enabled when we drop support for 3.9
-ignore = E203, E266, E501, W503, B905, B907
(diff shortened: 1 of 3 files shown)


cooperlees (changes requested):
We'll should probably edit our own `.flake8` config (or delete it + move to pyproject.yaml which I think is supported now)

hauntsaninja:
Thanks, updated! As far as I can tell, flake8 does not support pyproject.toml, e.g. see https://github.com/PyCQA/flake8/pull/1867 from three days ago

cooperlees (approved):
I guess time to update my many projects with this too and live with this formatting change.

Issue #3897, opened 2023-09-20
Add Python 3.11 as part of the `-t` configuration


iamibi:
**Is this related to a problem? Please describe.**

The documentation currently doesn't share that python 3.11 is supported as part of the `-t, --target-version` flag.

**Describe the solution you'd like**

Add the relevant changes to display `py311` as the supported version.

**Describe alternatives you've considered**

**Additional context**

PR #3898, opened 2023-09-20
added the py311 to target-version config

iamibi:
### Description

The documentation currently doesn't share that python 3.11 is supported as part of the `-t, --target-version` flag. This PR updates the documentation to support that.

### Checklist - did you ...

Issue: https://github.com/psf/black/issues/3897

### DIFF
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -52,18 +52,19 @@ See also [the style documentation](labels/line-length).
 
 #### `-t`, `--target-version`
 
-Python versions that should be supported by Black's output. You should include all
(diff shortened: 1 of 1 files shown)


JelleZijlstra (changes requested):
Thanks! Let's remove py37 from the list while we're at it.

Also, let's add a sentence like "You can run `black --help` and look for the `--target-version` option to  see the full list of supported versions." so it's less likely we have to update this again.

iamibi:
> Thanks! Let's remove py37 from the list while we're at it.
> 
> Also, let's add a sentence like "You can run `black --help` and look for the `--target-version` option to see the full list of supported versions." so it's less likely we have to update this again.

sure.

JelleZijlstra (approved):

Issue #2316, opened 2021-06-08
PEP 604 and line breaks


srittau:
**Describe the style change**

Long PEP 604 annotations (`X | Y`) in function definitions (and probably other contexts) are currently broken into multiple lines without extra indentation. This makes it hard to recognize the arguments in the signature. My suggestion would be to treat those annotations similar to regular bitwise or operators inside type annotations: Indent followup lines and possibly use parentheses.

**Examples in the current _Black_ style**

```python
def foo(
    i: int,
    x: Loooooooooooooooooooooooong
    | Looooooooooooooooong
    | Looooooooooooooooooooong
    | Looooooong,
    *,
    s: str
) -> None:
    pass
```

A practical example (with shorter line width):

```python
def create_medical_record_entries(
    context: RequestContext,
    db_entries: Iterable[CharlyMedicalRecordEntry]
    | Iterable[CharlyLabRecordEntry],
    *,
    ignore_signs: bool = True
) -> list[MedicalRecordEntry]:
    ...
```

**Desired style (without extra arguments)**

```python
def foo(
    x: (
        Loooooooooooooooooooooooong
        | Looooooooooooooooong
        | Looooooooooooooooooooong
        | Looooooong
    ),
) -> None:
    pass
```

Or maybe:

```python
def foo(
    x:
        Loooooooooooooooooooooooong
        | Looooooooooooooooong
        | Looooooooooooooooooooong
        | Looooooong,
) -> None:
    pass
```

Or:

```python
def foo(
    x: Loooooooooooooooooooooooong
        | Looooooooooooooooong
        | Looooooooooooooooooooong
        | Looooooong,
) -> None:
    pass
```

felix-hilden:
To me 1 and 3 look good, 2 is a bit weird. But I agree variable annotations should work similarly.

srittau:
Option 3 looks clearer to me in function annotations, but I don't know if there should be separate styles for function arguments and other annotations.

Zeta611:
Current black line break is awkward with short type annotation with a long default value as well:
```python
@app.get("/path/")
async def foo(
    q: str
    | None = Query(None, title="Some long title", description="Some long description")
):
    pass
```

IMHO the following is better:
```python
@app.get("/path/")
async def foo(
    q: str | None = Query(
        None, title="Some long title", description="Some long description"
    )
):
    pass
```

It is how it is done with out the type annotation.

MarcDufresne:
[...]

PR #3899, opened 2023-09-20
fix indentation of line breaks in long type hints by adding parens

jakkdl:
### Description
The simple way of resolving #2316
I would personally prefer to resolve it by formatting it without added parentheses (i.e. option 3 in #2316) to save two unnecessary lines, but that will likely require a bunch of messing around in `_maybe_split_omitting_optional_parens` and could also impact some other cases - so I think I'll hold off on that one for the moment.

I had to extend `maybe_make_parens_invisible_in_atom` to also allow expressions, I was quite afraid this would have side effects elsewhere in the code - but it seems to be fine? I could be defensive and add a parameter that enables checking `syms.expr`, and only set it when called from `visit_tname`.

There's minor changes to two existing test cases, but they both seem fine to me.

### Checklist - did you ...
- [x] Add an entry in `CHANGES.md` if necessary? [TODO]
- [x] Add new / update outdated documentation? [I don't think this needs to be done]

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@
 
 ### Preview style
 
+- Long type hints are now wrapped in parentheses and properly indented when split across
+  multiple lines (#3899)
+
 <!-- Changes that affect Black's preview style -->
 
 ### Configuration
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -397,6 +397,23 @@ class LineGenerator(Visitor[Line]):
             node.insert_child(index, Node(syms.atom, [lpar, operand, rpar]))
         yield from self.visit_default(node)
 
+    def visit_tname(self, node: Node) -> Iterator[Line]:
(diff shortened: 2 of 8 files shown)


JelleZijlstra:
Thanks. This looks generally good, but we'll want to make the change only in the preview style for now. Look for existing code like `if Preview.parenthesize_conditional_expressions in self.mode:` for examples.

jakkdl:
> Thanks. This looks generally good, but we'll want to make the change only in the preview style for now. Look for existing code like `if Preview.parenthesize_conditional_expressions in self.mode:` for examples.

Thanks, fixed!

JelleZijlstra (approved):

JelleZijlstra:
Looking at the diff-shades output, all changes seem like improvements: https://github.com/psf/black/actions/runs/6275527706/job/17044773640#step:10:1

PR #3902, opened 2023-09-25
add support for printing the diff of AST trees when running tests

jakkdl:
### Description
As I've been working more on Black, I found the full printing of AST trees to be cumbersome and *insanely long*, but it regardless often containing necessary info, so I ended up modifying the value of that variable a lot and/or temporarily splitting tests into shorter files. But realized that diff-printing would solve that issue, and while not as simple as I'd initially hoped it wasn't too tricky to work out and is turning out very useful. 

### Checklist - did you ...

      I don't think changes to test QoL warrants entries in `CHANGES.md`
      Given the precedence of `test_assertFormatEqual` I should probably write one.
      `"SKIP_AST_PRINT"` isn't documented anywhere, but it and this should probably be in `CONTRIBUTING.md` and/or the contributing documentation on RTD - it took me a while to find myself.

### DIFF
--- a/docs/contributing/the_basics.md
+++ b/docs/contributing/the_basics.md
@@ -37,6 +37,44 @@ the root of the black repo:
 (.venv)$ tox -e run_self
 ```
 
+### Development
+
+Further examples of invoking the tests
+
+```console
+# Run all of the above mentioned, in parallel
+(.venv)$ tox --parallel=auto
+
+# Run tests on a specific python version
+(.venv)$ tox -e py39
+
+# pass arguments to pytest
+(.venv)$ tox -e py -- --no-cov
+
+# print full tree diff, see documentation below
+(.venv)$ tox -e py -- --print-full-tree
+
+# disable diff printing, see documentation below
+(.venv)$ tox -e py -- --print-tree-diff=False
+```
+
+`Black` has two pytest command-line options for QoL while developing features that uses
+test files in `tests/data/` and are split into an input part, and an output part,
+separated by a line with`# output`. These can be passed to `pytest` through `tox`, or
+directly into pytest if not using `tox`.
+
+#### `--print-full-tree`
+
+Upon a failing test, print the full CST tree as it is after processing the input
+("actual"), and the tree that's yielded after parsing the output ("expected"). Note that
+a test can fail with different output with the same CST tree. This used to be the
+default, but now defaults to `False`.
+
+#### `--print-tree-diff`
+
+Upon a failing test, print the diff of the trees as described above. This is the
+default, to turn it off pass `--print-tree-diff=False`.
+
 ### News / Changelog Requirement
 
 `Black` has CI that will check for an entry corresponding to your PR in `CHANGES.md`. If
--- a/docs/contributing/the_basics.md
+++ b/docs/contributing/the_basics.md
@@ -58,16 +58,16 @@ Further examples of invoking the tests
 (.venv)$ tox -e py -- --print-tree-diff=False
 ```
 
-`Black` has two pytest command-line options affecting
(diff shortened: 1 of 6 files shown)


JelleZijlstra:
This is a good, idea, thanks! The full AST is indeed insanely long. A few thoughts:

- Maybe we should just remove the environment variables and make this the only option during tests? Developers who really want the full AST can always modify the code, but hidden env vars aren't a great user experience in any case.
- "AST" is really a misnomer here; lib2to3 produces a CST, not an AST.

jakkdl:
> Maybe we should just remove the environment variables and make this the only option during tests? Developers who really want the full AST can always modify the code, but hidden env vars aren't a great user experience in any case.

I just realized, this would be much easier to use if added as a CLI option in `tests/conftest.py`. That way we also get visibility in `pytest --help` for free.
I think keeping ability to print the full tree is still valuable, but we could probably default that to off, and default tree-diff-printing to on.


> "AST" is really a misnomer here; lib2to3 produces a CST, not an AST.

yeah, let's rename them at the same time.

docs/contributing/the_basics.md
```diff
@@ -37,6 +37,44 @@ the root of the black repo:
...
+
+# print full tree diff, see documentation below
+(.venv)$ tox -e py -- --print-full-tree
+
+# disable diff printing, see documentation below
+(.venv)$ tox -e py -- --print-tree-diff=False
+```
+
+`Black` has two pytest command-line options for QoL while developing features that uses
+test files in `tests/data/` and are split into an input part, and an output part,
```
JelleZijlstra, line 62:
```suggestion
`Black` has two pytest command-line options affecting
test files in `tests/data/` that are split into an input part, and an output part,
```
Avoid the potentially obscure abbreviation "QoL" and make it a bit more concise

docs/contributing/the_basics.md
```diff
@@ -37,6 +37,44 @@ the root of the black repo:
...
+separated by a line with`# output`. These can be passed to `pytest` through `tox`, or
+directly into pytest if not using `tox`.
+
+#### `--print-full-tree`
+
+Upon a failing test, print the full CST tree as it is after processing the input
+("actual"), and the tree that's yielded after parsing the output ("expected"). Note that
+a test can fail with different output with the same CST tree. This used to be the
```
JelleZijlstra, line 70:
```suggestion
Upon a failing test, print the full concrete syntax tree (CST) as it is after processing the input
("actual"), and the tree that's yielded after parsing the output ("expected"). Note that
a test can fail with different output with the same CST. This used to be the
```

docs/contributing/the_basics.md
```diff
@@ -37,6 +37,44 @@ the root of the black repo:
...
+default, but now defaults to `False`.
+
+#### `--print-tree-diff`
+
+Upon a failing test, print the diff of the trees as described above. This is the
+default, to turn it off pass `--print-tree-diff=False`.
```
JelleZijlstra, line 76:
```suggestion
default. To turn it off pass `--print-tree-diff=False`.
```

tests/conftest.py
```diff
@@ -1 +1,28 @@
...
+        "--print-full-tree",
+        action="store_true",
+        default=False,
+        help="print full CST trees on failed tests",
+    )
+    parser.addoption(
+        "--print-tree-diff",
+        action="store_true",
+        default=True,
+        help="print diff of CST trees on failed tests",
```
JelleZijlstra, line 20:
```suggestion
        help="print diff of syntax trees on failed tests",
```

JelleZijlstra (commented):
Thanks! Left some comments mostly on wording.

jakkdl:
The CI fail seem unrelated to this PR

Issue #1553, opened 2020-07-16
Misleading formatting of long parenthesized yields of tuples


JelleZijlstra:
When code of the form `(yield a, b)` is split into multiple lines, Black puts the `yield a` and `b` each  on its own line. This is misleading, because in fact this code yields the tuple `(a, b)`, while Black's output makes it look like we yield only `a`.

**Examples in the current _Black_ style**

```python
def f():
    (
        yield abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_,
        abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_,
    )
```

**Desired style**

```python
def f():
    yield (
        abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_,
        abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_abcdefg_,
    )
```

I encountered this in code using github.com/quora/asynq, which frequently contains statements of the form `a, b = yield get_a.asynq(), get_b.asynq()`.

PR #3912, opened 2023-10-02
Parenthesize singleton yield tuples

### LINKED ISSUE #3851: single element tuples in yield expressions are not parenthesized
**Describe the style change**

Unlike other expression level single element tuples, single element tuples in yield expressions (e.g. `yield 1,`) are currently not parenthesized. This is inconsistent, they should also be parenthesized.

Input:
```python
for a, in b:
    pass

c = d,


def e():
    yield f,
    return g,
```

Output:
```python
for (a,) in b:
    pass

c = (d,)


def e():
    yield f,
    return (g,)
```

**Desired style**

```python
for (a,) in b:
    pass

c = (d,)


def e():
    yield (f,) 
    return (g,)
```

**Additional context**

black 23.7.0

spagh-eddie:
### Description

Closes #3851

Please let me know if I should be going along a different path c^: 

I had originally intended to make a `visit_yield_expr = partial(v, keywords={"yield"}, parens={"yield"})` to be more similar to `return` alongside this, but that doesn't work since `yield` can be an expression but `return` can't.  https://github.com/psf/black/blob/1b08cbc63400a8b8e0fbb620b6b2a4dab35e1e7b/src/black/linegen.py#L527 

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@
 <!-- Changes that affect Black's stable style -->
 
 - Fix comments getting removed from inside parenthesized strings (#3909)
+- Parenthesize singleton tuples in `yield` expressions as they already are in `return` statements,
+  e.g. `yield 5,` ➡️ `yield (5,)` (#3912)
 
 ### Preview style
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,8 +11,8 @@
 <!-- Changes that affect Black's stable style -->
 
 - Fix comments getting removed from inside parenthesized strings (#3909)
-- Parenthesize singleton tuples in `yield` expressions as they already are in `return` statements,
-  e.g. `yield 5,` ➡️ `yield (5,)` (#3912)
+- Parenthesize singleton tuples in `yield` expressions as they already are in `return`
+  statements, e.g. `yield 5,` ➡️ `yield (5,)` (#3912)
 
 ### Preview style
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -11,6 +11,8 @@
 <!-- Changes that affect Black's stable style -->
 
 - Fix comments getting removed from inside parenthesized strings (#3909)
+- Parenthesize singleton tuples in `yield` expressions as they already are in `return`
```
JelleZijlstra, line 14: This needs to go in the preview style, not the stable style. (That's also why CI is failing.)

JelleZijlstra (commented):
I ran this diff also on our internal codebase, which very heavily uses yield because of https://github.com/quora/asynq. Most of the changes are clear improvements, but I noticed a few cases like this:

```
-        content_type = yield (
-            a.model.tribe.item.TribeItem(tribe_item_id).get_content_type.asynq()
-        )
+        content_type = yield a.model.tribe.item.TribeItem(
+            tribe_item_id
+        ).get_content_type.asynq()
```

Here the old formatting seems better.

There are also cases like this:

```
                 cant_edit = (
-                    yield a.controller.tribe.item.is_tribe_answer_and_cant_edit.asynq(
-                        self.aid, self.viewer
+                    yield (
+                        a.controller.tribe.item.is_tribe_answer_and_cant_edit.asynq(
+                            self.aid, self.viewer
+                        )
                     )
                 )
```

Where ideally we should remove the outer set of parentheses.

Linking #1553, another issue with how we manage yields; it may make sense to tackle them together.

cobaltt7:

PR #3914, opened 2023-10-02
Update output display to job summary

csalerno-asml:
### Description

Black default job summary isn't really readable, since Python code will be converted 1:1 in Markdown, leading to something like this: https://github.com/csalerno-asml/test-black-repo/actions/runs/6380983558#summary-17316480770 

With this proposal, the diff will be wrapped in a triple-quoted python box: https://github.com/csalerno-asml/test-black-repo/actions/runs/6380983558#summary-17316481072

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,8 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- The action output displayed in the job summary is now wrapped in Markdown (#3914)
+
 ### Documentation
 
 <!-- Major changes to documentation and policies. Small docs changes
--- a/action.yml
+++ b/action.yml
@@ -36,10 +36,14 @@ runs:
     - name: black
       run: |
         if [ "$RUNNER_OS" == "Windows" ]; then
-          python $GITHUB_ACTION_PATH/action/main.py | tee -a $GITHUB_STEP_SUMMARY
+          out=$(python $GITHUB_ACTION_PATH/action/main.py)
         else
-          python3 $GITHUB_ACTION_PATH/action/main.py | tee -a $GITHUB_STEP_SUMMARY
+          out=$(python3 $GITHUB_ACTION_PATH/action/main.py)
         fi
+
+        echo "\`\`\`python" >> $GITHUB_STEP_SUMMARY
+        echo "${out}" >> $GITHUB_STEP_SUMMARY
+        echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
       env:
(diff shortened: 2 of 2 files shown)


cooperlees (changes requested):
I like this idea. Makes a lot of sense.

I do wonder tho, can we break people who might be already parsing the output? If so, should we consider making a config parameter (if possible) to disable this code block new default?

csalerno-asml:
Hi @cooperlees , thanks for the feedback!

Regarding your concern, I am not really sure that somebody can/will parse the output?
I think the only use case with summaries is: an action/step write to `GITHUB_STEP_SUMMARY`, and a human read it via UI.

If a user wants to read the output of black, usually he/she will use `black` via CLI or with a pre-commit hook

cooperlees (approved):
Ok, I thought this was for human consumption only, was just checking we were not missing anything.

In that case lets add a change log entry so we show somewhere we've decided to make this change and rebase this and I'll merge.

Thankyou for your contribution!

csalerno-asml:
Hey @cooperlees , I noticed that the exit code was not handled correctly

cooperlees:
Many thanks again. This will help people see the changes to the code etc.

heiner:
I believe this change might have caused the GitHub action to fail without actionable output in some cases, e.g. in a way where the only output is:

```
Run psf/black@stable
  with:
    version: 23.10.0
    options: --check --diff
    src: .
    jupyter: false
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.18/x64
Run if [ "$RUNNER_OS" == "Windows" ]; then
  if [ "$RUNNER_OS" == "Windows" ]; then
    runner="python"
  else
    runner="python3"
  fi
  
  out=$(${runner} $GITHUB_ACTION_PATH/action/main.py)
  exit_code=$?
  
  # Display the raw output in the step
  echo "${out}"
  
  # Display the Markdown output in the job summary
  echo "\`\`\`python" >> $GITHUB_STEP_SUMMARY
  echo "${out}" >> $GITHUB_STEP_SUMMARY
  echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
  
  # Exit with the exit-code returned by Black
  exit ${exit_code}
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.18/x64
    INPUT_OPTIONS: --check --diff
    INPUT_SRC: .
    INPUT_JUPYTER: false
    INPUT_BLACK_ARGS: 
    INPUT_VERSION: 23.10.0
    pythonioencoding: utf-8
Error: Process completed with exit code 1.
```

PR #3917, opened 2023-10-03
exclude tests/data/.* from mypy

jakkdl:
### Description
I run pylsp-mypy in my dev environment, which screams loudly about all the illegal stuff being done in `tests/data/.*`. So excluding those files in the mypy config would be a nice QoL for me, and probably other users as well, while making no difference to the CI.

### Checklist - did you ...
None of the below

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -19,6 +19,10 @@ no_implicit_reexport = False
 # to avoid 'em in the first place.
 warn_unreachable=True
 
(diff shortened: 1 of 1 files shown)


jakkdl:
wait no, this isn't enough as LSP (and probably mypyd) will explicitly ask for the filename which bypasses the exclude check.

Issue #3919, opened 2023-10-03
Docker builds failing on 3.12


cooperlees:
**Describe the bug**

Docker is failing to build aiohttp with python 3.12.

**To Reproduce**

Build docker with a 3.12 base image

**Expected behavior**

Docker to build

**Environment**

Black's version: [main]
OS and Python version: Debian/3.12

**Additional context**

This is due to aiohttp being used for blackd. THere are 3.12 wheels coming: https://github.com/aio-libs/aiohttp/issues/7639

PR #3927, opened 2023-10-06
Set Docker to use 3.11 for now

cooperlees:
Until we get new aiohttp wheels we need to build with 3.11. You can see an example of a fail here:

Workaround for #3919 - Will leave it open until we can move to 3.12

EDIT: Forgot to add I ran this locally on my box from the root of the black repo:
- `docker build -t black_test --network host .`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3-slim AS builder
+FROM python:3.11-slim AS builder
 
 RUN mkdir /src
 COPY . /src/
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

JelleZijlstra:
> until we can move to 3.21

I will remind you in 2033.

cooperlees:
haha - Good call - I also noticed we don't compile black in the docker container. I guess we should :) I'll open a dedicated issue.

Issue #3834, opened 2023-08-11
PEP 604 type hints in a class definition are not broken up when hitting the line length limit


mmohrhard:
I have checked with the online formatter that it is not fixed in the most up to date development branch.

**Describe the bug**

The following code has a line that is longer than 88 characters and yet is not split by black:

```
class LongNameLongName:
    pass


class LongNameLongName2:
    pass


class LongNameLongName3:
    pass


class LongNameLongName4:
    pass


class LongNameLongName5:
    pass


class LongNameLongName6:
    pass


class ClassWithTypeDefs:
    name: str
    field: LongNameLongName | LongNameLongName2 | LongNameLongName3 | LongNameLongName4 | LongNameLongName5 | LongNameLongName6


def func(
    a: LongNameLongName
    | LongNameLongName2
    | LongNameLongName3
    | LongNameLongName4
    | LongNameLongName5
    | LongNameLongName6,
):
    pass
```

As can be seen, for a function black breaks the typing statement into multiple lines but not for the class definition.

```
$ flake8 --max-line-length 88 test.py
test.py:27:89: E501 line too long (127 > 88 characters)
$ black -l 88 -t py311 test.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

**Expected behavior**

black should break the line or at least not reformat the code to a line longer than the line width limit if the line was broken up before.

**Environment**

```
Linux
$ black --version
black, 23.7.0 (compiled: no)
Python (CPython) 3.11.4
```

jakkdl:
This gets wrapped with parentheses and no longer breaks the line length after https://github.com/psf/black/pull/3899

JelleZijlstra:
Indeed

```
% black --preview -c '''class ClassWithTypeDefs:
    name: str
    field: LongNameLongName | LongNameLongName2 | LongNameLongName3 | LongNameLongName4 | LongNameLongName5 | LongNameLongName6
'''
class ClassWithTypeDefs:
    name: str
    field: (
        LongNameLongName
        | LongNameLongName2
        | LongNameLongName3
        | LongNameLongName4
        | LongNameLongName5
        | LongNameLongName6
    )
```

PR #3930, opened 2023-10-07
split long function parameter type hints without parentheses

### LINKED ISSUE #2316: PEP 604 and line breaks

jakkdl:
### Description

Resolves #2316 according to the `nextline` strategy. The poll is currently 10 to 6 of favor of this style, which certainly isn't a slam-dunk, but enough for me to implement it and we can progress the discussion after seeing impacts on existing code.

This is a pretty messy implementation, as can be seen by the sheer length and the complexity check having made me reformat my code several times to break out functions.
I don't love the existence of `_split_first_typehint` and feel like that should be handled on a second pass by `delimiter_split` or something. And likewise I ideally wouldn't need any changes in `visit_tname` after #3899 - I don't really remember why I gave up on resolving it while keeping invisible lpars.

I started out with `append_to_line` being a subfunction like in `delimiter_split` and `standalone_comment_split`, but when the complexity check for good reason wanted me to break up `_split_tname` I ended up having to add some helper functions to `Line` and `BracketTracker`, and while this is not *super* pretty it's probably better than having `append_to_line` repeated three times over in the code - and the other uses of it should probably be changed to use `_append_to_line` as well.

~~I stumbled upon #3834 just as I thought myself done, and realized that case should clearly also be handled without parentheses. I'm not entirely sure why it isn't, but will probably add that to this PR.~~ derp, it can't use the `nextline` strategy as per normal type hints since they're not inside parentheses and not `tname`s. This differentiation being one of the downsides of special-casing parameter lists.

### Checklist - did you ...

- [ ] Add new / update outdated documentation? (this is maybe noteworthy enough to be documented?)

### DIFF
--- a/tests/data/preview_py_310/pep604_union_types_line_breaks.py
+++ b/tests/data/preview_py_310/pep604_union_types_line_breaks.py
@@ -83,6 +83,54 @@ def f(
     ...
 
 
+# don't lose comments
+def f(  # 1
+        looooooooooooooooooooooooooong  # 2
+        :  # 3
+        Loooooooooooooooooooooooooooooooooooooooooooooooong  # 4
+        |  # 5
+        Loooooooooooooooooooooooooooooooooooooooooooooooong  # 6
+        =  # 7
+        3  # 8
+        ):  # 9
+    ...  # 10
+
+
+loooooooooooooooooooooooooooooong: (
+    Loooooooooooooooooooooooooooong  # aoeuaoeuaoeuaoeuaoeuaoeu
+)
+
+
+def foo(
+    loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong:
+        Loooooooooooooooooooooooooooong,
+): ...
+
+
+def foo(
+    loooooooooooooooong: Tuple[
+        Looooooooooooooooooooooooooooong, Loooooooooooooong
+    ],  # aoeuaoeuaoeu
+): ...
+
+
+def foo(a: loooooooooooooong| loooooooooooooooooooooooooooong| looooooooooong| looooooooooong = 3): ...
+def foo(c: (loooooooooooooong| loooooooooooooooooooooooooooong| looooooooooong| looooooooooong) = 4): ...
+def foo(bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb: bbbbbbbbbbbbbbbbbbbbloooooooooooooong| loooooooooooooooooooooooooooong| looooooooooong| looooooooooong = 5): ...
+def foo(not_too_long_param_name: not_too_long_typehint # but there's a reasonably long comment here
+        ): ...
+def foo(not_too_long_param_name: not_too_long_typehint # but there's a reasonably long comment here
+        | loooooooooooooooooooooooooooong| looooooooooong| looooooooooong = 5): ...
+def foo(a: tuple[int, int,] = (7, 3,)): ...
+def foo(a: (
+    # comment_inside_paren
+            loooooooooooooong
+            # comment2
+            |
+            # comment3
+            loooooooooooooong
+            )): ...
+
 # output
 # This has always worked
 z = (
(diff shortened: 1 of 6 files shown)


jakkdl:
Hmm, looking at the preview diffs I think I might always want a newline after the colon. So we get
```python

     extLst: Typed[ExtensionList, Literal[True]]
     __elements__: ClassVar[tuple[str, ...]]
     def __init__(
         self,
         custUnit: _HasTagAndGet[_ConvertibleToFloat | None] | _ConvertibleToFloat | None = None,
         builtInUnit:
             _HasTagAndGet[_DisplayUnitsLabelListBuiltInUnit]
             | _DisplayUnitsLabelListBuiltInUnit
             | Literal["none"]
             | None
             = None,
         dispUnitsLbl: DisplayUnitsLabel | None = None,
         extLst: Unused = None,
     ) -> None: ...
 
 class NumericAxis(_BaseAxis):
```
instead of https://github.com/psf/black/actions/runs/6442045741/job/17492459124?pr=3930#step:10:570

though I'm also personally starting to lean towards this maybe not being much of an improvement, if at all.

tests/data/preview_py_310/pep604_union_types_line_breaks.py
```diff
@@ -172,16 +218,100 @@ def foo(
 
 @app.get("/path/")
 async def foo(
-    q: str | None = Query(
-        None, title="Some long title", description="Some long description"
-    )
+    q: str
+        | None
```
JelleZijlstra, line 222: The old style was much better here.

tests/data/preview_py_310/pep604_union_types_line_breaks.py
```diff
@@ -172,16 +218,100 @@ def foo(
...
     pass
 
 
 def f(
-    max_jobs: int | None = Option(
-        None, help="Maximum number of jobs to launch. And some additional text."
-    ),
+    max_jobs: int
+        | None
+        = Option(
```
JelleZijlstra, line 231: And here

tests/data/preview_py_310/pep604_union_types_line_breaks.py
```diff
@@ -172,16 +218,100 @@ def foo(
...
+    ],  # aoeuaoeuaoeu
+): ...
+
+
+def foo(
+    a: loooooooooooooong
+        | loooooooooooooooooooooooooooong
+        | looooooooooong
+        | looooooooooong
+        = 3,
```
JelleZijlstra, line 271: All these examples with defaults look bad. It's as if we're doing long division.

JelleZijlstra (commented):
Thanks, but several of the test cases look quite bad to me. I'd prefer option (1) in Sebastian's issue: parenthesize it.

Zac-HD:
Having now looked at the diffs, I've changed my reaction back on the issue to prefer our current-preview parenthesized style 😓

While I really appreciate @jakkdl's work on this issue, perhaps we should close the issue as solved by https://github.com/psf/black/pull/3899 and call it a day?

jakkdl:
Sounds like we're all in agreement - a shame to toss out the work I did on this, but I'm quite comfortable in the black codebase now at least! Closing the PR~

PR #3931, opened 2023-10-07
Use inline flags for test cases

### LINKED ISSUE #3922: Support inline flags in test cases
Currently, our test cases are executed with different flags based on some combination of their filename and what directory they are in. I would prefer a system where we define the flags in a directive inline, e.g. like this:

```
# flags: --target-version=py38 --preview
1+1

# output
1 + 1
```

JelleZijlstra:
Fixes #3922.


tests/test_black.py
```diff
@@ -209,8 +211,8 @@ def test_piping_diff(self) -> None:
             r"(STDIN|STDOUT)\t\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d\.\d\d\d\d\d\d"
             r"\+\d\d:\d\d"
         )
-        source, _ = read_data("simple_cases", "expression.py")
-        expected, _ = read_data("simple_cases", "expression.diff")
+        source, _ = read_data("cases", "expression.py")
```
JelleZijlstra, line 214: Possibly the test cases in this file that use `read_data()` should be cleaned up, especially those using the `cases` directory, but that can wait for another PR.

hauntsaninja (approved):
Looks good!

Issue #1872, opened 2020-12-16
Enforce blank line after module docstring


lsorber:
Black does not have an opinion on whether there should be a blank line immediately after a module's docstring. This causes inconsistencies across blackened files because whatever the author chooses, stays.

Both examples below satisfy black:
```python
"""Module docstring."""

import typing
```

```python
"""Module docstring."""
import typing
```

My preference would be to enforce a blank line.

Related issue: https://github.com/psf/black/issues/450

JelleZijlstra:
Agree that we should add a blank line here.

felix-hilden:
A moderate +1 for the proposal. To get a feel for multiline docstrings and some more imports:

```python
"""
Multiline docstring.

That explains details about the module.
"""
import stuff
import antigravity

from __future__ import braces

# VS

"""
Multiline docstring.

That explains details about the module.
"""

import stuff
import antigravity

from __future__ import braces
```

I don't have a strong preference either way. Personally I don't add the empty line because the `"""` is enough of a separator, but that's not a hill I'm willing to die on 😄 I'd be happy with either 👍

PR #3287, opened 2022-09-26
Standardise newlines after module-level docstrings

ichard26:
### Description

I resolved the merge conflicts in #2996. This PR is that PR with a few tweaks on top. 

Closes #2996, resolves #1872.

### Checklist - did you ...

- [x] Add a CHANGELOG entry if necessary?
- [x] Add / update tests if necessary?
- [x] Add new / update outdated documentation?

PR #3932, opened 2023-10-08
Standardise newlines after module-level docstrings

DanielNoord:
### Description

Resolves https://github.com/psf/black/issues/1872.

This is a rebase of https://github.com/psf/black/pull/2996 and therefore most credit should go to @jpy-git.

I have excluded the refactoring that was included in the rebase attempt at https://github.com/psf/black/pull/3287. I feel the refactoring makes the PR unnecessarily hard to review and rebase. It's probably a very desired refactor (I can't comment on this since I don't know the codebase too well), but it can be taken up in a specific PR.

_I won't autoclose those PRs with this PR as I think the refactoring should probably be rebased on its own and not lost to the closed PR list._

I would appreciate a review as I'm not sure this is the right approach for the checks, it seemed to be the best compromise between `main` and the various versions of the check that @jpy-git wrote in their original PR. Happy to change it to another check if desired. I'd really like to see this change make it into the preview so happy to help wherever needed to get this to pass.

### Checklist - did you ...


tests/data/module_docstring_1.py
```diff
@@ -0,0 +1,25 @@
+"""Single line module-level docstring should be followed by single newline."""
```
JelleZijlstra, line 1: I don't think we'll run this test as it's not in `simple_cases/`.

DanielNoord: Moved the tests to ``preview/`` to make sure they run under preview mode. They executed now with `pytest -k module_docstring`.

Also added another two tests cases of which I thought the original PR didn't really have a good test:
1. A file that already has a new line in-between the docstring and the rest of code and
2. A file that has no lines in between the docstring and the rest of code and should thus get one

JelleZijlstra (approved):
Thanks! Pushed a change moving the tests as I just changed the structure of the tests.

chalecado:
how can i exclude this from being enforced on my project? its caused 100s of files to need to change

PR #2740, opened 2022-01-01
Remove Python 2 support

ichard26:
### Description

Should unblock GH-2297 and mypyc.

### Checklist - did you ...

- [x] Add a CHANGELOG entry if necessary?
- [x] Add / update tests if necessary?
- [x] Add new / update outdated documentation?

### Review notes

Right now this PR uses `22.0` as the cutoff version but this is highly up to debate / subject to change. It might just be better to remove these references until we know for certain.

PR #3933, opened 2023-10-09
Drop support for parsing Python 2

hauntsaninja:
Support for formatting Python 2 was dropped in 22.0 / https://github.com/psf/black/pull/2740

I think it's time that we do some cleanup. In particular, I was investigating some issue with the Grammar pickles and seems worth simplifying pygram.initialize

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,6 +40,9 @@
 
 <!-- Changes to Black's terminal output and error messages -->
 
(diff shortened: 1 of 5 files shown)


hauntsaninja:
mypyc builds catching something useful!

JelleZijlstra (approved):

PR #3936, opened 2023-10-09
Migrate mypy config to pyproject.toml

acharles7:
### Description

This PR is focused on streamlining the project's configuration files. Instead of maintaining separate `mypy.ini` file, this PR help us consolidate remaining mypy configuration file within the `pyproject.toml` file as per [PEP 518](https://peps.python.org/pep-0518/). This modification leads to a more concise and organized config files.

Relevant PR https://github.com/psf/black/pull/3632 (Closed)

### Checklist


src/blackd/__init__.py
```diff
@@ -74,7 +74,7 @@ def main(bind_host: str, bind_port: int) -> None:
     app = make_app()
     ver = black.__version__
     black.out(f"blackd version {ver} listening on {bind_host} port {bind_port}")
-    web.run_app(app, host=bind_host, port=bind_port, handle_signals=True, print=None)
+    web.run_app(app, host=bind_host, port=bind_port, handle_signals=True, print=None)  # type: ignore[arg-type]
```
JelleZijlstra, line 77: What was the issue here?

pyproject.toml
```diff
@@ -223,3 +223,33 @@ omit = [
...
+python_version = "3.8"
+mypy_path = "src"
+strict = true
+warn_unused_configs = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_return_any = true
+# Unreachable blocks have been an issue when compiling mypyc, let's try to avoid 'em in the first place.
+warn_unreachable = true
+disallow_untyped_calls = true
```
JelleZijlstra, line 240: Many of these are included in `--strict`:

```
  --strict                  Strict mode; enables the following flags: --warn-unused-configs, --disallow-any-generics, --disallow-subclassing-any, --disallow-untyped-calls, --disallow-untyped-defs,
                            --disallow-incomplete-defs, --check-untyped-defs, --disallow-untyped-decorators, --warn-redundant-casts, --warn-unused-ignores, --warn-return-any, --no-implicit-
                            reexport, --strict-equality, --extra-checks
```

JelleZijlstra (commented):
Thanks!

src/blackd/__init__.py
acharles7: `print` argument accepts `Callable[..., None] = print` 
i.e. 
```zsh
src/blackd/__init__.py:77:81: error: Argument "print" to "run_app" has incompatible type "None"; expected "Callable[..., None]"  [arg-type]

```

pyproject.toml
acharles7: Nice, They are now removed, Thanks

pyproject.toml
```diff
@@ -223,3 +223,26 @@ omit = [
...
+show_column_numbers = true
+
+[[tool.mypy.overrides]]
+module = ["pathspec.*", "IPython.*", "colorama.*", "tokenize_rt.*", "uvloop.*", "_black_version.*"]
+ignore_missing_imports = true
+
+# CI only checks src/, but in case users are running LSP or similar we explicitly ignore
+# errors in test data files.
+[[tool.mypy.overrides]]
+module = ["tests.data.*", "scripts.*"]
```
hauntsaninja, line 247: We shouldn't ignore errors in scripts. It's fine to have a looser config apply to scripts though, like ignore missing imports

acharles7: Good point, I thought about it but then skipped as it'll add few extra ignores in repo.

- Added `type: ignore` where necessary in `scripts.*`
-  Added `types-commonmark`, `urllib3`, `hypothesmith` as `additional_dependencies` in mypy hook to avoid adding `type: ignore` (let me know if its not the right way)

acharles7:
@JelleZijlstra @hauntsaninja Appreciate any tip you can provide for resolving these failed checks!

acharles7:
@JelleZijlstra @hauntsaninja Appreciate any tip you can provide for resolving these failed checks!

src/blackd/__init__.py
JelleZijlstra: Seems like this is fixed on master: https://github.com/aio-libs/aiohttp/blob/30850babb43a8e28dd2df036776c62fd613d3d89/aiohttp/web.py#L453C5-L453C5. Can you add a comment saying that aiohttp had an incorrect annotation and that it will be fixed once aiohttp releases that code?

JelleZijlstra (commented):
The CI failures look like mypyc isn't picking up the `ignore_missing_imports` in the config file. Not sure why that is, but possibly the hatch-mypyc config needs to be tweaked.

src/blackd/__init__.py
acharles7: Added, thanks

acharles7:
> @JelleZijlstra @hauntsaninja Appreciate any tip you can provide for resolving these failed checks!

I looked into it and seems like adding 

```toml
[tool.hatch.build.targets.wheel.hooks.mypyc]
mypy-args = ["--ignore-missing-imports"]
```

https://github.com/ofek/hatch-mypyc#mypy-arguments fixes the pipeline. However, I feel like its wildcard option and can ignore all imports errors. But, this makes CICD happy. 
Moreover, `Lint` check will work as expected using mypy config from `pyproject.toml` and can catch imports errors if any. 

I'm not sure why hatch mypyc hook is not using default `pyproject.toml`, maybe hatch mypyc limitation?

If you have any better way, Please suggest, Thanks!

JelleZijlstra (approved):

PR #3938, opened 2023-10-09
Report all stacktraces in verbose mode

hauntsaninja:
Previously these were swallowed; make the reporting more as like `black/__init__.py`

### DIFF
--- a/src/black/concurrency.py
+++ b/src/black/concurrency.py
@@ -173,7 +173,9 @@ async def schedule_formatting(
                 cancelled.append(task)
             elif task.exception():
                 if report.verbose:
-                    traceback.print_exception(task.exception())
+                    exc = task.exception()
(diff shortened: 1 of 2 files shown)


src/black/concurrency.py
```diff
@@ -171,7 +172,11 @@ async def schedule_formatting(
             if task.cancelled():
                 cancelled.append(task)
             elif task.exception():
-                report.failed(src, str(task.exception()))
+                exc = task.exception()
+                assert exc is not None
```
JelleZijlstra, line 176: Use a walrus so you can avoid the assert?

`elif exc := task.exception():`

hauntsaninja: Oh right, we dropped 3.7

JelleZijlstra (approved):

PR #3939, opened 2023-10-10
Fix test that was not being run

JelleZijlstra:


hauntsaninja (approved):

JelleZijlstra:
I also looked at the rest of the files in `tests/data/` for similar problems. Most files outside cases are in test directories for gitignore/file discovery, so it makes sense to keep those separate. `miscellaneous/` has a bunch of other files that probably make sense to keep by themselves for various reasons.

Issue #3889, opened 2023-09-13
Lines with power operator `**` with simple operands are wrapped too early


andersk:
**Describe the bug**

Black miscomputes the horizontal space required for lines containing the power operator `**` with simple operands, and rewraps them much earlier than the expected 88-character line length. Perhaps that computation wasn’t adjusted for #538/#2726.

**To Reproduce**

Run Black with no arguments on this code ([playground link](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADWAE5dAD2IimZxl1N_WmyLkeQ19WvnMnMLiOdunzNME7LTg9q9ETmvUb3zADAfwfc_ujwsAWOwtuqfHF_058CbOyGNssGXm6mbe5uTjBx_QytrAAAAACmo9-vGjRMyAAFq1wEAAADNPFPpscRn-wIAAAAABFla)):

```python
x = 1**1**1**1**1**1**1**1**1**1**1**1**1**1**1**1**1
y = 1**1**1**1**1**1**1**1**1**1**1**1**1**1**1**1**1**1
```

Black leaves the first line unaffected, but unexpectedly wraps the second vertically, even though its horizontal layout is only 56 characters long.

```python
x = 1**1**1**1**1**1**1**1**1**1**1**1**1**1**1**1**1
y = (
    1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
    ** 1
)
```

**Expected behavior**

Both lines should be unchanged.

**Environment**

- Black's version: 23.9.1
- OS and Python version: Linux/Python 3.10.12

PR #3942, opened 2023-10-13
[3889] Fix long lines with power operator(s) getting splitted before line length

henriholopainen:
### Description
Hugging power operators had a bug, where operators that end up hugging were still counted as not hugging when deciding to split the line. The bug is described [here](https://github.com/psf/black/issues/3889). The whole issue is a bit circular as we need to know the line length before we start to apply transformations. Calculating the hugging of power operators at the beginning of `transform_line` seems to do the trick. Ideally the `hug_power_op` transform would be done only once, but switching the order of transforms broke a lot of other places, so I don't know if there's any other good options here.

### Checklist - did you ...


henriholopainen:
> **diff-shades** results comparing this PR ([43fcdee](https://github.com/psf/black/commit/43fcdee7c29ab3f30f4d440afd6c7b337e306e94)) to main ([935f303](https://github.com/psf/black/commit/935f303a0a7b794e722c7df00c906be285884874)). The full diff is [available in the logs](https://github.com/psf/black/actions/runs/6510146111/job/17686088372#step:10:1) under the "Generate HTML diff report" step.
> 
> ```
> ╭─────────────────────── Summary ────────────────────────╮
> │ 3 projects & 10 files changed / 56 changes [+14/-42]   │
> │                                                        │
> │ ... out of 2 496 979 lines, 11 745 files & 23 projects │
> ╰────────────────────────────────────────────────────────╯
> 
> Differences found.
> ```
> 
> [**What is this?**](https://black.readthedocs.io/en/latest/contributing/gauging_changes.html#diff-shades) | [Workflow run](https://github.com/psf/black/actions/runs/6510146111) | [diff-shades documentation](https://github.com/ichard26/diff-shades#readme)

These all seem to be cases of erroneous line splitting getting fixed.

henriholopainen:
> Thanks. This looks correct, but we'll have to make the change in the preview style only for now.

Ok, will do. Does this mean that the tests should also be reverted, as they will fail after the fix is behind `--preview`?

JelleZijlstra:
You can add `# flags: --preview` to the top of the test so they get run in preview mode.

JelleZijlstra (approved):

PR #3950, opened 2023-10-16
Fix parser bug where "type" was misinterpreted as a keyword inside a match

### LINKED ISSUE #3790: Black 23.7.0 thinks accessing an attribute named `type` inside brackets in a switch statement is invalid syntax
**Describe the bug**

When black encounters access of a `.type` attribute inside brackets, inside a switch statement, it thinks that it is invalid syntax, even though python 3.11 executes it just fine.

**To Reproduce**

For example, take this code:

```python
class SomeClass:
    type = ""


match (SomeClass.type):
    case _:
        print("Hello there ...")
```

```sh
$ pipx run --python python3.11 --spec=black==23.7.0 black --target-version py311 --diff file.py
```

The resulting error is:

> error: cannot format file.py: Cannot parse: 5:0: match (SomeClass.type):

**Expected behavior**

It should leave the file unchanged or format it, as black 23.3.0 leaves it unchanged, and it is indeed valid syntax:

```console
$ python3.11 --version
Python 3.11.4
$ python3.11 file.py
Hello there ...
$ pipx run --python python3.11 --spec=black==23.3.0 black --target-version py311 --diff file.py
⚠️  black is already on your PATH and installed at /home/iwana/.local/bin/black. Downloading and running anyway.
All done! ✨ 🍰 ✨
```

**Environment**

- Black's version: 23.3.0
- OS and Python version: Linux/Python 3.11.4

JelleZijlstra:
Fixes #3790

Slightly hacky, but I think this is correct and it should also improve performance somewhat.


hauntsaninja (approved):
Not particularly familiar with blib2to3, but heuristic seems fine

Issue #3924, opened 2023-10-06
Preview style unnecessarily moves comment


JelleZijlstra:
The preview style moves a comment on one line of an implicitly concatenated string to a different line.

```
$ python -m black --version
python -m black, 23.9.1 (compiled: yes)
Python (CPython) 3.9.14
$ python -m black --diff --preview movecomment.py 
--- movecomment.py      2023-10-06 01:40:10.153647+00:00
+++ movecomment.py      2023-10-06 01:40:41.077230+00:00
@@ -1,7 +1,7 @@
 xxxxxxx.xxfx(
-    f"xxxx_xxxxxxxx: "  # xxxxxxx
+    "xxxx_xxxxxxxx: "
     f"xxxxx={xxxx_xxxxxxxx.xxxxx}, "
     f"_xxx_xxxxxx={xxxx_xxxxxxxx._xxx_xxxxxx}, "
     f"xxx_xxxxxx={xxxx_xxxxxxxx.xxx_xxxxxx}, "
-    f"xxx_xxxxx={xxxx_xxxxxxxx.xxx_xxxxx}"
+    f"xxx_xxxxx={xxxx_xxxxxxxx.xxx_xxxxx}"  # xxxxxxx
 )
would reformat movecomment.py

All done! ✨ 🍰 ✨
1 file would be reformatted.
$ cat movecomment.py 
xxxxxxx.xxfx(
    f"xxxx_xxxxxxxx: "  # xxxxxxx
    f"xxxxx={xxxx_xxxxxxxx.xxxxx}, "
    f"_xxx_xxxxxx={xxxx_xxxxxxxx._xxx_xxxxxx}, "
    f"xxx_xxxxxx={xxxx_xxxxxxxx.xxx_xxxxxx}, "
    f"xxx_xxxxx={xxxx_xxxxxxxx.xxx_xxxxx}"
)
```

henriholopainen:
This seems to be happening also with the stable style. Should it still be considered as a bug?

```
% black --version
black, 23.9.1 (compiled: yes)
Python (CPython) 3.11.3
% black --diff movecomment.py
--- movecomment.py	2023-10-14 08:29:58.442254+00:00
+++ movecomment.py	2023-10-14 08:29:59.965634+00:00
@@ -1,7 +1,7 @@
 xxxxxxx.xxfx(
-    f"xxxx_xxxxxxxx: "  # xxxxxxx
+    "xxxx_xxxxxxxx: "
[...]

PR #3956, opened 2023-10-19
[3924] Fix merging implicit multiline strings that have inline comments

henriholopainen:
### Description

After some discussions in the [issue](https://github.com/psf/black/issues/3924), we decided not to merge implicit multiline strings if there are inline comments on the string lines. Added description of behaviour to `Preview style` documentation.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@
 
 ### Preview style
 
-- Fix merging implicit multiline strings that have inline comments (#3952)
+- Fix merging implicit multiline strings that have inline comments (#3956)
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@
 
 ### Preview style
 
-<!-- Changes that affect Black's preview style -->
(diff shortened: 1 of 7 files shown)


henriholopainen:
The documentation build error looks like it's not rooted in these changes:

```
Configuration error:
There is a programmable error in your configuration file:

-------

pkg_resources.DistributionNotFound: The 'black' distribution was not found and is required by the application
```

Will look into the `diff-shades`

henriholopainen:
Hmm, can't reproduce the diff-shades locally:

```
% black --version
black, 0.1.dev1687+g9edba85.d20231019 (compiled: no)
Python (CPython) 3.11.3
% black --diff ../django/tests/admin_scripts/tests.py
All done! ✨ 🍰 ✨
```

```
% black --version
black, 0.1.dev1694+g74d46c8.d20231019 (compiled: no)
Python (CPython) 3.11.3
% black --diff ../django/tests/admin_scripts/tests.py
All done! ✨ 🍰 ✨
```

Is there some special configuration in the diff Action?

JelleZijlstra (approved):

Issue #925, opened 2019-07-11
Add note about dict indentation style to docs


strokirk:
Hi! 

I've been searching a bit on why this code containing a single inline-dict is expanded into multiple lines with hanging braces by black, but couldn't find anything about configuring it, if it's expected, or a bug. 

```
# In
func({
	# expand me
    'a':37,
    'b':42,
    'c':927,
})

# Out
func(
    {
		# expand me
        "a": 37,
        "b": 42,
        "c": 927,
    }
)
```

I personally would prefer the dict to keep it's original formatting for better readability, and lose the extra indentation and [Allman-style](https://en.wikipedia.org/wiki/Indentation_style#Allman_style) hanging bracket. Essentially we would keep the input formatting in this example. 

This pattern is also much more common in the python code I regularly read, including the standard library, django and a couple of other codebases. Grepping `\({$` vs `\(\n\s*{$` shows usages.

However, I expect this pattern might be fairly set in stone by now and would maybe cause a lot of unnecessary diffs for existing `black` users. What do you think?

Either way, I think something like this would be a useful example to add to [the Black code style](https://black.readthedocs.io/en/stable/the_black_code_style.html). That way people like me who are wondering why the function call is expanded can find out that it is to be expected. 
(I'm opening this issue partly to be able to google for it myself in coming years.)

Hopefully the feedback is somewhat useful!

Jma353:
If the inline `dict` was actually large and had to be split over several lines, it would be re-formatted in the above way as well, e.g.
````python
# In
func({"a": 37, "b": 42, "c": 927, "aaaaaaaaaaaaaaaaaaaaaaaaa": 11111111111111111111111111})

# Out
func(
    {
        "a": 37,
        "b": 42,
        "c": 927,
        "aaaaaaaaaaaaaaaaaaaaaaaaa": 11111111111111111111111111,
    }
)
````

I believe changing this would likely cause some diff-based churn.  In your above example, I believe the comment causes the inline `dict` to be restructured in this way, omitting the comment results in the `dict` fitting on one line: 
````python
# In
func({
    'a':37,
    'b':42,
    'c':927,
})

# Out
func({"a": 37, "b": 42, "c": 927})
````
This kinda is captured in the above linked documentation: 

[...]

PR #3964, opened 2023-10-22
[925] Improve multiline dictionary and list indentation for sole function parameter

henriholopainen:
### Description

Making multiline lists and dictionaries less indented when passed as a function parameter and pairing parens with brackets/braces was discussed at in [#925](https://github.com/psf/black/issues/925). This PR introduces a conservative first step, where a multiline list or dict as the sole parameter to a function call will have special treatment for better readability. The feature is behind `hug_parens_with_braces_and_square_brackets` preview toggle.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@
 ### Preview style
 
 - Fix merging implicit multiline strings that have inline comments (#3956)
-- Improve multiline dictionary and list indentation for sole function parameter (#XXXX)
+- Improve multiline dictionary and list indentation for sole function parameter (#3964)
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 ### Preview style
 
 - Fix merging implicit multiline strings that have inline comments (#3956)
+- Improve multiline dictionary and list indentation for sole function parameter (#XXXX)
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,8 @@
 ### Preview style
 
 - Fix merging implicit multiline strings that have inline comments (#3956)
-- Improve multiline dictionary and list indentation for sole function parameter (#3964)
+- Multiline dictionaries and lists that are the sole argument to a function are now
+  indented less (#3964)
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,10 +12,8 @@
 
 ### Preview style
 
-- Fix merging implicit multiline strings that have inline comments (#3956)
 - Multiline dictionaries and lists that are the sole argument to a function are now
   indented less (#3964)
-- Allow empty first line after block open before a comment or compound statement (#3967)
 
 ### Configuration
 
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -113,6 +113,32 @@ my_dict = {
 }
 ```
 
+### Improved multiline dictionary and list indentation for sole function parameter
+
+For better readability and less verticality, _Black_ now pairs parantheses ("(", ")")
+with braces ("{", "}") and square brackets ("[", "]") on the same line for single
+parameter function calls. For example:
+
+```python
+foo(
+    [
+        1,
+        2,
+        3,
+    ]
+)
+```
+
+will be changed to:
+
+```python
+foo([
+    1,
+    2,
+    3,
+])
+```
+
 ### Improved multiline string handling
 
 _Black_ is smarter when formatting multiline strings, especially in function arguments,
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -815,6 +815,20 @@ def _first_right_hand_split(
     tail_leaves.reverse()
     body_leaves.reverse()
     head_leaves.reverse()
+
+    if Preview.hug_parens_with_braces_and_square_brackets in line.mode:
+        if (
+            tail_leaves[0].type == token.RPAR
+            and tail_leaves[0].value
+            and head_leaves[-1].type == token.LPAR
+            and head_leaves[-1].value
+            and body_leaves[0].type in [token.LBRACE, token.LSQB]
+            and body_leaves[-1].type in [token.RBRACE, token.RSQB]
+        ):
+            head_leaves = head_leaves + body_leaves[:1]
+            tail_leaves = body_leaves[-1:] + tail_leaves
+            body_leaves = body_leaves[1:-1]
+
     head = bracket_split_build_line(
         head_leaves, line, opening_bracket, component=_BracketSplitComponent.head
     )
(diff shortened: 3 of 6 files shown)


CHANGES.md
```diff
@@ -13,6 +13,7 @@
 ### Preview style
 
 - Fix merging implicit multiline strings that have inline comments (#3956)
+- Improve multiline dictionary and list indentation for sole function parameter (#3964)
```
JelleZijlstra, line 16:
```suggestion
- Multiline dictionaries and lists that are the sole argument to a function are now indented less (#3964)
```
The changelog entry should just describe the change; it's up to the reader to judge whether it's an improvement :)

docs/the_black_code_style/future_style.md
```diff
@@ -113,6 +113,32 @@ my_dict = {
 }
 ```
 
+### Improved multiline dictionary and list indentation for sole function parameter
```
JelleZijlstra, line 116: Thanks for adding this documentation!

tests/data/cases/preview_hug_parens_with_braces_and_square_brackets.py
```diff
@@ -0,0 +1,79 @@
+# flags: --preview
+def foo_brackets(request):
+    return JsonResponse(
```
JelleZijlstra, line 3: Ideas for additional tests:

- Test that comments in various places are preserved (e.g., after the opening `(` or before the closing `)`)
- Test where the argument is a list comprehension
- Test where the argument is a set literal

JelleZijlstra (approved):

JelleZijlstra:
@henriholopainen seems like there's a crash on some code in Hypothesis (from the diff-shades failure). Can you take a look?

henriholopainen:
Located the issue in Hypothesis (and other places); opening and closing brackets/braces don't necessary belong to the same entity. For example:

python```
foooooooooooooooooooo([{c: n + 1 for c in range(256)} for n in range(100)] + [{}], {size})
```

becomes currently:

python```
foooooooooooooooooooo([
    {c: n + 1 for c in range(256)} for n in range(100)] + [{}], {size
})
```

which is clearly very wrong. I'll add some test cases and look for a sophisticated approach for a fix tomorrow.

henriholopainen:
Test cases added and logic fixed

CHANGES.md
```diff
@@ -12,7 +12,10 @@
 
 ### Preview style
 
-<!-- Changes that affect Black's preview style -->
+- Fix merging implicit multiline strings that have inline comments (#3956)
+- Multiline dictionaries and lists that are the sole argument to a function are now
+  indented less (#3964)
+- Allow empty first line after block open before a comment or compound statement (#3967)
```
```suggestion
- Multiline dictionaries and lists that are the sole argument to a function are now
  indented less (#3964)
```

The other two were released

src/black/linegen.py
```diff
@@ -815,6 +815,19 @@ def _first_right_hand_split(
     tail_leaves.reverse()
     body_leaves.reverse()
     head_leaves.reverse()
+
+    if Preview.hug_parens_with_braces_and_square_brackets in line.mode:
+        if (
+            tail_leaves[0].type == token.RPAR
+            and tail_leaves[0].value
+            and tail_leaves[0].opening_bracket is head_leaves[-1]
```
JelleZijlstra, line 823:
```suggestion
            and tail_leaves[0].opening_bracket == head_leaves[-1]
```
I don't think we can guarantee identity

src/black/linegen.py
```diff
@@ -815,6 +815,19 @@ def _first_right_hand_split(
...
+            and body_leaves[-1].type in [token.RBRACE, token.RSQB]
+            and body_leaves[-1].opening_bracket is body_leaves[0]
```
JelleZijlstra, line 825:
```suggestion
            and body_leaves[-1].opening_bracket == body_leaves[0]
```

CHANGES.md
henriholopainen: Sorry, bad merge on this

src/black/linegen.py
henriholopainen: If we don't check for identity here, cases like

```
baaaaaaaaaaaaar([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], {x}, "a string", [1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
```

will format to

```
baaaaaaaaaaaaar([
    1, 2, 3, 4, 5, 6, 7, 8, 9, 10], {x}, "a string", [1, 2, 3, 4, 5, 6, 7, 8, 9, 10
])
```

which is not something we want.

src/black/linegen.py
henriholopainen: Why not? The whole `omit` system in `_first_right_hand_split` seems to rely on `id(leaf)`, so the leaf objects should be stable and thus identity comparisons should be safe also here. Or am I missing something?

JelleZijlstra: Thanks, I was confused

hauntsaninja:
Thanks for the PR! While I like the style, it's a fairly big change for black and goes against the philosophy of diff reduction (in the case you add a new argument). I wonder therefore if it makes sense to have magic trailing commas after the single arg prevent this reformatting.

henriholopainen:
> Thanks for the PR! While I like the style, it's a fairly substantive change for black and goes against the philosophy of diff reduction (in the case you add or remove arguments). I wonder therefore if it makes sense to have magic trailing commas after the single arg prevent this reformatting.

Do you mean that we should instead format

```
foo({0: "looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong line"},)
```

as it would be formatted without the trailing comma:

```
foo({
    0: "looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong line"
})
```

?

henriholopainen:
> ```python
> foo(*[
> 	pretend,
> 	many,
> 	and,
> 	long,
> 	arguments,
> 	# ... ..................................................................
> 	*another_list
> ])
> ```

Currently it doesn't, but adding support for that would make sense IMO.

hauntsaninja:
It looks like Black already does what I suggested, I went ahead and added a test case and docs for it here: https://github.com/psf/black/pull/3991

JelleZijlstra:
Maybe I was too hasty in merging this. I think this makes things generally look better, but we should definitely think more about whether we want to go through before we move this change to stable.

charliermarsh:
One question, as both a user and a Ruff maintainer: is it intended that this also applies to parenthesized lists, sets, etc., outside of function calls? For example, with this commit:

```python
# Input
([1, 2, 3,])

# HEAD^
(
    [
        1,
        2,
        3,
    ]
)

# HEAD
([
    1,
    2,
    3,
])
```

cthoyt:
This makes me so very happy, thanks @JelleZijlstra :)

Issue #3975, opened 2023-10-24
Changed semantics for matching files in project root with `--include`


sth:
**Describe the bug**

If I have as file named `somefile` in the root of the project and I want to include it with `--include`, traditionally the appropriate regex would contain a leading slash:

    black --include '^/somefile$' ...

In #3846 this changed, now the leading slash prevents the file from being matched. Instead `--include '^somefile$'` is required.
The related `exclude` regexes still need the leading slash.

The cause are the path processing changes in #3846. Slashes are only added to `root_relative_path` when checking excludes and are now missing from `normalized_path`, which is later used to check includes.

**To Reproduce**

```sh
$ python3 -m venv venv
$ source venv/bin/activate
$ touch pyproject.toml
$ echo 'print( )' > somefile
$ pip install black==23.7.0
...
$ black --check --include '^/somefile$' .
would reformat /home/sth/tst/somefile

Oh no! 💥 💔 💥
1 file would be reformatted.
$ pip install black==23.10.1
...
$ black --check --include '^/somefile$' .
No Python files are present to be formatted. Nothing to do 😴
```

**Expected behavior**

The file name matching for `--include` should work as it did previously, since the change was most likely unintentional and the matching of excludes still works the old way.

**Environment**

- Black's version: 23.10.1
- OS and Python version: Ubuntu 22.04, Python 3.10.12

sth:
Fixed in #3976

PR #3976, opened 2023-10-24
Fix matching of absolute paths in `--include`

sth:
### Description

This is a fix for #3975.

It modifies `normalized_path` in the way it was done before #3846, so that the check for included files still matches the same files.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,7 @@
 
 <!-- Changes to how Black can be configured -->
 
-- Fix a bug in the matching of absolute path names in `--include` (#3975)
+- Fix a bug in the matching of absolute path names in `--include` (#3976)
 
 ### Packaging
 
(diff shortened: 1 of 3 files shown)


hauntsaninja (commented):
Thanks for catching this and opening a PR! I think we should use `root_relative_path` on L396 , so we apply include logic before symlink resolution

sth:
I agree it makes sense to check include and exclude against the same type of path and not use resolved symlinks for one and unresolved symlinks for the other.

It now checks the path with unresolved symlinks against includes and excludes and the symlink target is not checked this way (but it is still checked if is under the root dir).

I also added `is_dir` and `is_file` to all the mocks in the symlink testcase, since if the items aren't files they won't be included no matter the name.

hauntsaninja (approved):
Thank you!

Issue #3608, opened 2023-03-16
INTERNAL ERROR when using multiple fmt: skip


BenjaminBossan:
**Describe the bug**

Using multiple `# fmt: skip` in the same expression causes an `INTERNAL ERROR`.

**To Reproduce**

The first example (`a`) works but all others fail:

```python
a = (
    "this should "  # fmt: skip
    "be fine"
)

b = (
    "this is "  # fmt: skip
    "not working"  # fmt: skip
)

c = (
    "and neither "  # fmt: skip
    "is this "  # fmt: skip
    "working"
)

d = (
    "nor "
    "is this "  # fmt: skip
    "working"  # fmt: skip
)

e = (
    "and this "  # fmt: skip
    "is definitely "
    "not working"  # fmt: skip
)
```

The resulting error is:

> `INTERNAL ERROR: Black produced code that is not equivalent to the source`

Log:

```
--- src
+++ dst
@@ -33,11 +33,11 @@
       value=
         Constant(
           kind=
             None,  # NoneType
           value=
-            'this is not working',  # str
+            'this is',  # str
         )  # /Constant
     )  # /Assign
     Assign(
       targets=
         Name(
@@ -52,11 +52,11 @@
       value=
         Constant(
           kind=
             None,  # NoneType
           value=
-            'and neither is this working',  # str
+            'and neither working',  # str
         )  # /Constant
     )  # /Assign
     Assign(
       targets=
         Name(
@@ -71,11 +71,11 @@
       value=
         Constant(
           kind=
             None,  # NoneType
           value=
-            'nor is this working',  # str
+            'nor is this',  # str
         )  # /Constant
     )  # /Assign
     Assign(
       targets=
         Name(
@@ -90,10 +90,10 @@
       value=
         Constant(
           kind=
             None,  # NoneType
           value=
-            'and this is definitely not working',  # str
+            'and this not working',  # str
         )  # /Constant
     )  # /Assign
   type_ignores=
 )  # /Module
\ No newline at end of file
```

**Expected behavior**

The lines should be left untouched.

**Environment**

- black, 22.12.0 (compiled: yes)
- Python (CPython) 3.10.9
- Ubuntu 22.04

Can also be reproduced on https://black.vercel.app.

**Additional context**

Maybe I'm using `fmt: skip` incorrectly, but documentation on it is very sparse.

max-wittig:
Maybe I got the same here:

```sh
[...]

Issue #3438, opened 2022-12-13
Black adds an extra blank line to an import line that ends with `# fmt: skip`


yilei:
**Describe the bug**

It reformats:

```python
import os
import re  # fmt: skip
import sys
```

to:

```python
import os

import re  # fmt: skip
import sys
```

Example: https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ACSAFtdAD2IimZxl1N_WlwxRhR7MSgE-ANVmAcXiiHH_vLVBv6H9oJYdLgh3uJSbWM1qliEV5h8ji5l4G2iqd5AHPniBCkwXU5Gi97RWyWgW8b3Yvc90k3x9wXSWeryKEAAAIvWais0zjPAAAF3kwEAAAAB02j7scRn-wIAAAAABFla

**Expected behavior**

`# fmt: skip` shouldn't make it add an extra line.

**Additional context**

The issue is Black internally converts the `import re  # fmt: skip` line to a `STANDALONE_COMMENT`, and for other "real" `STANDALONE_COMMENT` lines it will add the blank line as designed. Maybe internally it needs to distinguish between real `STANDALONE_COMMENT` nodes v.s. converted from fmt off/skip?

ichard26:
I didn't know we added a blank before a comment preceding imports, that's a bit surprising. Anyway, distinguishing between real and internal STANDALONE_COMMENTS sounds like a reasonable fix, but I haven't looked at the code. This also affects `fmt: off/on` FWIW:

```diff
--- test.py	2022-12-14 04:17:08.331940 +0000
+++ test.py	2022-12-14 04:17:22.664904 +0000
@@ -1,5 +1,6 @@
 import os
+
 # fmt: off
 import re
 # fmt: on
 import sys
```

julian-west:
I have also come across this bug in the situation where I have a long import statement which black reformats across multiple lines. I tried to skip formatting on this import statement but instead an extra line is created. This causes problems with tools like `isort` which then tries to remove the new blank line.

For example:
```python
from my_package.my_module import function
from my_package.my_module_with_a_very_very_long_name import function_with_a_very_very_long_name  # fmt: skip
```
Reformatted to:
```python
from my_package.my_module import function

from my_package.my_module_with_a_very_very_long_name import function_with_a_very_very_long_name  # fmt: skip
```
---
[...]

PR #3978, opened 2023-10-26
[3608] Fix using multiple `fmt:skip` pragmas in a single block

henriholopainen:
### Description

This PR contains some refactoring to `comments.py`, as the code was not the cleanest and comments/naming were not up to date. This also fixes bugs with using multiple `fmt: skip` pragmas inside a single block, originally reported [here](https://github.com/psf/black/issues/3608).
Due to the refactoring, this PR is in small commits, so it's easier to review the changes piece by piece.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@
 
 ### Stable style
 
-<!-- Changes that affect Black's stable style -->
+- Fix bug where multiple fmt:skip pragmas inside single block fails (#XXXX)
 
 ### Preview style
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -35,7 +35,7 @@ from pathspec.patterns.gitwildmatch import GitWildMatchPatternError
 
 from _black_version import version as __version__
 from black.cache import Cache
-from black.comments import normalize_fmt_off
+from black.comments import normalize_format_skipping
 from black.const import (
     DEFAULT_EXCLUDES,
     DEFAULT_INCLUDES,
@@ -1099,7 +1099,7 @@ def _format_str_once(src_contents: str, *, mode: Mode) -> str:
         for feature in {Feature.PARENTHESIZED_CONTEXT_MANAGERS}
         if supports_feature(versions, feature)
     }
-    normalize_fmt_off(src_node, mode)
+    normalize_format_skipping(src_node, mode)
     lines = LineGenerator(mode=mode, features=context_manager_features)
     elt = EmptyLineTracker(mode=mode)
     split_line_features = {
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -166,8 +166,6 @@ def _convert_one_fmt_off_or_skip(node: Node, mode: Mode) -> bool:
                 if prev:
                     if found_fmt_off and prev.type not in WHITESPACE:
                         continue
-                    if found_fmt_skip and prev.type in WHITESPACE:
-                        continue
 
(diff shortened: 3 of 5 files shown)


henriholopainen:
Apparently one line in Hypothesis would be changed due to this. Are bug fixes also supposed to go behind `--preview`?

JelleZijlstra:
> Apparently one line in Hypothesis would be changed due to this. Are bug fixes also supposed to go behind `--preview`?

Part of the point of the stable style is so we don't have to decide whether something is a bug or a feature. When you upgrade from one version of Black 23.* to another, you shouldn't see any formatting changes on already-formatted code. However, in this case it looks like the change is made only in preview style, so you're fine!

I'll look at the rest of the diff over the next few days.

henriholopainen:
After looking at the format skipping code for a while now, it is starting to feel like modifying the AST by rewriting nodes to standalone comments might have not been the optimal approach and [PR #3610](https://github.com/psf/black/pull/3610/files) might have been a step into problems in the future. Format skipping could instead be a property of the node and leave the tree intact. Then it would be the responsibility of `transform_line` to skip lines that contain nodes that have format skipping enabled.

There might be something I still don't understand about the code, but I think giving this change a shot could tell if it's feasible quite fast. Or if you @JelleZijlstra have some insight why the current approach should not be changed, please let me know! 🙏

JelleZijlstra:
I'm open to alternative approaches for implementing `# fmt: skip`. We'll just have to try it and see if it works.

Is there a concrete problem that the current approach can't solve?

henriholopainen:
> I'm open to alternative approaches for implementing `# fmt: skip`. We'll just have to try it and see if it works.
> 
> Is there a concrete problem that the current approach can't solve?

I guess the current approach can solve cases, but it leads to the transformation code being contaminated with checks for standalone comment created by `fmt: skip`. Case example (from [issue 3438](https://github.com/psf/black/issues/3438)):

```
if (
    self.previous_line
    and self.previous_line.is_import
    and not current_line.is_import     <---
    and not current_line.is_fmt_pass_converted(first_leaf_matches=is_import)    <---
    and depth == self.previous_line.depth
):
```

This would be relevant to

```
import foo
import bar  # fmt: skip
```

If the nodes are not transformed to standalone comments, we wouldn't have to check for it everywhere we might be interested in the fact. Currently there probably are edge cases where the code doesn't check both for node type and standalone comment containing that node type, but those are not tested against and nobody has reported them. And if they surface, I don't think fixing them with the current approach will be a battle that can be won.

henriholopainen:
@JelleZijlstra should this PR be split into smaller ones to make reviewing more inviting?

JelleZijlstra:
It's probably fine as is, I'll need a bit of time to review it though. (And I'd welcome reviews from anyone else following along.)

CHANGES.md
```diff
@@ -11,11 +11,9 @@
 
 ### Stable style
 
-- Fix crash on formatting bytes strings that look like docstrings (#4003)
+- Fix bug where multiple fmt:skip pragmas inside single block fails (#3978)
```
JelleZijlstra, line 14:
```suggestion
- Fix bug where multiple fmt:skip pragmas inside single block fails (#3978)
- Fix crash on formatting bytes strings that look like docstrings (#4003)
```
I messed up the merge, sorry

cobaltt7:
Both issues mentioned in the description have been fixed.

PR #3984, opened 2023-10-27
Fix CI by running on Python 3.11

hauntsaninja:
aiohttp doesn't yet support 3.12


cooperlees (approved):
Thanks. Make me sad we could not be ready before GitHub flicked the switch and we create this tech debt, but such is life.

PR #3987, opened 2023-10-27
Check excludes before symlink resolution for command line parameters

### LINKED ISSUE #3826: Force Exclude does not work with symlinks
**Describe the bug**
I use the `force-exclude` feature to exclude files living in symlinked directories.
However, it looks like `black` uses the regexp from `force-exclude` **after** the symlink is resolved (not the symlinked path).

```
$ black -v /path_2_file/application/api/lib/dir_symlinked/__init__.py
Identified `/path_2_file` as project root containing a .git directory.
line_length: 135
target_version: ['py39']
force_exclude: (?:.*?application/api(?:-ext)?/lib/dir_symlinked/.+)

skip_string_normalization: True
Found input source: "application/lib/ref_dir/__init__.py"
/path_2_file/application/api/lib/dir_symlinked/__init__.py wasn't modified on disk since last run.

All done! ✨ 🍰 ✨
1 file left unchanged.
```

The resulting error is:

It should have ignored the file based on this path `/path_2_file/application/api/lib/dir_symlinked/__init__.py` rather than including it based on this path `application/lib/ref_dir/__init__.py`.
Said differently it used the resolved/resulting path rather than the path through the symlink

**Environment**

- Black's version: 23.7.0
- OS and Python version: MacOS Ventura 13.4.1, Python 3.9.12, 

**Additional context**

I would be equally happy if I could simply tell `black` **not** to follow symlinks.

Thank you in advance for your help!

sth:
### Description

This adjusts exclusion checking for command line parameters the same way as #3846 did for other files. Specifically, `--force-exclude` is now checked before symlink resolution. This way symlink handling is consistent between command line parameters and other collected files.
    
Fixes #3826.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,7 +19,7 @@
 
 - Add support for single line format skip with other comments on the same line (#3959)
 - File exclusion and inclusion logic is now consistently applied before symlink
-  resolution. (#3987, #3976)
+  resolution. (#3987) (#3976)
 - Fix a bug in the matching of absolute path names in `--include` (#3976)
 
 ### Packaging
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,8 @@
 ### Configuration
 
 - Add support for single line format skip with other comments on the same line (#3959)
-
+- File exclusion and inclusion logic is now consistently applied before symlink
+  resolution. (#3987, #3976)
 - Fix a bug in the matching of absolute path names in `--include` (#3976)
 
 ### Packaging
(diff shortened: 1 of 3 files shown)


sth:
The current problem in the test case is probably that the `root` path went through a `resolve()` while the source file path didn't. So if there are symlinks or other path changes that get normalized in `resolve()`, the two kinds of paths might not match anymore, causing errors in `relative_to()`.

sth:
This is obsolete now, #4015 fixed the problems :tada:

PR #3992, opened 2023-10-27
Hug parens also with multiline unpacking

henriholopainen:
### Description

As noted by @Avasam [here](https://github.com/psf/black/pull/3964#issuecomment-1783318353), hugging parens with brackets on multiline dictionaries and lists as sole function parameters didn't take into account list/dict unpacking. This PR fixes that shortcoming.

### Checklist - did you ...

### DIFF
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -139,7 +139,7 @@ foo([
 ])
 ```
 
-This also applies to list unpacking:
+This also applies to list and dictionary unpacking:
 
 ```python
 foo(
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -144,9 +144,8 @@ This also applies to list unpacking:
 ```python
 foo(
     *[
-        1,
-        2,
-        3,
+        a_long_function_name(a_long_variable_name)
+        for a_long_variable_name in some_generator
     ]
 )
 ```
@@ -155,9 +154,8 @@ will become:
 
 ```python
 foo(*[
-    1,
-    2,
-    3,
+    a_long_function_name(a_long_variable_name)
+    for a_long_variable_name in some_generator
 ])
 ```
 
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -139,6 +139,28 @@ foo([
 ])
 ```
 
+This also applies to list unpacking:
+
+```python
+foo(
+    *[
+        1,
+        2,
+        3,
+    ]
+)
+```
+
+will become:
+
+```python
+foo(*[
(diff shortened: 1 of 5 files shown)


docs/the_black_code_style/future_style.md
```diff
@@ -139,6 +139,28 @@ foo([
 ])
 ```
 
+This also applies to list unpacking:
+
+```python
+foo(
+    *[
+        1,
```
JelleZijlstra, line 147: This example isn't very compelling, as it could equivalently be written as `foo(1, 2, 3)`. I'd argue that the previous formatting is better, because it emphasizes that something weird is going on.

A more convincing variant might be if the argument is a comprehension, e.g. `foo(*[str(i) for i in range(3)])`. How do we handle that?

henriholopainen: Good point! Comprehensions work as any list would, so your example fitting on one line would not be formatted to anything. But I can come up with a suitable one for the docs.

Avasam:
In only gave list as an example. But dictionaries support unpacking too:
```python
def foo(**kwargs):
    print(kwargs)


other = {"d": 4}

foo(**{
    "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa": 1,
    "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb": 2,
    "ccccccccccccccccccccccccccccccccc": 3,
    **other,
})
```

tests/data/cases/preview_hug_parens_with_braces_and_square_brackets.py
```diff
@@ -137,6 +137,10 @@ def foo_square_brackets(request):
     [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], {x}, "a string", [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
 )
 
+foo(*["long long long long long line", "long long long long long line", "long long long long long line"])
```
JelleZijlstra, line 140: Can you also add tests for `**` unpacking (as mentioned by @Avasam)?

henriholopainen:
Hmm, the linting is failing. Should we format Black with the `--preview` style? If so, shouldn't that be caught with `tox -e run_self`?

tests/data/cases/preview_hug_parens_with_braces_and_square_brackets.py
```diff
@@ -137,6 +137,19 @@ def foo_square_brackets(request):
...
+
+foo(*[str(i) for i in range(100000000000000000000000000000000000000000000000000000000000)])
+
+foo(
```
JelleZijlstra, line 144: Also test a dict comprehension for completeness (like the one in cache.py).

PR #3994, opened 2023-10-28
Fix crash on await (a ** b)

### LINKED ISSUE #3676: bug formatting await (a**b)
**Describe the bug**

black converts `await (a**b)` into `await a ** b` (if run with --fast) 

**To Reproduce**

```python
async def name_5():
    await (name_1 ** name_4)
```

```sh
$ black file.py
```

The resulting error is:

> cannot format file.py: INTERNAL ERROR: ...

**Expected behavior**

black should preserve the braces and be able to format the file

**Environment**

python -m black, 23.3.1.dev10+geb32729 (compiled: no)
Python (CPython) 3.11.3

**Additional context**

error log:
``` diff
--- src
+++ dst
@@ -14,34 +14,34 @@
             None,  # NoneType
         )  # /arguments
       body=
         Expr(
           value=
-            Await(
-              value=
-                BinOp(
-                  left=
+            BinOp(
+              left=
+                Await(
+                  value=
                     Name(
                       ctx=
                         Load(
                         )  # /Load
                       id=
                         'name_1',  # str
                     )  # /Name
-                  op=
-                    Pow(
-                    )  # /Pow
-                  right=
-                    Name(
-                      ctx=
-                        Load(
-                        )  # /Load
-                      id=
-                        'name_4',  # str
-                    )  # /Name
-                )  # /BinOp
-            )  # /Await
+                )  # /Await
+              op=
+                Pow(
+                )  # /Pow
+              right=
+                Name(
+                  ctx=
+                    Load(
+                    )  # /Load
+                  id=
+                    'name_4',  # str
+                )  # /Name
+            )  # /BinOp
         )  # /Expr
       decorator_list=
       name=
         'name_5',  # str
       returns=


```

For context:
I currently write a generator wich generates random python code. This is still work in progress, but finding bugs in black shows me that I am on the right track.
I will provide a script similar to `scripts/fuzz.py` as soon as it is ready.

JelleZijlstra:
Fixes #3676

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -1354,14 +1354,18 @@ def remove_await_parens(node: Node) -> None:
                 if bracket_contents.type != syms.power:
                     ensure_visible(opening_bracket)
                     ensure_visible(closing_bracket)
-                elif (
-                    bracket_contents.type == syms.power
-                    and bracket_contents.children[0].type == token.AWAIT
-                ):
-                    ensure_visible(opening_bracket)
-                    ensure_visible(closing_bracket)
-                    # If we are in a nested await then recurse down.
-                    remove_await_parens(bracket_contents)
+                elif bracket_contents.type == syms.power:
+                    if bracket_contents.children[0].type == token.AWAIT:
+                        ensure_visible(opening_bracket)
+                        ensure_visible(closing_bracket)
+                        # If we are in a nested await then recurse down.
+                        remove_await_parens(bracket_contents)
+                    elif (
+                        len(bracket_contents.children) == 3
+                        and bracket_contents.children[1].type == token.DOUBLESTAR
+                    ):
+                        ensure_visible(opening_bracket)
+                        ensure_visible(closing_bracket)
 
 
 def _maybe_wrap_cms_in_parens(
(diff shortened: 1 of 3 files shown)


tests/data/cases/remove_await_parens.py
```diff
@@ -80,6 +80,15 @@ async def main():
 async def main():
     await (yield)
 
+async def main():
+    await (a ** b)
+
+async def main():
```
JelleZijlstra, line 86: These two cases were already fine, but I added them for completeness

JelleZijlstra:
This isn't enough, it still crashes on `await (a[b] ** c)`.

src/black/linegen.py
```diff
@@ -1354,14 +1354,18 @@ def remove_await_parens(node: Node) -> None:
...
-                    ensure_visible(opening_bracket)
-                    ensure_visible(closing_bracket)
-                    # If we are in a nested await then recurse down.
-                    remove_await_parens(bracket_contents)
+                elif bracket_contents.type == syms.power:
+                    if bracket_contents.children[0].type == token.AWAIT:
+                        ensure_visible(opening_bracket)
+                        ensure_visible(closing_bracket)
+                        # If we are in a nested await then recurse down.
+                        remove_await_parens(bracket_contents)
```
hauntsaninja, line 1362: When do we need to manually make this recursive call?

JelleZijlstra: Seems like we don't, all the tests pass if I remove it. We'll visit nested awaits anyway because this is invoked from a visitor. It also isn't necessary for removing multiple pairs of parens; we have tests for that.

hauntsaninja (approved):

Issue #3860, opened 2023-09-03
INTERNAL ERROR: Black produced code that is not equivalent to the source. (with --preview)


alexveden:
Black produced code that is not equivalent to the source. (with --preview) 

**Describe the bug**

**To Reproduce**

For example, take this code:

```python
log.info(f'Skipping: {desc["db_id"]=} {desc["ms_name"]} {money=} {dte=} {pos_share=} {desc["status"]=} {desc["exposure_max"]=}')
```

And run it with these arguments:

```sh
$ black --preview file.py
```

The resulting error is:

> cannot format file.py: INTERNAL ERROR: ...

```

--- src
+++ dst
@@ -8,11 +8,11 @@
               values=
                 Constant(
                   kind=
                     None,  # NoneType
                   value=
-                    'Skipping: desc["db_id"]=',  # str
+                    "Skipping: desc['db_id']=",  # str
                 )  # /Constant
                 FormattedValue(
                   conversion=
                     114,  # int
                   format_spec=
@@ -134,11 +134,11 @@
                 )  # /FormattedValue
                 Constant(
                   kind=
                     None,  # NoneType
                   value=
-                    'desc["status"]=',  # str
+                    "desc['status']=",  # str
                 )  # /Constant
                 FormattedValue(
                   conversion=
                     114,  # int
                   format_spec=
@@ -167,11 +167,11 @@
                 )  # /FormattedValue
                 Constant(
                   kind=
                     None,  # NoneType
                   value=
-                    'desc["exposure_max"]=',  # str
+                    "desc['exposure_max']=",  # str
                 )  # /Constant
                 FormattedValue(
                   conversion=
                     114,  # int
                   format_spec=

```
**Expected behavior**

**Environment**

- Black's version: 23.7.0
- OS and Python version: OS Linux, Python 3.10

**Additional context**

paketb0te:
I can reproduce with the example given (black, 23.7.0 (compiled: yes), Python (CPython) 3.11.5) - interestingly, the error does _not_ occur when swapping the quotes around (using double quotes for the "outer" f-string, and single quotes for the "inner" strings:

```py
log.info(f"Skipping: {desc['db_id']=} {desc['ms_name']} {money=} {dte=} {pos_share=} {desc['status']=} {desc['exposure_max']=}")
```

PR #4005, opened 2023-10-31
Preserve visible quote types for f-string debug expressions

henriholopainen:
### Description

Don't touch visible quotes in f-string debug expressions. From this [issue](https://github.com/psf/black/issues/3860).

### Checklist - did you ...


src/black/__init__.py
```diff
@@ -1343,6 +1343,37 @@ def get_imports_from_children(children: List[LN]) -> Generator[str, None, None]:
     return imports
 
 
+def _is_diff_from_string_quote_types(diff: str) -> bool:
+    """
+    Returns True if the diff contains only changes in strings where quotes change from
+    ' -> " or " -> '. For example
+
+    -                    'desc["status"]=',  # str
+    +                    "desc['status']=",  # str
```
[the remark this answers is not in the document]
henriholopainen: Valid point! Would inconsistent behaviour between

```
f'foo {x["bar"]}'
-->
f"foo {x['bar']}"
```

and

```
f'foo {x["bar"]=}'
-->
f'foo {x["bar"]=}'
```

be acceptable?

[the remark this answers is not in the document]
JelleZijlstra: Yes, I think that's fine.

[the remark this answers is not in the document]
henriholopainen: Excellent point, I think the second revision's solution is way better. 🙏

hauntsaninja (commented):
I don't think the regex is right, need to care about things like `{foo("asdf")=}` as well. Also maybe worth using `iter_fexpr_spans`?

henriholopainen:
> I don't think the regex is right, need to care about things like `{foo("asdf")=}` as well. Also maybe worth using `iter_fexpr_spans`?

Thanks, will fix! 🙏

src/black/trans.py
```diff
@@ -590,11 +590,21 @@ def make_naked(string: str, string_prefix: str) -> str:
...
-                string = _toggle_fexpr_quotes(string, QUOTE)
-                # After quotes toggling, quotes in expressions won't be escaped
-                # because quotes can't be reused in f-strings. So we can simply
-                # let the escaping logic below run without knowing f-string
-                # expressions.
+                f_expressions = (
+                    string[span[0] : span[1]] for span in iter_fexpr_spans(string)
+                )
+                expressions_contain_printable_quotes = any(
+                    re.search(r".*[\'\"].*=", expression)
```
hauntsaninja, line 597: Something like this could avoid false triggering on `f"{x == y}"`
```suggestion
                    re.fullmatch(r".*[\'\"].*=\s*", expression)
```

hauntsaninja (approved):
Thanks, this looks pretty good!

src/black/trans.py
henriholopainen: `f"{foo["bar"]=:.2f"` would still be a problem. Maybe `re.fullmatch(r"{.*[\'\"][^=]*=[^=]*}", expression)`?

hauntsaninja: Oh good point. I think that won't work with `!=` and `:=` and so on. I think maybe let's just keep this as in the current version of the PR and revisit when PEP 701 support is landed

JelleZijlstra, line 597: This catches a little too much:

```
% black --line-length=20 --preview -c "log.info(f'{\"a\" == \"b\"} aaaaaa aa aa aa')"                  
log.info(
    f'{"a" == "b"} aaaaaa'
    " aa aa aa"
)
```

Here I'd expect the quotes to get reversed.

tests/data/cases/preview_long_strings.py
```diff
@@ -305,6 +305,15 @@ def foo():
...
 
+log.info(f'Skipping: {desc["db_id"]=} {desc["ms_name"]} {money=} {dte=} {pos_share=} {desc["status"]=} {desc["exposure_max"]=}')
+
+log.info(f"Skipping: {desc['db_id']=} {desc['ms_name']} {money=} {dte=} {pos_share=} {desc['status']=} {desc['exposure_max']=}")
+
+log.info(f'Skipping: {desc["db_id"]} {desc["ms_name"]} {money=} {dte=} {pos_share=} {desc["status"]} {desc["exposure_max"]}')
+
+log.info(f'Skipping: {foo("asdf")=} {desc["ms_name"]} {money=} {dte=} {pos_share=} {desc["status"]} {desc["exposure_max"]}')
+
+log.info(f'Skipping: {  longer_longer_longer_longer_longer_longer_name   [ "db_id" ]      [ "another_key" ]   =  :  .3f }')
```
JelleZijlstra, line 317: Add some test cases involving triple quotes, e.g.

```
% black --line-length=20 --preview -c "log.info(f'''{\"a\" == 'b'} aaaaaa aa aa aa aa aa aa aa''')"
log.info(
    f"""{"a" == 'b'} aaaaaa aa aa aa aa aa aa aa"""
)
```

CHANGES.md
```diff
@@ -17,6 +17,7 @@
   indented less (#3964)
 - Multiline list and dict unpacking as the sole argument to a function is now also
   indented less (#3992)
+- Preserve printable quote types for debug f-strings (#4005)
```
JelleZijlstra, line 20: Not sure why you call this "printable" quotes, aren't they just string quotes?

JelleZijlstra (commented):
Thank you!

CHANGES.md
henriholopainen: Only quotes that are visible in the the final string's debug replacement fields are of interest to us, as those affect the AST. I agree that 'printable' probably isn't the best name here. Maybe "In f-string debug replacement fields preserve quote types that are visible in the final string"?

src/black/trans.py
henriholopainen: Nice catch! Also `{ "a" == "b"= }` is valid, as is `{ "a" == "b" == "c" ... = }`. I'll try to find a better regex.

henriholopainen: And even `{ foo("bar",x=123) }` 🙃

hauntsaninja: Like in https://github.com/psf/black/pull/4005/files#r1378259931 I'm fine with the regex as is — we can revisit with PEP 701 support. Although we should add some test cases for these.

I don't see a way to do it without negative lookbehind, something like `.*[\'\"].*(?<![=!:])=` maybe

henriholopainen: Yeah, currently looking at `.*[\'\"].*(?<![!:=])={1}(?!=)(?![^\s:])`

henriholopainen: And maybe it's better to be more consistent, even if the regex becomes a bit more complex

tests/data/cases/preview_long_strings.py
henriholopainen: I don't think your example is valid, as f-strings are not supposed to have backslashes inside the expressions. I wonder why the `-c` option allows this.

JelleZijlstra: These backslashes are handled by the shell, not Python. (This was on a Mac; on Windows you'd probably need different quoting.)

JelleZijlstra (approved):

hauntsaninja (commented):
I wouldn't mind a test involving `:=`

hauntsaninja (approved):
Thank you!!!!

JelleZijlstra:
diff-shades keeps failing, but it passed previously and we haven't changed anything substantial since, so I'm just going to merge.

henriholopainen:
Can we close the issue also 🙏

JelleZijlstra:
@henriholopainen it's easier if you put "Fixes #3860" in your PR description :)

PR #4010, opened 2023-10-31
Prefer more equal signs before a break when splitting chained assignments

### LINKED ISSUE #4007: Preview: Improved line breaks with multiple targets
**Describe the bug**

The "Improved line breaks" preview style should break the value before the target according to the preview style documentation. This is true for the first target, but doesn't seem to be the case when using assignments with multiple targets:

**To Reproduce**
```python
a["a"] = b["b"] = aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

Gets formatted as 

```python
a["a"] = b[
    "b"
] = aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

```

[Playground](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADMAFVdAD2IimZxl1N_WlPES01_qCOT7awAfNGVJ9IWJHgdr0d2gMGKSpT3UMUMo5J6ECjUnb1jNhRjYKWZ3fsosIWsXGVDUtIk1F-eVokrlu_cNCY1wSFZAQAAAAAA-lFsvxNMpXAAAXHNAQAAALZBEqqxxGf7AgAAAAAEWVo=)

**Expected behavior**

Black to parenthesize the value before splitting the targets:

```python
a["a"] = b["b"] = (
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
)
```

henriholopainen:
### Description

This PR makes rhs processing prefer more equal signs before breaking the line.

Fixes #4007 

### Checklist - did you ...


tests/data/cases/preview_prefer_rhs_split.py
```diff
@@ -84,3 +84,14 @@
 ) or (
     isinstance(some_other_var, BaseClass) and table.something != table.some_other_thing
 )
+
+# Multiple targets
+a = b = (
```
JelleZijlstra, line 89: We should also consider cases where one of the assignment targets is a long string and the RHS is short (`a = bbbbbbbbbbbbbb = c`).

MichaReiser:
Thanks

The following diff seems interesting

```diff
         elif constraint is None:
-            self.constraint_target = (
-                self.inferred_target_elements
-            ) = self.inferred_target_whereclause = None
+            self.constraint_target = self.inferred_target_elements = (
+                self.inferred_target_whereclause
+            ) = None
```

src/black/linegen.py
```diff
@@ -919,10 +931,20 @@ def _maybe_split_omitting_optional_parens(
...
+    rhs_oop: RHSResult, rhs: RHSResult, mode: Mode
+) -> bool:
     """
-    Returns whether we should prefer the result from a split omitting optional parens.
+    Returns whether we should prefer the result from a split omitting optional parens
+    (rhs_oop) over the original (rhs).
     """
+    # If we have multiple targets, we prefer more `=`s on the head vs pushing them to
+    # the body
+    lhs_equal_count = [leaf.type for leaf in rhs.head.leaves].count(token.EQUAL)
```
JelleZijlstra, line 943:
```suggestion
    rhs_equal_count = [leaf.type for leaf in rhs.head.leaves].count(token.EQUAL)
```
Not sure where the "lhs" comes from here?

JelleZijlstra:
I'm not quite sure about this. Looking at the diff-shades output, there's a few good ones:

```diff
-                            collected_attributes[name] = column_copies[
-                                obj
-                            ] = ret
+                            collected_attributes[name] = column_copies[obj] = (
+                                ret
+                            )
```

It's nicer to keep the subscript on one line.

```diff
-        self.selectable = (
-            self.persist_selectable
-        ) = self.local_table = selectable
+        self.selectable = self.persist_selectable = self.local_table = (
+            selectable
+        )
```

It's better to keep all the assignment targets next to each other, instead of randomly enclosing one in parentheses.

But there's also a lot of cases like the one @MichaReiser highlighted, where we just change which of the LHS assignment targets get parenthesized. In those cases, I don't think either formatting is clearly better; it looks ugly either way. I'd rather have Black not make a change there, because every time we change the formatting of existing code, that's disruptive to users. I'd want to make changes only if we believe the new formatting is better. Therefore, I'd prefer to stick with the current formatting in those cases.

Perhaps the rule could be that we split around the RHS if possible, and if that doesn't work, we preserve the current behavior and split around the second assignment target from the left.

src/black/linegen.py
henriholopainen: The thought process was, that rhs.head becomes the lhs in the cases where we have multiple assignments. But yeah, better to stick with `rhs_head`. 👍

henriholopainen:
> I'm not quite sure about this. Looking at the diff-shades output, there's a few good ones:
> 
> ```diff
> -                            collected_attributes[name] = column_copies[
> -                                obj
> -                            ] = ret
> +                            collected_attributes[name] = column_copies[obj] = (
> +                                ret
> +                            )
> ```
> 
> It's nicer to keep the subscript on one line.
> 
> ```diff
> -        self.selectable = (
> -            self.persist_selectable
> -        ) = self.local_table = selectable
> +        self.selectable = self.persist_selectable = self.local_table = (
> +            selectable
> +        )
> ```
> 
> It's better to keep all the assignment targets next to each other, instead of randomly enclosing one in parentheses.
> 
> But there's also a lot of cases like the one @MichaReiser highlighted, where we just change which of the LHS assignment targets get parenthesized. In those cases, I don't think either formatting is clearly better; it looks ugly either way. I'd rather have Black not make a change there, because every time we change the formatting of existing code, that's disruptive to users. I'd want to make changes only if we believe the new formatting is better. Therefore, I'd prefer to stick with the current formatting in those cases.
> 
> Perhaps the rule could be that we split around the RHS if possible, and if that doesn't work, we preserve the current behavior and split around the second assignment target from the left.

For me even @MichaReiser's example is better with the new formatting. Somehow it feels more natural when as many assignments as possible end up formatted as early as possible. Consider for example this (rather extreme) example:

```
self.selectable = self.persist_selectable = self.local_table = self.foo_bar = self.foo = self.longlonglong_key = self.another_long_long_long_key = self.yet_again_a_long_key = value
```

On current main it becomes:

```
self.selectable = (
    self.persist_selectable
) = (
    self.local_table
) = (
    self.foo_bar
) = (
    self.foo
) = (
    self.longlonglong_key
) = self.another_long_long_long_key = self.yet_again_a_long_key = value
```

and with this PR:

```
self.selectable = self.persist_selectable = self.local_table = self.foo_bar = (
    self.foo
) = self.longlonglong_key = self.another_long_long_long_key = (
    self.yet_again_a_long_key
) = value
```

I don't think a hybrid model would make sense here: 

```
self.selectable = (
    self.persist_selectable
) = (
    self.local_table
) = (
    self.foo_bar
) = (
    self.foo
) =  self.longlonglong_key = self.another_long_long_long_key = self.yet_again_a_long_key = (
    value
)
```

I think there is some inherent value to consistency and easy to follow rules. Thus I feel like we should stick with either splitting early or late, but not make it depend on the case, and in general it seems splitting late >= splitting early. And while I do appreciate avoiding introducing unnecessary diffs, this is quite the corner case and doesn't touch that many lines of code.

All that being said, as a contributor I trust your view on what is good and desired. Also, even though intuitively I feel consistency should weigh in cases where two formatting options are both ugly, I'm not fully sure if it should be enough to justify changing the formatting.

JelleZijlstra:
@hauntsaninja do you have any opinion here? I think this is mostly an improvement but I'm a bit hesitant to make style changes in existing code that aren't clear improvements.

hauntsaninja (approved):
I think it's an improvement! Splitting the one in the middle is not usually a choice a human would make.
Moreover, the change here isn't super common or controversial, so I think churn costs aren't too high.

JelleZijlstra:
Sounds good! Unfortunately diff-shades failed again, I'll retry to see if it succeeds.

JelleZijlstra (approved):
Let's do it then, I'd like to see if I can get diff-shades to work first though.

JelleZijlstra:
I think the diff-shades errors miraculously fixed themselves.

Issue #563, opened 2018-10-15
Black fails when comment exists inside of list comprehension


digitalresistor:
Howdy! Sorry you're having trouble. To expedite your experience,
provide some basics for me:

Operating system:
Python version: 3.7
*Black* version: 18.9b0
Does also happen on master: yes

```
class Something(object):
    def acceptable_offers(self, offers):
        """
        Return the offers that are acceptable according to the header.

        Any offers that cannot be parsed via
        :meth:`.Accept.parse_offer` will be ignored.

        :param offers: ``iterable`` of ``str`` media types (media types can
                       include media type parameters)
        :return: When the header is invalid, or there is no ``Accept`` header
                 in the request, all `offers` are considered acceptable, so
                 this method returns a list of (media type, qvalue) tuples
                 where each offer in `offers` is paired with the qvalue of 1.0,
                 in the same order as in `offers`.
        """
        return [
            (offers[offer_index], 1.0)
            for offer_index, _
            # avoid returning any offers that don't match the grammar so
            # that the return values here are consistent with what would be
            # returned in AcceptValidHeader
            in self._parse_and_normalize_offers(offers)
        ]
```

Is mis-formatted as:

```
class Something(object):
    def acceptable_offers(self, offers):
        """
        Return the offers that are acceptable according to the header.

        Any offers that cannot be parsed via
        :meth:`.Accept.parse_offer` will be ignored.

        :param offers: ``iterable`` of ``str`` media types (media types can
                       include media type parameters)
        :return: When the header is invalid, or there is no ``Accept`` header
                 in the request, all `offers` are considered acceptable, so
                 this method returns a list of (media type, qvalue) tuples
                 where each offer in `offers` is paired with the qvalue of 1.0,
                 in the same order as in `offers`.
        """
        return [
            (offers[offer_index], 1.0)
            for offer_index, _# avoid returning any offers that don't match the grammar so# that the return values here are consistent with what would be# returned in AcceptValidHeader in self._parse_and_normalize_offers(
                offers
            )
        ]
```

Which is not valid Python:

```
error: cannot format /Users/xistence/Projects/Pylons/webob/src/webob/acceptparse.py: INTERNAL ERROR: Black produced invalid code: invalid syntax (<unknown>, line 1346). Please report a bug on https://github.com/ambv/black/issues.  This invalid output might be helpful: /var/folders/jy/3csvdt_s3ys8sdn_123w89t00000gp/T/blk_ogrf5m4o.log
```

[...]

PR #4016, opened 2023-11-01
[563] Fix standalone comments inside complex blocks crashing Black

henriholopainen:
### Description

As described [here](https://github.com/psf/black/issues/563), and in referenced issues thereof, Black crashes when standalone comments are sprinkled inside complex multiline blocks. This PR takes a first step in fixing the most obvious cases and aims to get rid of blatant crashing. There are known issues with more nested expressions not being indented correctly, e.g.

```
[
    (
        lambda
        # a comment                                                                                                                     
        : None
    )
]
```

currently becomes

```
[
    (lambda
    # a comment                                                                                                                         
    : None)
]
```

I suggest we tackle those more cosmetic problems in a separate PR.

### Checklist - did you ...


henriholopainen:
...aaand there seems to be some formatting changes in the stable style, so I'll convert this into a draft until those are resolved.

src/black/lines.py
```diff
@@ -103,7 +102,11 @@ def append_safe(self, leaf: Leaf, preformatted: bool = False) -> None:
         Raises ValueError when any `leaf` is appended after a standalone comment
         or when a standalone comment is not the first leaf on the line.
         """
-        if self.bracket_tracker.depth == 0:
+        if (
+            self.bracket_tracker.depth == 0
+            or self.bracket_tracker._for_loop_depths
```
JelleZijlstra, line 107: We should make these attributes public if we're using them outside the class.

JelleZijlstra (approved):

src/black/lines.py
henriholopainen: Agreed, wrapped in a function call

henriholopainen:
Can we also close the issue?

PR #4019, opened 2023-11-03
Fix crash with f-string docstrings

### LINKED ISSUE #4018: Internal error on a specific file
**Describe the bug**
`black` reports an internal error when formatting a specific file.

```
error: cannot format /home/nicpa/codes/sisl/src/sisl_toolbox/siesta/minimizer/_metric_siesta.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk__3mh1ucd.log
```

**To Reproduce**

1. Download [this file](https://github.com/zerothi/sisl/blob/5a63302b57fcb38d7460507bf000f077655ac664/src/sisl_toolbox/siesta/minimizer/_metric_siesta.py)
2. Run `black` on the file. (I have done `pip install -U black` as of today!)

My pyproject.toml configuration file has this:
```toml
[tool.black]
line-length = 88
target-version = ["py38", "py39", "py310", "py311", "py312"]
```

The resulting error is:
```
error: cannot format /home/nicpa/codes/sisl/src/sisl_toolbox/siesta/minimizer/_metric_siesta.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk__3mh1ucd.log
```

Here is the attached diff-log file:
```diff
--- src
+++ dst
@@ -2307,16 +2307,10 @@
                   body=
                     Expr(
                       value=
                         JoinedStr(
                           values=
-                            Constant(
-                              kind=
-                                None,  # NoneType
-                              value=
-                                '',  # str
-                            )  # /Constant
                             FormattedValue(
                               conversion=
                                 -1,  # int
                               format_spec=
                                 None,  # NoneType
@@ -3263,16 +3257,10 @@
                   body=
                     Expr(
                       value=
                         JoinedStr(
                           values=
-                            Constant(
-                              kind=
-                                None,  # NoneType
-                              value=
-                                '',  # str
-                            )  # /Constant
                             FormattedValue(
                               conversion=
                                 -1,  # int
                               format_spec=
                                 None,  # NoneType
@@ -4273,16 +4261,10 @@
                   body=
                     Expr(
                       value=
                         JoinedStr(
                           values=
-                            Constant(
-                              kind=
-                                None,  # NoneType
-                              value=
-                                '',  # str
-                            )  # /Constant
                             FormattedValue(
                               conversion=
                                 -1,  # int
                               format_spec=
                                 None,  # NoneType
```

**Expected behavior**

**Environment**

- Black's version: 
  > black --version                                                          
  > black, 23.10.1 (compiled: yes)
  > Python (CPython) 3.11.6

- OS and Python version: Linux/debian, Python 3.11.6

hauntsaninja:
Fixes #4018

Python does not consider f-strings to be docstrings, so we probably shouldn't be formatting them as such. This is the easiest way to fix the issue with the AST check.

Despite its name, I think docstring_preview.py might be testing stable style, which this unfortunately does change


src/black/nodes.py
```diff
@@ -529,7 +529,7 @@ def is_docstring(leaf: Leaf) -> bool:
         return False
 
     prefix = get_string_prefix(leaf.value)
-    if "b" in prefix or "B" in prefix:
+    if "b" in prefix or "B" in prefix or "f" in prefix or "F" in prefix:
```
AlexWaygood, line 532: Might be slightly cleaner?

```suggestion
    if set(prefix).intersection("bBfF"):
```

JelleZijlstra:
Thanks, it does look like this changes the stable style. The case it changes is rather odd, but we should still follow our policy and make the change only in the stable style.

Some options:

- Just wait two months and merge this PR in January so it can go straight into the stable style, without waiting for preview. Seems OK since the change is rather obscure.
- For now, keep formatting f-strings without any placeholders, and turn off formatting for f-strings with placeholders. This would fix the crash while preserving behavior. In the preview style we can also disable formatting for f-strings without placeholders.
- Fix the crash instead by loosening the AST safety check (which already special cases docstrings). I'd rather not do that though.

hauntsaninja:
Option 2 would also change the stable style right?

I would also like to avoid Option 3. Note that the AST safety check special cases all strings, not just docstrings, which makes me even more loathe to take Option 3.

So Option 1 it is?

JelleZijlstra:
I thought Option 2 was safe because any change we make to f-strings with placeholders would cause an AST safety crash. Though I guess it would change the stable style under `--fast`? I would say that's OK but we haven't specified it explicitly.

hauntsaninja:
Oh, it's not any f-string with placeholders. It's f-strings with leading or trailing space with placeholders adjacent to that space.

hauntsaninja:
Actually wait, there is no stability concern. The stability guarantee is just that code that was previously formatted with Black will not change, not that formatting unformatted code will lead to the same results.

JelleZijlstra (approved):

Issue #3918, opened 2023-10-03
Formatting one-tuples as multi-line if already multi-line


bluetech:
**Describe the style change**

Black excludes one-tuples `(1,)` and single-item subscripts with trailing comma `tuple[int,]` from magic comma handling, because unlike in list literals etc. the comma here is required, so cannot be used to distinguish between the user's desire to single-line or multi-line.

The single-line format chosen by Black is the desired behavior for "actual" one-tuples, but is not the desired behavior for "currently 1 item but maybe more in the future" tuples.

**Examples in the current _Black_ style**

Given **input**:

```python
class WidgetAdmin(admin.ModelAdmin):
    readonly_fields = (
        'id',
    )
    fields = ('foo',)
```

the formatting is:

```python
class WidgetAdmin(admin.ModelAdmin):
    readonly_fields = ("id",)
    fields = ("foo",)
```

**Desired style**

I would like Black to have a special case for one-tuples (and one-subscripts), distinguishing between the newline and no-newline cases. Black would use the multiline format if the input is multiline.

```python
class WidgetAdmin(admin.ModelAdmin):
    readonly_fields = (
        "id",
    )
    fields = ("foo",)
```

This adds a new form of context sensitivity in addition to magic trailing comma, but I think it makes sense since it plugs a hole in the magic trailing comma handling.

**Additional context**

Working on moving some large projects to use Black, this is my major gripe. The problem with the current style is that it removes the git-diff friendliness of the multi-line format, which magic trailing comma normally handles nicely.

Common examples for us are in Django Admin classes (like above), another is in `Literal`s, but we have them in *a lot* of different cases.

Known workarounds:

1. Switch to list literals. Not great because lists are mutable, heavier than tuples, and sometimes a tuple specifically is needed.
2. Add a "forcing" comment:

    ```py
        readonly_fields = (
           "id",
            #
       )
    ```

    I think it's not very pretty.

Backward compat: it breaks compat in the sense that arbitrary input will get different output. But given already-formatted input, the output isn't changed. This adheres to the [Stability Policy](https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy) if I understand it correctly.

bluetech:
I looked into this briefly. AFAICT, the main problem with implementing this is that the AST does not preserve non-logical newlines, and there is no way to go from an AST node to the original tokens either. The `NL` token necessary for detecting this case is gone. So implementing this proposal requires changes to blib2to3.

[...]

PR #4023, opened 2023-11-04
Remove redundant condition from `has_magic_trailing_comma`

bluetech:
Simplify the `has_magic_trailing_comma` logic (it will make #3918 a bit easier to implement).

The second `if` cannot be true at its execution point, because it is already covered by the first `if`. The condition
`comma.parent.type == syms.subscriptlist` always holds if `closing.parent.type == syms.trailer` holds, because `subscriptlist` only appears inside `trailer` in the grammar:

```
trailer: '(' [arglist] ')' | '[' subscriptlist ']' | '.' NAME
subscriptlist: (subscript|star_expr) (',' (subscript|star_expr))* [',']
```

NOTE: I split the PR to two commits, the first one is just a refactor with no logical changes, but makes the argument above easier to follow. The commits should be squashed when merging.

### DIFF
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -363,25 +363,6 @@ class Line:
             ):
                 return False
 
-            if not ensure_removable:
-                return True
-
-            comma = self.leaves[-1]
-            if comma.parent is None:
-                return False
-
-            if (
-                comma.parent.type == syms.subscriptlist
-                and closing.opening_bracket is not None
-                and is_one_sequence_between(
-                    closing.opening_bracket,
-                    closing,
-                    self.leaves,
-                    brackets=(token.LSQB, token.RSQB),
-                )
-            ):
-                return False
-
             return True
 
         if self.is_import:
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -351,9 +351,9 @@ class Line:
 
         if closing.type == token.RSQB:
             if (
-                closing.parent
+                closing.parent is not None
                 and closing.parent.type == syms.trailer
-                and closing.opening_bracket
+                and closing.opening_bracket is not None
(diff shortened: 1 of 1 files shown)


bluetech:
cc @hauntsaninja - you added some of the condition in 4ebf14d17ed544be893be5706c02116fd8b83b4c, maybe you can poke a hole in my logic :)

Also c4bd2e31ceeac84d68592986fe70920f3d3d0443 made a change here (removing the `return comma.parent.type == syms.listmaker` part), I haven't looked into why though.

JelleZijlstra:
> Also [c4bd2e3](https://github.com/psf/black/commit/c4bd2e31ceeac84d68592986fe70920f3d3d0443) made a change here (removing the `return comma.parent.type == syms.listmaker` part), I haven't looked into why though.

Looks like the listmaker condition was only used if preview style for that year was off. Turning on the preview style by default meant that that branch was dead.

JelleZijlstra (approved):
Seems like you're correct, and diff-shades also found nothing changed by this refactor.

bluetech:
The branch is only relevant when `--skip-magic-trailing-comma` is used. I'm not sure whether diff-shades uses each project own config, and if so if any of the [projects](https://github.com/ichard26/diff-shades/blob/a53999fbec3ca34a7375cb2ec4e41269d0164388/src/diff_shades/config.py) use the flag (I can check if necessary).

There's always a risk with changes like this of unexpected breakage, so it'd be nice to get some assurance like diff-shades. If diff-shades doesn't cover it, I'll look for some projects which use `skip-magic-trailing-comma` and check them.

JelleZijlstra:
I'll run it on our internal codebase to check.

Issue #4033, opened 2023-11-08
`--line-ranges` bad behavior and internal errors


tartley:
**Describe the bug**

I can't get the recently-merged `--line-ranges` command line option to work at all. It either produces incorrect results, or produces an internal error.

**To Reproduce**

**Five lines of input causes an internal error**

(long lines of output, the error message & .log content, wrapped by me for readability)

```
$ cat unformatted.py 
print ( "format me" )
print ( "format me" )
print ( "format me" )
print ( "format me" )
print ( "format me" )
$ black --line-ranges=2-3 unformatted.py 
error: cannot format unformatted.py: INTERNAL ERROR: Black produced
 different code on the second pass of the formatter.  Please report a bug
 on https://github.com/psf/black/issues.  This diff might be helpful:
 /tmp/blk_fn8p09n6.log

Oh no! 💥 💔 💥
1 file failed to reformat.
$ echo $?
123
$ cat /tmp/blk_fn8p09n6.log
Mode(target_versions=set(),
 line_length=88,
 string_normalization=True,
 is_pyi=False,
 is_ipynb=False,
 skip_source_first_line=False,
 magic_trailing_comma=True,
 experimental_string_processing=False,
 python_cell_magics=set(),
 preview=False)
--- source
+++ first pass
@@ -1,5 +1,5 @@
 print ( "format me" )
+print("format me")
+print("format me")
 print ( "format me" )
-print ( "format me" )
-print ( "format me" )
-print ( "format me" )
+print("format me")
--- first pass
+++ second pass
@@ -1,5 +1,5 @@
 print ( "format me" )
 print("format me")
 print("format me")
-print ( "format me" )
 print("format me")
+print("format me")
```

**4 lines of input completes, but formats the wrong line range**

```bash
$ cat unformatted.py 
print ( "format me" )
print ( "format me" )
print ( "format me" )
print ( "format me" )
$ black --line-ranges=2-3 unformatted.py 
reformatted unformatted.py

All done! ✨ 🍰 ✨
1 file reformatted.
$ cat unformatted.py 
print ( "format me" )
print("format me")
print("format me")
print("format me")
```

**Expected behavior**

I expect only lines 2 and 3 to be formatted, but instead 2-4 have been formatted.

**Environment**

I'm using a version of Black cloned from Github's latest main:

```bash
$ black --version
black, 23.11.1.dev1+g58f31a7 (compiled: no)
Python (CPython) 3.10.12
```

[...]

PR #4034, opened 2023-11-09
Disable the stability check with --line-ranges for now.

yilei:
### Description

When performing a second formatting pass with `--line-ranges`, we can't simply use the original lines specified by the user.

For example, unformatted:

```python
def restrict_to_this_line(arg1,
  arg2,
  arg3):
    print  ( "This should not be formatted." )
    print  ( "This should not be formatted." )
```

If we let it format lines `1-3`, after the first pass it becomes:

```python
def restrict_to_this_line(arg1, arg2, arg3):
    print  ( "This should not be formatted." )
    print  ( "This should not be formatted." )
```

If we use the original `1-3` lines in a second pass, it would format all the lines now.

To solve this, we have an [`adjusted_lines`](https://github.com/psf/black/blob/58f31a70efe6509ce8213afac998bc5d5bb7e34d/src/black/ranges.py#L48) function to calculate the new line ranges for the second pass. It uses the diffing algorithm from `difflib.SequenceMatcher`. Unfortunately it could produce undesired results for certain edge cases like a list of unformatted lines with the same content.

I _think_ we could use a custom algorithm that takes advantage of the diff is produced by a formatter (it rarely adds/removes lines of code, except blank lines, optional parenthesis). But it's complicated so I'm proposing to simply disable the stability check for now to avoid the crash mentioned in #4033.

This could still lead to formatting more lines than requested, but it should be very rare in practice. Even when that happens, it usually doesn't hurt too much if a few extra lines are reformatted. Also, if we could get rid of the two-pass formatting, this will no longer be an issue. (We still need a better `adjusted_lines` to re-enable the stability check, but it should be a smaller issue.)

### Checklist - did you ...

### DIFF
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -196,7 +196,11 @@ Example: `black --line-ranges=1-10 --line-ranges=21-30 test.py` will format line
 
 This option is mainly for editor integrations, such as "Format Selection".
 
-Currently it also disables _Black_'s formatting stability check in `--safe` mode.
+```{note}
+Due to #4052, `--line-ranges` might format extra lines outside of the ranges when there
+are unformatted lines with the exact content. It also disables _Black_'s formatting
+stability check in `--safe` mode.
+```
 
 #### `--color` / `--no-color`
 
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -192,6 +192,8 @@ Example: `black --line-ranges=1-10 --line-ranges=21-30 test.py` will format line
 
 This option is mainly for editor integrations, such as "Format Selection".
 
+Currently it also disables _Black_'s formatting stability check in `--safe` mode.
+
 #### `--color` / `--no-color`
 
 Show (or do not show) colored diff. Only applies when `--diff` is given.
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -197,7 +197,7 @@ Example: `black --line-ranges=1-10 --line-ranges=21-30 test.py` will format line
 This option is mainly for editor integrations, such as "Format Selection".
 
 ```{note}
-Due to #4052, `--line-ranges` might format extra lines outside of the ranges when there
+Due to [#4052](https://github.com/psf/black/issues/4052), `--line-ranges` might format extra lines outside of the ranges when there
 are unformatted lines with the exact content. It also disables _Black_'s formatting
 stability check in `--safe` mode.
 ```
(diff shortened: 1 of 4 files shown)


tests/data/cases/line_ranges_diff_edge_case.py
```diff
@@ -0,0 +1,28 @@
...
+# flag above as it's formatting specifically these lines.
+
+# Reproducible example for https://github.com/psf/black/issues/4033.
+# This can be fixed in the future if we use a better diffing algorithm, or make Black
+# perform formatting in a single pass.
+
+print ( "format me" )
+print("format me")
+print("format me")
+print("format me")
```
JelleZijlstra, line 27: Why are these lines formatted? Shouldn't it only format the second and third prints?

yilei: This is caused exactly by the diff algorithm used in `adjusted_lines`. In the second formatting pass, the adjusted lines becomes 12-13 where it should stayed as 10-11.

This PR doesn't fix this underlying issue (I'll file a separate issue if this PR is merged). Until it's really fixed, I'm proposing to skip the stability check (affected by the same issue) to avoid the crash.

The edge case is very rare in real code as it requires many same unformatted lines.

JelleZijlstra: I see, thanks. Maybe in that case we should put a warning in the docs for now about this bug.

JelleZijlstra: Agree that this doesn't feel very likely in real code, though @tartley apparently did run into this.

yilei: Updated the doc.

@tartley can correct me, but #4033 appears to be a "toy" example for just trying out the new `--line-ranges` feature. #4052 can also serves the purpose of collecting real world example in case it happens.

What do you think?

docs/usage_and_configuration/the_basics.md
```diff
@@ -196,6 +196,12 @@ Example: `black --line-ranges=1-10 --line-ranges=21-30 test.py` will format line
 
 This option is mainly for editor integrations, such as "Format Selection".
 
+```{note}
+Due to #4052, `--line-ranges` might format extra lines outside of the ranges when there
```
JelleZijlstra, line 200:
```suggestion
Due to [#4052](https://github.com/psf/black/issues/4052), `--line-ranges` might format extra lines outside of the ranges when there
```
I don't think this will autolink

JelleZijlstra (approved):

tests/data/cases/line_ranges_diff_edge_case.py
tartley: Agreed that I only encountered the problem using toy example code. By bad luck the toy example was the first thing I tried `--line-ranges` out on, while trying to integrate the use of it into my editor (e.g. so Vim would call Black on the selected range of lines). Although only a toy example it did genuinely confuse the heck out of me, and convinced me at the time that `--line-ranges` wasn't working for me at all.

PR #4038, opened 2023-11-09
Format one-tuples and one-subscripts as multi-line if already multi-line

bluetech:
This implement the proposal in #3918.

I wasn't sure if this needs to go as a preview option; it shouldn't affect already formatted code (so adheres to the Stability Policy), but it does change how unformatted input is formatted.

Split to multiple commits to make reviewing a bit easier. Can be squashed to a single commit. The actual meaningful change is the last commit.

I've used `opening.lineno != closing.lineno` as an easy way to detect "multiline". I think it's correct and simple but maybe it's not appropriate.

Tested on some of our projects and it has the intended effect. For `skip-magic-trailing-comma = true` configurations it should have no effect at all.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,10 +10,9 @@
 
 <!-- Changes that affect Black's stable style -->
 
-- Magic trailing commas are now respected in one-tuples and single-element
-  subscripts when they are already split to multiple lines. So a multiline
-  one-tuple is no longer collapsed to a single line when magic trailing
-  commas are enabled. (#4038)
+- Magic trailing commas are now respected in one-tuples and single-element subscripts
+  when they are already split to multiple lines. So a multiline one-tuple is no longer
+  collapsed to a single line when magic trailing commas are enabled. (#4038)
 
 ### Preview style
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,11 @@
 
 <!-- Changes that affect Black's stable style -->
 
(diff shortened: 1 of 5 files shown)


bluetech:
I started looking into the SQLAlchemy diff-shades stability failure. I minimized it to the following failure:

```diff
Mode(target_versions={<TargetVersion.PY38: 8>}, line_length=79, string_normalization=True, is_pyi=False, is_ipynb=False, skip_source_first_line=False, magic_trailing_comma=True, experimental_string_processing=False, python_cell_magics=set(), preview=True)
--- source
+++ first pass
@@ -1,8 +1,6 @@
 if True:
     if True:
         if ooooooooooooooooooooooooo and oooooooooooooooooo[ooo].oooooooo == (
             0,
         ):
-            Baz(
-                ""
-            )
+            Baz("")
--- first pass
+++ second pass
@@ -1,6 +1,6 @@
 if True:
     if True:
-        if ooooooooooooooooooooooooo and oooooooooooooooooo[ooo].oooooooo == (
-            0,
-        ):
+        if ooooooooooooooooooooooooo and oooooooooooooooooo[
+            ooo
+        ].oooooooo == (0,):
             Baz("")
```

However when I try to understand what happens in the second pass by running black on the output of the first pass, I get no changes. So I'm a bit stumped; maybe I misunderstood how it works and the second pass is *not* just running black again on the output of the first pass? I'll try figure it out, in the meantime I'll mark the PR as draft.

PR #3940, opened 2023-10-10
confine pre-commit to stages

davidculley:
### Description

See https://pre-commit.com/#confining-hooks-to-run-at-certain-stages

> If you are authoring a tool, it is usually a good idea to provide an appropriate `stages` property. For example a reasonable setting for a linter or code formatter would be `stages: [pre-commit, pre-merge-commit, pre-push, manual]`.

### Checklist - did you ...

- [x] Add an entry in `CHANGES.md` if necessary?
- [ ] Add / update tests if necessary?
- [ ] Add new / update outdated documentation?

PR #4041, opened 2023-11-13
Fix support for pre-commit versions < 3.2.0

andrewmwilson:
### Description

This [previous change](https://github.com/psf/black/pull/3940/files) broke support for versions of pre-commit earlier than 3.2.0.

As far as I can tell, there's no reason _not_ to leave these as `commit`, `merge-commit` and `push` for now. Eventually they will need to be updated, but the current version of pre-commit supports both sets of values.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,7 +40,7 @@
 
 ### Integrations
 
-- Fix support for pre-commit versions < 3.2.0 (#4040)
+- Fix support for pre-commit versions < 3.2.0 (#4041)
 
 ### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,7 +40,7 @@
 
 ### Integrations
 
-<!-- For example, Docker, GitHub Actions, pre-commit, editors -->
+- Fix support for pre-commit versions < 3.2.0 (#4040)
 
 ### Documentation
 
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,7 +4,7 @@
   name: black
   description: "Black: The uncompromising Python code formatter"
   entry: black
-  stages: [pre-commit, pre-merge-commit, pre-push, manual]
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -40,7 +40,7 @@
 
 ### Integrations
 
-<!-- For example, Docker, GitHub Actions, pre-commit, editors -->
+- Fix support for pre-commit versions < 3.2.0 (#4040)
```
cooperlees, line 43:
```suggestion
- Fix support for pre-commit versions < 3.2.0 (#4041)
```

davidculley:
> As far as I can tell, there's no reason _not_ to leave these as `commit`, `merge-commit` and `push` for now.

Agreed.

> the current version of pre-commit supports both sets of values.

True.

> Eventually they will need to be updated

Yes, probably sometime in mid-to-late 2024, see https://github.com/psf/black/pull/4067#issuecomment-1822673095

owenlamont:
Can this PR be merged soon? I'm keen to get this fix and aren't in a position to update pre-commit in the environment I'm using.

hauntsaninja:
Thanks for the PR! However, this situation seems pretty annoying, since specifying these values sounds like it will just break again with a newer pre-commit in a few months. I think we should just revert https://github.com/psf/black/pull/3940 , afaict users who care about specific stages can just specify that directly in their own config. Does that sound reasonable, @davidculley? Maybe there's some motivation for that change that I'm missing

hauntsaninja:
Also note for anyone currently broken by this, you can try using https://github.com/psf/black-pre-commit-mirror instead. This never received that change, so should still work. It'll also make black like 2x faster :-)

davidculley:
> I think we should just revert #3940 , afaict users who care about specific stages can just specify that directly in their own config. Does that sound reasonable, @davidculley? Maybe there's some motivation for that change that I'm missing

Yes, that sounds reasonable to me.

PR #4045, opened 2023-11-13
add trunk instructions

joshmarinacci:
### Description

Hi. I'm an engineer at Trunk, which uses and automates Black. This PR is just to add instructions for how to use Black if you are using our free CLI. Please let me know if this isn't appropriate. Thanks!


hauntsaninja:
Thanks, this isn't appropriate

PR #4049, opened 2023-11-18
Upgrade mypy to 1.6.1

AlexWaygood:
### Description

This PR bumps the pinned version of mypy to 1.7.0. Since mypy 1.7.0 includes https://github.com/python/mypy/commit/940fceb59a55513913c3c9c1eaf89a7f5ee774d6, this should unblock us from creating mypyc-compiled wheels for Python 3.12 (though I don't do that here; it's deferred to a followup PR).

### Checklist - did you ...

### DIFF
--- a/tests/optional.py
+++ b/tests/optional.py
@@ -28,7 +28,7 @@ except ImportError:
     # pytest < 7
     #
     # "isort: skip" is needed or it moves the "type: ignore" to the following line
-    # becuase of the line length, and then mypy complains.
+    # because of the line length, and then mypy complains.
     # Of course, adding the "isort: skip" means that
     # flake8-bugbear then also complains about the line length,
     # so we *also* need a "noqa" comment for good measure :)
--- a/tests/optional.py
+++ b/tests/optional.py
@@ -26,7 +26,15 @@ try:
     from pytest import StashKey
 except ImportError:
     # pytest < 7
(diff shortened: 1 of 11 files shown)


AlexWaygood:
The CI failures look like https://github.com/python/mypy/issues/16480. I guess we'll have to wait for mypy 1.7.1, which will (I assume) include https://github.com/python/mypy/pull/16484

AlexWaygood:
Okay, for now I've changed the PR so that it only upgrades mypy to 1.6.1. That means that it no longer unblocks building mypycified wheels for Python 3.12, since mypy 1.6.1 doesn't include https://github.com/python/mypy/commit/940fceb59a55513913c3c9c1eaf89a7f5ee774d6. But merging this PR will mean there will be fewer changes necessary to upgrade mypy to 1.7.1, once mypy 1.7.1 is released.

tests/optional.py
```diff
@@ -26,7 +26,15 @@
     from pytest import StashKey
 except ImportError:
     # pytest < 7
-    from _pytest.store import StoreKey as StashKey  # type: ignore[import, no-redef]
+    #
+    # "isort: skip" is needed or it moves the "type: ignore" to the following line
+    # becuase of the line length, and then mypy complains.
```
JelleZijlstra, line 31:
```suggestion
    # because of the line length, and then mypy complains.
```

JelleZijlstra (approved):

PR #4051, opened 2023-11-18
Block aiohttp==3.9.0 from being installed in CI on Windows/pypy

AlexWaygood:
### Description

`aiohttp==3.9.0` fails to import if you're using PyPy on Windows: see https://github.com/aio-libs/aiohttp/issues/7848. This is currently causing our Windows/PyPy CI job to fail on every PR. We're using PyPy3.8 in CI, but bumping to PyPy3.9 in CI doesn't fix the problem: see the failing CI on the latest commit of #4035.

This PR just blocks `aiohttp==3.9.0` from being installed in CI, to make things green again.

### Checklist - did you ...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,7 @@ dynamic = ["readme", "version"]
 colorama = ["colorama>=0.4.3"]
 uvloop = ["uvloop>=0.15.2"]
 d = [
-  "aiohttp>=3.7.4",
+  "aiohttp>=3.7.4, !=3.9.0",
 ]
 jupyter = [
   "ipython>=7.8.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,8 @@ dynamic = ["readme", "version"]
 colorama = ["colorama>=0.4.3"]
 uvloop = ["uvloop>=0.15.2"]
 d = [
-  "aiohttp>=3.7.4, !=3.9.0",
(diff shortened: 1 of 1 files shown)


cooperlees:
I think it's better to just not test for pypy (even just on windows) for now if we can easily in the action instead ... Is that doable?

AlexWaygood:
I could do something like this, so that aiohttp==3.9.0 is only banned for the Windows/PyPy combination?

```diff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,8 @@ dynamic = ["readme", "version"]
 colorama = ["colorama>=0.4.3"]
 uvloop = ["uvloop>=0.15.2"]
 d = [
-  "aiohttp>=3.7.4, !=3.9.0",
+  "aiohttp>=3.7.4; sys_platform != 'win32' or implementation_name != 'pypy'",
+  "aiohttp>=3.7.4, !=3.9.0; sys_platform == 'win32' and implementation_name == 'pypy'",
 ]
 jupyter = [
   "ipython>=7.8.0",
```

cooperlees:
Damn, that's using all the specifiers! Nice.

I'll merge that so we block it from as little as possible to maybe help others + potentially find more bugs ...

cooperlees (approved):
Version PEP440 specifier soup.

Accepting as this will be temporary until we get a fix into aiohttp 3.9.X.

PR #4053, opened 2023-11-18
[ci] Move 'lint' to 3.12

### LINKED ISSUE #3983: Make blackd + CI work in 3.12
**Describe the bug**

Our CI needs [d] extra install which brings in aiohttp. Due to this we don't work in 3.12

aiohtp are working on their 3.12 release: https://github.com/aio-libs/aiohttp/issues/7675
- They have found bugs due to things being fixed in 3.12 ans a regression in one of their deps

**To Reproduce**

Install black[d] in python 3.12

**Expected behavior**

We install and CI passes

**Environment**

- Black's version: main
- OS and Python version: Any OS + 3.12

This will then unblock us to completely work with 3.12

cooperlees:
- Add to run on MacOS + Windows too
- Do not install [d] depdendecies as blackd is not actually ran / checked
- Move to default GitHub action version - which is 3.12 today

Fixes #3983

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,4 +1,4 @@
-name: Lint
+name: Lint + format ourselves
 
 on: [push, pull_request]
 
@@ -11,7 +11,11 @@ jobs:
       github.event_name == 'push' || github.event.pull_request.head.repo.full_name !=
       github.repository
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macOS-latest, windows-latest]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):

.github/workflows/lint.yml
```diff
@@ -11,7 +11,11 @@ jobs:
       github.event_name == 'push' || github.event.pull_request.head.repo.full_name !=
       github.repository
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macOS-latest, windows-latest]
```
JelleZijlstra, line 18: It doesn't actually work on Windows: https://github.com/psf/black/actions/runs/6939494268/job/18876939284?pr=4060. I don't think we gain much from running this on multiple OSes so I'll just revert this part of the change.

cooperlees: I thought I saw this pass ...

The benefit is the run self running on the different OS's and it also lets us know our linting has a good chance to run for developers on every OS ...

PR #4060, opened 2023-11-21
Allow empty lines at beginning of blocks (again)

### LINKED ISSUE #4043: Newlines after code block open, once and for all?
Background:
- https://github.com/psf/black/issues/902
  Original issue, with some follow-up comments post merge
- https://github.com/psf/black/pull/3035
  Original implementation
- https://github.com/psf/black/issues/3551
  Follow-up issue
- https://github.com/psf/black/pull/3967 
  Recent preview change partially walking things back
- https://github.com/psf/black/issues/619#issuecomment-1792996633
  Discussion about PEP 8

---

tldr; the 2023 stable style strips newlines after code block opens. This has proven controversial (and was most of the change from 22 to 23 by volume). Recently, we walked this back a little bit in preview style, but we should try to get it right for 24.

The main argument against is it can hurt readability, see https://github.com/psf/black/issues/902#issuecomment-1425713025 . I'm quite sympathetic to this. For instance, functions with a complicated return type and first statement can get quite dense. Similarly, there are examples with comments in a new code block where things can get too dense.

An auxiliary argument against it is that stripping new lines between `class` and first method is in violation of a close reading and changelog analysis of PEP 8, see https://github.com/psf/black/issues/619#issuecomment-1792996633. I don't care much for PEP 8 absolutism (and certainly not at the expense of churning established behaviour), but I consider it a nice-to-have to not reformat away from it.

Recently in https://github.com/psf/black/pull/3967, Henri contributed a nice change that walks this back in some cases, involving comments or compound statements. However, I'm not sure this is enough. [Here are two examples from the first few pages of Black 22 -> 23 diff in my work codebase where I feel readability really suffers](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ARDAb9dAD2IimZxl1N_WlbvK5V-fnaWIwL5TYBadve5dORVlMpn2CM7Ex6036-fmBvZZxxU7M4SzueDveXQyyVCyVj-_SI8G5MhyXK3nNLHUmR1qpfo3F4ciXU1BMotQs0XgWgkBVKnr_rKC7aIr11sgQsmB5Vap7pzA0R4M5007ghia_366U1-hEjwpaB66-Lkb9ZsEQGd_I3HS8XCUyn1O9pe47U_JM9qrygevkU8y8MHi42r1FAGT-KAUg8Rw7JMVMPhspr1cpS3BzKbGToLSkpjYcCJCj70GUsfc-SKPPQf6n9nHSnwHTurwtoO8zChj8UkwEuEK_PzkD2G_pjTZTi2pwy8tqqwW4_Sg7NqddQaPn1JDfnMvajayo16d3g90jDv7alCLxL4CghJEpl5sU2gaWINat_wzfhsHjDOd77doe6GceCBcWuklNcdz1KAT8N7e_h68QFQcENH51Z8nzi6Nsk0ViOJ7WcYti9iYfIqrura2wKKSmLKz1FVErFWWojvd07WdSB1M2nGrsld_aGwOC7R7wb84ks5wpovY1jSZdnLj1QgrWWAB8ns3Zvi7HnLsmBGxIFV2EtqyAyKkQswAAAAs1goWJRFiskAAdsDxAgAAAhqMAuxxGf7AgAAAAAEWVo=) (scrolling through I see many others). Henri's change won't save us here. And fundamentally this code is quite simple, it just has slightly long type and variable names, i.e. it's not the kind of code where there should have been a comment or something.

---

Jelle nudged me to think about this, and I now feel we should just walk this change back entirely. We should allow users to have either 0 or 1 blank newlines. This is also simpler and more explainable than the partially walked back logic.

I am fine with continuing to strip blank newlines between a function declaration and its docstring (and preview does this for classes as well). We currently enforce one blank newline after class and module docstrings. We don't have an opinion on blank newlines after function docstrings (see https://github.com/psf/black/issues/2370). I think this is probably fine as-is and not having an opinion would be consistent with the proposed no docstring case. (I'll grant that it feels easier to come up with a good opinion for the post docstring blank line in functions. At the very least, we certainly should not resolve that issue by always removing blank lines).

This will not create any churn going from 23 -> 24 (in fact, most of this change technically even meets our stability policy (although any change here should certainly only be made in preview style)).

---

As a meta thing, as a relatively new maintainer of Black, I'm still unsure of my decision making framework. I feel the original change had at most marginal benefit, especially given its churn (I've never browsed code and been like ugh one extra blank line, nor has any code reviewer or linter ever yelled at me for it. These are the things Black usually saves me from. All the motivating changes I found have multiple blank newlines or are toy code). On the other hand, it's very costly to revert changes in stable style. The only thing that makes it possible in this case is that both unreverted and reverted formatting will be valid and inconsistency in this specific case is not costly.

JelleZijlstra:
Fixes #4043

### DIFF
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -684,7 +684,8 @@ class EmptyLineTracker:
                 return 0, 1
             return before, 1
 
-        # In preview mode, always allow blank lines, except right before a function docstring
+        # In preview mode, always allow blank lines, except right before a function
+        # docstring
         is_empty_first_line_ok = (
             Preview.allow_empty_first_line_in_block in current_line.mode
             and (
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -684,14 +684,17 @@ class EmptyLineTracker:
                 return 0, 1
             return before, 1
 
-        is_empty_first_line_ok = Preview.allow_empty_first_line_in_block in current_line.mode and (
-            not is_docstring(current_line.leaves[0])
-            # If it's a triple quote comment (but not at the start of a funcdef)
-            or (
-                self.previous_line
-                and self.previous_line.leaves[0]
-                and self.previous_line.leaves[0].parent
-                and not is_funcdef(self.previous_line.leaves[0].parent)
+        # In preview mode, always allow blank lines, except right before a function docstring
+        is_empty_first_line_ok = (
+            Preview.allow_empty_first_line_in_block in current_line.mode
+            and (
+                not is_docstring(current_line.leaves[0])
+                or (
+                    self.previous_line
+                    and self.previous_line.leaves[0]
+                    and self.previous_line.leaves[0].parent
+                    and not is_funcdef(self.previous_line.leaves[0].parent)
+                )
             )
         )
 
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -684,18 +684,14 @@ class EmptyLineTracker:
                 return 0, 1
             return before, 1
 
-        is_empty_first_line_ok = (
-            Preview.allow_empty_first_line_in_block
-            in current_line.mode
-            and (
-                not is_docstring(current_line.leaves[0])
-                # If it's a triple quote comment (but not at the start of a funcdef)
-                or (
-                    self.previous_line
-                    and self.previous_line.leaves[0]
-                    and self.previous_line.leaves[0].parent
-                    and not is_funcdef(self.previous_line.leaves[0].parent)
-                )
+        is_empty_first_line_ok = Preview.allow_empty_first_line_in_block in current_line.mode and (
+            not is_docstring(current_line.leaves[0])
+            # If it's a triple quote comment (but not at the start of a funcdef)
+            or (
+                self.previous_line
+                and self.previous_line.leaves[0]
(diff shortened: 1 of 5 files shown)


src/black/lines.py
```diff
@@ -684,18 +684,14 @@ def _maybe_empty_lines(self, current_line: Line) -> Tuple[int, int]:
                 return 0, 1
             return before, 1
 
+        # In preview mode, always allow blank lines, except right before a function
```
hauntsaninja, line 687: Btw a few lines up we have `and current_line.is_triple_quoted_string`. Should that also use `is_docstring`? Seems preferable to have consistent behaviour for single quote docstrings as well

hauntsaninja (approved):

src/black/lines.py
JelleZijlstra: Yeah I don't really understand why we sometimes use one and sometimes the other. I'll merge this first and then maybe make another preview style change harmonizing between these two.

return42:
@JelleZijlstra thank you for taking the efforts / with your help we now _"back to black"_ :)

- https://github.com/searxng/searxng/pull/3309

Issue #4063, opened 2023-11-21
Preview style: Don't collapse dummy implementation with comment


JelleZijlstra:
I found this change in our test suite while trying to stabilize the preview style:

```
 match command.split():
     case ["north"] | ["go", "north"]:
         current_room = current_room.neighbor("north")
-    case ["get", obj] | ["pick", "up", obj] | ["pick", obj, "up"]:
-        ...  # Code for picking up the given object
+    case (
+        ["get", obj] | ["pick", "up", obj] | ["pick", obj, "up"]
+    ): ...  # Code for picking up the given object
```

This is from the `dummy_implementation` preview change, #3796.

I think the new code looks worse here; we shouldn't use the dummy_implementation branch if there is a comment after the ellipsis.

JelleZijlstra:
I tried to change this but it causes some problems in typeshed:

```
-    def __mul__(cls: type[_CT], other: int) -> type[Array[_CT]]: ...  # type: ignore[misc]  # pyright: ignore[reportGeneralTypeIssues]
-    def __rmul__(cls: type[_CT], other: int) -> type[Array[_CT]]: ...  # type: ignore[misc]  # pyright: ignore[reportGeneralTypeIssues]
+    def __mul__(cls: type[_CT], other: int) -> type[Array[_CT]]:
+        ...  # type: ignore[misc]  # pyright: ignore[reportGeneralTypeIssues]
+
+    def __rmul__(cls: type[_CT], other: int) -> type[Array[_CT]]:
+        ...  # type: ignore[misc]  # pyright: ignore[reportGeneralTypeIssues]
```

Will have to think more about the right approach here.

Also considering to turn off the dummy_implementations change for suites that aren't functions or classes.

JelleZijlstra:
Sounds good, I updated the PR to apply the dummy impl logic only to classes and functions.

Issue #4062, opened 2023-11-21
Crash with preview style + line length 1


JelleZijlstra:
On Black 23.11.0:

```
% black -v --line-length 1 -c 'def f() -> tuple[int, int]: pass' --preview
def f() -> tuple[int, int]: pass
Traceback (most recent call last):
  File "src/black/__init__.py", line 1444, in assert_equivalent
  File "src/black/parsing.py", line 140, in parse_ast
SyntaxError: expected ':' (<unknown>, line 1)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "src/black/__init__.py", line 782, in reformat_code
  File "src/black/__init__.py", line 945, in format_stdin_to_stdout
  File "src/black/__init__.py", line 1011, in format_file_contents
  File "src/black/__init__.py", line 985, in check_stability_and_equivalence
  File "src/black/__init__.py", line 1447, in assert_equivalent
AssertionError: INTERNAL ERROR: Black produced invalid code: expected ':' (<unknown>, line 1). Please report a bug on https://github.com/psf/black/issues.  This invalid output might be helpful: /var/folders/hh/hdb5qd2x3v51v6l0gknpfjfr0000gp/T/blk_wad8_c6m.log
error: cannot format <string>: INTERNAL ERROR: Black produced invalid code: expected ':' (<unknown>, line 1). Please report a bug on https://github.com/psf/black/issues.  This invalid output might be helpful: /var/folders/hh/hdb5qd2x3v51v6l0gknpfjfr0000gp/T/blk_wad8_c6m.log
Oh no! 💥 💔 💥
```

Works without preview:

```
% black -v --line-length 1 -c 'def f() -> tuple[int, int]: pass'

def f() -> (
    tuple[
        int,
        int,
    ]
):
    pass
reformatted <string>
All done! ✨ 🍰 ✨
```

You can also make it crash with a slightly more reasonable line length if you make the identifiers longer:

```
% black -v --line-length 20 -c 'def ffffffffffffff() -> tupleeeeee[int, iiiiiiint]: pass' --preview
def ffffffffffffff() -> tupleeeeee[int, iiiiiiint]: pass
Traceback (most recent call last):
  File "src/black/__init__.py", line 1444, in assert_equivalent
  File "src/black/parsing.py", line 140, in parse_ast
SyntaxError: expected ':' (<unknown>, line 1)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

Issue #4042, opened 2023-11-13
Setting the 2024 stable style


hauntsaninja:
See issue for 2023 here: https://github.com/psf/black/issues/3407

Here's what this issue is:
- I've listed all the features we have enabled under `--preview`
- ...in descending order of how controversial I think they are
- I've also drawn a dividing line; past "uncontroversial" is where I feel we don't need to think twice about it
- I've listed the number of digits in the diff-shade lines changed count and linked original PR
- I tried to list any open issues next to the relevant preview feature, but it's possible I missed something
  - <s>Issues tagged with `preview` I haven't yet associated with a feature: #3614 and #3802</s> (sorted these out–Jelle)

Here's the current call to action:
- Any comments / opinions / vibes on these features is very welcome
- Please test your codebase with `--preview` and let us know your thoughts!
- If you have a specific problem with a preview feature, especially something that feels like an implementation issue, it's best to open a separate issue (that we can then link to from here)

For many of the changes, diff-shades logs have expired. We should find a way to keep these around for longer!

It's currently November, so there's still time to make more uncontroversial changes and hopefuly iron out some issues in the not uncontroversial ones.

#### string_processing

This one is controversial, vast majority of preview style issues are related to string processing
Open issues I believe are caused by string_processing: #2334, #2314, #3466, #3497, #3409, #3424, #3428, #3210, #3410, #3747, #3665, #3663, #2553, #3855, #3802

#### hug_parens_with_braces_and_square_brackets (#3964, #3992)

Causes O(10000) changes, although counting lines changed is unfavourable to this change.
A very substantial change in style for Black. We can't revert once it's in stable.
Outstanding crash: #4036

#### parenthesize_conditional_expressions (#2278)

Causes O(1000) changes. Need to look at diff
~Has unfortunate interactions with comments: #3555~

#### wrap_long_dict_values_in_parens (#3440)

Causes O(100) changes. Need to look at diff
~Outstanding crash: #3758~
Report of worse splitting: #3452
~Unnecessary nesting: #4129~
Removes parentheses: #4158

#### multiline_string_handling (#1879)

Causes O(1000) changes. Need to look at diff
Maybe has some synergy with `hug_parens_with_braces_and_square_brackets`.
Unfortunate white space preservation: #4159

#### dummy_implementations (#3796)

I can't find up-to-date diff-shades, but maybe O(1000) changes
Requires a change in flake8 config: #3887
Can result in an extra line, but no change expected to be made: #3872

[...]

PR #4064, opened 2023-11-21
First draft for 2024 stable style

JelleZijlstra:
See #4042.

I enabled almost all current preview features based on the discussion in #4042.

I didn't yet:
- Fix the entire test suite. Partly because I found a few bugs while doing this (#4062, #4063), partly because I was lazy.
- Remove `# flags: --preview` from test cases that no longer need it

However, this should be enough to get useful diff-shades feedback.


JelleZijlstra:
Moving to #4106

Issue #4065, opened 2023-11-21
pre-commit: "Expected one of commit, commit-msg, manual, ..." since 23.11.0


martin-thoma:
**Describe the bug**

When the pre-commit hook is installed and I commit, I get:

```
An error has occurred: InvalidManifestError: 
==> File /home/martin/.cache/pre-commit/repolc31_gpb/.pre-commit-hooks.yaml
==> At Hook(id='black')
==> At key: stages
==> At index 0
=====> Expected one of commit, commit-msg, manual, merge-commit, post-checkout, post-commit, post-merge, post-rewrite, prepare-commit-msg, push but got: 'pre-commit'
```

**To Reproduce**

```
$ python --version
3.11.1

$ python -m venv venv
$ source venv/bin/activate

$ vim .pre-commit-config.yaml
# Add this:
-   repo: https://github.com/psf/black
    rev: 23.11.0
    hooks:
    -   id: black  

$ echo "a = 'b'" > c.py

$ git init
$ git add .pre-commit-config.yaml c.py
$ pip install pre-commit
$ pre-commit install

$ pre-commit --version
pre-commit 3.5.0

$ git commit -m "Example"
```

The resulting error is:

```
An error has occurred: InvalidManifestError: 
==> File /home/martin/.cache/pre-commit/repolc31_gpb/.pre-commit-hooks.yaml
==> At Hook(id='black')
==> At key: stages
==> At index 0
=====> Expected one of commit, commit-msg, manual, merge-commit, post-checkout, post-commit, post-merge, post-rewrite, prepare-commit-msg, push but got: 'pre-commit'
```

**Expected behavior**

A pre-commit fixes `c.py`

**Environment**

- Black's version: 23.11.0
- OS and Python version: Linux / Python 3.11.1

**Additional context**

https://github.com/psf/black/pull/3940

MT-Cash:
The [`minimum_pre_commit_version`](https://pre-commit.com/#top_level-minimum_pre_commit_version) should get updated to 3.2.0 at least.

davidculley:
Does executing [`pre-commit clean`](https://pre-commit.com/#pre-commit-clean) fix the issue?

davidculley:
[...]

PR #4067, opened 2023-11-21
fix(pre-commit): raise minimum_pre_commit_version (#4065)

davidculley:
### Description

The changes made in Pull Request #3940 require version 3.2.0 of `pre-commit` (or newer). See https://github.com/psf/black/pull/3940#issuecomment-1820908057

We should also raise `minimum_pre_commit_version` if we want to go from the proposal made in #4041

```yaml
stages: [commit, merge-commit, push, manual]
```

to

```yaml
stages: [pre-commit, pre-merge-commit, pre-push, manual]
```

### Checklist - did you ...

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -6,7 +6,7 @@
   entry: black
   stages: [pre-commit, pre-merge-commit, pre-push, manual]
   language: python
-  minimum_pre_commit_version: 2.9.2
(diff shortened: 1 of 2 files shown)


MT-Cash:
This would fix https://github.com/psf/black/issues/4065

AlexWaygood:
#4041 is the better solution, no?

PR #4076, opened 2023-11-28
docs: Unify option descriptions between `--help` and `the_basics.md`

RedGuy12:
### Description

The descriptions for some options were inconsistent between the `--help` command and the `the_basics` documentation page. This PR unifies the two, as well as making the order consistant. Finally, the `-x` option was previously not mentioned at all in the documentation, so I added it.

### Checklist - did you ...

- [x] Add / update tests if necessary? (N/A)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@
 
 - `--line-ranges` now skips _Black_'s internal stability check in `--safe` mode. This
   avoids a crash on rare inputs that have many unformatted same-content lines. (#4034)
+- Clarify some changelog entries (#4075)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,7 +23,6 @@
 
 - `--line-ranges` now skips _Black_'s internal stability check in `--safe` mode. This
   avoids a crash on rare inputs that have many unformatted same-content lines. (#4034)
-- Clarify some changelog entries (#4076)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,7 +23,7 @@
 
 - `--line-ranges` now skips _Black_'s internal stability check in `--safe` mode. This
   avoids a crash on rare inputs that have many unformatted same-content lines. (#4034)
-- Clarify some changelog entries (#4075)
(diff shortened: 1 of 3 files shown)


src/black/__init__.py
```diff
@@ -270,7 +271,7 @@ def validate_regex(
     "-x",
     "--skip-source-first-line",
     is_flag=True,
-    help="Skip the first line of the source code.",
+    help="Skip formatting the first line of the source code.",
```
JelleZijlstra, line 274: I think the previous wording is better. We don't just skip formatting the first line; we skip even parsing it.

CHANGES.md
```diff
@@ -23,6 +23,7 @@
 
 - `--line-ranges` now skips _Black_'s internal stability check in `--safe` mode. This
   avoids a crash on rare inputs that have many unformatted same-content lines. (#4034)
+- Clarify some changelog entries (#4075)
```
JelleZijlstra, line 26:
```suggestion
- Clarify some changelog entries (#4076)
```
Should be the PR number, not issue number

src/black/__init__.py
```diff
@@ -404,26 +416,30 @@ def validate_regex(
...
 )
 @click.option(
     "-q",
     "--quiet",
     is_flag=True,
     help=(
-        "Don't emit non-error messages to stderr. Errors are still emitted; silence"
-        " those with 2>/dev/null."
+        "Stop emitting all non-critical output. Error messages will still be emitted"
+        " (which can Error messages will still be emitted (which can silenced by"
```
JelleZijlstra, line 431: This looks garbled

src/black/__init__.py
```diff
@@ -404,26 +416,30 @@ def validate_regex(
...
 )
 @click.option(
     "-v",
     "--verbose",
     is_flag=True,
     help=(
-        "Also emit messages to stderr about files that were not changed or were ignored"
-        " due to exclusion patterns."
+        "Emit messages about files that were not changed or were ignored due to"
+        " exclusion patterns. If Black is using a configuration file, a blue message"
```
JelleZijlstra, line 441: I feel like we shouldn't specify color in our text.

JelleZijlstra (approved):

Issue #4068, opened 2023-11-21
`--line-ranges` dedents a `# fmt: off` in the middle of a decorator


yilei:
**Describe the bug**

Given the following code:

```python
@decor(
    a = 1,
    # fmt: off
    b = (2, 3),
    # fmt: on
)
def func():
    pass
```

Run black with

```sh
$ black --line-ranges=8-8 file.py
```

Line 8 is `pass`, and the file will be formatted to:

```python
@decor(
    a = 1,
# fmt: off
    b = (2, 3),
    # fmt: on
)
def func():
    pass
```

**Expected behavior**

No formatting changes.

**Environment**

- Black's version: [main]

**Additional context**

I'll take a look at this, I think we need to save additional info when code is converted to `STANDALONE_COMMENT` for `# fmt: off/skip` markers before we convert lines outside of `--line-ranges=` to `STANDALONE_COMMENT`.

PR #4084, opened 2023-12-03
[4068] Fix:  --line-ranges dedents a # fmt: off in the middle of a decorator

riyaz489:
### Description

When transforming `fmt:off` into **standalone** **comments**, a new node is generated for `fmt:off`, and in this procedure, the initial indentation of `fmt:off` is not retained. Black code consistently converts `fmt:off/on` into standalone comments before proceeding with additional formatting. Consequently, the formatting of `fmt:off` is disrupted, even when it is not within the specified line range.

I **addressed** this issue by retaining the indentation of `fmt:off` and incorporating it into the prefix of the new node

**Additionally**, I observed that black consistently formats comments, regardless of whether `fmt:off` is present or if the comments fall outside the specified line range. 
I refrained from raising an issue for this because I was uncertain whether it was intended behaviour or not.

Fixes:  [#4068](https://github.com/psf/black/issues/4068)

### Checklist 

- [ x] Add an entry in `CHANGES.md` if necessary?
- [ x] Add / update tests if necessary?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,8 @@
 
 ### Stable style
 
-<!-- Changes that affect Black's stable style -->
+- Fix bug where `# fmt: off` automatically dedents even when it is not within the
+  specified line range. (#4084)
 
 ### Preview style
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,8 @@
 
 ### Stable style
 
-- Fix bug where `# fmt: off` automatically dedents even when it is not within the
-  specified line range. (#4084)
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
Thanks! cc @yilei

yilei:
Thank you for the PR, the change looks good to me.

(My original thought would be to add extra info in the [`Leaf`](https://github.com/psf/black/blob/a0e270d0f246387202e676b25abbf7a02ddcbc71/src/blib2to3/pytree.py#L395) attributes since we already had to do it for a previous change, but this PR also works.)

JelleZijlstra (approved):

PR #4091, opened 2023-12-09
perf: increasing the performance to remove unnecessary for loops

Krupeshgithub:
1) Removing the for loop and using the filter, which increase little performance when opening the ```.git_archival.txt``` file in ```black/action/main```. 
2) Removing unnecessary for loops in ```black/src/black/files```.

### Checklist - did you ...

### DIFF
--- a/action/main.py
+++ b/action/main.py
@@ -26,12 +26,12 @@ else:
 if version_specifier:
     req = f"black{extra_deps}{version_specifier}"
 else:
-    describe_name = ""
     with open(ACTION_PATH / ".git_archival.txt", encoding="utf-8") as fp:
-        for line in fp:
-            if line.startswith("describe-name: "):
-                describe_name = line[len("describe-name: ") :].rstrip()
-                break
+        prefix = "describe-name: "
+        describe_name = "".join(
+            filter(lambda x: x.startswith(prefix), fp)
+        ).split(maxsplit=1)
+
     if not describe_name:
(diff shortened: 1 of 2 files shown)


action/main.py
```diff
@@ -26,12 +26,12 @@
...
     req = f"black{extra_deps}{version_specifier}"
 else:
-    describe_name = ""
     with open(ACTION_PATH / ".git_archival.txt", encoding="utf-8") as fp:
-        for line in fp:
-            if line.startswith("describe-name: "):
-                describe_name = line[len("describe-name: ") :].rstrip()
-                break
+        prefix = "describe-name: "
+        describe_name = "".join(
```
JelleZijlstra, line 31: This looks incorrect, `describe_name` is now a list instead of a string. I'm also skeptical that this improves performance, and even if it does, I doubt the minor performance difference is worth making the code harder to understand.

src/black/files.py
```diff
@@ -63,17 +63,18 @@ def find_project_root(
     if not srcs:
         srcs = [str(Path.cwd().resolve())]
 
-    path_srcs = [Path(Path.cwd(), src).resolve() for src in srcs]
+    src_parents = list()
+    for src in srcs:
```
JelleZijlstra, line 67: Not clear that this improves performance, and if it does, not clear the change is worth the cost in decreased readability.

hauntsaninja:
Closing, as per Jelle's comments

PR #4095, opened 2023-12-10
Unify docstring detection

### LINKED ISSUE #4094: Docstring formatting does not apply to module docstrings
https://github.com/psf/black/pull/4060#discussion_r1404646891 made me look into unifying some of our docstring logic. I discovered this bug:

```python
"""

has trailing space   

"""
```

I would expect Black to remove the trailing space, which it does for class and function docstrings, but it doesn't.

JelleZijlstra:
Fixes #4094.

We had two slightly different ways to detect docstrings. This preview feature unifies the two.

There are at least two concrete behavior changes:
- Module docstrings are now correctly recognized and formatted.
- We now remove blank lines before class docstrings that do not use triple quotes; we already did that before class docstrings that do use triple quotes.


src/black/strings.py
```diff
@@ -63,6 +63,8 @@ def lines_with_leading_tabs_expanded(s: str) -> List[str]:
             )
         else:
             lines.append(line)
+    if s.endswith("\n"):
```
JelleZijlstra, line 66: This was necessary so that we don't remove the newline at the end of a module docstring.

hauntsaninja (approved):

PR #4096, opened 2023-12-10
Add --unstable flag

### LINKED ISSUE #4093: Split preview into `--preview` and `--unstable`?
Black's `--preview` mode is supposed to give a preview of what next year's stable style will look like. But for the last few years, it hasn't been able to do that job, because `--preview` includes experimental string processing, which produces numerous bugs and likely again won't make it into the 2024 stable style (#4042). Several other current preview features likely won't make it into the 2024 stable style.

I propose to add a new style, `--unstable`. The idea would be that `--preview` is only for new style features that we currently think would make it into next year's stable style, and `--unstable` is for new features that we would like to eventually stabilize, but that are currently still buggy. `--unstable` would include all `--preview` features. This way, `--preview` will truly be a preview.

Concrete rules:
- Any preview feature that leads to new crashes (e.g., stability bugs) should be demoted to `--unstable` until the issue is fixed.
- Similarly, if a preview feature is shown to lead to bad formatting in some cases (in our judgment as maintainers), we demote the feature to unstable. For example, #3555 should make us demote the related preview feature to unstable.
- New preview features merged in November or December would by default go into `--unstable`, so that there is enough time for the ecosystem to try out the new feature. We can make exceptions for sufficiently simple and uncontroversial changes. We'd move these features into `--preview` once the new year starts.

Alternative solutions considered:

- We could simply be more aggressive in deleting preview features that cause trouble. I'd personally be hesitant to do this though, especially for features that have had significant effort invested into them: it will be a lot more effort to apply them to the Black codebase again later once the original issue is fixed. However, I do think we should delete unstable features that have remained in that state for too long.
- We could move unstable features out of `--preview`, but not add a separate `--unstable` flag. Instead, we could add a private flag like `-X string_processing` to enable individual preview features. This would allow more targeted testing, but it opens the door to more configurability than I'd like, and might lead to users expecting us to continue supporting such fine-grained configuration control.

JelleZijlstra:
Closes #4093.

We'd need more discussion before we actually implement that idea, but this also allows me to test whether string processing is responsible for diff-shades crashing all the time.


JelleZijlstra:
> but this also allows me to test whether string processing is responsible for diff-shades crashing all the time.

Seems like that's not it.

src/black/mode.py
```diff
@@ -197,6 +197,11 @@ class Preview(Enum):
     allow_form_feeds = auto()
 
 
+UNSTABLE_FEATURES: Set[Preview] = {
+    Preview.string_processing,
+}
```
amyreese, line 202: Would it be more prudent to instead have a set of "approved" preview features, and check that, rather than expecting someone to remember to add new features to this set before they're "ready" to include in the preview style?

JelleZijlstra: The intention is that new features go into the "approved" set by default (the approval is that we merge the PR). We'd put features in "unstable" only in the hopefully rare case where they are problematic in some way.

JelleZijlstra:
Merged the stable style branch (#4106) into this one, and added documentation. I think we should merge this in time for 24.1.0.

docs/the_black_code_style/current_style.md
```diff
@@ -449,6 +449,12 @@ file that are not enforced yet but might be in a future version of the formatter
 _Black_ will normalize line endings (`\n` or `\r\n`) based on the first line ending of
 the file.
 
+### Form feed characters
```
JelleZijlstra, line 452: Moved from "future style", this change is in the 2024 stable style

docs/the_black_code_style/future_style.md
```diff
@@ -1,117 +1,38 @@
 # The (future of the) Black code style
 
-```{warning}
-Changes to this document often aren't tied and don't relate to releases of
```
JelleZijlstra, line 4: This warning isn't really true any more now that we explicitly list preview features.

docs/the_black_code_style/future_style.md
```diff
@@ -1,117 +1,38 @@
...
-_Black_. It's recommended that you read the latest version available.
-```
-
-## Using backslashes for with statements
```
JelleZijlstra, line 8: Moved this to the end as it's a speculative feature that we haven't even implemented. The real-world preview style is more important.

src/black/mode.py
```diff
@@ -207,12 +218,15 @@ def __contains__(self, feature: Preview) -> bool:
         """
         Provide `Preview.FEATURE in Mode` syntax that mirrors the ``preview`` flag.
 
-        The argument is not checked and features are not differentiated.
-        They only exist to make development easier by clarifying intent.
+        In unstable mode, all features are enabled. In preview mode, all features
+        except those in UNSTABLE_FEATURES are enabled. For legacy reasons, the
+        string_processing feature has its own flag, which is deprecated.
```
RedGuy12, line 223: Question- what's the policy around deprecated flags? When is `--experimental-string-processing` expected to be removed?

JelleZijlstra: There isn't a real policy. That flag has been deprecated since 2022 so I think it's about time it goes away.

JelleZijlstra:
@hauntsaninja I added a separate `--enable-unstable-feature` flag as you suggested (you called it `--enable-unstable-hidden-feature` but that felt redundant).

I took the opportunity to remove `--experimental-string-processing`; it's been deprecated for two years and now you can use `--preview --enable-feature string_processing` instead.

CHANGES.md
```diff
@@ -53,6 +61,7 @@ release:
 
 <!-- Changes that affect Black's preview style -->
 
+- Add `--unstable` style (#4096)
```
hauntsaninja, line 64: Can we add a quick link / blurb here for preview style codebases that want hug_parens?

src/black/__init__.py
```diff
@@ -300,6 +306,26 @@ def validate_regex(
...
+    is_flag=True,
+    help=(
+        "Enable potentially disruptive style changes that have known bugs or are not"
+        " currently expected to make it into the stable style Black's next major"
+        " release. Implies --preview."
+    ),
+)
+@click.option(
+    "--enable-unstable-feature",
+    type=click.Choice([v.name for v in UNSTABLE_FEATURES]),
```
hauntsaninja, line 320: It could be worth it to be more permissive here, so folks can set `--enable-unstable-feature hug_parens` and forget about it for a year. But I think that much bother is probably fine for preview...

hauntsaninja (approved):
Thanks for making this happen!

src/black/__init__.py
JelleZijlstra: I think that's actually fine since we only allow `--enable-unstable-feature` if you also have `--preview`, so setting it for a flag that's in preview is a no-op. I'll change this.

CHANGES.md
JelleZijlstra: I kept the hugging feature (`hug_parens_with_braces_and_square_brackets`) in preview. Or maybe you're referring to `multiline_string_handling`?

JelleZijlstra: Also possibly `hug_parens_with_braces_and_square_brackets` should go into unstable; I haven't thought much about it beyond not putting it in 2024 stable. We have 11 months left to do that though.

hauntsaninja: Ah sorry, I was wrong on the specifics. Just thought more of a mention in changelog could be nice for whatever unstable features users may want

JelleZijlstra: Good point, I'll add something.

PR #4104, opened 2023-12-12
Fix feature detection for parenthesized context managers

JelleZijlstra:
This showed up in #4064.

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -25,6 +25,7 @@ from typing import (
     List,
     Optional,
     Sequence,
+    Set,
     Type,
     TypeVar,
     Union,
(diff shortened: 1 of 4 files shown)


tests/test_black.py
```diff
@@ -874,71 +875,88 @@ def test_get_features_used_decorator(self) -> None:
...
+        self.check_features_used("try: pass\nexcept (*Something,): pass", set())
+        self.check_features_used(
+            "try: pass\nexcept *Group: pass", {Feature.EXCEPT_STAR}
+        )
+
+        self.check_features_used("a[*b]", {Feature.VARIADIC_GENERICS})
+        self.check_features_used("a[x, *y(), z] = t", {Feature.VARIADIC_GENERICS})
+        self.check_features_used("def fn(*args: *T): pass", {Feature.VARIADIC_GENERICS})
+
+        self.check_features_used("with a: pass", set())
```
JelleZijlstra, line 928: These are the new tests. Several of them would have previously failed. The case that I ran into in one of our tests was

```
with (x := await a, y := await b):
        pass
```

hauntsaninja (approved):

PR #4114, opened 2023-12-16
Simplify chained expressions and remove unused param

Anupya:
### Description

- `previous` was unused in `is_split_after_delimiter` so removed it
- rename `reversed` in `enumerate_with_length` to `is_reversed` because `reversed` is a [Python built-in function](https://docs.python.org/3/library/functions.html#reversed)

Screenshot 2023-12-15 at 10 11 54 PM

- Simpify if statements to make them easier to read
- Fix docstrings (`well-formed` and `fast-forward` is one word, `Therefore` and `Otherwise` needs a comma after, remove `be be`, should be `a part` instead of `apart`)

### Checklist - did you ...

None of the below are necessary!

### DIFF
--- a/src/black/brackets.py
+++ b/src/black/brackets.py
@@ -115,7 +115,7 @@ class BracketTracker:
             if delim and self.previous is not None:
                 self.delimiters[id(self.previous)] = delim
             else:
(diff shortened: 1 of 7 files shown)


hauntsaninja:
Thanks!

PR #4128, opened 2023-12-25
Do not round cache mtimes

### LINKED ISSUE #4116: Changes to files happing quickly after running `black` are not detected
**Describe the bug**

It seems `black` does not detect that a file has been changed and needs to be re-checked, if the change happens quickly after `black` has been run on the file.

I'm talking about this feature: [Ignoring unmodified files](https://black.readthedocs.io/en/stable/usage_and_configuration/file_collection_and_discovery.html#ignoring-unmodified-files)

**To Reproduce**

I'm using the following shell script to reproduce the issue:

```bash
#! /usr/bin/env bash

set -eux

echo 'print (1)' > file.py
black file.py
echo 'print ()' > file.py
black file.py
cat file.py
```

The script writes overwrites the file and then runs `black`, doing this twice in a row. In the second run, `black` should reformat `print ()` to `print()`, but this does not happen:

```
$ ./bug.sh 
+ echo 'print (1)'
+ black file.py
reformatted file.py

All done! ✨ 🍰 ✨
1 file reformatted.
+ echo 'print ()'
+ black file.py
All done! ✨ 🍰 ✨
1 file left unchanged.
+ cat file.py
print ()
```

Even running it manually after a few seconds does not fix the issue, but removing the cache directory does:

```
$ black file.py
All done! ✨ 🍰 ✨
1 file left unchanged.
$ cat file.py
print ()
$ rm -r ~/Library/Caches/black/
$ black file.py
reformatted file.py

All done! ✨ 🍰 ✨
1 file reformatted.
$ cat file.py
print()
```

**Expected behavior**

I think black should not get confused by changes to files that happen quickly after it has formatted a file. The file should be checked again if it is possible that its content has changed without also changing its timestamp.

**Environment**

- Black's version:

    ```
    $ black --version
    black, 23.12.0 (compiled: yes)
    Python (CPython) 3.11.6
    ```

- OS and Python version: macOS 12.7.1, Python 3.11.6 installed via Homebrew.

hauntsaninja:
Fixes #4116

This logic was introduced in #3821, I believe as a result of copying logic inside mypy that I think isn't relevant to Black

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@
 ### Configuration
 
 <!-- Changes to how Black can be configured -->
+- Fix cache mtime logic that resulted in false positive cache hits (#4128)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@
 ### Configuration
 
 <!-- Changes to how Black can be configured -->
+
 - Fix cache mtime logic that resulted in false positive cache hits (#4128)
 
 ### Packaging
--- a/src/black/cache.py
+++ b/src/black/cache.py
@@ -101,7 +101,7 @@ class Cache:
         st = res_src.stat()
         if st.st_size != old.st_size:
             return True
-        if int(st.st_mtime) != int(old.st_mtime):
+        if st.st_mtime != old.st_mtime:
             new_hash = Cache.hash_digest(res_src)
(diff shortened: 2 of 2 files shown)


cdce8p (approved):
Removing the int conversion should be fine. See my comment on the issue https://github.com/psf/black/issues/4116#issuecomment-1868845509

PR #4132, opened 2023-12-28
Remove empty lines before docstrings in async functions

hauntsaninja:
Relates to #4043

### DIFF
--- a/src/black/nodes.py
+++ b/src/black/nodes.py
@@ -742,10 +742,6 @@ def is_multiline_string(leaf: Leaf) -> bool:
     return has_triple_quotes(leaf.value) and "\n" in leaf.value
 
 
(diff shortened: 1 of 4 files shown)


src/black/nodes.py
```diff
@@ -742,10 +742,6 @@ def is_multiline_string(leaf: Leaf) -> bool:
     return has_triple_quotes(leaf.value) and "\n" in leaf.value
 
 
-def is_funcdef(node: Node) -> bool:
-    return node.type == syms.funcdef
-
-
 def is_function_or_class(node: Node) -> bool:
```
hauntsaninja, line 745: This function is suspicious since many async defs don't have async_funcdef, but this can be subject for different PR

JelleZijlstra (approved):

PR #4133, opened 2023-12-29
Remove is_function_or_class helper footgun

hauntsaninja:
This is a no-op change.

That function was not a good way to tell if something is a function or a class, since it basically only worked for async functions by accident (the parent of a suite / simple_stmt of async function body is a funcdef).


JelleZijlstra (approved):

JelleZijlstra:
Thanks for all the improvements!

PR #4146, opened 2024-01-08
fix: Don't normalize whitespace before fmt:skip comments

### LINKED ISSUE #4143: Whitespace before `fmt: skip` comment is normalised
**Describe the bug**

Apologies if this is a duplicate; I have read:

- https://github.com/psf/black/issues/3330
- https://github.com/psf/black/pull/3959/files
- https://github.com/psf/black/issues/2213

Based on https://github.com/psf/black/pull/3959, Black should support e.g. this: `# pylint: disable=redefined-outer-name # fmt: skip` and yet I cannot get it to work on the newest Black version.

I have the following line:

```
client = billingdataservice.app.test_client() # pylint: disable=redefined-outer-name
```

The line is too long for black, so when I run black, I get this:

```
client = (
    billingdataservice.app.test_client()
)  # pylint: disable=redefined-outer-name
```

We don't want that, so I tried to use `fmt: skip`. I tried the following comments:

- `client = billingdataservice.app.test_client() # pylint: disable=redefined-outer-name; fmt: skip`
- `client = billingdataservice.app.test_client() # pylint: disable=redefined-outer-name # fmt: skip`
- `client = billingdataservice.app.test_client() # fmt: skip # pylint: disable=redefined-outer-name`
- `client = billingdataservice.app.test_client() # fmt: skip; pylint: disable=redefined-outer-name`

as well as mixing number of spaces before/after the hash sign, e.g. `...test_client()  #  fmt: skip; pylint: ...` or `...test_client()  # fmt: skip; pylint: ...` etc.

**Reproducer**

`file.py`:

```
def client():
    client = billingdataservice.app.test_client() # fmt: skip; pylint: disable=redefined-outer-name
```

Command:

```sh
$ black --preview file.py
```

(The `--preview` flag seems not to matter for this issue; behavior is identical with or without it)

Black always modifies the line, e.g.:

```
def client():
    client = ( 
        billingdataservice.app.test_client()
    )  # fmt: skip; pylint: disable=redefined-outer-name
```

Which means `fmt: skip` did not get honored.

**Environment**

```
$ black --version
black, 23.12.1 (compiled: yes)
Python (CPython) 3.11.6
```

- Black's version: 23.12.1
- OS and Python version: OpenSUSE Tumbleweed, Python 3.11

RedGuy12:
### Description

Resolves #4143. I attempted to fix https://github.com/psf/black/pull/2970#pullrequestreview-932691073 while I was at it. However, that's a larger undertaking since `_contains_fmt_skip_comment()` requires a `mode` parameter and `make_comment()` doesn't - I'd have to add it to every caller of `make_comment()` and each of its callers, etc. Let me know if it's desired anyway.

### Checklist - did you ...


JelleZijlstra:
Thanks. This will have to go into the preview style as it changes existing formatting, and it'll probably have to wait until the 2024 stable style is out.

I don't understand the statement that you're not addressing https://github.com/psf/black/pull/2970#pullrequestreview-932691073—what else is there in that comment that this PR is not fixing?

RedGuy12:
> what else is there in that comment that this PR is not fixing?

#4143 talks about normalizing whitespace before the `#`. https://github.com/psf/black/pull/2970#pullrequestreview-932691073 is about whitespace immediately after the `#`. I thought I might as well just fix both, but like I said, the content after the `#` is formatted separately and would be harder to code.

> This will have to go into the preview style as it changes existing formatting

Done!

src/black/comments.py
```diff
@@ -88,7 +89,8 @@ def list_comments(prefix: str, *, is_endmarker: bool) -> List[ProtoComment]:
     form_feed = False
     for index, full_line in enumerate(re.split("\r?\n", prefix)):
         consumed += len(full_line) + 1  # adding the length of the split '\n'
-        line = full_line.lstrip()
+        match = re.match(r"^(\s*)(\S.*|)$", full_line)
+        whitespace, line = ("  ", full_line) if match is None else match.groups()
```
JelleZijlstra, line 93: The regex should always match, right? If so, would be clearer to add an assert.

RedGuy12: @JelleZijlstra Done

hauntsaninja (approved):
This looks great, thank you!

(technically this change does meet the stability policy, since it won't cause changes in already blackened code, but always nice to let things bake in preview...)

PR #4152, opened 2024-01-13
Fixed the empty spaces of '# type:'

### LINKED ISSUE #2097: is_type_comment() is wrong

gawandeabhishek:
Fixes #2097

### Description

I have added the `join` and `split` methods so that the `split` method breaks up the string into a list by separating it with spaces, and then the `join` method joins the list back into one string with a single space between each word.

### DIFF
--- a/src/black/nodes.py
+++ b/src/black/nodes.py
@@ -875,7 +875,7 @@ def is_type_comment(leaf: Leaf) -> bool:
     used in modern version of Python, this function may be deprecated in the future."""
     t = leaf.type
     v = leaf.value
-    return t in {token.COMMENT, STANDALONE_COMMENT} and v.startswith("# type:")
+    return t in {token.COMMENT, STANDALONE_COMMENT} and v.startswith(" ".join("# type:".split()))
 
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Thanks, but this is not correct. (it would check only for `#   t y p e :` or something like that.)

gawandeabhishek:
Hi @JelleZijlstra , I am a first-year student. I want to participate in Google Summer of Code (GSoC) and contribute to the Python Software Foundation (PSF). I am familiar with the MERN stack, HTML, CSS, JavaScript, and C++. I am currently learning Data Structures and Algorithms (DSA) in Java. Recently, I acquired knowledge in Python, Docker, and I am eager to start contributing to PSF. However, I am facing some confusion on how to run the code locally. In web development, I use Node.js to run websites on my local machine. I have also read the documentation (md) but I am still unclear on the process. Additionally, despite learning Python, I am struggling to solve any of the issues. Could you please guide me on what I should learn to become capable of contributing to PSF without making mistakes?

PR #4153, opened 2024-01-13
Improve call chain

### LINKED ISSUE #571: Strange formatting of  fluent interface
Operating system: Mac 
Python version: 3.7
*Black* version: 18.9b0
Does also happen on master: 

Following on https://twitter.com/llanga/status/1052631100290420736

I have a code which uses `yarl.URL`
```python
a = str(
    my_very_very_very_long_url_name
    .with_user(and_very_long_username)
    .with_password(and_very_long_password)
)
```

*Black* formats it like this
```python
a = str(
    my_very_very_very_long_url_name.with_user(and_very_long_username).with_password(
        and_very_long_password
    )
)
```

I understand that first call is not on a new line because of [reasons](https://github.com/ambv/black/issues/67#issuecomment-389681164), but the rest still looks quite odd, especially considering that when there are three calls in the chain, *Black* produces

```python
a = str(
    my_very_very_very_long_url_name.with_user(and_very_long_username)
    .with_user(and_very_long_username)
    .with_password(and_very_long_password)
)
```

Which is still not perfect, because first `with_user` is a separate logical step, but at least it's pretty.

Booome:
Fixes #571

### DIFF
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -399,8 +399,9 @@ def _contains_fmt_skip_comment(comment_line: str, mode: Mode) -> bool:
             ],
             *[
                 _COMMENT_PREFIX + comment.strip()
-                for comment in comment_line.strip(_COMMENT_PREFIX)
-                .split(_COMMENT_LIST_SEPARATOR)
+                for comment in comment_line.strip(_COMMENT_PREFIX).split(
+                    _COMMENT_LIST_SEPARATOR
+                )
             ],
         ]
         if Preview.single_line_format_skip_with_multiple_comments in mode
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -399,9 +399,8 @@ def _contains_fmt_skip_comment(comment_line: str, mode: Mode) -> bool:
             ],
             *[
                 _COMMENT_PREFIX + comment.strip()
-                for comment in comment_line.strip(_COMMENT_PREFIX).split(
-                    _COMMENT_LIST_SEPARATOR
-                )
+                for comment in comment_line.strip(_COMMENT_PREFIX)
+                .split(_COMMENT_LIST_SEPARATOR)
             ],
         ]
         if Preview.single_line_format_skip_with_multiple_comments in mode
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -242,7 +242,11 @@ class LineGenerator(Visitor[Line]):
                 if i == 0:
                     continue
                 if node.children[i - 1].type == token.COLON:
-                    if child.type == syms.atom and child.children[0].type == token.LPAR:
+                    if (
+                        child.type == syms.atom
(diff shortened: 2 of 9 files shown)


tungol:
Seems like a lot of the diff in diff shades only have one method call. This in particular seems especially bad:

```diff
     assert (
         makeFY5253NearestEndMonthQuarter(
             weekday=1, startingMonth=3, qtr_with_extra_week=3
-        ).freqstr
+        )
+        .freqstr
         == "REQ-N-MAR-TUE-3"
     )
```

src/black/comments.py
```diff
@@ -399,9 +399,8 @@ def _contains_fmt_skip_comment(comment_line: str, mode: Mode) -> bool:
             ],
             *[
                 _COMMENT_PREFIX + comment.strip()
-                for comment in comment_line.strip(_COMMENT_PREFIX).split(
-                    _COMMENT_LIST_SEPARATOR
-                )
+                for comment in comment_line.strip(_COMMENT_PREFIX)
```
JelleZijlstra, line 402: This feels worse than the previous formatting.

src/black/linegen.py
```diff
@@ -1180,9 +1181,12 @@ def delimiter_split(
     except ValueError:
         raise CannotSplit("No delimiters found") from None
 
-    if delimiter_priority == DOT_PRIORITY:
-        if bt.delimiter_count_with_priority(delimiter_priority) == 1:
-            raise CannotSplit("Splitting a single attribute from its owner looks wrong")
+    if (
```
JelleZijlstra, line 1184: We might want to keep this one as is, see the comment on the PR.

JelleZijlstra:
Looking at the diff-shades output, a lot of changes feel like mild negatives, e.g.

```diff
         self.assertTrue(
-            logger.messages[0].startswith(
+            logger.messages[0]
+            .startswith(
                 "route matched for url http://localhost:8080"
                 "/archives/action1/article1; "
                 "route_name: 'foo', "
                 "path_info: "
             )
```

```diff
-        assert repr_locals.lines[0].startswith(
-            "x          = <[NotImplementedError() raised in repr()] ObjWithErrorInRepr"
+        assert (
+            repr_locals.lines[0]
+            .startswith(
+                "x          = <[NotImplementedError() raised in repr()]"
+                " ObjWithErrorInRepr"
+            )
         )
```

src/black/comments.py
Booome: I also don't like the new one, I personaly like is: 
```python
                for comment in (
                    comment_line.strip(_COMMENT_PREFIX).split(_COMMENT_LIST_SEPARATOR)
                )
```
anyway, I will revert this change.

src/black/linegen.py
Booome: Sorry, I'm not quite sure which COMMENT you're referring to?

PR #4154, opened 2024-01-16
fix: bug where the doublestar operation had inconsistent formatting.

### LINKED ISSUE #4149: Inconsistent formatting of `**` in tenary expression.
```python
m2 = None if not isinstance(dist, Normal) else μ**2 + σ**2
m2 = None if ... else μ**2 + σ**2
```

becomes ([black-playground](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADWAIBdAD2IimZxl1N_WmA--ISCB5EifcHtD-SC05S16MrO2s_CTIlU7kJn3hUYYEAR2UQPTBTwi80_HP-MWRwv-fM86Wk4t0tumjrAkl6y_aJWDqESGKOTm19NXn-bXgtU_nNq-Ina2D_Vmb_KpKm_IeMxherTcuswNovlvqb4_kcDY1DbAN1sYibT_XliAAGcAdcBAABCwNAuscRn-wIAAAAABFla))

```python
m2 = None if not isinstance(dist, Normal) else μ ** 2 + σ**2
m2 = None if ... else μ**2 + σ**2
```

There are spaces added to `μ**2` depending on unrelated code.

wannieman98:
### Description

Fixes #4149, which according to the ternary operation expression was not correctly being formatted. This was caused by the existing black's logic to determine whether the `base` of the exponential expression is `simple` or not, meaning whether it does not have parenthesis or brackets and such and just accepting naming or dotted lookups. Essentially, the existing logic was to loop over the direction (base backwards, exponential forward) and find whether a bracket or parenthesis existed. This logic worked for the `exponential` case since it doesn't need to care about chained expressions. This PR is to involve checking whether base is a chained expression and check the expression is simple.  

Just to make sure, I added some more complex cases that would cause the bug other than the case mentioned in the issue. 

### Checklist - did you ...


wannieman98:
Hi! Seems like diff-shades check failed with "6 semaphore objects missing" user warning. From my experience this was a result of memory issue rather than the test actually failing. Could we try running this step again? Thanks!

CHANGES.md
```diff
@@ -67,6 +67,11 @@
 
 - Fixed a bug that included dependencies from the `d` extra by default (#4108)
 
+### Parser
```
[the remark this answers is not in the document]
wannieman98: I see, thanks for letting me know!

JelleZijlstra:
Thanks! I haven't looked in detail yet, but this likely needs to go into the preview style only if it changes formatting for existing code. (diff-shades didn't find any, but that might just be because the patterns where this makes a difference are rare.)

wannieman98:
Review would be appreciated!

JelleZijlstra:
Thanks, I'm planning to look at this after the 24.1 release goes out (hopefully today/tomorrow). There's enough in that release already :)

tests/data/cases/power_op_spacing.py
```diff
@@ -1,3 +1,8 @@
+m2 = None if not isinstance(dist, Normal) else μ** 2 + σ * 2
+m3 = None if not isinstance(dist, Normal) else μ**2 + σ * 2
+m4 = None if not isinstance(dist, Normal) else s**2 + σ * 2
```
JelleZijlstra, line 3:
```suggestion
m2 = None if not isinstance(dist, Normal) else m** 2 + s * 2
m3 = None if not isinstance(dist, Normal) else m**2 + s * 2
m4 = None if not isinstance(dist, Normal) else s**2 + s * 2
```
Replace the Greek letters with Latin ones. If there is something unusual going on in a test, then readers will naturally think the test is about that unusual thing. In this case the Greek doesn't matter, so use less unusual characters.

tests/data/cases/power_op_spacing.py
```diff
@@ -78,6 +83,11 @@ def function_dont_replace_spaces():
 
 
 # output
+m2 = None if not isinstance(dist, Normal) else μ**2 + σ * 2
+m3 = None if not isinstance(dist, Normal) else μ**2 + σ * 2
+m4 = None if not isinstance(dist, Normal) else s**2 + σ * 2
+m5 = obj.method(another_obj.method()).attribute ** 2
+m6 = None if ... else μ**2 + σ**2
```
JelleZijlstra, line 90:
```suggestion
m2 = None if not isinstance(dist, Normal) else m**2 + s * 2
m3 = None if not isinstance(dist, Normal) else m**2 + s * 2
m4 = None if not isinstance(dist, Normal) else s**2 + s * 2
m5 = obj.method(another_obj.method()).attribute ** 2
m6 = None if ... else m**2 + s**2
```

tests/data/cases/power_op_spacing.py
```diff
@@ -1,3 +1,8 @@
+m2 = None if not isinstance(dist, Normal) else μ** 2 + σ * 2
+m3 = None if not isinstance(dist, Normal) else μ**2 + σ * 2
+m4 = None if not isinstance(dist, Normal) else s**2 + σ * 2
+m5 = obj.method(another_obj.method()).attribute **2
+m6 = None if ... else μ**2 + σ**2
```
JelleZijlstra, line 5:
```suggestion
m6 = None if ... else m**2 + s**2
```

src/black/trans.py
```diff
@@ -94,43 +94,30 @@ def hug_power_op(
...
         # An operand is considered "simple" if's a NAME, a numeric CONSTANT, a simple
         # lookup (see above), with or without a preceding unary operator.
         start = line.leaves[index]
         if start.type in {token.NAME, token.NUMBER}:
-            return is_simple_lookup(index, step=(1 if kind == "exponent" else -1))
+            return is_simple_lookup(index, kind)
 
         if start.type in {token.PLUS, token.MINUS, token.TILDE}:
             if line.leaves[index + 1].type in {token.NAME, token.NUMBER}:
                 # step is always one as bases with a preceding unary op will be checked
```
JelleZijlstra, line 117:
```suggestion
                # kind is always one as bases with a preceding unary op will be checked
```

src/black/trans.py
```diff
@@ -162,6 +149,73 @@ def is_simple_operand(index: int, kind: Literal["base", "exponent"]) -> bool:
...
+
+    while 0 <= index < len(line.leaves):
+        current = line.leaves[index]
+        chain.append(current)
+        if not contains_disallowed and current.type in disallowed:
+            contains_disallowed = True
+        if not is_expression_chained(chain):
+            return not contains_disallowed
+
+        index += -1
```
JelleZijlstra, line 169:
```suggestion
        index -= 1
```

src/black/trans.py
```diff
@@ -162,6 +149,73 @@ def is_simple_operand(index: int, kind: Literal["base", "exponent"]) -> bool:
...
+
+    return True
+
+
+def handle_is_simple_lookup_forward(
+    line: Line, index: int, disallowed: Set[int]
+) -> bool:
+    """
+    Handling decision is_simple_lookup for the lines prior to the doublestar token.
```
JelleZijlstra, line 178: This docstring looks wrong, shouldn't it say behind the doublestar?

src/black/trans.py
```diff
@@ -162,6 +149,73 @@ def is_simple_operand(index: int, kind: Literal["base", "exponent"]) -> bool:
...
+    This function is simplified to keep consistent with the prior logic and the forward
+    case are more straightforward and do not need to care about chained expressions.
+    """
+    while 0 <= index < len(line.leaves):
+        current = line.leaves[index]
+        if current.type in disallowed:
+            return False
+        if current.type not in {token.NAME, token.DOT} or current.value == "for":
```
JelleZijlstra, line 186: Check that the leaf is of the right type before checking its value.

JelleZijlstra (changes requested):
Thanks!

This will need to go in the preview style, so gated behind an `if Preview... in mode` check (you can see examples elsewhere in the code).

src/black/trans.py
wannieman98: I would have thought `prior` and `behind` would mean the same to be honest. But I can change it to suggested.

src/black/trans.py
wannieman98: I wanted to reserve the behavior logic for the previous logic implemented. But I can go ahead and ensure it will be a `token.NAME` before checking the value.

wannieman98:
Hi @JelleZijlstra, I have addressed the comments, could you kindly review again? Thanks!

src/black/trans.py
```diff
@@ -162,6 +155,99 @@ def is_simple_operand(index: int, kind: Literal["base", "exponent"]) -> bool:
...
+        index -= 1
+
+    return True
+
+
+def handle_is_simple_lookup_forward(
+    line: Line, index: int, disallowed: Set[int]
+) -> bool:
+    """
+    Handling decision is_simple_lookup for the lines behind to the doublestar token.
```
JelleZijlstra, line 208:
```suggestion
    Handling decision is_simple_lookup for the lines behind the doublestar token.
```

src/black/mode.py
```diff
@@ -177,6 +177,7 @@ class Preview(Enum):
     wrap_long_dict_values_in_parens = auto()
     multiline_string_handling = auto()
     typed_params_trailing_comma = auto()
+    is_simple_lookup_for_doublestar_expression = auto()
```
JelleZijlstra, line 180: Could you add this feature to the documentation in `future-style.md`, where we now list all preview features? The description can just be something like "fix line length computation for certain expressions that involve the power operator".

tests/data/cases/is_simple_lookup_for_doublestar_expression.py
```diff
@@ -0,0 +1,14 @@
+# flags: --preview
+m2 = None if not isinstance(dist, Normal) else m** 2 + s * 2
+m3 = None if not isinstance(dist, Normal) else m**2 + s * 2
+m4 = None if not isinstance(dist, Normal) else s**2 + s * 2
```
JelleZijlstra, line 4: This test seems the same as the previous one.

JelleZijlstra (commented):
Thanks! Getting very close.

src/black/mode.py
wannieman98: Sure thing, I'll add it!

tests/data/cases/is_simple_lookup_for_doublestar_expression.py
wannieman98: You are right, they are testing the same thing I'll remove the duplicate test.

src/black/trans.py
wannieman98: Let me add it with the changes to the test cases

wannieman98:
@JelleZijlstra Thanks for the review! Addressed the comments hope this works out :)

JelleZijlstra (approved):

PR #4155, opened 2024-01-17
fix: Don't allow unparenthesizing walruses

### LINKED ISSUE #3758: cannot parse assignment expression in preview style since 23.1
Hello,

I seem to have found a bug in black's preview style regarding assignment expressions, that has been present since version 23.1.0.

**Describe the bug**

When using an assignment expression in my code example, black with preview=true complains it cannot parse the line. Black preview=false accepts the code happily and leaves no changes. 

**To Reproduce**

Here's my code example:

```python
# file.py
from pydriller import Commit

commits: list[Commit] = []
update_hashes: list[str] = []
upstream_messages: list[str] = []
parsed = [
    {
        "hash": commit.hash,
        "author": f"{commit.author.name} <{commit.author.email}>",
        # black 23.1 --preview can't parse the following line:
        "is_update": (up := commit.hash in update_hashes),
        "is_upstream": up and commit.msg in upstream_messages,
    }
    for commit in commits
]
```

```sh
$ black file.py --target-version py311 --preview
```

The resulting error is:

> `cannot format file.py: Cannot parse: 12:24:         "is_update": up := commit.hash in update_hashes,`

**Expected behavior**

It should parse the line like it does with preview=false.

Also, look at the error above. The line of code shown doesn't include the parentheses like it does in my source. Without the parens, cpython can't parse it either!

**Environment**

- Black's version: 23.3, also tested on main branch at g839ef35.
- OS and Python version: Linux/Python 3.11.4

**Additional context**

I love you lots <3

RedGuy12:
### Description

Resolves #3758.

### Checklist - did you ...

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -246,7 +246,6 @@ class LineGenerator(Visitor[Line]):
                         if maybe_make_parens_invisible_in_atom(
                             child,
                             parent=node,
-                            mode=self.mode,
                             remove_brackets_around_comma=False,
                         ):
                             wrap_in_parentheses(node, child, visible=False)
@@ -268,7 +267,6 @@ class LineGenerator(Visitor[Line]):
                     if maybe_make_parens_invisible_in_atom(
                         child,
                         parent=node,
-                        mode=self.mode,
                         remove_brackets_around_comma=False,
                     ):
                         wrap_in_parentheses(node, child, visible=False)
@@ -373,7 +371,7 @@ class LineGenerator(Visitor[Line]):
             ):
                 wrap_in_parentheses(node, leaf)
 
-        remove_await_parens(node, mode=self.mode)
+        remove_await_parens(node)
 
         yield from self.visit_default(node)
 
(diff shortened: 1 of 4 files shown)


src/black/linegen.py
```diff
@@ -1545,6 +1555,10 @@ def maybe_make_parens_invisible_in_atom(
             and max_delimiter_priority_in_atom(node) >= COMMA_PRIORITY
         )
         or is_tuple_containing_walrus(node)
+        or (
+            Preview.wrap_long_dict_values_in_parens in mode
```
JelleZijlstra, line 1559: Do we need the preview check? Intuitively I'd think we shouldn't because we'll almost always cause a crash if we remove the parens around a walrus.

RedGuy12: Good point, resolved

RedGuy12:
It turns out that adding that check to `maybe_make_parens_invisible_in_atom` causes Black to add parenthesis around walruses where it's unnecessary. See [the test failure](https://github.com/RedGuy12/black/actions/runs/7558997228/job/20581801069?pr=4155#step:6:447) (we didn't notice it previously since that test doesn't run with `--preview`). I moved the check to `visit_dictsetmaker()` where we can be sure it won't affect anything else.

JelleZijlstra (approved):
Thank you!

wizpig64:
Thanks @RedGuy12 for fixing my bug!

PR #4161, opened 2024-01-22
fix: pathlib exception handling with symlinks

### LINKED ISSUE #4077: pathlib exception when symlinks are involved
**Describe the bug**

Hello! First of all, thank you for this invaluable tool.

I first ran into this issue on Windows when using black to format code located on a mapped network drive. 

This line in [get_sources()](https://github.com/psf/black/blob/a0e270d0f246387202e676b25abbf7a02ddcbc71/src/black/__init__.py#L656C17-L656C17) doesn't ignore symlinks pointing outside of `root`, which has symlinks resolved by [find_project_root()](https://github.com/psf/black/blob/a0e270d0f246387202e676b25abbf7a02ddcbc71/src/black/files.py#L46):
https://github.com/psf/black/blob/a0e270d0f246387202e676b25abbf7a02ddcbc71/src/black/__init__.py#L687

**To Reproduce on Linux**

```shell
mkdir test_black
ln -s test_black test_black_sym
cd test_black
git init
echo "print('hello')" > app.py
black ~/test_black_sym/app.py
```

`git init` is needed so that [find_project_root()](https://github.com/psf/black/blob/main/src/black/files.py#L46) doesn't return the root of the file system. 

The resulting error is:

```shell
  <some omitted>
  File "/home/tg2648/black/src/black/__init__.py", line 600, in main
    sources = get_sources(
              ^^^^^^^^^^^^
  File "/home/tg2648/black/src/black/__init__.py", line 687, in get_sources
    root_relative_path = path.absolute().relative_to(root).as_posix()
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tg2648/.pyenv/versions/3.11.0/lib/python3.11/pathlib.py", line 730, in relative_to
    raise ValueError("{!r} is not in the subpath of {!r}"
ValueError: '/home/tg2648/test_black_sym/app.py' is not in the subpath of '/home/tg2648/test_black' OR one path is relative and the other is absolute.
```

**Environment**

- Black's version: 23.11.0
- OS and Python version: Windows and Linux / Python 3.11.0

**Ideas**

[normalize_path_maybe_ignore()](https://github.com/psf/black/blob/main/src/black/files.py#L250), which is used later in `get_sources()`, also calculates the root relative path but catches the `ValueError` exception:

https://github.com/psf/black/blob/46be1f8e54ac9a7d67723c0fa28c7bec13a0a2bf/src/black/files.py#L262-L269

Maybe one solution is to extract these lines into a function like `get_root_relative_path` and then use it in `get_sources()` and `normalize_path_maybe_ignore()`. I can submit a PR if this sounds reasonable.

dankrzeminski32:
Fixes #4077 

### Description

Added additional exception handling when calculating the root relative file path.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,8 @@
 
 <!-- Changes to how Black can be configured -->
 
+- Fix symlink handling, properly catch and ignore symlinks that point outside of root
+  (#4161)
 - Fix cache mtime logic that resulted in false positive cache hits (#4128)
 
 ### Packaging
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -49,6 +49,7 @@ from black.files import (
     find_user_pyproject_toml,
     gen_python_files,
     get_gitignore,
+    get_root_relative_path,
     normalize_path_maybe_ignore,
     parse_pyproject_toml,
     path_is_excluded,
@@ -700,7 +701,10 @@ def get_sources(
 
         # Compare the logic here to the logic in `gen_python_files`.
         if is_stdin or path.is_file():
-            root_relative_path = path.absolute().relative_to(root).as_posix()
+            root_relative_path = get_root_relative_path(path, root, report)
+
+            if root_relative_path is None:
+                continue
 
             root_relative_path = "/" + root_relative_path
 
--- a/src/black/files.py
+++ b/src/black/files.py
@@ -259,20 +259,28 @@ def normalize_path_maybe_ignore(
     try:
         abspath = path if path.is_absolute() else Path.cwd() / path
         normalized_path = abspath.resolve()
-        try:
-            root_relative_path = normalized_path.relative_to(root).as_posix()
-        except ValueError:
-            if report:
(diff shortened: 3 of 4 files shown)


JelleZijlstra (approved):

lopopolo:
I was impacted by this bug. I run `black` in `--check` mode in a Bazel test. Bazel tests execute in a sandbox with symlinks that point elsewhere to sources in the actual repository.

When I pass a fully qualified path to `black`, I want it to read the contents at that path. The concept of a "root" is only useful when asking black to traverse a directory tree.

lopopolo:
As an aside, `prettier` recently made a very similar change:

- https://github.com/prettier/prettier/pull/14627
- https://github.com/prettier/prettier/pull/15533
- https://github.com/prettier/prettier/issues/15843

In general, when invoking a code formatter as `fmt path/to/file.type`, I expect the formatter to read the file and format it, without having opinions on the path itself.

PR #4164, opened 2024-01-23
fix: Consistently add trailing comma on typed parameters

### LINKED ISSUE #4080: Single-argument function includes trailing comma inconsistently
**Describe the bug**

When a single-argument function definition breaks across multiple lines, Black will add a trailing comma (after the argument) in some cases but not others.

**To Reproduce**

For example, given the following Black adds a trailing comma after the annotation on `x` in the first case, but not the second, third, or fourth:

```python
# Black inserts a trailing comma after the first argument's annotation here.
def long_function_name_goes_here(
    x: Callable[List[int]]
) -> Union[List[int], float, str, bytes, Tuple[int]]:
    pass

# But not here, or in any of the below.
def long_function_name_goes_here(
    x: Callable[[str, Any], int]
) -> Union[List[int], float, str, bytes, Tuple[int]]:
    pass

def long_function_name_goes_here(
    x: Callable[List[int, str]]
) -> Union[List[int], float, str, bytes, Tuple[int]]:
    pass

def long_function_name_goes_here(
    x: Callable[List[int], str]
) -> Union[List[int], float, str, bytes, Tuple[int]]:
    pass
```

So the inclusion of the trailing comma seems to depend on the _content_ of the annotation (but not its length). One observation: Black omits the trailing comma when the _annotation_ itself contains a comma.

(Here's the same example in the [playground](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ALeARBdAD2IimZxl1N_Wg0-BMXRAiYxkRLgmwJgPbucmCxkyRhldo2eOEP-o3QLvZCqxRaHz-pxcr5e3vmEsE5PexxyCAYlXUhuHdiFWHwBfgyaZvRwMoj2wQ47aKPkoaoLpBx_292oceqgoACgNq4DptlCROTZX4Eln6rV3lGi5g7Um9MqWVVSz_1GDauhC_vI-6bkxdvRzj5cV36sYL4CPmW3fLYyM7uoexyWDuQeReA2kZtfSGON0Q5YljRYNYMui5uN8TEzCh10LXJKOx-7YxN5UkfH0vF0DLyZnL_Aezfs1NQuNXMHjtms5vCJ4VLZ14R7hxT5f2xZXmin_qRA4qElJuYA1ZQ5Zj-gIT1of6fgPaf_AHHe-wyOt36MAAGsAt8FAAAMPbevscRn-wIAAAAABFla).)

RedGuy12:
### Description

Resolves #4080

### Checklist - did you ...


src/black/linegen.py
```diff
@@ -1074,7 +1075,14 @@ def bracket_split_build_line(
             no_commas = (
                 original.is_def
                 and opening_bracket.value == "("
-                and not any(leaf.type == token.COMMA for leaf in leaves)
+                and not any(
+                    leaf.type == token.COMMA
+                    and (
+                        Preview.typed_params_trailing_comma not in original.mode
+                        or not is_part_of_annotation(leaf)
```
hauntsaninja, line 1082: Can we use `is_part_of_annotation` to simplify some of the logic in the next check too? Lines 1092-1099?

hauntsaninja (commented):
Thank you for all you recent PRs!

src/black/linegen.py
RedGuy12: I tried, but it ended up adding a lot of unrelated changes to this PR since `is_part_of_annotation` currently checks for both return and parameter annotations. I can do that refactor in a future PR.

RedGuy12:
No idea why diff-shades is detecting changes in stable style now, I only merged upstream in and didn't change any logic. Additionally, it's only finding changes when linting Black itself, past diff-shades runs showed changes in other projects on preview style (compare [stable run today](https://github.com/psf/black/actions/runs/7670505321/job/20906853963?pr=4164#step:10:17) and [preview run Wednesday](https://github.com/psf/black/actions/runs/7643567492/job/20825924076?pr=4164#step:10:17))

JelleZijlstra:
I suspect it's because Black itself is now formatted with `unstable = true` and diff-shades maybe doesn't handle that correctly.

hauntsaninja:
https://github.com/ichard26/diff-shades/pull/15 as a quick patch

CHANGES.md
```diff
@@ -112,6 +112,7 @@ release:
 - Fix crash when using a walrus in a dictionary (#4155)
 - Fix unnecessary parentheses when wrapping long dicts (#4135)
 - Stop normalizing spaces before `# fmt: skip` comments (#4146)
+- Consistently add trailing comma on typed parameters (#4164)
```
JelleZijlstra, line 115: Need to move this up to the unreleased section

JelleZijlstra (commented):
Looks good. I'll fix the one minor issue, then hopefully diff-shades will succeed if I retry.

JelleZijlstra:
Maybe ichard26/diff-shades#16 helps, trying again.

JelleZijlstra:
diff-shades succeeded (not sure it's because of my changes, as it was apparently already using 4 processes). Quite a few changes, but they all look expected.

PR #4165, opened 2024-01-23
Raise exception on invalid toml configuration

### LINKED ISSUE #4140: Error on invalid configuration
Currently if you make a typo when configuring Black you don't get an error

dankrzeminski32:
### Description

Resolves #4140, raises an exception when toml config file contains a key that doesn't exist in the set of possible cli arguments.

### Checklist 


src/black/__init__.py
```diff
@@ -181,6 +186,15 @@ def read_pyproject_toml(
     return value
 
 
+def spellcheck_pyproject_toml_keys(
+    ctx: click.Context, config_keys: List[str], config_file_path: str
+) -> None:
+    available_config_options = {param.name for param in ctx.command.params}
+    for key in config_keys:
+        if key not in available_config_options:
+            raise InvalidConfigKey(f"Invalid key {key} in {config_file_path}")
```
JelleZijlstra, line 195: If there are multiple invalid keys, could you make the error message list them all at once? That makes for a better debugging experience, since otherwise if you have N invalid keys you have to run Black N times to see them all.

JelleZijlstra (commented):
Thanks! This is probably OK but I left a comment on the issue.

src/black/__init__.py
dankrzeminski32: Agreed, just updated PR.

JelleZijlstra (approved):
Thanks, this looks good. I pushed a slight wording change.

dankrzeminski32:
Great, thanks for the help.

hauntsaninja:
Thank you for adding this feature!

PR #4167, opened 2024-01-25
Simplify code in lines.py

hauntsaninja:
This has been getting a little messy. These changes neaten things up, we don't have to keep guarding against `self.previous_line is not None`, we make it clearer what logic has side effects, we reduce the amount of code that tricky `before` could touch, etc

This change is intended to be a no-op

### DIFF
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -565,13 +565,7 @@ class EmptyLineTracker:
         )
         before, after = self._maybe_empty_lines(current_line)
         previous_after = self.previous_block.after if self.previous_block else 0
-        before = (
-            # Black should not insert empty lines at the beginning
-            # of the file
-            0
-            if self.previous_line is None
-            else before - previous_after
-        )
+        before = max(0, before - previous_after)
         if (
             # Always have one empty line after a module docstring
             self.previous_block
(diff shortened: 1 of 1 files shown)


src/black/lines.py
```diff
@@ -623,9 +619,23 @@ def _maybe_empty_lines(self, current_line: Line) -> Tuple[int, int]:
         user_had_newline = bool(before)
         depth = current_line.depth
 
+        # Mutate self.previous_defs, remainder of this function should be pure
         previous_def = None
         while self.previous_defs and self.previous_defs[-1].depth >= depth:
             previous_def = self.previous_defs.pop()
+        if current_line.is_decorator or current_line.is_def or current_line.is_class:
+            if not current_line.is_decorator:
+                self.previous_defs.append(current_line)
```
JelleZijlstra, line 628:
```suggestion
        if current_line.is_def or current_line.is_class:
            self.previous_defs.append(current_line)
```
This should be equivalent unless I'm missing something

hauntsaninja: Thanks, fixed. Logic was previously inside `_maybe_empty_lines_for_class_or_def` and I'd just copied the guard around that function call.

JelleZijlstra (approved):

PR #4169, opened 2024-01-25
Remove reference

hauntsaninja:
This is out-of-date and just a chore. I don't think this is useful to contributors / can't matter to users since Black doesn't even have a public API.


JelleZijlstra (approved):
Good call, I was thinking of removing this too

Issue #4178, opened 2024-01-26
Deprecate unusual ways to set things in pyproject.toml


JelleZijlstra:
@henryiii shared some useful data on common Black configurations in pyproject.toml files found on PyPI. A few pieces of weirdness stood out:

- Some people set line-length to a string, e.g. `line-length = "100"`. This is very rare, about 25 instances total out of almost 20k (~0.1%).
- We allow both `line_length` and `line-length`. In Henry's sample, `line_length` was used about 3% as much as `line-length`. Our documentation only uses the hyphenated versions.

This might cause problems when we add a JSON schema (#4160), because it's harder to provide useful feedback to users if anything is allowed. For the typing issue, it also just seems confusing to allow setting an integer value to a string.

Proposed actions:
- In the documentation, state clearly that the hyphenated versions are preferred and that you should use the right type.
- Print a warning if we encounter the incorrect versions, and possibly remove support for them in Black 25 next year. I feel this is definitely appropriate for values of the wrong type. For the underscored versions, I feel deprecation is less justifiable as they're more common in practice and less likely to lead to confusion. Curious to hear other opinions on whether that is worth deprecating.

henryiii:
IMO, you could deprecate the unusual types, and produce a warning, and "verbally" deprecate (as in, no warning, but docs discourage it). I think that would be enough to leave the alternate forms out of the Schema, or at the very least have them deprecated there (though most tools don't do much with the info, since it's technically not part of schema v7). Then if usage of the `_` versions decreases, maybe eventually a warning could be produced.

Just a thought.

FYI, SchemaStore used to encourage strictness, but has gone toward encouraging matching what tools accept. See https://github.com/SchemaStore/schemastore/issues/3467#issuecomment-1866413005.

PR #4181, opened 2024-01-27
feat: add schema and validate-pyproject support

### LINKED ISSUE #4160: Can we add a json schema to complete pyproject.toml's `[tool.black]`?

[json schema](https://json-schema.org) can let editor which support LSP to complete some configuration file. Such as:

Screenshot from 2024-01-20 23-24-42

There are many tools have supported json schema: <https://json.schemastore.org/pyproject.json>

Screenshot from 2024-01-20 23-36-38

Can [tool.black](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-format) be supported?

Refer <https://github.com/cmhughes/latexindent.pl/issues/206>:

The steps are following:

1. Create a json schema
2. Add its URL to [schemastore](https://www.schemastore.org/json/).

henryiii:
### Description

This adds a script (Python 3.10+ required) to produce a JSONSchema. I've added it to the package, and provided a helper function / entrypoint to allow [validate-pyproject](https://github.com/abravalheri/validate-pyproject) to use it. You can try it out like this:

```console
$ virtualenv .venv
$ py -m pip install -e . validate-pyproject[all]
$ validate-pyproject -v pyproject.toml
[INFO] black.schema.get_schema defines `tool.black` schema
[INFO] validate_pyproject.api.load_builtin_plugin defines `tool.distutils` schema
[INFO] validate_pyproject.api.load_builtin_plugin defines `tool.setuptools` schema
[WARNING] `blackd:patched_main [d]` - using extras for entry points is not recommended
Valid file: pyproject.toml
```

You can also adjust the tool.black section and see the validation errors. :)

I've also included a `--schemastore` option, which adjusts the output a little for SchemaStore. After this goes in, I'll make a PR to SchemaStore with the schema. The main difference is `enable-unstable-feature` is simplified to just be any string. If someone is getting the schema directly from black, then they get the exact enum of options, but SchemaStore is not versioned.

Closes #4160.

This goes along with #4178 and doesn't include all the interesting ways to set things that are technically currently supported, like using strings instead of bools and using `_` instead of `-`.

### Checklist - did you ...

I'll adjust those things after I see how much/little is needed here. I could add (re)generation of the schema file to CI, some tests using validate-pyproject, etc. Or I could remove parts too.

I also realize I forgot to see if this works on Python 3.8, as the importlib logic is different there.

### DIFF
--- a/scripts/generate_schema.py
+++ b/scripts/generate_schema.py
@@ -9,31 +9,37 @@ import black
 def generate_schema_from_click(
     cmd: click.Command,
 ) -> dict[str, Any]:
-    result = {}
+    result: dict[str, dict[str, Any]] = {}
     for param in cmd.params:
         if not isinstance(param, click.Option) or param.is_eager:
             continue
+
         assert param.name
         name = param.name.replace("_", "-")
-        default = {"default": param.default} if param.default is not None else {}
-        json_type: dict[str, Any]
+
+        result[name] = {}
+
         match param.type:
             case click.types.IntParamType():
-                json_type = {"type": "integer"}
+                result[name]["type"] = "integer"
             case click.types.StringParamType() | click.types.Path():
-                json_type = {"type": "string"}
+                result[name]["type"] = "string"
             case click.types.Choice(choices=choices):
-                json_type = {"enum": choices}
+                result[name]["enum"] = choices
             case click.types.BoolParamType():
-                json_type = {"type": "boolean"}
+                result[name]["type"] = "boolean"
             case _:
                 msg = f"{param.type!r} not a known type for {param}"
                 raise TypeError(msg)
 
         if param.multiple:
-            json_type = {"type": "array", "items": json_type}
+            result[name] = {"type": "array", "items": result[name]}
 
-        result[name] = {**default, **json_type, "description": param.help}
+        result[name]["description"] = param.help
+
+        if param.default is not None and not param.multiple:
+            result[name]["default"] = param.default
+
     return result
 
 
@@ -43,6 +49,7 @@ def generate_schema_from_click(
 def main(schemastore: bool, outfile: IO[str]) -> None:
     properties = generate_schema_from_click(black.main)
     del properties["line-ranges"]
+
     schema: dict[str, Any] = {
         "$schema": "http://json-schema.org/draft-07/schema#",
         "$id": (
@@ -53,6 +60,7 @@ def main(schemastore: bool, outfile: IO[str]) -> None:
         "additionalProperties": False,
         "properties": properties,
     }
+
     if schemastore:
         schema["$id"] = ("https://json.schemastore.org/partial-black.json",)
         schema["properties"]["enable-unstable-feature"]["items"] = {"type": "string"}
--- a/scripts/generate_schema.py
+++ b/scripts/generate_schema.py
@@ -1,0 +1,64 @@
+import json
+from typing import IO, Any
+
+import click
+
+import black
+
+
+def generate_schema_from_click(
+    cmd: click.Command,
+) -> dict[str, Any]:
+    result = {}
+    for param in cmd.params:
+        if not isinstance(param, click.Option) or param.is_eager:
+            continue
+        assert param.name
+        name = param.name.replace("_", "-")
+        default = {"default": param.default} if param.default is not None else {}
+        json_type: dict[str, Any]
+        match param.type:
+            case click.types.IntParamType():
+                json_type = {"type": "integer"}
+            case click.types.StringParamType() | click.types.Path():
+                json_type = {"type": "string"}
+            case click.types.Choice(choices=choices):
+                json_type = {"enum": choices}
+            case click.types.BoolParamType():
+                json_type = {"type": "boolean"}
+            case _:
+                msg = f"{param.type!r} not a known type for {param}"
+                raise TypeError(msg)
+
+        if param.multiple:
+            json_type = {"type": "array", "items": json_type}
+
+        result[name] = {**default, **json_type, "description": param.help}
+    return result
+
+
+@click.command(context_settings={"help_option_names": ["-h", "--help"]})
+@click.option("--schemastore", is_flag=True, help="SchemaStore format")
+@click.option("--outfile", type=click.File(mode="w"), help="Write to file")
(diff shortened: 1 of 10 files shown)


scripts/generate_schema.py
```diff
@@ -0,0 +1,72 @@
...
+        ),
+        "$comment": "black table in pyproject.toml",
+        "type": "object",
+        "additionalProperties": False,
+        "properties": properties,
+    }
+
+    if schemastore:
+        schema["$id"] = ("https://json.schemastore.org/partial-black.json",)
+        schema["properties"]["enable-unstable-feature"]["items"] = {"type": "string"}
```
JelleZijlstra, line 66:
```suggestion
        # The precise list of unstable features may change frequently, so don't bother putting it in SchemaStore
        schema["properties"]["enable-unstable-feature"]["items"] = {"type": "string"}
```

scripts/generate_schema.py
```diff
@@ -0,0 +1,72 @@
...
+def main(schemastore: bool, outfile: IO[str]) -> None:
+    properties = generate_schema_from_click(black.main)
+    del properties["line-ranges"]
+
+    schema: dict[str, Any] = {
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "$id": (
+            "https://github.com/psf/black/blob/main/black/resources/black.schema.json"
+        ),
+        "$comment": "black table in pyproject.toml",
```
JelleZijlstra, line 58:
```suggestion
        "$comment": "tool.black table in pyproject.toml",
```

src/black/resources/black.schema.json
```diff
@@ -0,0 +1,148 @@
...
+      "description": "Python versions that should be supported by Black's output. You should include all versions that your code supports. By default, Black will infer target versions from the project metadata in pyproject.toml. If this does not yield conclusive results, Black will use per-file auto-detection."
+    },
+    "pyi": {
+      "type": "boolean",
+      "description": "Format all input files like typing stubs regardless of file extension. This is useful when piping source on standard input.",
+      "default": false
+    },
+    "ipynb": {
+      "type": "boolean",
+      "description": "Format all input files like Jupyter Notebooks regardless of file extension.This is useful when piping source on standard input.",
```
JelleZijlstra, line 42: Fixing the missing space in #4182

src/black/schema.py
```diff
@@ -0,0 +1,19 @@
+import importlib.resources
+import json
+import sys
+from typing import Any
+
+
+def get_schema(tool_name: str = "cibuildwheel") -> Any:
+    "Get the stored complete schema for black's settings."
+    assert tool_name == "black", "Only black is supported."
```
JelleZijlstra, line 9: Then why is the default `cibuildwheel`?

src/black/schema.py
```diff
@@ -0,0 +1,19 @@
+import importlib.resources
+import json
+import sys
+from typing import Any
+
+
+def get_schema(tool_name: str = "cibuildwheel") -> Any:
+    "Get the stored complete schema for black's settings."
```
JelleZijlstra, line 8:
```suggestion
    """Get the stored complete schema for black's settings."""
```

tox.ini
```diff
@@ -96,3 +96,11 @@ skip_install = True
 commands =
     pip install -e .
     black --check {toxinidir}/src {toxinidir}/tests
+
+[testenv:generate_schema]
+setenv = PYTHONWARNDEFAULTENCODING =
+skip_install = True
```
JelleZijlstra, line 102: Question: Why skip the install and then install anyway below with `pip install -e .`? What's the difference?

JelleZijlstra (commented):
Thank you! Some nits and questions below.

src/black/schema.py
```diff
@@ -0,0 +1,19 @@
+import importlib.resources
+import json
+import sys
+from typing import Any
+
+
+def get_schema(tool_name: str = "cibuildwheel") -> Any:
```
henryiii, line 7:
```suggestion
def get_schema(tool_name: str = "black") -> Any:
```

src/black/schema.py
henryiii: There’s a small, tiny chance I might have started from the file I wrote for cibuildwheel…

tox.ini
henryiii: Mostly because the ones above it do this. Ideally this should build the wheel with PEP 517 & reuse it, but that is a bigger change. (Or use nox, still a little annoyed tox required me to implement `--outfile`)

I can do it with installs, though, that’s how I did it at first before noticing the others in the file do it this way.

JelleZijlstra: Oops, didn't realize we did that elsewhere. I don't personally use the tox targets much. In that case, let's just keep this consistent with what's already there.

henryiii: FYI, the pip install is twice as fast as the default way (which can be fixed with settings I could look up, I think). Might have something to do with why it's that way. My original commit has the default way, so I was able to compare when rebasing.

src/black/resources/black.schema.json
henryiii: Rebased and picked it up (will push soon).

JelleZijlstra (approved):

JelleZijlstra:
@henryiii I fixed a few CI issues but the entry points check appears broken on 3.9 and below:

```
    def test_schema_entrypoint() -> None:
        if sys.version_info < (3, 10):
            eps = importlib.metadata.entry_points()["validate_pyproject.tool_schema"]
>           (black_ep,) = [ep for ep in eps if ep.name == "wheel"]
E           ValueError: not enough values to unpack (expected 1, got 0)
```

henryiii:
diff-shades reporting "Opération was cancelled" is normal?

JelleZijlstra:
It's a known issue, unfortunately. This change is highly unlikely to affect Black's formatting, so I'll merge. Thank you!

henryiii:
Awesome, thanks!

> This change is highly unlikely to affect Black's formatting

I would hope so. 😆

I'll make the SchemaStore contribution next & mention this PR.

hauntsaninja:
So this is somewhat embarrassing, but Black actually fails to parse this file...

PR #4184, opened 2024-01-28
chore: ignore node_modules (produced by a pre-commit check)

henryiii:
### Description

This directory gets created by a pre-commit check (prettier, probably), so adding it to the ignores.

### Checklist - did you ...

(I don't think any of these are needed for this change?)


JelleZijlstra:
Nice consequence is that if I merge this I don't have to approve CI for you any more :)

Issue #4180, opened 2024-01-27
additional newline added to docstring when the previous line length is less than the line length limit minus 1


DetachHead:
# before
```py
"""
87 characters ............................................................................
"""
```

# after
```py
"""
87 characters ............................................................................

"""
```

# playground

https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADOAFldAD2IimZxl1N_Wk8JgdvQ8tSq0OhVZoEptWlmdC__omMouK3ooOylgkoYGVUTpRL_A_26nUoIFOzsQcUzX_N1N5WllahS-T5qlleGaKsl1U08Mc7wSjEUdEm8AAAAADdi2bh7NF1_AAF1zwEAAADAUENLscRn-wIAAAAABFla

veryslowcode:
I believe [4185](https://github.com/psf/black/pull/4185) will fix this.

PR #4185, opened 2024-01-28
fix: additional newline added to docstring when the previous line length is less than the line length limit minus 1

veryslowcode:
### Description

     This resolves the issue mentioned in issue #4180, where a newline will
     be added to a docstring comment (that is approaching the line limit),
     even if the docstring contains a newline before the closing quotes.

### Checklist - did you ...

- [✓] Add an entry in `CHANGES.md` if necessary?
- [✓] Add / update tests if necessary?
- [✓] Add new / update outdated documentation?


src/black/linegen.py
```diff
@@ -477,13 +477,15 @@ def visit_STRING(self, leaf: Leaf) -> Iterator[Line]:
...
-                # closing quotes
+                # the maximum line length, and the closing quote is not
+                # prefixed by a newline then put a line break before
+                # the closing quotes
                 if (
                     len(lines) > 1
                     and last_line_length + quote_len > self.mode.line_length
                     and len(indent) + quote_len <= self.mode.line_length
                     and not has_trailing_backslash
+                    and not leaf.value[-4] == "\n"
```
JelleZijlstra, line 488: This will need to be gated to the preview style (see #4164 for a recent example).

veryslowcode: Thank you for the example. With the most recent push, I believe this should be gated now.

veryslowcode:
I see that diff-shades detected changes, but they appear to be the intended changes based on the original issue. If there is anything remaining that needs to be done/changed, please let me know and I am happy to do so.

src/black/mode.py
```diff
@@ -177,6 +177,7 @@ class Preview(Enum):
     wrap_long_dict_values_in_parens = auto()
     multiline_string_handling = auto()
     typed_params_trailing_comma = auto()
+    docstring_check_for_newline = auto()
```
JelleZijlstra, line 180: Could you document this in future-style.md? (I have a PR in #4187 that will enforce this.)

src/black/linegen.py
```diff
@@ -477,15 +477,22 @@ def visit_STRING(self, leaf: Leaf) -> Iterator[Line]:
...
                 if (
                     len(lines) > 1
                     and last_line_length + quote_len > self.mode.line_length
                     and len(indent) + quote_len <= self.mode.line_length
                     and not has_trailing_backslash
                 ):
-                    leaf.value = prefix + quote + docstring + "\n" + indent + quote
+                    if (
+                        Preview.docstring_check_for_newline in self.mode
+                        and leaf.value[-4] == "\n"
```
JelleZijlstra, line 491:
```suggestion
                        and leaf.value[-1 - quote_len] == "\n"
```

JelleZijlstra (commented):
Thanks, this is close to being ready.

src/black/mode.py
veryslowcode: Yep, it is in there now. I do apologize for missing that.

PR #4187, opened 2024-01-28
Test that preview/unstable features are documented

JelleZijlstra:
In #4096 I added a list of current preview/unstable features to the docs. I think
this is important for publicizing what's in our preview style. This PR adds an
automated test to ensure the list stays up to date in the future.

This doesn't work with pre-commit (which is why I'm making this a draft). Haven't been able to figure out a way to fix that. I considered making it a unit test instead, but that will probably get redistributors unhappy.


Pozaranch14 (approved):
.

hauntsaninja (approved):
This looks good! I think if you want pre-commit to work you have to do repo: local (and also probably avoid any non-stdlib deps)

JelleZijlstra:
Thanks for reviewing!

> I think if you want pre-commit to work you have to do repo: local (and also probably avoid any non-stdlib deps)

I tried that, but the dependency on `black.mode` was hard to get rid of :)

PR #4194, opened 2024-01-29
docs: Refactor pycodestyle/Flake8 compatibility docs

### LINKED ISSUE #4173: Black and flake8 no longer in sync
**Describe the bug**

Empty classes now format differently between black and what flake8 expects:

**To Reproduce**

```python3
class Foo:
    ...
```

Reformats with the latest black to:

```python3
class Foo: ...
```

Which then flails flake8 with:

E701 multiple statements on one line (colon)

**Expected behavior**

Flake8 and black should be compatible in accordance with guidance at: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8

**Environment**

- Black's version: 24.1.0
- OS and Python version: Python 3.10.13
- flake8 --version:  6.1.0 (mccabe: 0.7.0, pycodestyle: 2.11.1, pyflakes: 3.1.0) CPython 3.10.13 on Linux

**Additional context**

RedGuy12:
### Description

Resolves #4173
Related to #3888

### Checklist - did you ...

### DIFF
--- a/docs/compatible_configs/pycodestyle/.flake8
+++ b/docs/compatible_configs/pycodestyle/.flake8
@@ -1,0 +1,3 @@
+[pycodestyle]
+max-line-length = 88
+ignore = E203,E701
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -79,6 +79,7 @@ following will not be formatted:
 
 ## Why does Flake8 report warnings?
 
+Some of Flake8's rules conflict with Black's style. We recommend disabling these rules.
 See [Using _Black_ with other tools](labels/why-pycodestyle-warnings).
 
 ## Which Python versions does Black support?
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -79,8 +79,7 @@ following will not be formatted:
 
 ## Why does Flake8 report warnings?
 
-See
-[Using _Black_ with other tools](guides/using_black_with_other_tools.md#why-pycodestyle-warnings).
+See [Using _Black_ with other tools](labels/why-pycodestyle-warnings).
 
 ## Which Python versions does Black support?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -77,13 +77,10 @@ following will not be formatted:
 - invalid syntax, as it can't be safely distinguished from automagics in the absence of
   a running `IPython` kernel.
 
-## Why are Flake8's E203 and W503 violated?
+## Why does Flake8 report warnings?
 
-Because they go against PEP 8. E203 falsely triggers on list
-[slices](the_black_code_style/current_style.md#slices), and adhering to W503 hinders
-readability because operators are misaligned. Disable W503 and enable the
-disabled-by-default counterpart W504. E203 should be disabled while changes are still
-[discussed](https://github.com/PyCQA/pycodestyle/issues/373).
+See
+[Using _Black_ with other tools](guides/using_black_with_other_tools.md#why-pycodestyle-warnings).
 
 ## Which Python versions does Black support?
 
--- a/docs/guides/using_black_with_other_tools.md
+++ b/docs/guides/using_black_with_other_tools.md
@@ -166,10 +166,10 @@ should be disabled.
 
 ##### `E701` / `E704`
 
-_Black_ will collapse implementations of classes and functions consisting solely of
-"..." or "pass" to a single line. This matches how such examples are formatted in PEP 8.
-It remains true that in all other cases Black will prevent multiple statements on the
-same line, in accordance with PEP 8 generally discouraging this.
+_Black_ will collapse implementations of classes and functions consisting solely of `..`
+to a single line. This matches how such examples are formatted in PEP 8. It remains true
+that in all other cases Black will prevent multiple statements on the same line, in
+accordance with PEP 8 generally discouraging this.
 
 However, `pycodestyle` does not mirror this logic and may raise
 `E701 multiple statements on one line (colon)` in this situation. Its
--- a/docs/guides/using_black_with_other_tools.md
+++ b/docs/guides/using_black_with_other_tools.md
@@ -167,9 +167,9 @@ should be disabled.
 ##### `E701` / `E704`
 
 _Black_ will collapse implementations of classes and functions consisting solely of
-"..." or "pass" on to a single line. This matches how such examples are formatted in
-PEP 8. It remains true that in all other cases Black will prevent multiple statements on
-the same line, in accordance with PEP 8 generally discouraging this.
+"..." or "pass" to a single line. This matches how such examples are formatted in PEP 8.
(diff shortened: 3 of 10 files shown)


docs/the_black_code_style/current_style.md
```diff
@@ -166,7 +166,7 @@ If you use Flake8, you have a few options:
    max-line-length = 80
    ...
    select = C,E,F,W,B,B950
-   extend-ignore = E203, E501, E704
+   extend-ignore = E203, E501, E701, E704
```
JelleZijlstra, line 169: I feel like this whole section should be replaced with a link to the "using Black with other tools" page.

docs/guides/using_black_with_other_tools.md
```diff
@@ -145,7 +145,7 @@ There are a few deviations that cause incompatibilities with _Black_.
 
 ```
 max-line-length = 88
-extend-ignore = E203, E704
+extend-ignore = E203, E701, E704
```
JelleZijlstra, line 148: The paragraph below should explain why E701 and E704 should be ignored too.

JelleZijlstra (commented):
Thanks, this is good, but I think there's some larger areas of possible improvement. If you're up for it, please make these changes; otherwise we can merge this as is and leave the rest for another PR.

docs/guides/using_black_with_other_tools.md
JelleZijlstra: Also I feel like we could be clearer about which paragraph refers to which option. Perhaps a bulleted list with the relevant error code at the beginning of each entry.

docs/the_black_code_style/current_style.md
RedGuy12: I can do these, working on them now.

RedGuy12 (commented):
@JelleZijlstra I've made the requested changes. My thought process:
- I updated the `compatible_configs` folder (and added some for pycodestyle). This was an oversight in #3888. Not sure how useful this folder is, but I didn't think removing it would be in scope here.
- It turns out that `E704` is already disabled by default, so #3888 actually didn't document it correctly. Fixed that here.
- I moved the pycodestyle section above Flake8 and moved the reasoning section there as well. Since Flake8 is a wrapper around pycodestyle, I felt like it made more sense to have the bulk of the docs in the pycodestyle section instead of under Flake8.
- The Bugbear information was previously only in current_style.md. I moved it to using_black_with_other_tools.md, and added a link in current_style.md and in the FAQ.

docs/guides/using_black_with_other_tools.md
```diff
@@ -134,78 +134,124 @@ profile = black
...
+
+##### `E701` / `E704`
+
+PEP 8 states (and has always stated), "Blank lines may be omitted between a bunch of
+related one-liners." _Black_ takes advantage of this by putting the pass statement of
+dummy implementations on the same line as the colon, reducing the number of lines while
+improving readability. This does go against pycodestyle's
+`E701 multiple statements on one line (colon)`. Its disabled-by-default
+`E704 multiple statements on one line (def)` rule may also raise warnings in this
+situation and should not be enabled.
```
hauntsaninja, line 175: This issue isn't really with blank lines, I'd maybe go for something like:
```suggestion
_Black_ will collapse implementations of classes and functions consisting solely
of "..." on to a single line. This matches how such examples are formatted in
PEP 8. It remains true that in all other cases Black will prevent multiple
statements on the same line, in accordance with PEP 8 generally
discouraging this.

However, `pycodestyle` does not mirror this logic and may raise
`E701 multiple statements on one line (colon)` in this situation.
Its disabled-by-default `E704 multiple statements on one line (def)` rule may
also raise warnings and should not be enabled.
```

hauntsaninja (commented):
Thanks for working on this!

docs/faq.md
```diff
@@ -77,13 +77,9 @@ following will not be formatted:
...
 
-## Why are Flake8's E203 and W503 violated?
+## Why does Flake8 report warnings?
 
-Because they go against PEP 8. E203 falsely triggers on list
-[slices](the_black_code_style/current_style.md#slices), and adhering to W503 hinders
-readability because operators are misaligned. Disable W503 and enable the
-disabled-by-default counterpart W504. E203 should be disabled while changes are still
-[discussed](https://github.com/PyCQA/pycodestyle/issues/373).
+See [Using _Black_ with other tools](labels/why-pycodestyle-warnings).
```
JelleZijlstra, line 82:
```suggestion
Some of flake8's rules conflict with Black's style. We recommend disabling these rules. See [Using _Black_ with other tools](labels/why-pycodestyle-warnings).
```
I feel like your wording doesn't really answer the question any more.

docs/guides/using_black_with_other_tools.md
```diff
@@ -134,78 +134,126 @@ profile = black
...
+### Flake8
 
-```ini
+[Flake8](https://pypi.org/p/flake8/) is a wrapper around multiple linters, including
+pycodestyle. As such, it has many of the same issues.
+
+#### Bugbear
+
+It's recommended to use [the Bugbear plugin](https://github.com/PyCQA/flake8-bugbear)
+and enabling
```
JelleZijlstra, line 210:
```suggestion
and enable
```

docs/the_black_code_style/current_style.md
```diff
@@ -143,45 +143,21 @@ significantly shorter files than sticking with 80 (the most popular), or even 79
 by the standard library). In general,
 [90-ish seems like the wise choice](https://youtu.be/wf-BqAjZb8M?t=260).
 
-If you're paid by the line of code you write, you can pass `--line-length` with a lower
-number. _Black_ will try to respect that. However, sometimes it won't be able to without
-breaking other rules. In those rare cases, auto-formatted code will exceed your allotted
-limit.
+If you're paid by the number of lines of code you write, you can pass `--line-length`
```
JelleZijlstra, line 146: I think the previous wording here is fine and more concise.

JelleZijlstra (commented):
Thanks! A few more comments.

docs/compatible_configs/pycodestyle/.flake8
```diff
@@ -0,0 +1,3 @@
+[pycodestyle]
+max-line-length = 88
+ignore = E203,E701
```
asottile-sentry, line 3: pycodestyle does not read the `.flake8` config file

this configuration also will result in quite a lot of frustration as it enables several pairs of incompatible codes

RedGuy12: ah, i just added the files based off what was previously documented in black

what further incompatible codes need to be disabled?

asottile-sentry: personally I would include (at least) all of the disabled-by-default codes

Issue #3952, opened 2023-10-17
Request: supporting sub-configs (aka per-directory config)


jamesbraza:
**Is your feature request related to a problem? Please describe.**

I like `black`'s default of 88-char line length, except for `docs/` renderings, where code blocks may be rendered via Sphinx with 79 char width. This leads to unnecessary scrolling left/right to see the last 9 chars.

It would be useful to enable `black` to have sub-configurations. This was requested in https://github.com/psf/black/issues/1370, and was postponed into the future.

**Describe the solution you'd like**

Enabling subdirectories to have their own `pyproject.toml`, and `black` will accept overrides from that sub-`pyproject.toml`.

**Describe alternatives you've considered**

`ruff` is a project that adopted sub-configs through use of the [`extend` keyword](https://docs.astral.sh/ruff/configuration/#pyprojecttoml-discovery). It has been great, I have:

- One repo-wide config
- Two sub-configs: `docs/` and `tests/`

This can be an example successful implementation of sub-configs.

**Additional context**

One counter argument is simplying invoking `black` once/config:

- Once for the repo root
- Once for the subdirectory with its special config

I think this complicates things (e.g. multiple `make` commands or multiple `pre-commit` hooks), though it is a valid workaround

maanijou:
+1 for this.
We have also usecases where we have different types of modules and scripts and for some of them it makes sense to have a different config. For example, formatting for a utility library, compared to a pyspark code can be different and one might need to customize the formatting for different modules in the code

luator:
+1. I just ran into the exact same use-case that for examples which are included in Sphinx documentation, a shorter line-length would be good.

Jswig:
+ 1 as well. I ran into this recently with Databricks, whose built-in formatter uses black with 88 line length, while the rest of our organization's codebase uses 100 lines

PR #4204, opened 2024-02-02
Ignore pyproject.toml missing tool.black section

### LINKED ISSUE #2863: Black stops looking for config at first pyproject.toml file even if there's no black section in it
Disclaimer: I was hesitant whether I should report this as a bug or rather a "Feature request".

**Describe the bug**

In a project dealing with multiple sub-projects in a single repository (aka. mono-repo style) having more than one `pyproject.toml` file confuses black.
In our example we wanted to have one top-level black configuration,
while also having sub-project specific `pyproject.toml` files (without any black config) in each sub-directory.
Unfortunately, running black on a sub-directory
uses default configuration instead of the one defined in the top-level config file.

**To Reproduce**

Given this file structure:
```
.
|-- pyproject.toml
|-- sub1
|   |-- pyproject.toml
|   `-- test.py
`-- sub2
    `-- test.py
```

With the following contents:
```
==> ./pyproject.toml <==
[tool.black]
line-length = 20

==> ./sub1/pyproject.toml <==
[tool.isort]
profile = "black"

==> ./sub1/test.py <==
a_very_long_line = {"that should be wrapped": "to fit 20 characters line length"}

==> ./sub2/test.py <==
a_very_long_line = {"that should be wrapped": "to fit 20 characters line length"}
```

Running black on both test.py files:
```sh
$ black -v sub1/test.py
Identified `/private/tmp/black-many-configs/sub1` as project root containing a pyproject.toml.
Sources to be formatted: "test.py"
sub1/test.py already well formatted, good job.

All done! ✨ 🍰 ✨
1 file left unchanged.

$ black -v sub2/test.py
Identified `/private/tmp/black-many-configs` as project root containing a pyproject.toml.
Sources to be formatted: "sub2/test.py"
reformatted sub2/test.py

All done! ✨ 🍰 ✨
1 file reformatted.
```

Modifies only the file inside `sub2` directory (the one without `pyproject.toml` file)
even though `sub1/pyproject.toml` doesn't contain any black config section.

**Expected behavior**

Both files should be modified
and black should continue looking for a file
that is relevant for its configuration (in this case, the one that has a `[tool.black]` section).

**Environment**

- Black's version: black, 22.1.0 (compiled: yes)
- OS and Python version: OSX - Python 3.9.9

**Additional context**

A variation of this problem has already been reported in https://github.com/psf/black/issues/1826, where the proposed solution is to introduce support for the `config` option inside the sub-project's `pyproject.toml` config file.

Other (potentially) relevant issues: https://github.com/psf/black/issues/2537, https://github.com/psf/black/issues/2850.

hauntsaninja:
Fixes #2863

This is pretty desirable in a monorepo situation where you have configuration in the root since it will mean you don't have to reconfigure every project.

The good news for backward compatibility is that `find_project_root` continues to stop at any git or hg root, so in all cases repo root coincides with a pyproject.toml missing tool.black, we'll continue to have the project root as before and end up using default config as before (i.e. we're unlikely to randomly start using the user config).

The other thing we need to be a little careful about is that changing find_project_root logic affects what `exclude` is relative to.  Since we only change in cases where there is no config, this only applies where users were using `exclude` via command line arg (and had pyproject.toml missing `tool.black` in a dir that was not repo root).

Finally, for the few who could be affected, the fix is to put an empty `[tool.black]` in pyproject.toml

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,16 @@
 # Change Log
 
+## Unreleased
+
+### Configuration
+
+- _Black_ now ignores `pyproject.toml` that is missing a `tool.black` section when
+  discovering project root and configuration. Since _Black_ continues to use version
+  control as an indicator of project root, this is expected to primarily change behavior
+  for users in a monorepo setup (desirably). If you wish to preserve previous behavior,
+  simply add an empty `[tool.black]` to the previously discovered `pyproject.toml`
+  (#4204)
+
 ## 24.1.1
 
 Bugfix release to fix a bug that made Black unusable on certain file systems with strict
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -457,10 +457,10 @@ of tools like [Poetry](https://python-poetry.org/),
 ### Where _Black_ looks for the file
 
 By default _Black_ looks for `pyproject.toml` containing a `[tool.black]` section
-starting from the common base directory of
-all files and directories passed on the command line. If it's not there, it looks in
-parent directories. It stops looking when it finds the file, or a `.git` directory, or a
-`.hg` directory, or the root of the file system, whichever comes first.
+starting from the common base directory of all files and directories passed on the
+command line. If it's not there, it looks in parent directories. It stops looking when
+it finds the file, or a `.git` directory, or a `.hg` directory, or the root of the file
+system, whichever comes first.
 
 If you're formatting standard input, _Black_ will look for configuration starting from
 the current working directory.
--- a/docs/usage_and_configuration/the_basics.md
+++ b/docs/usage_and_configuration/the_basics.md
@@ -456,7 +456,8 @@ of tools like [Poetry](https://python-poetry.org/),
 
 ### Where _Black_ looks for the file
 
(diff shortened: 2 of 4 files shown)


hauntsaninja:
This has some relation to #1826 in that this gives some folks affected by that issue an option (assuming it's all within the same repo, have pyproject.toml in a parent directory). However, I don't really like #1826 on its own — without #3952 I think it's a little footgun-y.

JelleZijlstra (approved):
Thanks!

alicederyn:
Note: this breaks for any CI process that downloads a git repo via zip, rather than the CLI. This is a pretty standard way to run tools in CI.

cobaltt7:
Sorry, I'm not sure I understand. Why would it be any different if all the files are the same, just downloaded differently?

alicederyn:
Because the zip download does not contain the .git directory.

alicederyn:
To be clear, the issue here is that when the `.git` directory is missing from a CI checkout (a standard optimization as typically linters should not be accessing git history or branch information), the black linter is not finding the `requires-python` line in the `pyproject.toml` unless a `[tool.black]` section has been added.

One way to fix would be for the tool to go back to the last `pyproject.toml` it found if it never finds one higher up.

PR #4206, opened 2024-02-03
Clearer error messages and handeling for different versions.

AshishNarne:
### Checklist - did you ...

- [x] Add an entry in `CHANGES.md` if necessary? Will be adding that now.
- [x] Add / update tests if necessary? N/A.
- [x] Add new / update outdated documentation? New comments on code.


JelleZijlstra:
Thanks, but this isn't a complete implementation. Could you make the code actually use the new logic and add some tests demonstrating how it works?

Issue #4209, opened 2024-02-05
stdin-filename doesn't support UNC path in python 3.8/3.9 on Windows


guotuofeng:
**Describe the bug**

In Windows, the UNC path is not supported when using as `--stdin-filename`. For example, let's say one path is: `\\?\D:\git\OLive\setup.py`

When running 
```bash
python.exe -mblack --stdin-filename \\?\D:\git\OLive\setup.py -v -
```
The following error message will be shown.
![image](https://github.com/psf/black/assets/817030/7b47d183-ec6f-46bc-9b53-e7e98e88d949)

**To Reproduce**

Windows Python pathlib doesn't support UNC path. 
**Expected behavior**
UNC path `\\?\D:\git\OLive\setup.py` should be supported in Windows allow as stdin-filename

**Environment**
* Windows
* Python 3.8.18
* black: 24.1.1
* 

- Black's version: 
- OS and Python version: 

**Additional context**

PR #4210, opened 2024-02-05
fix unc path bug as stdin-filename in Windows python 3.8

guotuofeng:
### Description
Fix https://github.com/psf/black/issues/4209. 

### Checklist - did you ...


guotuofeng:
@dankrzeminski32, would you please take a look?

src/black/__init__.py
```diff
@@ -749,7 +749,12 @@ def get_sources(
 
         # Compare the logic here to the logic in `gen_python_files`.
         if is_stdin or path.is_file():
-            root_relative_path = get_root_relative_path(path, root, report)
+            # From https://stackoverflow.com/questions/42513056/how-to-get-absolute-path-of-a-pathlib-path-object  # noqa: B950
+            # suggest to use resolve instead of absolute prior to Python 3.11
+            # to get the absolute path.
+            # This change is to fix bug:
+            # https://github.com/psf/black/issues/4209
+            root_relative_path = get_root_relative_path(path.resolve(), root, report)
```
JelleZijlstra, line 757: Looks like `.resolve()` will raise an error in some cases if the path doesn't exist or is a broken symbolic link. Is that the right behavior here?

I'm less concerned about using `.absolute()` than that StackOverflow answer. If it's documented in recent versions, we can use `.absolute()` here; it's not going to change behavior now in Python 3.8 and 3.9.

JelleZijlstra: Maybe this is the right solution; I don't know much about Windows paths so I can't tell. I just want to make sure we use a solution that's unlikely to cause new problems in other contexts.

guotuofeng: the resolve raise error before python 3.6, which is already deprecate. Need we support Python prior to 3.7 inclusively?

tests/test_black.py
```diff
@@ -2623,6 +2624,35 @@ def test_symlinks(self) -> None:
         outside_root_symlink.resolve.assert_called_once()
         ignored_symlink.resolve.assert_not_called()
 
+    def test_unc_path(self) -> None:
+        if system() != "Windows":
+            return
+
+        # during my testing, I found the UNC bug happens in both Python 3.8 and 3.9
```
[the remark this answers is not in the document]
guotuofeng: sure, will do

src/black/__init__.py
JelleZijlstra: We don't care about 3.6 or 3.7.

tests/test_black.py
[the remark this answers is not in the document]
guotuofeng: done

dankrzeminski32 (approved):

src/black/__init__.py
hauntsaninja: `resolve` hurts performance and resolving symlinks can result in unexpected behaviour. I'd rather e.g. inline the right version of relative_to on 3.8

guotuofeng: what's your suggestion to fix this bug?

guotuofeng: just tried  Path("\\\\?\\d:\\git\\Olive\\setup.py").absolute().relative_to(Path("d:\\git\\Olive")) on Python 3.10 and 3.11. 
both of them complains like:
image

hauntsaninja: Would you mind testing https://github.com/psf/black/pull/4222 and seeing if it works well for your use case?

guotuofeng: sorry for late response since Chinese New Year vacation. will try and update the thread.

guotuofeng: PR https://github.com/psf/black/pull/4222 fix the issue. will close this PR.

PR #4214, opened 2024-02-06
Remove redundant parentheses in `case` statement `if` guards

dosisod:
### Description

A follow up to #4024 but for `if` guards in `case` statements. I noticed this when #4024 was made stable, and noticed I had some code that had extra parens around the `if` guard.

There might be some edge cases I am not thinking of, but all tests are passing. I put these new changes under the `remove_redundant_guard_parens` preview name.

### Checklist

### DIFF
--- a/tests/data/cases/remove_redundant_parens_in_case_guard.py
+++ b/tests/data/cases/remove_redundant_parens_in_case_guard.py
@@ -1,4 +1,4 @@
-# flags: --minimum-version=3.10 --preview
+# flags: --minimum-version=3.10 --preview --line-length=79
 
 match 1:
     case _ if (True):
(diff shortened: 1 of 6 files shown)


tests/data/cases/remove_redundant_parens_in_case_guard.py
```diff
@@ -0,0 +1,24 @@
+# flags: --minimum-version=3.10 --preview
+
+match 1:
+    case _ if (True):
+        pass
+
+
+match 1:
+    case _ if (
+        True
```
JelleZijlstra, line 10: Add tests with a comment inside the parentheses in various places

JelleZijlstra (approved):
Thanks!

Issue #4205, opened 2024-02-02
`black` ignores file in current folder if current folder is `git` workspace and (parent of) current folder is a symbolic link


bersbersbers:
**Describe the bug**

`black` incorrectly ignores a file in the current folder

**To Reproduce**

```
(.venv) C:\>dir

12/21/2023  10:19 AM    <JUNCTION>     Code [C:\ws]
...
02/02/2024  08:33 AM    <DIR>          ws

(.venv) C:\>cd Code

(.venv) C:\Code>mkdir Bug

(.venv) C:\Code>cd Bug

(.venv) C:\Code\Bug>git init
Initialized empty Git repository in C:/ws/Bug/.git/

(.venv) C:\Code\Bug>echo 1 > bug.py

(.venv) C:\Code\Bug>black --verbose bug.py
Identified `C:\ws\Bug` as project root containing a .git directory.
bug.py ignored: is a symbolic link that points outside C:\ws\Bug
No Python files are present to be formatted. Nothing to do 😴
```

**Expected behavior**
`black` should not ignore `bug.py`

**Environment**

black, 24.1.2.dev11+g632f44bd68 (compiled: no)
Python (CPython) 3.12.1
Windows 10 22H2

hauntsaninja:
Hm, I tried the following and couldn't repro:
```
. λ mkdir actual
. λ ln -sf actual symlink
. λ cd symlink
./symlink λ ls
./symlink λ mkdir bug
./symlink λ cd bug
./symlink/bug λ git init
Initialized empty Git repository in /Users/shantanu/dev/black/actual/bug/.git/
./symlink/bug λ echo "x     = 'asdf'" > bug.py
./symlink/bug λ black --verbose bug.py
Identified `/Users/shantanu/dev/black/actual/bug` as project root containing a .git directory.
Found input source: "bug.py"
reformatted bug.py

All done! ✨ 🍰 ✨
1 file reformatted.
```
Perhaps something Windows/junction-specific is happening?

hauntsaninja:
Okay, there is a `--stdin-filename` variant here that is an issue:
```
λ cat bug.py | black --stdin-filename bug.py -v -
Identified `/Users/shantanu/dev/black/actual/bug` as project root containing a .git directory.
Found input source: "__BLACK_STDIN_FILENAME__bug.py"
x = "asdf"
reformatted bug.py

All done! ✨ 🍰 ✨
1 file reformatted.
[...]

Issue #3552, opened 2023-02-06
Tests do not reflect real-world CLI file-discovery behavior


aaossa:
**Describe the bug**

> This issue is more a heads-up than a bug report.

Follow up of conversation at https://github.com/psf/black/pull/3385. Seems like the test suite does not emulate real-world CLI usage. The paths get mixed up somewhere. **This may be hiding some bugs**: while writing #3385 I wrote a test to detect a confirmed bug (#3384 ) but the test was not failing, so I had to write a test with some patches to emulate the target behavior.

**More details**

When trying to reproduce #3384 in the test suite I wasn't able to because of an exception. I explained a bit in this comment:

> Seems like this can't be tested easily because the exception can't be reproduced in the test suite:
> 
> ```python
> black.get_sources(
>     ctx=ctx,
>     src=("./dir",),
>     quiet=False,
>     verbose=True,
>     include=DEFAULT_INCLUDE,
>     exclude=DEFAULT_EXCLUDE,
>     report=report,
>     extend_exclude=None,
>     force_exclude=None,
>     stdin_filename=None,
> )
> ```
> 
> This function call is similar to the call from the CLI, but the test suite fails because `src` is an invalid value (`is_dir()` returns false), because every test uses `assert_collected_sources` and passes an absolute path to the source/target files. But, in reality, the CLI receives relative paths. This means that the test suite is not reproducing the CLI behavior 100% correctly. This seems like another issue, related to https://github.com/psf/black/issues/3040#issuecomment-1157792654 but not the same.
> 
> This reported bug does not happen in the test, because (as you can see in the patch) the solution uses the absolute path to the target file.
> 
> _Originally posted by @aaossa in https://github.com/psf/black/issues/3384#issuecomment-1314295118_
>             
                
This resulted in an ugly (but working and correct) test at #3385 . 

As you can see, I'm not the first person that noticed that the tests do not reflect the actual CLI usage. This comment from https://github.com/psf/black/issues/3040 reports another issue related to file-collection:

> I've been trying to write a test for this and I'm concerned that the existing exclude tests do not reflect real-world usage.
> 
[...]

PR #4215, opened 2024-02-07
fix: resolve symlinks when checking which files to format

nfvs:
### Description

Currently `find_project_root()` resolves symlinks, but `get_sources()` doesn't, so trying to format a file that contains symlinks will fail:
```
$ python -m black --check /home/nfvs/code/black/src/black/files.py
```

This commit changes methods used in `get_sources()` to, like `find_project_root()`, use `Path.resolve()` instead of `Path.absolute()`, which will follow symlinks.

Also fixes:
- https://github.com/microsoft/vscode-black-formatter/issues/444
- https://github.com/psf/black/issues/4205

### Checklist - did you ...

### DIFF
--- a/src/black/files.py
+++ b/src/black/files.py
@@ -283,7 +283,7 @@ def get_root_relative_path(
 ) -> Optional[str]:
     """Returns the file path relative to the 'root' directory"""
     try:
-        root_relative_path = path.absolute().relative_to(root).as_posix()
+        root_relative_path = path.resolve().relative_to(root).as_posix()
     except ValueError:
         if report:
             report.path_ignored(path, f"is a symbolic link that points outside {root}")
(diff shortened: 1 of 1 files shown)


nfvs:
Test failures may be due to another bug: https://github.com/psf/black/issues/3552

hauntsaninja (changes requested):
Thanks for the PR, but the tests that are failing test important behaviour. Do you have a small repro of the case where you want current behaviour to be different?

hauntsaninja:
If you get a moment, would you mind testing https://github.com/psf/black/pull/4222 and see if it works for your use case?

nfvs:
> If you get a moment, would you mind testing #4222 and see if it works for your use case?

It does fix it, I'll close this PR. Thanks!

PR #4217, opened 2024-02-08
Use W503 instead of E704 in examples

YoannQDQ:
The docs explains why E203 and W503 are used, but then uses E203 and E704 in the config snippets


RedGuy12:
The main branch has updated docs already.

JelleZijlstra:
Yes, your PR targets `stable`; you should target `main` instead, and there I just merged a bigger update to this area from @RedGuy12.

PR #4221, opened 2024-02-11
Simplify check for symlinks that resolve outside root

hauntsaninja:
This should not change any behaviour.

This would have made the #4015 / #4161 bug much more obvious. This makes a case where we resolve more explicit and prevent a resolved path from leaking out via the return.

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -50,9 +50,9 @@ from black.files import (
     gen_python_files,
     get_gitignore,
     get_root_relative_path,
-    normalize_path_maybe_ignore,
     parse_pyproject_toml,
     path_is_excluded,
+    resolves_outside_root_or_cannot_stat,
     wrap_stream_for_windows,
 )
 from black.handle_ipynb_magics import (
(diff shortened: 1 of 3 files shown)


src/black/__init__.py
```diff
@@ -780,7 +777,7 @@ def get_sources(
                 continue
 
             if verbose:
-                out(f'Found input source: "{normalized_path}"', fg="blue")
```
hauntsaninja, line 783: We don't use the value of normalized_path, we do `sources.add(path)`, so no point logging it

tests/test_black.py
```diff
@@ -1760,12 +1760,15 @@ def test_bpo_33660_workaround(self) -> None:
...
         root = Path("/")
         with change_directory(root):
             path = Path("workspace") / "project"
             report = black.Report(verbose=True)
-            normalized_path = black.normalize_path_maybe_ignore(path, root, report)
-            self.assertEqual(normalized_path, "workspace/project")
+            resolves_outside = black.resolves_outside_root_or_cannot_stat(
+                path, root, report
+            )
+            self.assertEqual(resolves_outside, False)
```
JelleZijlstra, line 1771:
```suggestion
            self.assertIs(resolves_outside, False)
```

JelleZijlstra (approved):

PR #4222, opened 2024-02-11
Fix ignoring input files for symlink reasons

### LINKED ISSUE #4077: pathlib exception when symlinks are involved

### LINKED ISSUE #4205: `black` ignores file in current folder if current folder is `git` workspace and (parent of) current folder is a symbolic link

### LINKED ISSUE #4209: stdin-filename doesn't support UNC path in python 3.8/3.9 on Windows

hauntsaninja:
This relates to #4015, #4161 and the behaviour of os.getcwd()

Black is a big user of pathlib and as such loves doing `.resolve()`, since for a long time it was the only good way of getting an absolute path in pathlib. However, this has two problems:

The first minor problem is performance, e.g. in #3751 I (safely) got rid of a bunch of `.resolve()` which made Black 40% faster on cached runs.

The second more important problem is that always resolving symlinks results in unintuitive exclusion behaviour. For instance, a gitignored symlink should never alter formatting of your actual code. This kind of thing was reported by users a few times.

In #3846, I improved the exclusion rule logic for symlinks in `gen_python_files` and everything was good.

But `gen_python_files` isn't enough, there's also `get_sources`, which handles user specified paths directly (instead of files Black discovers). So in #4015, I made a very similar change to #3846 for `get_sources`, and this is where some problems began.

The core issue was the line:
```
root_relative_path = path.absolute().relative_to(root).as_posix()
```
The first issue is that despite root being computed from user inputs, we call `.resolve()` while computing root (likely unecessarily). Which means that `path` may not actually be relative to `root`. So I started off this PR trying to fix that, when I ran into the second issue. Which is that `os.getcwd()` (as called by `os.path.abspath` or `Path.absolute` or `Path.cwd`) also often resolves symlinks!
```
>>> import os
>>> os.environ.get("PWD")
'/Users/shantanu/dev/black/symlink/bug'
>>> os.getcwd()
'/Users/shantanu/dev/black/actual/bug'
```
This also meant that the breakage often would not show up when input relative paths.

This doesn't affect `gen_python_files` / #3846 because things are known to be relative to `root` and always absolute.

Anyway, it looks like #4161 fixed the crash by just swallowing the error and ignoring the file. Instead, we should just try to compute the actual relative path. I think this PR should be quite safe, but we could also consider reverting some of the previous changes; the associated issues weren't too popular.

At the same time, I think there's still behaviour that can be improved and I kind of want to make larger changes, but maybe I'll save that for if we do something like #3952

Fixes #4205, fixes #4209, original report in #4077

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -44,12 +44,12 @@ from black.const import (
     STDIN_PLACEHOLDER,
 )
 from black.files import (
+    best_effort_relative_path,
     find_project_root,
     find_pyproject_toml,
     find_user_pyproject_toml,
     gen_python_files,
     get_gitignore,
-    get_root_relative_path,
     parse_pyproject_toml,
     path_is_excluded,
     resolves_outside_root_or_cannot_stat,
@@ -734,6 +734,7 @@ def get_sources(
     """Compute the set of files to be formatted."""
     sources: Set[Path] = set()
 
+    assert root.is_absolute(), f"INTERNAL ERROR: `root` must be absolute but is {root}"
     using_default_exclude = exclude is None
     exclude = re_compile_maybe_verbose(DEFAULT_EXCLUDES) if exclude is None else exclude
     gitignore: Optional[Dict[Path, PathSpec]] = None
@@ -749,11 +750,12 @@ def get_sources(
 
         # Compare the logic here to the logic in `gen_python_files`.
         if is_stdin or path.is_file():
-            root_relative_path = get_root_relative_path(path, root, report)
-
-            if root_relative_path is None:
+            if resolves_outside_root_or_cannot_stat(path, root, report):
+                if verbose:
+                    out(f'Skipping invalid source: "{path}"', fg="red")
                 continue
 
+            root_relative_path = best_effort_relative_path(path, root).as_posix()
             root_relative_path = "/" + root_relative_path
 
             # Hard-exclude any files that matches the `--force-exclude` regex.
@@ -761,11 +763,6 @@ def get_sources(
                 report.path_ignored(
                     path, "matches the --force-exclude regular expression"
                 )
-                continue
-
-            if resolves_outside_root_or_cannot_stat(path, root, report):
-                if verbose:
-                    out(f'Skipping invalid source: "{path}"', fg="red")
                 continue
 
             if is_stdin:
--- a/src/black/files.py
+++ b/src/black/files.py
@@ -49,6 +49,11 @@ def _load_toml(path: Union[Path, str]) -> Dict[str, Any]:
 
 
 @lru_cache
+def _cached_resolve(path: Path) -> Path:
+    return path.resolve()
+
+
+@lru_cache
 def find_project_root(
     srcs: Sequence[str], stdin_filename: Optional[str] = None
 ) -> Tuple[Path, str]:
@@ -67,9 +72,9 @@ def find_project_root(
     if stdin_filename is not None:
         srcs = tuple(stdin_filename if s == "-" else s for s in srcs)
     if not srcs:
-        srcs = [str(Path.cwd().resolve())]
+        srcs = [str(_cached_resolve(Path.cwd()))]
 
-    path_srcs = [Path(Path.cwd(), src).resolve() for src in srcs]
+    path_srcs = [_cached_resolve(Path(Path.cwd(), src)) for src in srcs]
 
     # A list of lists of parents for each 'src'. 'src' is included as a
     # "parent" of itself if it is a directory
@@ -236,7 +241,7 @@ def find_user_pyproject_toml() -> Path:
     else:
         config_root = os.environ.get("XDG_CONFIG_HOME", "~/.config")
         user_config_path = Path(config_root).expanduser() / "black"
-    return user_config_path.resolve()
+    return _cached_resolve(user_config_path)
 
 
 @lru_cache
@@ -266,27 +271,31 @@ def resolves_outside_root_or_cannot_stat(
     try:
         if sys.version_info < (3, 8, 6):
             path = path.absolute()  # https://bugs.python.org/issue33660
-        resolved_path = path.resolve()
-        return get_root_relative_path(resolved_path, root, report) is None
+        resolved_path = _cached_resolve(path)
     except OSError as e:
         if report:
             report.path_ignored(path, f"cannot be read because {e}")
         return True
-
-
-def get_root_relative_path(
-    path: Path,
-    root: Path,
-    report: Optional[Report] = None,
-) -> Optional[str]:
-    """Returns the file path relative to the 'root' directory"""
     try:
-        root_relative_path = path.absolute().relative_to(root).as_posix()
+        resolved_path.relative_to(root).as_posix()
     except ValueError:
         if report:
             report.path_ignored(path, f"is a symbolic link that points outside {root}")
-        return None
-    return root_relative_path
+        return True
+    return False
+
+
+def best_effort_relative_path(path: Path, root: Path) -> Path:
+    # Precondition: resolves_outside_root_or_cannot_stat(path, root) is False
+    try:
+        return path.absolute().relative_to(root)
+    except ValueError:
+        pass
+    root_parent = next((p for p in path.parents if _cached_resolve(p) == root), None)
+    if root_parent is not None:
+        return path.relative_to(root_parent)
+    # something adversarial, fallback to path guaranteed by precondition
+    return _cached_resolve(path).relative_to(root)
 
 
 def _path_is_ignored(
@@ -339,7 +348,8 @@ def gen_python_files(
 
     assert root.is_absolute(), f"INTERNAL ERROR: `root` must be absolute but is {root}"
     for child in paths:
-        root_relative_path = child.absolute().relative_to(root).as_posix()
+        assert child.is_absolute()
+        root_relative_path = child.relative_to(root).as_posix()
 
         # First ignore files matching .gitignore, if passed
         if gitignore_dict and _path_is_ignored(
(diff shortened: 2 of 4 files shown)


src/black/__init__.py
```diff
@@ -734,6 +734,7 @@ def get_sources(
     """Compute the set of files to be formatted."""
     sources: Set[Path] = set()
 
+    assert root.is_absolute(), f"INTERNAL ERROR: `root` must be absolute but is {root}"
```
hauntsaninja, line 737: We have a similar assert in `gen_python_files` (which this usually calls)

src/black/__init__.py
```diff
@@ -749,11 +750,12 @@ def get_sources(
 
         # Compare the logic here to the logic in `gen_python_files`.
         if is_stdin or path.is_file():
-            root_relative_path = get_root_relative_path(path, root, report)
-
-            if root_relative_path is None:
+            if resolves_outside_root_or_cannot_stat(path, root, report):
```
hauntsaninja, line 753: As you may have guessed from the name, the difference between this and `get_root_relative_path` (from #4161) is the `.resolve()`

src/black/files.py
```diff
@@ -266,27 +271,31 @@ def resolves_outside_root_or_cannot_stat(
...
+    return False
+
+
+def best_effort_relative_path(path: Path, root: Path) -> Path:
+    # Precondition: resolves_outside_root_or_cannot_stat(path, root) is False
+    try:
+        return path.absolute().relative_to(root)
+    except ValueError:
+        pass
+    root_parent = next((p for p in path.parents if _cached_resolve(p) == root), None)
```
hauntsaninja, line 294: I added `_cached_resolve` mainly in case this loop is expensive

src/black/files.py
```diff
@@ -339,7 +348,8 @@ def gen_python_files(
 
     assert root.is_absolute(), f"INTERNAL ERROR: `root` must be absolute but is {root}"
     for child in paths:
-        root_relative_path = child.absolute().relative_to(root).as_posix()
+        assert child.is_absolute()
```
hauntsaninja, line 351: Not new, this should be guaranteed by `path = root / (path.resolve().relative_to(root))` in `get_sources`

tests/test_black.py
```diff
@@ -2626,39 +2626,100 @@ def test_symlinks(self) -> None:
...
             assert_collected_sources(
-                root=Path("target_directory/"),
+                root=root,
                 src=["-"],
                 expected=[],
                 stdin_filename=stdin_filename,
             )
 
-    @patch("black.find_project_root", lambda *args: (THIS_DIR.resolve(), None))
     def test_get_sources_with_stdin(self) -> None:
```
hauntsaninja, line 2700: Not really related, but while I was here I removed some monkeypatching since it's not needed. We probably do a little too much mocking, so the new tests touch the file system

PabloLION:
(fake@)hauntsaninja, Thank you for your continuous and beautiful work, I was waiting for this for a while, and I thought #4221 solves this problem. But sadly, no.
An idea: How about adding a work around flag like `--experimental-force-ignore-symlink-check`, since it's blocking many including me myself? If approved, I'll be glad to PR on this.

hauntsaninja:
PabloLION, thanks for taking a look at this PR. Does this PR solve the thing blocking you? If it does, we can make a release relatively quickly once it's merged.

src/black/__init__.py
PabloLION: This helped me located where the filtering happens and I just added `sources.add(path)` before the old "continue"in line 755/756 and my problem was solved.
Tho this, if you need, I can still test it against my case where my home folder `~` was set to a symlink like `/homes/user` -> `/external/homes/user`. Please don't hesitate to tell me if you need any help! Thanks for the great work again!

PabloLION:
I replied between the code https://github.com/psf/black/pull/4222#discussion_r1485559255
please let me know if assist is needed.

tests/test_black.py
```diff
@@ -2626,39 +2626,100 @@ def test_symlinks(self) -> None:
...
+                    expected=[symlink_proj / "nested" / "another.py"],
+                )
+
     def test_get_sources_with_stdin_symlink_outside_root(
         self,
     ) -> None:
         path = THIS_DIR / "data" / "include_exclude_tests"
         stdin_filename = str(path / "b/exclude/a.py")
         outside_root_symlink = Path("/target_directory/a.py")
+        root = Path("target_dir/").resolve().absolute()
```
JelleZijlstra, line 2691: target_dir/ or target_directory?

tests/test_black.py
JelleZijlstra: Agree we should prefer to avoid mocking.

tests/test_black.py
hauntsaninja: `target_dir` is right. Rename doesn't affect semantics, I did it to make the test clearer, the relevant difference is `/target_directory` is absolute and so is outside root

JelleZijlstra: I see, it would have been slightly clearer if you used two names that didn't look like one was a variant of the other, e.g. `target_dir1` and `target_dir2`. What you have now is fine too though.

src/black/files.py
```diff
@@ -266,27 +271,31 @@ def resolves_outside_root_or_cannot_stat(
...
-
-def get_root_relative_path(
-    path: Path,
-    root: Path,
-    report: Optional[Report] = None,
-) -> Optional[str]:
-    """Returns the file path relative to the 'root' directory"""
     try:
-        root_relative_path = path.absolute().relative_to(root).as_posix()
+        resolved_path.relative_to(root).as_posix()
```
JelleZijlstra, line 280:
```suggestion
        resolved_path.relative_to(root)
```
Doesn't look like `.as_posix()` should ever throw.

JelleZijlstra (approved):
Looks good, thanks. Some optional nits.

PR #4236, opened 2024-02-15
Fix 4227: Black does not print errors with --quiet --check

### LINKED ISSUE #4227: Black does not print errors with --quiet --check
Running Black with both `--quiet` and `--check` on a non-valid file leads to an exit status of 1 (correct) but no error being printed (unexpected).

Easy to reproduce, e.g. with

```python
def foo(): print("bar")
foo()
```

`python -m black --quiet --check test.py` exits with status 1 but does not print anything.

Reproduced with Black 23.12.1 and 24.1.1.

For `--quiet`, help output states:

> Stop emitting all non-critical output. Error messages will still be emitted

This is obviously incorrect.

`--quiet` should be the way to get no output on success and only errors on failure.

dankrzeminski32:
### Description

This closes #4227 

black was previously not printing any information when called with the `--quiet` and `--check` flags. 

the related issue makes a valid point that black should stay quiet on successful runs and notify the user when changes are found. 

I've omitted part of the output in an attempt to keep the output still quiet, but I can add it in if desired. 

For example, when called without the `--quiet` flag: 

```
would reformat ../test_black/test.py

Oh no! 💥 💔 💥
1 file would be reformatted.
```

When called with the `--quiet` flag:

```

Oh no! 💥 💔 💥
1 file would be reformatted.
```

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,7 +44,7 @@
 
 ### Documentation
 
-- Note what happens when --check is called with --quiet (#4236)
+- Note what happens when `--check` is used with `--quiet` (#4236)
 <!-- Major changes to documentation and policies. Small docs changes
      don't need a changelog entry. -->
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,8 @@
 
 ### Output
 
+- print reformat info when black is called with --quiet and --check and potential
+adjustments are found (#4236)
 <!-- Changes to Black's terminal output and error messages -->
 
 ### _Blackd_
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,8 +32,6 @@
 
 ### Output
 
-- print reformat info when black is called with --quiet and --check and potential
-adjustments are found (#4236)
 <!-- Changes to Black's terminal output and error messages -->
 
 ### _Blackd_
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
Thanks, but I think we don't need to change anything here (except perhaps the docs); see the issue.

dankrzeminski32:
@JelleZijlstra Sounds good, I can make a small doc adjustment in this PR or create a new PR if desired.

JelleZijlstra:
Either works!

docs/usage_and_configuration/the_basics.md
```diff
@@ -183,6 +183,8 @@ Don't write the files back, just return the status. _Black_ will exit with:
 - code 1 if some files would be reformatted; or
 - code 123 if there was an internal error
 
+If used in combination with `--quiet` then only the exit code will be returned.
+
```
JelleZijlstra, line 187:
```suggestion
If used in combination with `--quiet` then only the exit code will be returned,
unless there was an internal error.

```

CHANGES.md
```diff
@@ -44,6 +44,7 @@
 
 ### Documentation
 
+- Note what happens when --check is called with --quiet (#4236)
```
JelleZijlstra, line 47:
```suggestion
- Note what happens when `--check` is used with `--quiet` (#4236)
```

PR #4242, opened 2024-02-19
Actually use result  of  best_effort_relative_path

reporter123:
### Description

Don't clobber result of best_effort_relative_path. Some how this was missed during but we had:
     root_relative_path = best_effort_relative_path(path, root).as_posix()
            root_relative_path = "/" + root_relative_path

Obviously not what was intended in PR #4222
### Checklist - did you ...

- [*] Add an entry in `CHANGES.md` if necessary?
- [*] Add / update tests if necessary?
- [*] Add new / update outdated documentation?

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -756,8 +756,6 @@ def get_sources(
                 continue
 
             root_relative_path = best_effort_relative_path(path, root).as_posix()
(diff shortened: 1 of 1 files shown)


hauntsaninja:
This change isn't correct, the result is used, it just needs a slash preprended.

Issue #2547, opened 2021-10-19
Plugin can't find black python module


scottwn:
**Describe the bug**
I have installed Black using vim-plug. Installation is successful and I can update the plugin no problem. However when I run `black` inside Vim I get the following error
```
Error detected while processing /Users/scott/.vim/plugged/black/autoload/black.vim:
line  195:
Traceback (most recent call last):
  File "<string>", line 97, in <module>
ModuleNotFoundError: No module named 'black'
Error detected while processing function black#Black:
line   10:
Traceback (most recent call last):
  File "<string>", line 1, in <module>
NameError: name 'Black' is not defined
```

**To Reproduce**
Install Black via vim-plug
Run `black` in Vim

**Expected behavior**

I expect Black to run successfully

**Environment (please complete the following information):**

- Version: main
- OS and Python version: Mac OS 11.6 Big Sur, Python 3.10

**Does this bug also happen on main?**
Yes, I've forced vim-plug to use `main` and we can see that in the `PlugUpdate` output.
```
  1 Updated. Elapsed time: 3.993817 sec.
  2 [============]
  3
  4 - Finishing ... Done!
  5 - neoformat: Already up to date.
  6 - MatchTagAlways: Already up to date.
  7 - vim-es2015-snippets: Already up to date.
  8 - vim-react-snippets: Already up to date.
  9 - vim-jsx: Already up to date.
 10 - vim-prettier: Already up to date.
 11 - tsuquyomi: Already up to date.
 12 - typescript-vim: Already up to date.
 13 - vim-snippets: Already up to date.
 14 - ultisnips: Already up to date.
 15 - black: Your branch is up to date with 'origin/main'.
 16 - vim-javascript: Already up to date.
```

**Additional context**

I am using Vim inside Fish shell.

AntiCompositeNumber:
Having the same issue, but with Vundle instead of vim-plug. Removing the `~/.vim/black/` folder worked.

samuel-emrys:
I'm having this issue, but nuking `~/.vim/plugged/black` doesn't seem to have any effect.

Contents of my `~/.vimrc`:
```
call plug#begin('~/.vim/plugged')
Plug 'psf/black', { 'branch': 'stable' }
call plug#end
```

`:PlugInstall` appears to successfully install black, and re-running it I get:

```
- black: Already installed
```
The traceback of the error I'm getting is:

```
Error detected while processing /home/user/.vim/plugged/black/autoload/black.vim:
line  202:
[...]

PR #4245, opened 2024-02-20
[WIP] Allow Black to use a different Python version than Vim

sbivol:
### Description

This branch should fix https://github.com/psf/black/issues/2547

It is sometimes the case that Vim was built with a different Python version than the one used by Black.
One such scenario is when the black virtualenv was created by an older version of Vim and then the system was upgraded in place.
Another scenario is when `g:black_use_virtualenv` is disabled and the manually created environment in `g:black_virtualenv` uses a different Python version. Right now Ubuntu 24.04 builds Vim with Python 3.12 but the system uses Python 3.11 as default, making this scenario more likely to happen:

    $ python3 --version
    Python 3.11.8

    $ vim --version | grep lpython
    Linking: gcc -Wl,-Bsymbolic-functions -flto=auto -ffat-lto-objects -Wl,-z,relro -Wl,-z,now -Wl,--as-needed -o vim -lm -ltinfo -lselinux -lsodium -lacl -lattr -lgpm -L/usr/lib/python3.12/config-3.12-x86_64-linux-gnu -lpython3.12 -ldl -lm

Currently, `:BlackVersion` reports Vim's Python version instead of the one found in the virtualenv. This PR fixes that inconsistency as well.

Finally, I added a few TODO comments for issues that I encountered while testing various scenarios but didn't have the time to fix in this PR.

### Checklist - did you ...

### DIFF
--- a/autoload/black.vim
+++ b/autoload/black.vim
@@ -55,10 +55,12 @@ def _get_python_binary(exec_prefix, pyver=sys.version_info[:3]):
     return exec_path
   raise ValueError("python executable not found")
 
-def _get_python_version(exec_path):
+def _get_python_version(exec_path, as_string=False):
   import subprocess
   version_proc = subprocess.run([exec_path, "--version"], stdout=subprocess.PIPE, text=True)
-  version = tuple(map(int, version_proc.stdout.split(" ")[1].strip().split(".")))
+  version = version_proc.stdout.split(" ")[1].strip()
+  if not as_string:
+    version = tuple(map(int, version.split(".")))
   return version
 
 def _get_pip(venv_path):
@@ -231,7 +233,9 @@ def BlackUpgrade():
   _initialize_black_env(upgrade=True)
 
 def BlackVersion():
-  print(f'Black, version {black.__version__} on Python {sys.version}.')
+  virtualenv_path = Path(vim.eval("g:black_virtualenv")).expanduser()
+  virtualenv_python_version = _get_python_version(_get_python_binary(virtualenv_path), as_string=True)
+  print(f'Black, version {black.__version__} on Python {virtualenv_python_version}.')
 
 EndPython3
 
--- a/autoload/black.vim
+++ b/autoload/black.vim
@@ -34,7 +34,7 @@ FLAGS = [
 ]
 
 
-def _get_python_binary(exec_prefix, pyver):
+def _get_python_binary(exec_prefix, pyver=sys.version_info[:3]):
   try:
     default = vim.eval("g:pymode_python").strip()
   except vim.error:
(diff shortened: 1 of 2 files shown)


autoload/black.vim
```diff
@@ -51,39 +53,52 @@ def _get_python_binary(exec_prefix, pyver):
...
       print("Either use your system package manager (or pip) to upgrade black separately,")
       print("or modify your vimrc to have 'let g:black_use_virtualenv = 1'.")
       return False
     else:
-      # Nothing needed to be done.
+      if virtualenv_site_packages not in sys.path:
+        sys.path.insert(0, virtualenv_site_packages)
       return True
 
+  # TODO: Is this check about the Vim plugin, or about Black?
```
JelleZijlstra, line 94: It's about Black primarily; Black itself requires 3.8+ (since Python 3.7 is no longer supported). By default I would want the same version requirement for the Vim plugin, but if there's some reason to keep supporting older versions in the plugin code, I'm open to it.

autoload/black.vim
```diff
@@ -120,7 +135,12 @@ def _initialize_black_env(upgrade=False):
   return True
 
 if _initialize_black_env():
-  import black
+  # TODO: Better handling of the case when import succeeds but Black doesn't
```
JelleZijlstra, line 138: I don't understand this comment; if Black is uninstalled, surely the import would fail?

JelleZijlstra (commented):
Thank you!

autoload/black.vim
sbivol: Unfortunately, [not always](https://github.com/pypa/pip/issues/11835). This can be triggered by Vim built with Py3.12 importing black from `env/lib/python3.11/site-packages`, which this PR makes possible.
If black is uninstalled after it was used once in such a setup, then `pip uninstall` leaves `site-packages/black/__pycache__/` behind and that causes `import black` to succeed:
```
$ env/bin/python3 -c "import black; print(dir(black)); print(black.__file__)"
['__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__']
None
```
It's a corner case of a corner case, but it's a pain to troubleshoot when it bites.

sbivol:
An alternative approach is to detect the Python version in the virtualenv, and if it doesn't match what Vim was built with, then go through these steps:
- inform the user that Vim won't use that virtualenv
- reset `g:black_virtualenv` to its default value
- create the default virtualenv as usual

This approach wouldn't enable mixing different Python versions and would use a bit of extra space for a new ~/.vim/black virtualenv, but should still leave the user with a working Black and no chance of confusion.

sbivol:
After giving this some more thought, and further discussion in https://github.com/pypa/pip/issues/11835, I'm now convinced that my approach isn't ideal. Mixing Python versions can have undesired side effects.
I am going to rework this PR.

sbivol:
I used a different approach in #4675 which should hopefully make the changes easier to review and to test.
Closing this PR.

PR #4248, opened 2024-02-22
fix: Don't move comments while splitting delimiters

### LINKED ISSUE #4220: `black` internal error on line-broken `==`
**Describe the bug**

`black` crashes when encountering a `type: ignore` comment on a line-broken `==`, where the `==` is on the _first_ line.

**To Reproduce**

For example, take this code:

```python
z = (
    x == # type: ignore[foo]
    y # type: ignore[foo]
  ) # <-- to trigger black
```

```sh
$ black file.py
```

The resulting error is:

> error: cannot format -: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /var/folders/0m/7yv6pgt57vzdqfbzcr8lztkh0000gp/T/blk_zliifnke.log

And the diff:
```diff
--- src
+++ dst
@@ -110,8 +110,6 @@
         )  # /Compare
     )  # /If
   type_ignores=
     TypeIgnore(
     )  # /TypeIgnore
-    TypeIgnore(
-    )  # /TypeIgnore
 )  # /Module
\ No newline at end of file
```

Note: if you move the `==` to the second line:

```python
z = (
    x # type: ignore[foo]
    == y # type: ignore[foo]
  ) # <-- to trigger black
```

then no problem.

**Environment**

- Black's version:
```sh
$ black --version
black, 24.1.1 (compiled: yes)
Python (CPython) 3.11.7
```
- OS and Python version:
```sh
$ python3 --version
Python 3.11.7
$ uname -v
Darwin Kernel Version 22.6.0: Sun Dec 17 22:13:25 PST 2023; root:xnu-8796.141.3.703.2~2/RELEASE_ARM64_T6020
```

RedGuy12:
### Description

Resolves #4220. Black moved the + at the end of the line to the start of the next line, then unintentionally moved any comments as well. This could cause comments to be merged. To fix this, I stopped Black from moving any comments with the delimiter. It only crashed when merging `type: ignore` comments, but I think it makes sense to keep *all* comments on their original line instead of following the delimiter. This does not break the stability policy since already-Blackened code will not change.

Depends on #4257 for lint to succeed.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Don't move comments along with delimiters, which could cause crashes (#4248)
+
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -1194,8 +1194,11 @@ def delimiter_split(
     for leaf_idx, leaf in enumerate(line.leaves):
         yield from append_to_line(leaf)
 
-        for comment_after in line.comments_after(leaf):
-            yield from append_to_line(comment_after)
+        previous_leaf = line.leaves[leaf_idx - 1] if leaf_idx > 0 else None
+        previous_priority = previous_leaf and bt.delimiters.get(id(previous_leaf))
+        if previous_priority != delimiter_priority:
+            for comment_after in line.comments_after(leaf):
+                yield from append_to_line(comment_after)
 
         lowest_depth = min(lowest_depth, leaf.bracket_depth)
         if leaf.bracket_depth == lowest_depth:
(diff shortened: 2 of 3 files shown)


RedGuy12:
Test failure is caused by
```diff
 bad_split3 = (
-    "What if we have inline comments on "  # First Comment
-    "each line of a bad split? In that "  # Second Comment
-    "case, we should just leave it alone."  # Third Comment
+    "What if we have inline comments on "  # First Comment  # Second Comment  # Third Comment
+    "each line of a bad split? In that "
+    "case, we should just leave it alone."
 )
```

RedGuy12:
I have extracted the unrelated refactorings to #4257 for ease of review. That needs to be merged before lint passes on this PR.

JelleZijlstra (approved):
Thank you!

PR #4251, opened 2024-02-26
Remove usage of `pkg_resources` in `docs/conf.py`

Viicos:
### Checklist - did you ...


JelleZijlstra (approved):
Looks like `importlib.metadata.version` was added in 3.8, so this should be safe.

Issue #4254, opened 2024-02-27
Trailing commas in function params cause line splits in type params instead


sumezulike:
**Describe the bug**

A function with type parameters gets formatted incorrectly if the function parameters have a magic trailing comma.

**To Reproduce**

For example, take this code:

```python
def func1[T](a,): pass
def func3[T,](a,): pass
def func3[T,](a,) -> R[X, Y,]: pass
```

And run it with these arguments:

```sh
$ black file.py --target-version py312
```

The resulting code is:

```python
def func1[
    T
](a,):
    pass


def func3[
    T,
](a,):
    pass


def func3[
    T,
](a,) -> R[
    X,
    Y,
]:
    pass

```
The type parameters are split instead of the function parameters.

**Expected behavior**

I would expect this result.
```python
def func1[T](
    a,
):
    pass


def func2[
    T,
](
    a,
):
    pass


def func3[
    T,
](
    a,
) -> R[
    X,
    Y,
]:
    pass

```

**Environment**

`$ black --version`
black, 24.2.0 (compiled: yes)
Python (CPython) 3.12.1

**Additional context**

I am currently drafting a PR to fix this bug.
My current approach is to rework `should_split_funcdef_with_rhs` to also check for the position of the trailing comma in type and function parameters.
If I am overlooking something with that approach or if you have a better idea on how to fix this, please feel free to let me know.

sth:
This now (25.1.0) works as expected, thanks to #4553.

dominiquegarmier:
this is still a bug if you use constrained generics:

```python
def func1[T: (int, str)](a,): pass
```

turns into:
```python
def func1[T: (
    int,
    str,
)](a,):
    pass
```

with
```
black --version
black, 25.1.0 (compiled: yes)
Python (CPython) 3.12.7
```
i opened #4740

Issue #4071, opened 2023-11-26
Wrapping lines of a function signature that uses the new python 3.12 generic type parameter syntax


Questionsdenombres:
**Context:**
- When a function has many parameters, it's more readable when on multiple lines, enforcing black to wrap lines adding a comma after the last parameter.
- See https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines
```python
def func(a: int, b: int) -> int:
    return a + b 

def func(
    a: int,
    b: int,
) -> int:
    return a + b
```

**python 3.12 new feature:**
- https://docs.python.org/3/whatsnew/3.12.html#summary-release-highlights
- A generic function can now have type parameters used by the static type checker.
- See [PEP 695](https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep695), [type](https://docs.python.org/3/reference/simple_stmts.html#type) parameter syntax and the type statement
- See https://peps.python.org/pep-0695/#summary-examples
```python
def func[T](a: T, b: T) -> T:
    return a
```

**Example in the current _Black_ style:**
Enforcing black to wrap lines gives the ugly:
```python
def func[
    T
](a: T, b: T,) -> T:
    return a
```

**Desired style:**
```python
def func[T](
    a: T,
    b: T,
) -> T:
    return a + b
```

**Additional context**
python==3.12
black==23.11.0
Windows==11
VScode==1.84.2
VScode extension! Black Formatter==v2023.6.0

**Credits:**
Thank you for your great tool.
Patrick, Paris, France.

JelleZijlstra:
Yes, we should change this. PR welcome!

JackismyShephard:
What is the ETA on this?

hauntsaninja:
Are you interested in making a PR?

JackismyShephard:
Well, sadly I don't quite have the time for that right now.

AndreaCensi:
@JackismyShephard if you need to undo what `black` did, try `ruff`: it will do exactly what you want

FlorinAsavoaie:
FYI: An attempt to fix this is found in #4272 but doesn't seem trivial.

Feuermurmel:
Just converted a large-ish code-base to the new syntax and found an even more egregious example. 😅 I thought that this example could maybe serve as an additional test case because the return type is also wrapped:

```python
def something_something_function[
    T: Model
](param: list[int], other_param: type[T], *, some_other_param: bool = True) -> QuerySet[
[...]

PR #4255, opened 2024-02-27
Fix type param formatting in funcdef with trailing comma

### LINKED ISSUE #3929: Line wrapping on generic functions using the type parameter syntax
**Describe the style change**

The suggested style change regards the type parameter syntax introduced in Python 3.12. 

When formatting lines too long, I think line-wrapping should occur in the arguments' parenthesis rather than in the type parameters' brackets.

This would work with any generic function using the new type parameter syntax with a long enough signature (longer than the `line-length`).

```python

type SupportedDataTypes = float
type Tensor = list

def sum_backward[
T: SupportedDataTypes
](a: Tensor[T], grad: Tensor[T]) -> dict[Tensor[T], Tensor[T]]:
    ...
```
**Desired style**

I think this kind of line-wrapping, keeping the type parameters list on a single line when possible, should be preferred over what currently happens as it is more readable.

```python

type SupportedDataTypes = float
type Tensor = list

def sum_backward[T: SupportedDataTypes](
    a: Tensor[T], 
    grad: Tensor[T]
) -> dict[Tensor[T], Tensor[T]]:
    ...
```

sumezulike:
### Description

This change adds additional checks to the decision whether to split a funcdef from the left or right-hand side.
Specifically, this adds the following condition to `should_split_funcdef_with_rhs`:
If a funcdef has type parameters and its parameters also have a trailing comma, return `True`, to indicate that the funcdef should not be split from the left first.

This fixes the bug described in #4254.

I am currently trying to also fix #3929 and #4071 since they seem related, but that seems to require a little more work. Any suggestions are appreciated.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Don't move comments along with delimiters, which could cause crashes (#4248)
+
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,12 +12,12 @@
 
 - Don't move comments along with delimiters, which could cause crashes (#4248)
 
-- Fixed a bug where lines were split incorrectly for function definitions with type
-  params and trailing commas (#4255)
-
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
+
+- Fixed a bug where lines were split incorrectly for function definitions with type
+  params and trailing commas (#4255)
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@
 
 - Don't move comments along with delimiters, which could cause crashes (#4248)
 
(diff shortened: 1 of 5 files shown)


sumezulike:
I thought it might be best to just submit a fix for the bug and work on the other changes in a separate PR.
Happy about any feedback!

CHANGES.md
```diff
@@ -10,6 +10,11 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Don't move comments along with delimiters, which could cause crashes (#4248)
+
+- Fixed a bug where lines were split incorrectly for function definitions with type
```
JelleZijlstra, line 15: This needs to be a change to the preview style instead.

JelleZijlstra: (And to be clear, the bigger ask here is that the style change itself should only be applied in the preview style, not just that the changelog entry be moved.)

sumezulike: Oh, so sorry about that, I misunderstood.

sumezulike:
Closing this in favor of #4272

PR #4270, opened 2024-03-09
Fix AST safety check false negative

### LINKED ISSUE #4268: AST safety check fails to catch incorrect f-string change
**Describe the style change**

I wonder if f-strings should be formatted at all.

```python
print(f"{"|".join(['a','b','c'])}")
```

**Desired style**

```python
print(f"{"|".join(['a','b','c'])}")
```

> [!NOTE]
> There's no change to the input here.

**Additional context**

Black formats the `"|"` to become `" | "` and that affects the output which cannot be desired.

Black formats the snippet like this currently:

```python
print(f"{" | ".join(['a','b','c'])}")
```

JelleZijlstra:
Fixes #4268

Previously we would allow whitespace changes in all strings, now
only in docstrings.

### DIFF
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -110,6 +110,10 @@ def lib2to3_unparse(node: Node) -> str:
     return code
 
 
+class ASTSafetyError(AssertionError):
+    """Raised when Black's generated code is not equivalent to the old AST."""
+
+
 def _parse_single_version(
     src: str, version: Tuple[int, int], *, type_comments: bool
 ) -> ast.AST:
@@ -154,9 +158,20 @@ def _normalize(lineend: str, value: str) -> str:
     return normalized.strip()
 
 
-def stringify_ast(node: ast.AST, depth: int = 0) -> Iterator[str]:
+def stringify_ast(node: ast.AST) -> Iterator[str]:
     """Simple visitor generating strings to compare ASTs by content."""
+    return _stringify_ast(node, [])
 
+
+def _stringify_ast_with_new_parent(
+    node: ast.AST, parent_stack: List[ast.AST], new_parent: ast.AST
+) -> Iterator[str]:
+    parent_stack.append(new_parent)
+    yield from _stringify_ast(node, parent_stack)
+    parent_stack.pop()
+
+
+def _stringify_ast(node: ast.AST, parent_stack: List[ast.AST]) -> Iterator[str]:
     if (
         isinstance(node, ast.Constant)
         and isinstance(node.value, str)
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
Seems like we got unlucky and someone added 3.12+ syntax to a diff-shades project just two hours ago (https://github.com/koaning/scikit-lego/commit/64485a9d99a24267a25a57dbd66214796d80b70a#diff-db41f509a7991f7b7579a72fe7bdba315b5ab82bdde2f237c6a53d03396c0081R97 line 97).

src/black/parsing.py
```diff
@@ -215,6 +238,9 @@ def stringify_ast(node: ast.AST, depth: int = 0) -> Iterator[str]:
                 normalized = value.rstrip()
             else:
                 normalized = value
-            yield f"{'  ' * (depth + 2)}{normalized!r},  # {value.__class__.__name__}"
+            yield (
+                f"{'    ' * (len(parent_stack) + 2)}{normalized!r},  #"
```
hauntsaninja, line 242: nit:
```suggestion
                f"{'    ' * (len(parent_stack) + 1)}{normalized!r},  #"
```

tests/test_black.py
```diff
@@ -2822,6 +2809,113 @@ def test_format_file_contents(self) -> None:
...
+            ''',
+        )
+
+    def test_assert_equivalent_fstring(self) -> None:
+        major, minor = sys.version_info[:2]
+        if major < 3 or (major == 3 and minor < 12):
+            pytest.skip("relies on 3.12+ syntax")
+        # https://github.com/psf/black/issues/4268
+        self.check_ast_equivalence(
+            """print(f"{"|".join(['a','b','c'])}")""",
```
hauntsaninja, line 2898: possibly this was meant:
```suggestion
            """print(f"{"|".join([a,b,c])}")""",
```

hauntsaninja (approved):
Thanks, two small things

tests/test_black.py
JelleZijlstra: I confused myself with shell quoting, I thought Black somehow removed the quotes. I'll keep your version so we have two tests.

src/black/parsing.py
```diff
@@ -110,6 +110,10 @@ def lib2to3_unparse(node: Node) -> str:
     return code
 
 
+class ASTSafetyError(Exception):
```
spyoungtech, line 113: This should probably inherit from `AssertionError` since you are replacing `raise AssertionError` with this new exception.

JelleZijlstra: We don't provide compatibility guarantees for this function. I don't think inheriting from AssertionError makes semantic sense.

spyoungtech: Thanks for the response. That makes sense considering that many people probably don't use this functionality directly and I maybe agree inheriting from `AssertionError` is odd from a semantic/purity standpoint, even if it's less practical for compatibility (which as you mention is a non-goal here).

PR #4272, opened 2024-03-10
Improve function definition wrapping

sumezulike:
### Description

This is intended to fix issues like #3929, #4071, and #4254, where the addition of the new type parameter syntax in Python 3.12 has caused suboptimal formatting of function definitions.

The new transformer for function definition inspects the given function definition, then splits at the first thing found from the following list:
  - function parameters with magic trailing comma
  - return type with magic trailing comma
  - type parameters with magic trailing comma
  - return type longer than max line length
  - type parameters longer than max line length
  - function parameters (as default)
  
For example, assuming a very small max line length
```py
def f[T](a): ...
```
is formatted as
```py
def f[T](
    a
): ...
```
but
```py
def f[T,](a): ...
```
becomes
```py
def f[
    T,
](a): ...
```
because the trailing comma gives the type parameters a higher priority.

This only really changes how functions with type parameters are split compared to the stable style.

A function that only has one of type params, function params or a return type is just passed to `left_hand_split`.

#### Possible issues

One test is currently failing, in `funcdef_return_type_trailing_comma.py`:
```py
def foo(a) -> list[
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
]:  # abpedeifnore
    pass
```
```py
# output
def foo(
    a,
) -> list[
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
]:  # abpedeifnore
    pass
```
With the new formatter, the test input is kept as is, which I feel is better.

I added the part with the long strings because it seems logical and some comments were considering going in that direction.
`line_to_string` has a warning about being computationally expensive, but I would think it should be fine for two lines.
A longer character width than the max line length might not be the most useful check though.

If the choice for splitting falls on return type, the line is passed to `right_hand_split`. That was my solution to not duplicate any more code, but that could maybe introduce inconsistencies? I couldn't think of any, but it feels a bit off.

I'm grateful for any feedback and suggestions!

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,8 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Improve wrapping of function definitions in edge-cases (#4272)
+
 ### Configuration
 
 <!-- Changes to how Black can be configured -->
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -34,6 +34,8 @@ Currently, the following features are included in the preview style:
   quotes of a docstring
 - `remove_redundant_guard_parens`: Removes redundant parentheses in `if` guards for
   `case` blocks.
+- `function_def_wrap_order`: extends function definition wrapping logic to avoid
+  suboptimal formatting, especially for PEP 695 type parameters
 
 (labels/unstable-features)=
 
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -7,7 +7,7 @@ import sys
 from dataclasses import replace
 from enum import Enum, auto
 from functools import partial, wraps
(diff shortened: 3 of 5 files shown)


jbms:
Why was this PR closed?

sumezulike:
I closed it because this issue turned out to require more extensive and complex changes than I initially assumed. Since I sadly didn't have the time to figure out all the intricacies in blacks inner workings I had to give it up. You're welcome to continue where I left off although I'm not sure I was on the right path.

PR #4273, opened 2024-03-12
Fix for #4264: --line-ranges formats entire file when ranges are at EOF

### LINKED ISSUE #4264: `--line-ranges` formats entire file when ranges are at EOF
Line ranges seem to format an entire file when the range goes to the end of the file

https://github.com/microsoft/vscode-black-formatter/issues/472

- Take a file that is unformatted
- Set the line ranges to any portion as long as the last line is EOF
- The entire file is formatted

For example, take this code:

** Ensure empty last line **
```python
abc = 'def'
ghi= 'jkl'
mno='pqr'
stu='vwx'

```

```sh
black --line-ranges 4-5 file.py
```

~The resulting error is:~

~> cannot format file.py: INTERNAL ERROR: ...~

**Expected behavior**

Only `stu='vwx'` being formatted, and not lines 1-3

**Environment**

- Black's version: 24.2.0
- OS and Python version: MacOS 14.3.1, Python 3.10.3

**Additional context**

sumezulike:
### Description

This fixes #4264. The issue was that the empty last line does not count as a line to `adjusted_lines` because it is not in the list returned by `str.splitlines`. Since `adjusted_lines` is only called on the second pass of `_format_str_once` in `format_str`, the first pass would format the code correctly, then the "invalid" line would get removed from `lines`, and the second pass would format the whole code.

I added a small change to `adjusted_lines` to cap the end value of any line tuple. For example, `--line-ranges 1-100` gets reduced to `(1, 4)` if the code is only four lines long.
One could change `if end > original_line_count` to `if end == original_line_count + 1` to only allow this one additional line, but I think allowing an oversized range to just cover the rest of the code is not surprising behavior, slices act similarly.

I also added a call to `adjusted_lines` with the unmodified source code before the first pass of `_format_str_once`. This is an additional computational expense but ensures consistency.

### Checklist - did you ...


JelleZijlstra:
cc @yilei

yilei:
Thanks for tagging, @JelleZijlstra!

> I also added a call to adjusted_lines with the unmodified source code before the first pass of _format_str_once. This is an additional computational expense but ensures consistency.

Why is this still necessary, after the change to adjusted_lines to cap the end value?

Could you also add a test file next to https://github.com/psf/black/blob/main/tests/data/cases/line_ranges_basic.py ?

This opens up a question on what happens when you specify a `--line-ranges=` that's outside of the unformatted file. This change makes it valid when you specify a larger `<END>`, but if the entire range is outside then the entire file is still formatted. How about make it format nothing if everything is outside of the range too? The implementation could be, instead of changing `adjusted_lines`, to call a new `sanitize_lines(lines, src_contents)` once in `format_str`:

```python
def format_str(...):
    if lines:
        lines = sanitize_lines(lines, src_contents)
        if not lines:
            return src_contents  # Nothing to format
    dst_contents = _format_str_once(src_contents, mode=mode, lines=lines)
    ...
```

sumezulike:
Thank you so much for the feedback!

Calling `adjusted_lines` beforehand is not really necessary. I just used it like one would `sanitize_lines` to make sure that any lines that would be removed on the second pass would already be removed on the first. Writing a new function for that definitely makes more sense!

Thank you also for noticing that moving the entire range out of the file still formats everything, I'll fix that as suggested.

tests/test_ranges.py
```diff
@@ -183,3 +183,60 @@ def test_diffs(lines: List[Tuple[int, int]], adjusted: List[Tuple[int, int]]) ->
...
+)
+def test_sanitize(
+    lines: List[Tuple[int, int]], sanitized: List[Tuple[int, int]]
+) -> None:
+    source = """\
+1. import re
+2. def func(arg1,
+3.   arg2, arg3):
+4.   pass
+"""
```
yilei, line 241: Can you also add a case for source not ending with a newline?

tests/data/cases/line_ranges_outside_source.py
```diff
@@ -0,0 +1,41 @@
+# flags: --line-ranges=5000-6000
+# NOTE: If you need to modify this file, pay special attention to the --line-ranges=
+# flag above as it's formatting specifically these lines, in this case none.
+def foo1(parameter_1, parameter_2, parameter_3, parameter_4, parameter_5, parameter_6, parameter_7): pass
+def foo2(parameter_1, parameter_2, parameter_3, parameter_4, parameter_5, parameter_6, parameter_7): pass
+def foo3(parameter_1, parameter_2, parameter_3, parameter_4, parameter_5, parameter_6, parameter_7): pass
+def foo4(parameter_1, parameter_2, parameter_3, parameter_4, parameter_5, parameter_6, parameter_7): pass
+
+# Adding some unformated code covering a wide range of syntaxes.
```
yilei, line 9: I would simply remove the lines below, as this test case just need to verify a completely out-of-range input doesn't format

src/black/ranges.py
```diff
@@ -45,6 +45,32 @@ def is_valid_line_range(lines: Tuple[int, int]) -> bool:
...
+
+    This removes ranges that are entirely outside the valid lines.
+
+    Other ranges are normalized so that the start values are at least 1 and the
+    end values are at most the (1-based) index of the last source line.
+    """
+    if not src_contents:
+        return []
+    good_lines = []
+    src_line_count = len(src_contents.splitlines())
```
yilei, line 61: Not too strong an opinion, it can be more efficient to do a `count("\n")` but then you also need to add `1` when it doesn't end with a new line.

yilei (commented):
Thanks! Looks good to me overall, left a few minor comments.

yilei (approved):

JelleZijlstra (approved):

sumezulike:
Thanks for reviewing and approving my PR! Glad to be able to contribute to one of my all-time favorite Python libraries :)

PR #4289, opened 2024-03-22
fix: Stop moving multiline strings to a new line unless inside brackets

### LINKED ISSUE #4247: assertion with multi-line string needlessly spans multiple lines
**Describe the bug**

When an `assert` statement has a multi-line string, the condition is split across 3 lines for no reason.

**To Reproduce**

For example, take this code:

```python
assert some_var == expected_result, f"""
expected: {expected_result}
actual: {some_var}"""
```

```sh
$ black file.py
```

The resulting output is:

```py
assert (
    some_var == expected_result
), f"""
expected: {expected_result}
actual: {some_var}"""
```

**Expected behavior**

The condition should remain as a single line.

**Environment**

- Black's version: [@ed770b](https://github.com/psf/black/commit/ed770b)
- OS and Python version: Whatever https://black.vercel.app uses

RedGuy12:
### Description

Resolves #4247

I fixed it by not wrapping around multiline strings outside brackets. This keeps all existing tests passing while allowing `assert` statements to not be reformatted. This is currently gated under the `multiline_string_handling` unstable style, since it's the smallest solution to just add it in there.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
 
 - `if` guards in `case` blocks are now wrapped in parentheses when the line is too long.
   (#4269)
+- Stop moving multiline strings to a new line unless inside brackets (#4289)
 
 ### Configuration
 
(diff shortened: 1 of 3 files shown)


hauntsaninja (approved):
Thanks you!

PR #4290, opened 2024-03-22
Improve AST safety check

### LINKED ISSUE #4288: When reformatting a triple-quote string, black fails with an INTERNAL ERROR: Black produced code that is not equivalent to the source
**Describe the bug**

When reformatting a triple-quote string, black fails with an internal error.

**To Reproduce**

Take this code:

```python
if __name__ == "__main__":
    """     This string...
    ... will confuse black.
    
    """
    pass
```

```sh
$ black file.py
```

The resulting error is:

```
error: cannot format blackerrdemo.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_d030my1_.log

Oh no! 💥 💔 💥
1 file failed to reformat.
```

**Expected behavior**

Black should leave the string untouched or, if it decides to reformat it, reformat it without failure.

**Environment**

- Black's version: [24.3.0 (compiled: yes), 24.3.1.dev5+gc9d2635 (compiled: no)]
- OS and Python version: [Debian Linux 12 (Bookworm)/Python 3.11.2]

**Additional context**

Reproduced with branch main at commit c9d2635b55d03cd125465cd6d7477e6c3cb8fba3

hauntsaninja:
Fixes #4288, regressed by #4270

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Fix unwanted crashes caused by AST equivalency check (#4290)
+
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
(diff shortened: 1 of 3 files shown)


JelleZijlstra (approved):
I feel ideally we should only reformat real docstrings, but this still preserves the invariant that we only change strings that aren't executed, so this is fine.

Thanks!

Issue #4285, opened 2024-03-19
GitHub Action: Option to parse version from pyproject.toml


srittau:
**Is your feature request related to a problem? Please describe.**

When using black's GitHub action "psf/black", you need to specify the black version to use manually as an option to the action, in addition to specifying it in pyproject.toml. This can easily lead to inconsistencies between the files, especially when using dependabot to keep the version current.

**Describe the solution you'd like**

It would be useful if the action could read the version to use from pyproject.toml. This would probably need a new option like "pyproject: true" to enable this behavior.

**Describe alternatives you've considered**

An alternative is to read pyproject.toml in a previous step in a GitHub workflow for example using SebRollen/toml-action, and provide the read value using a variable, but just using a flag to this action would be much easier and less error prone.

PR #4294, opened 2024-03-30
GitHub Action: Allow reading version from pyproject.toml

srittau:
### Description

This adds an `use_pyproject` input to the GitHub Action that – when set to `true` – will read the black version from pyproject.toml file. Either from the `project.dependencies` array or one of the arrays from the `project.optional-dependencies` table.

Closes: #4285

### Checklist - did you ...

Trial runs at: https://github.com/srittau/python-asserts/actions/runs/8490172420, using this branch of one of my projects: https://github.com/srittau/python-asserts/tree/black-test

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,9 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- Add a new option `use_pyproject` to the GitHub Action `psf/black`. This will read the
+  Black version from `pyproject.toml`.
+
 ### Documentation
 
 <!-- Major changes to documentation and policies. Small docs changes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,7 +49,7 @@
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
 - Add a new option `use_pyproject` to the GitHub Action `psf/black`. This will read the
-  Black version from `pyproject.toml`.
+  Black version from `pyproject.toml`. (#4294)
 
 ### Documentation
 
--- a/action/main.py
+++ b/action/main.py
@@ -63,20 +63,33 @@ def read_version_specifier_from_pyproject() -> str:
         )
         sys.exit(1)
 
-    try:
-        deps = pyproject["project"]["dependencies"]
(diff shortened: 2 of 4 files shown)


JelleZijlstra (commented):
Can we look for `required-version` in the `tool.black` table first? https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#required-version

That seems better than attempting to parse the dependency metadata.

hugovk (commented):
Docs nits.

action/main.py
```diff
@@ -13,12 +15,101 @@
...
+def find_black_version_in_array(array: object) -> Union[str, None]:
+    if not isinstance(array, list):
+        return None
+    try:
+        for item in array:
+            # Rudimentary PEP 508 parsing.
+            item = item.split(";")[0]
+            item = EXTRAS_RE.sub("", item).strip()
+            if item == "black":
+                return ""
```
JelleZijlstra, line 101: Would it make more sense to continue here? It looks like below we ultimately end up ignoring the empty string anyway and using the action version instead.

JelleZijlstra (commented):
Thanks! Could you also add a changelog entry?

srittau:
> Thanks! Could you also add a changelog entry?

How to best do this? This doesn't really fit the existing structure of the `CHANGES` file, as this is independent of the versions of black.

action/main.py
srittau: If you trace it all the way back to `read_version_specifier_from_pyproject()`, there is a difference: Due to returning `""`, we don't generate an error in the action, but use the action's default behavior (use the latest version). If we'd return `None`, this would cause an error when using just `black` as a requirement.

CHANGES.md
```diff
@@ -48,6 +48,9 @@
 
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
+- Add a new option `use_pyproject` to the GitHub Action `psf/black`. This will read the
+  Black version from `pyproject.toml`.
```
JelleZijlstra, line 52:
```suggestion
  Black version from `pyproject.toml`. (#4294)
```

JelleZijlstra (approved):

PR #4295, opened 2024-03-31
pipx instruction added

Samuel12451:
### Added the pipx instruction to install black

I have included the detailed instruction to install pipx first if it isn't installed already then install black using pipx.

#### Checklist:
- [ ] Added an entry in CHANGES.md if necessary.
- [ ] Added/updated tests if necessary.
- [x] Added new/updated outdated documentation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # Change Log
 
 ### Added Instruction
+
 - Adding the pipx instruction to install black (#4295)
 
 ## Unreleased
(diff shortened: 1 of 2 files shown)


hauntsaninja (changes requested):
Thanks, but this is way too much detail. It's longer than the actual introduction to Black itself!

I'd be okay with a single sentence like "If you use pipx, you can install Black with `pipx install black`".

But even that probably has very limited value...
- If you know what pipx is, you know that you can do that
- If you don't know what pipx is, there isn't a reason why Black should teach you and why that should be a thing standing between you and formatted code

Issue #4263, opened 2024-03-03
Add pipx installation in documentation


lebedevdes:
Using pipx (https://pipx.pypa.io/stable/) to install Black globally can be a convenient approach for applying it across multiple projects, without the need to install it separately in each virtual environment or directly into the system's Python. Incorporating this suggestion into the installation instructions could benefit many users by providing a more versatile setup option.

IsmartGaurav:
1. To get started install pipx on your system. For macOS you can use Homebrew by typing in the command brew install pipx. On Linux the installation steps differ based on the distribution. For Ubuntu you can use sudo install pipx while for Fedora you can use sudo dnf install pipx.

2. After installing pipx make sure that the paths are properly configured by running ensurepath. You may also run sudo ensurepath  global to enable actions, with the  global option.

3. Once you have set up pipx you can globally install Black using pipx. Simply enter the command pipx install black. This will set up an environment, for Black install the package and allow you to access the command globally.

Samuel12451:
I have added the installation instruction to the README file.

PR #4301, opened 2024-04-05
pipx Black installation instruction Added #4263 Fixed

Abhinav1326:
Issue #4263 Fixed

### Description
Issue #4263 Fixed as it says you should add a pipx command to add install Black. So i just added it.


hauntsaninja:
Same feedback as https://github.com/psf/black/pull/4295#pullrequestreview-1970401280

PR #4312, opened 2024-04-17
fix: Remove parenthesis around sole list items

### LINKED ISSUE #3545: A single conditional expression inside a list is unnecessarily parenthesized
**Describe the style change**

**Examples in the current _Black_ _future_ style**

```python
items = [
    (
        {"key1": "val1", "key2": "val2", "key3": "val3"}
        if some_var == "longstring"
        else {"key": "val"}
    )
]
```

**Desired style**

```python
items = [
    {'key1': 'val1', 'key2': 'val2', 'key3': 'val3'}
    if some_var == 'longstring'
    else {'key': 'val'}
]
```

**Additional context**

This is a change from https://github.com/psf/black/pull/2278.

I think for this edge case, it shouldn't be parenthesized which also adds an extra indentation level.

Playground link: https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADgAI9dAD2IimZxl1N_Wlw1a7WWyFbt4o765uThnUap6YXDKg-y2hZjBQk1tHkIJUmRf3ubZW6qh6rwYiTz123BJex-bSSddR0SO6n6vZGd9BpWqXFwR2_6FGijjM_Dne5AzdiE1FlCq8_2Dsx5LlvLchM8KOQsFr8WMFH9R1i7Ks2Es0D6bZmm7Cl6fnGsY_a6uEkAAACpTN_YLIf7gwABqwHhAQAAKlDv4rHEZ_sCAAAAAARZWg==

cobaltt7:
Fixes #3545, but it affects more places as well.

### Description

I added the `remove_lone_list_item_parens` preview style to remove parentheses around items in lists and sets that aren't accompanied by anything else. It's important to note that it depends on the unstable `hug_parens_with_braces_and_square_brackets` style to remove parentheses around multiline list items, such as the ones reported in #3545.

### Checklist - did you ...


docs/the_black_code_style/future_style.md
```diff
@@ -124,6 +124,55 @@ foo(
 )
 ```
 
+Parenthesises around sole list items will also be removed for similar reasons. Trailing
```
JelleZijlstra, line 127:
```suggestion
Parentheses around sole list items will also be removed for similar reasons. Trailing
```

JelleZijlstra (changes requested):
Thanks, these changes look good.

However, I think I'd prefer to keep feature names stable through the year. The main motivation for adding `--enable-unstable-feature` was to allow users who want to keep some feature on to just turn on the feature in their config and have it keep working.

Also, I don't want to accept a PR that directly adds a new unstable style. I see the unstable style as a holding pen for style changes that we'd originally put into preview but later ran into issues with. New style features should go into preview instead.

RedGuy12:
Apologies for the delay @JelleZijlstra, I've had other priorities recently.

The main issue was `hug_parens_with_braces_and_square_brackets` was required for the parens to be removed around multi-line expressions. I tried to identify exactly what parts it depended on, but I don't think it would be an easy refactor. I settled on just documenting this limitation instead.

I put `remove_lone_list_item_parens` in the unstable style due to this limitation, although it does work on its own for simple cases. If desired, I could move it to the preview style so preview users could get the benefits in those cases, or we could wait for `hug_parens_with_braces_and_square_brackets`. The original issue #3545 was for the more complex cases.

cobaltt7:
Hey @JelleZijlstra just returned to this PR, sorry for letting it go stale

I removed this change from the unstable style, because it doesn't add any issues on its own, it just doesn't work in all cases without `hug_parens_with_braces_and_square_brackets`. I also resolved the conflicts.

Do these changes look good?

CHANGES.md
```diff
@@ -21,6 +21,7 @@
 
 - Fix type annotation spacing between * and more complex type variable tuple (i.e. `def
   fn(*args: *tuple[*Ts, T]) -> None: pass`) (#4440)
+- Remove parenthesis around sole list items (#4312)
```
JelleZijlstra, line 24:
```suggestion
- Remove parentheses around sole list items (#4312)
```

tests/data/cases/preview_remove_lone_list_item_parens.py
```diff
@@ -0,0 +1,156 @@
+# flags: --unstable
+items = [(123)]
```
JelleZijlstra, line 2: Add a test containing a generator expression, to make sure we don't turn it into a listcomp:

```
>>> [(x for x in [1])]
[<generator object <genexpr> at 0x104277400>]
>>> [x for x in [1]]
[1]
```

Also add tests with comments in various places (e.g., after the opening paren, before the closing paren). Such cases tend to expose bugs in Black.

JelleZijlstra (commented):
Thank you!

JelleZijlstra (approved):

JelleZijlstra:
Maybe we should also extend this to single-element sets? The code should be very similar.

cobaltt7:
@JelleZijlstra Done!

JelleZijlstra (approved):

Issue #4291, opened 2024-03-25
Empty line and `...` and "unindent does not match any outer indentation level"


Azureblade3808:
**Describe the bug**

If the first line inside an indented function is empty and the body is `...`, black fails to format it.

**To Reproduce**
https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ACQAFxdAD2IimZxl1N_WlXnON2nzNJyA7rxEz3ITq0Rr8lcyuT-7lJq6-30yBPNldub44lXXOBKYonCShOJuW7cuZ_9TCCgKPjpZqqogwanJNRK0YrZc4gHqvyNL22TtygAACOiGQzKhfHyAAF4kQEAAAC0Mv5wscRn-wIAAAAABFla

```python
class C:
    def f(self):

        ...
```

**Expected behavior**

The snippet above can be formatted.

**Environment**

- Black's version:  Stable: 24.1.0 | Main: @ed770b
- OS and Python version:  irrelevant

**Additional context**

None.

Edit
-
Function body being `...` is a key condition.

JelleZijlstra:
Confirmed:

```
% black -c '''class C:
    def f(self):

        ...
'''
class C:
    def f(self):

        ...
error: cannot format <string>: unindent does not match any outer indentation level (<tokenize>, line 5)
```

Black should instead format this as:
```
class C:
    def f(self): ...
```

harvinmumick:
Hi, my friend, @peterkra25 , and I are two students from the University of Michigan enrolled in a software engineering class (EECS 481), and our final project is to contribute to an open-source Github project. We came across Black and saw this issue, and reached out to @JelleZijlstra regarding this. We're looking into this issue and will try to resolve it over the next few weeks.

JelleZijlstra:
One more thought: I haven't looked at the code, but I suspect the changes in #3796 are closely related to this bug, so looking at the code affected by that change will be a good way to start finding the location of the bug.

hauntsaninja:
Thank you for resolving this issue!

PR #4318, opened 2024-04-22
Add support to style function definitions with newlines before function stubs

peterkra25:
### Description

Per Issue #4291, Black previously failed to format function definitions that contained an empty line before an indented function stub. This change resolves this bug by removing any newlines in the prefix of the first dot in a function stub after a stub suite is found. We added this change in the visit_default() function of linegen.py.

Through this change, functions defined as below:

```
class C:
    def f(self):

        ...
```

are formatted as:

```
class C:
    def f(self): ...
```
### Checklist - did you ...

- [Y] Add an entry in `CHANGES.md` if necessary?
- [Y] Add / update tests if necessary?
- [Y] Add new / update outdated documentation?

This change was added by Peter Kraszulyak and Harvin Mumick.


peterkra25:
We are unsure as to why the “Lint + format ourselves / build” test fails on flake8 when running pre-commit. After first cloning the repo and before making any changes, we ran

```pre-commit run -a```

and it reported that flake8 failed. This message persisted after we made our changes, so we restarted by deleting our project folder and cloning the entire repo again, but we still received the same error. If anybody knows what may be causing flake8 to fail, please let us know!

**Comment by hauntsaninja** [comment]
It looks like flake8-bugbear made a new release including a new check that has some false positives. I filed https://github.com/PyCQA/flake8-bugbear/issues/467 , but we should just pin bugbear or maybe disable that error code

**Review by hauntsaninja — COMMENTED** [review]
Thanks for the PR! It's a little scary to me to be changing `visit_default`. Can we change it in a more targeted way (e.g. specifically at the call sites in #3796 when we know we're dealing with a stub body)?

As it stands, I think this will change formatting in other situations as well, for instance, if there are real statements after `...`.

One nit: let's move the test cases into `tests/data/cases/dummy_implementations.py`?

File: tests/data/cases/stub_empty_line.py [hunk 1]
```diff
@@ -0,0 +1,7 @@
+class C:
+    def f(self):
+
```
JelleZijlstra, line 3: Maybe add test cases (in the same file) where there are comments in various places around the `...`, e.g. on a line by itself or right after `:`.

harvinmumick:
> Thanks for the PR! It's a little scary to me to be changing `visit_default`. Can we change it in a more targeted way (e.g. specifically at the call sites in #3796 when we know we're dealing with a stub body)?
> 
> As it stands, I think this will change formatting in other situations as well, for instance, if there are real statements after `...`.
> 
> One nit: let's move the test cases into `tests/data/cases/dummy_implementations.py`?

We are struggling to find a more targeted way to modify the prefix at the call site. The only call site at which we know we are dealing with a stub body in the provided test cases is here: https://github.com/psf/black/blob/40458d7568099f6ec249f2582c1cc4f6f1b4dfe0/src/black/linegen.py#L317-L320

This call site verifies a stub suite when visiting the "suite" node in the syntax tree, which is 2 levels higher than the node containing the first DOT token – the node whose prefix needs to be stripped of newlines. Between this call site and the point at which we visit the first DOT token in `visit_default`, the suite's prefix is stripped of all indentations, which enables further stripping of newline characters by the time at which the first DOT token is visited. 

Since the indents are not cleaned at the call site, we don't see a way to clean the prefix at the call site.

hauntsaninja:
Why can't we do something like this?
```diff
diff --git a/src/black/linegen.py b/src/black/linegen.py
index 4b29a04..f80140b 100644
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -308,8 +308,11 @@ class LineGenerator(Visitor[Line]):
                 yield from self.line(-1)
 
         else:
-            if not node.parent or not is_stub_suite(node.parent):
-                yield from self.line()
+            if node.parent and is_stub_suite(node.parent):
+                node.prefix = ""
+                yield from self.visit_default(node)
+                return
+            yield from self.line()
             yield from self.visit_default(node)
 
     def visit_async_stmt(self, node: Node) -> Iterator[Line]:
```
Note `Node.prefix` has a setter that will set its `children[0]`'s prefix

peterkra25:
That change does work, thanks for the tip! We didn't realize the prefix setter was recursive, and the prefix will eventually get the space re-appended between the function def and the suite. Will commit the refactored change and the added tests as recommended by Jelle.

peterkra25:
Upon further review, that change works for the code snippet reported in https://github.com/psf/black/issues/4291. However, for a test as @JelleZijlstra recommended containing a comment on the line immediately after the function stub such as below:

```
class Class:
    def f(self):

        ...
        # Comment 3
```
The newline in between the function definition and stub is not removed and the file is left unchanged. In our previous implementation, the newline is stripped for this case, although the stub is not folded onto the same line as the function definition when possible (i.e: when there are no comments on the same line as the function definition).

JelleZijlstra:
It's fine to leave that case unchanged; in fact, per our stability policy, we must maintain the current formatting for the rest of the year.

I'm asking for those additional tests mostly because comments in unusual places tend to trigger bugs in Black. I want to make sure:
- Black doesn't crash on any inputs
- Black doesn't remove any comments
- Black preserves its existing formatting for any code that is currently formatted successfully (following the stability policy)

harvinmumick:
Thank you both for your feedback and clarifications. We've relocated the implementation and added the extra tests to `tests/data/cases/dummy_implementations.py`

hauntsaninja (approved):
Thank you for the fix!

PR #4319, opened 2024-04-22
Pin to old flake8-bugbear

hauntsaninja:
See https://github.com/PyCQA/flake8-bugbear/issues/467 or https://github.com/psf/black/pull/4318

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,7 @@ repos:
     hooks:
       - id: flake8
         additional_dependencies:
-          - flake8-bugbear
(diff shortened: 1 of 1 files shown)


cooperlees:
We could have just disabled B909 ...

hauntsaninja:
That would have worked! But B909 is a useful check and I think the new behaviour is just buggy (unless there's something I don't understand). Also pinning here seems fine... `main` breaking was confusing for the folks in #4318 and we pin several other deps in pre-commit / have the autoupdate thing.

But obviously will do whatever you suggest here, just let me know :-)

PR #4325, opened 2024-04-23
Prevent wrapping of multiline fstrings in parens

### LINKED ISSUE #4324: PEP 701 support breaks stability policy
https://github.com/psf/black/pull/3822 is amazing work. I've been sort of behind on all open source stuff, so I only just got to running it on my codebase at work. I found some reformatting, which violates the stability policy.

```
def foo():
    class aaa(bbbb):
        value: str = f"""
        fdsa
        asdf
        ```section
        {value}
        ```
        """
```
There were eleven instances of things that got changed

tusharsadhwani:
Resolves #4324

### Description

The function `is_multiline_string` was only being run for `Leaf` nodes, but in that context, multiline f-strings should also have been respected. The code has been adjusted to respect f-strings as well.

The `STRING` leaves created by `fstring_to_string` were also missing `lineno` fields which are used in parts of the codebase.


tusharsadhwani:
The one other edge case in the issue is also likely around this function:

https://github.com/psf/black/blob/551ede2825e4a92e9880b90ef517cce333a5cbfc/src/black/lines.py#L276

I'm seeing if that's a small fix that can go in the same PR.

tusharsadhwani:
^ So that other edge case is not related to #3822, as I'm able to reproduce it even on black v24.4.0:

```console
$ black --version                 
black, 24.4.0 (compiled: yes)
Python (CPython) 3.12.2
$ cat asd.py
log(
    f"Received operation {server_operation.name} from "
    f"{self.writer._transport.get_extra_info('peername')}",  # type: ignore[attr-defined]
)
$ black asd.py
reformatted asd.py

All done! ✨ 🍰 ✨
1 file reformatted.
$ cat asd.py 
log(
    f"Received operation {server_operation.name} from {self.writer._transport.get_extra_info('peername')}",  # type: ignore[attr-defined]
)
```
cc @hauntsaninja, does that seem right?

hauntsaninja:
No, that doesn't seem right.

Make sure you're not getting fooled by `unstable = true` (which you'll pick up by default if the file is in black project dir).

```
(black-drsy) ~/dev/black 551ede2 λ cat /tmp/b.py 
def a():
    def b():
        def c():
            def d():
                log(
                    f"Received operation {server_operation.name} from "
                    f"{self.writer._transport.get_extra_info('peername')}",  # type: ignore[attr-defined]
                    level=0,
                )
(black-drsy) ~/dev/black 551ede2 λ python -m black /tmp/b.py -l 100 --check --diff
--- /tmp/b.py	2024-04-23 20:23:56.523454+00:00
+++ /tmp/b.py	2024-04-23 20:24:23.464522+00:00
@@ -1,9 +1,8 @@
 def a():
     def b():
         def c():
             def d():
                 log(
-                    f"Received operation {server_operation.name} from "
-                    f"{self.writer._transport.get_extra_info('peername')}",  # type: ignore[attr-defined]
+                    f"Received operation {server_operation.name} from " f"{self.writer._transport.get_extra_info('peername')}",  # type: ignore[attr-defined]
                     level=0,
                 )
would reformat /tmp/b.py

Oh no! 💥 💔 💥
1 file would be reformatted.
(black-drsy) ~/dev/black 551ede2 λ gco HEAD~
Previous HEAD position was 551ede2 Add PEP 701 support (#3822)
HEAD is now at 944b99a Bump sphinx from 7.2.6 to 7.3.7 in /docs (#4322)
(black-drsy) ~/dev/black 944b99a λ python -m black /tmp/b.py -l 100 --check --diff
All done! ✨ 🍰 ✨
(black-drsy) ~/dev/black 944b99a λ 
```

hauntsaninja:
And sorry for not fully minimising it, e.g. this repros without `-l 100` for me:
```
log(
    f"Received operation {server_operation.name} from "
    f"{self.writer._transport.get_extra_info('peername')}",  # type: ignore[attr-defined]
    level=0,
)
```

tusharsadhwani:
Thanks for clarification, both cases should be fixed now.

hauntsaninja (approved):
This looks great. Thank you for the fixes (and thank you again for the original feature)!

hauntsaninja:
(also confirmed this commit is clean on my current work codebase)

JelleZijlstra (approved):
Thanks!

PR #4327, opened 2024-04-24
Support PEP 696

JelleZijlstra:


hauntsaninja (commented):
I pushed some more test cases. These may be preexisting, but:
- I think I'd prefer if `simple` / `longer` split the arguments instead of the generic part.
- The formatting in `trailing_comma2` looks buggy to me

hauntsaninja (approved):
Looks like we have an issue for those too :-)

JelleZijlstra:
Yes, #4071 for reference.

PR #4338, opened 2024-04-26
really drop python<=3.7 support

kloczek:
### Description
Filter all code (except tests/cases) over `pyupgrade --py38-plus`.


ichard26:
The `profiling` directory should be left alone as well.

tests/test_schema.py
```diff
@@ -5,7 +5,7 @@
 def test_schema_entrypoint() -> None:
     if sys.version_info < (3, 10):
         eps = importlib.metadata.entry_points()["validate_pyproject.tool_schema"]
-        (black_ep,) = [ep for ep in eps if ep.name == "black"]
+        (black_ep,) = (ep for ep in eps if ep.name == "black")
```
JelleZijlstra, line 8: I know pyupgrade does this but I don't like this change. It feels harder to understand (generator expressions are rarer than listcomps). The genexp might sometimes be faster, but it's unlikely to matter.

In this case, the new code appears to be slower for me (benchmark on Linux on 3.9):

```
In [8]: eps = importlib.metadata.entry_points()["validate_pyproject.tool_schema"]

In [9]: eps
Out[9]: (EntryPoint(name='black', value='black.schema:get_schema', group='validate_pyproject.tool_schema'),)

In [10]: %timeit (black_ep,) = [ep for ep in eps if ep.name == "black"]
170 ns ± 1.32 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [11]: %timeit (black_ep,) = (ep for ep in eps if ep.name == "black")
232 ns ± 1.61 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

scripts/migrate-black.py
```diff
@@ -51,7 +51,7 @@ def blackify(base_branch: str, black_command: str, logger: logging.Logger) -> in
                 "diff",
                 "--binary",
                 "--find-copies",
-                "%s-black..%s-black" % (last_commit, commit),
+                "{}-black..{}-black".format(last_commit, commit),
```
AlexWaygood, line 54:
```suggestion
                f"{last_commit}-black..{commit}-black",
```

kloczek: If you have any complains about how pyupgarde works please discuss that with pypgrade maintainer(s).
IMO cherry-picking "what I like/what I don't" does not make to much sense.

It is great value that it is GoodEnough™️ tool allowing to clean some code cleanups between major python versions.

You know that `black` code can be filtered to cut off some legacy tails.
If you don't like this PR just close it .. really no one will be hurt.

FYI: in only few weeks 3.8 will be EOSed and with that version can be dropped much more.

AlexWaygood: Hi, sorry to hear you're not a fan of my suggestion!

> IMO cherry-picking "what I like/what I don't" does not make to much sense.

Not sure I agree there -- in my opinion, any proposed change should be considered on its own merits, regardless of whether the change has come about as a result of an autofix or codemod. But I'm not a Black maintainer, so the decision on how best to review this isn't up to me 😊

> If you don't like this PR just close it .. really no one will be hurt.

Well, again, I'm not a maintainer, so I don't personally in fact have this power

AlexWaygood: > FYI: in only few weeks 3.8 will be EOSed

Python 3.8 is scheduled to go End Of Life in October 2024, FWIW, so we have at least six months to go yet

AlexWaygood (approved):

PR #4343, opened 2024-04-30
tokenizer: skip lines that are just slash and whitespace

### LINKED ISSUE #4261: Fails to parse backslash on line by itself
```
% cat parse.py 
class Plotter:
\
    pass
% black parse.py 
error: cannot format parse.py: Cannot parse: 3:4:     pass

Oh no! 💥 💔 💥
1 file failed to reformat.
% black --version
black, 24.2.0 (compiled: yes)
Python (CPython) 3.12.0
% python parse.py
% 
```

Saw this in astral-sh/ruff#10099; the example is minified from the repro case in that issue.

tusharsadhwani:
### Description

Resolves #4261 

### Checklist - did you ...


tusharsadhwani:
Ah. it affects multiline strings. Should be an easy fix...

tests/data/cases/comment_after_escaped_newline.py
```diff
@@ -14,5 +14,7 @@ def bob():  # pylint: disable=W9016
     pass
 
 
-def bobtwo():  # some comment here
+def bobtwo():
```
JelleZijlstra, line 17: hm so the behavior changes here? That's technically against the stability policy.

tusharsadhwani: That is true, but it is a very weird edge case (you need spaces a bit less than the next indent level, followed by a slash) and primer shows zero cases.

Should I try and preserve behaviour nonetheless?

JelleZijlstra: Actually, this falls under the exception where we can change the formatting of code that hasn't already been Black-formatted. Normally I'm hesitant to use that exception, but this pattern seems unlikely to appear in any real-world code.

src/blib2to3/pgen2/tokenize.py
```diff
@@ -608,6 +608,12 @@ def generate_tokens(
         except StopIteration:
             line = ""
         lnum += 1
+
+        # skip lines that are just a slash, to avoid storing that line's
+        # indent information.
+        if not contstr and line.rstrip("\n").strip(" \t") == "\\":
```
JelleZijlstra, line 614: Is this equivalent to just `line.strip() == "\\"`? Or do we need to care about exotic whitespace characters that are not newline/space/tab?

tusharsadhwani: I'm not entirely sure, so I tried to be conservative. Does Python normally treat characters like `\r` or `\v` as part of indentation?

JelleZijlstra: Not too sure, there is code in our tokenizer that deals with `\f` at least. In CPython (`Parser/lexer/lexer.c`) I see some code dealing with `\r` and with `\f` (`\014`).

tusharsadhwani:
```console
$ printf 'def foo():\n \t pass' | python -m tokenize
1,0-1,3:            NAME           'def'
1,4-1,7:            NAME           'foo'
1,7-1,8:            OP             '('
1,8-1,9:            OP             ')'
1,9-1,10:           OP             ':'
1,10-1,11:          NEWLINE        '\n'
2,0-2,3:            INDENT         ' \t '
2,3-2,7:            NAME           'pass'
2,7-2,8:            NEWLINE        ''
3,0-3,0:            DEDENT         ''
3,0-3,0:            ENDMARKER      ''

$ printf 'def foo():\n \f pass' | python -m tokenize
1,0-1,3:            NAME           'def'
1,4-1,7:            NAME           'foo'
1,7-1,8:            OP             '('
1,8-1,9:            OP             ')'
1,9-1,10:           OP             ':'
1,10-1,11:          NEWLINE        '\n'
2,0-2,3:            INDENT         ' \x0c '
2,3-2,7:            NAME           'pass'
2,7-2,8:            NEWLINE        ''
3,0-3,0:            DEDENT         ''
3,0-3,0:            ENDMARKER      ''

$ printf 'def foo():\n \r pass' | python -m tokenize
1,0-1,3:            NAME           'def'
1,4-1,7:            NAME           'foo'
1,7-1,8:            OP             '('
1,8-1,9:            OP             ')'
1,9-1,10:           OP             ':'
1,10-1,11:          NEWLINE        '\n'
2,1-2,3:            OP             '\r '
2,3-2,7:            NAME           'pass'
2,7-2,8:            NEWLINE        ''
3,0-3,0:            ENDMARKER      ''
```

So `\v` is infact legitimate indentation, how interesting.

tusharsadhwani:
```console
$ python -c 'print("def foo():\n \v pass")' | python -m tokenize
1,0-1,3:            NAME           'def'
1,4-1,7:            NAME           'foo'
1,7-1,8:            OP             '('
1,8-1,9:            OP             ')'
1,9-1,10:           OP             ':'
1,10-1,11:          NEWLINE        '\n'
2,0-2,1:            INDENT         ' '
<stdin>:2:2: error: invalid non-printable character U+000B
```

`\v` is unparseable.

So editing the PR to do `.lstrip(' \t\f')` should take care of all cases I believe.

tusharsadhwani: Made changes. The reason I can't do `.strip()` is because `\` must be at the end of the line. If there's spaces after the backslash it's no longer escaping the newline:

```console
$ printf 'print(2 + \\\n3)' | python3
5

$ printf 'print(2 + \\ \n3)' | python3
  File "<stdin>", line 1
    print(2 + \
               ^
SyntaxError: unexpected character after line continuation character
```

tusharsadhwani: Well a `form_feeds` test broke. Can you see if that affects us?

tusharsadhwani: The file contains a `\f\` line that is getting preserved, while the line was being deleted entirely before. Which I think is fine.

tests/data/cases/form_feeds.py
```diff
@@ -156,6 +156,7 @@ def something(self):
 
 #
 
+
```
tusharsadhwani, line 159: this is just a `\n`

JelleZijlstra (approved):

PR #4346, opened 2024-05-04
Convert PR numbers in docs/change_log to clickable links

### LINKED ISSUE #4258: Create URL links to actual PRs rather than simply referencing the PR number
The change log is immensely useful, however it makes reference to specific Github PRs and only mentions the PR number without actually linking to the issue.

I would like for the changelog to actually include links to the issue. It would be of great help to readers to be able to see the contextual discussion for a change, particularly for style changes.

https://black.readthedocs.io/en/stable/change_log.html

sumezulike:
### Description
Closes #4258.

This adds a listener to the sphinx `include-read` event to find all PR numbers `(#X)` with regex and replace them with the corresponding link `[(#X)](https://github.com/psf/black/pull/X)`.

I didn't add an entry to `CHANGES.md` since this seems like a rather minor docs change.

### Checklist - did you ...

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -38,19 +38,19 @@ def replace_pr_numbers_with_links(content: str) -> str:
     base_url = "https://github.com/psf/black/pull/"
     pr_num_regex = re.compile(r"\(#(\d+)\)")
 
-    def num_to_link(match: re.Match) -> str:
-        full_match, number = match.group(0, 1)
+    def num_to_link(match: re.Match[str]) -> str:
+        number = match.group(1)
         url = f"{base_url}{number}"
-        return f"[{full_match}]({url})"
+        return f"([#{number}]({url}))"
 
     return pr_num_regex.sub(num_to_link, content)
 
 
 def handle_include_read(
-        app: Sphinx,
-        relative_path: Path,
-        parent_docname: str,
-        content: list[str],
+    app: Sphinx,
+    relative_path: Path,
+    parent_docname: str,
+    content: list[str],
 ) -> None:
     """Handler for the include-read sphinx event."""
     if parent_docname == "change_log":
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,9 +13,12 @@
 #
 
 import os
+import re
 import string
 from importlib.metadata import version
 from pathlib import Path
+
+from sphinx.application import Sphinx
 
 CURRENT_DIR = Path(__file__).parent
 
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -29,6 +32,36 @@ def make_pypi_svg(version: str) -> None:
...
+    def num_to_link(match: re.Match) -> str:
+        full_match, number = match.group(0, 1)
+        url = f"{base_url}{number}"
+        return f"[{full_match}]({url})"
+
+    return pr_num_regex.sub(num_to_link, content)
+
+
+def handle_include_read(
+        app: Sphinx,
```
JelleZijlstra, line 50: I guess we're not formatting this file with Black? Let's fix that in a separate PR though.

docs/conf.py
```diff
@@ -29,6 +32,36 @@ def make_pypi_svg(version: str) -> None:
...
+def replace_pr_numbers_with_links(content: str) -> str:
+    """Replaces all PR numbers with the corresponding GitHub link."""
+
+    base_url = "https://github.com/psf/black/pull/"
+    pr_num_regex = re.compile(r"\(#(\d+)\)")
+
+    def num_to_link(match: re.Match) -> str:
+        full_match, number = match.group(0, 1)
+        url = f"{base_url}{number}"
+        return f"[{full_match}]({url})"
```
JelleZijlstra, line 44:
```suggestion
        return f"([#{number}]({url}))"
```
Let's not link the parentheses.

docs/conf.py
```diff
@@ -29,6 +32,36 @@ def make_pypi_svg(version: str) -> None:
         f.write(svg)
 
 
+def replace_pr_numbers_with_links(content: str) -> str:
+    """Replaces all PR numbers with the corresponding GitHub link."""
+
+    base_url = "https://github.com/psf/black/pull/"
+    pr_num_regex = re.compile(r"\(#(\d+)\)")
+
+    def num_to_link(match: re.Match) -> str:
```
JelleZijlstra, line 41:
```suggestion
    def num_to_link(match: re.Match[str]) -> str:
```

JelleZijlstra (commented):
Thanks!

JelleZijlstra:
The preview is at https://black--4346.org.readthedocs.build/en/4346/change_log.html, looks great!

RedGuy12:
Looks very clean! I skimmed the changelog and I did find a few inconsistencies that weren't always linked properly:
```
Add extra blank lines in stubs in a few cases (#3564, #3862)
Upgrade to mypy 1.7.1 (#4049) (#4069)
Revert the change ... (#3940) for better compatibility ...  (#4137)
... This fixes issue #1629 and all of its many many duplicates. (#2126)
added `--extend-exclude` argument (PR #2005)
re-implemented support for explicit trailing commas ... (#1288 and duplicates)
```
It may be easiest to linkify all mentions, even if outside parenthesis.

JelleZijlstra (approved):

JelleZijlstra:
Oops sorry, I merged just as @RedGuy12's comment appeared. Agree it makes sense to linkify all instances of `#NNN` even if they are not immediately within parentheses.

sumezulike:
Good point, I overlooked that there were inconsistencies, didn't scroll that far :smile: 
I initially decided to make it strict because the docs dictate the format like
```md
- `Black` is now more awesome (#X)
```
Also, some people mention issues in the descriptions which would then cause invalid links.
I can make the matching more lenient though, maybe all #numbers inside parentheses?
So `Fixes stuff (#123 and #456)` would match both but `Fixes #123 (#456)` only matches 456?

JelleZijlstra:
Issue links are actually fine, GitHub redirects them. I put up an implementation in #4347.

sumezulike:
Alright, thanks for the review!

PR #4365, opened 2024-05-16
Fix formatting of backslash escaped quote inside f-string

### LINKED ISSUE #4350: "EOF in multi-line string" on string containing same quote nested in f-string
**Describe the bug**

`black` cannot format this type of code: `f"{'\''}"`

```
$ black foo.py
error: cannot format foo.py: Cannot parse: 1:3: EOF in multi-line string

Oh no! 💥 💔 💥
1 file failed to reformat.
```

I was thinking that it might be the same bug as https://github.com/psf/black/issues/4334, but the example in that issue (`f'{'a'}'`) doesn't produce an error in black 24.4.2 anymore, and Python 3.12.2 accepts the above code:

```python
import ast
from pathlib import Path

print(f"{'a'}")
print(ast.parse(Path(__file__).read_text()))
```

```
$ python3 example.py 
a
<ast.Module object at 0x1014a9410>
```

**To Reproduce**

[Playground](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ABzAEpdAD2IimZxl1N_WlkB95OWvLc9zhiUiMvzcK31n0pZiP0hl421SAnZjfQPZ6wnLpoyVWEx6zvZljj6uyel4KXqoGdHtewh35anCp7gAAAAaYh7eWTcxyQAAWZ0L8xJRB-2830BAAAAAARZWg==)

**Expected behavior**

I have no opinion on how this should be formatted. Leaving this code as it is would be fine.

**Environment**

```
$ black --version
black, 24.4.2 (compiled: yes)
Python (CPython) 3.12.2
```

macOS 12.7.4 (21H1123)

hauntsaninja:
Fixes #4350

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -510,7 +510,18 @@ class LineGenerator(Visitor[Line]):
         # currently we don't want to format and split f-strings at all.
         string_leaf = fstring_to_string(node)
         node.replace(string_leaf)
-        yield from self.visit_STRING(string_leaf)
+        if not (
+            "\\" in string_leaf.value
+            and any(
+                "\\" not in str(child)
+                for child in node.children
+                if node.type == syms.fstring_replacement_field
+            )
+        ):
+            yield from self.visit_STRING(string_leaf)
+            return
+        yield from self.visit_default(string_leaf)
+        return
 
         # TODO: Uncomment Implementation to format f-string children
         # fstring_start = node.children[0]


src/black/linegen.py
```diff
@@ -510,7 +510,18 @@ def visit_fstring(self, node: Node) -> Iterator[Line]:
         # currently we don't want to format and split f-strings at all.
         string_leaf = fstring_to_string(node)
         node.replace(string_leaf)
-        yield from self.visit_STRING(string_leaf)
+        if not (
+            "\\" in string_leaf.value
+            and any(
+                "\\" not in str(child)
+                for child in node.children
+                if node.type == syms.fstring_replacement_field
```
tushar-deepsource, line 518:
```suggestion
                if child.type == syms.fstring_replacement_field
```

tusharsadhwani, line 518:
```suggestion
                if child.type == syms.fstring_replacement_field
```

tusharsadhwani:
Personally I'd rewrite it like so:

```py
if (
    "\\" in string_leaf.value
    and any(
        "\\" in str(child)
        for child in node.children
        if child.type == syms.fstring_replacement_field
    )
):
    yield from self.visit_default(string_leaf)
    return
yield from self.visit_STRING(string_leaf)
```

PR #4378, opened 2024-06-07
Added Python and grammar versions to a parser error message

smb55:
### Description

Improved the parsing error message to make the user aware of the Python version/s and associated grammar version/s that generated the error. This fixes [#3294](https://github.com/psf/black/issues/3294)

### Checklist - did you ...


hauntsaninja (approved):
Thanks! I simplified the error message to explicitly mention target version, since it points users directly to the setting affecting them (users don't need to know details about grammar sets)

JelleZijlstra (approved):

PR #4394, opened 2024-07-04
Fix wheel build

hauntsaninja:


JelleZijlstra (approved):

hauntsaninja:
Looks like it worked, and through the magic of auditwheel we don't even change the set of platforms users can install on

PR #4401, opened 2024-07-10
fix: don't try to format strings inside f-strings

### LINKED ISSUE #4350: "EOF in multi-line string" on string containing same quote nested in f-string

tusharsadhwani:
Resolves #4350 , closes #4365

If a string like `'\''` is inside a double-quoted f-string, like `f"{'\''}"` the formatter would break trying to remove the escaped `\'` quote.

This is because `normalize_string_quotes` looks at the entire `f"{'\''}"` input, and the outermost quotes `"`, hence determining that the `\'` escape is unnecessary.

To avoid these cases, we don't try to normalize inner strings inside f-string braces.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,9 @@
 - Fix regression where Black failed to parse a multiline f-string containing another
   multiline string (#4339)
 
+- Fix regression where Black failed to parse an escaped single quote inside an f-string
+  (#4401)
+
 - Fix bug with Black incorrectly parsing empty lines with a backslash (#4343)
 
 ### Performance
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -510,13 +510,10 @@ class LineGenerator(Visitor[Line]):
         # currently we don't want to format and split f-strings at all.
         string_leaf = fstring_to_string(node)
         node.replace(string_leaf)
-        if (
-            "\\" in string_leaf.value
-            and any(
-                "\\" in str(child)
-                for child in node.children
-                if child.type == syms.fstring_replacement_field
-            )
+        if "\\" in string_leaf.value and any(
+            "\\" in str(child)
+            for child in node.children
+            if child.type == syms.fstring_replacement_field
         ):
             # string normalization doesn't account for nested quotes,
             # causing breakages. skip normalization when nested quotes exist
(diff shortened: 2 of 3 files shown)


hauntsaninja (approved):
Thanks, looks like an actually correct version of https://github.com/psf/black/pull/4365 :-)

This has an amusing interaction with the stability policy, but better to have not the final formatting we want than have a crash

saksham-sak:
The formatter should not attempt to normalize inner strings (the content inside the {} braces of f-strings). It should only normalize the outermost string.
This ensures the inner content is preserved as written, avoiding any incorrect simplifications.

tusharsadhwani:
@saksham-sak that's what we do right now, but it's not ideal in practice since Python 3.12 and above.

Since the expression inside f-string braces can be any valid Python expression now, it should be formatted just like any other Python expression outside the f-strings, and it won't break any code.

The only exception is if the braces end with a `=` token, in which case it should not be tampered with.

PR #4402, opened 2024-07-13
Fix Type hint

ZhipengXue97:
## Description
Fix type check warning reported by Pyre@Google, which was outdated after code modification.

## Detail
update the return type of `test` from `None` to `str`, since the function return a string after commit https://github.com/psf/black/commit/87b8df28c48353d3d08d7e88d178c7a567de816a


JelleZijlstra:
The test cases are meant purely for formatting; they are not meaningful Python code and should not be type checked.

Issue #4405, opened 2024-07-18
black could more efficiently skip directories listed in .gitignore


rmcloughlin:
Say I have this file structure:

```
my-project
├── large_ignored_dir
│   ├── a.py
│   ├── inner
│   │   └── b.py
│   ├── inner2
│   │   └── c.py
│   └── inner3
│       └── d.py
└── z.py
```
The only file I want black to pay attention to is `z.py`.

Then I run `black --exclude='large_ignored_dir/' --check -v .` which shows me:
```
Identified `/` as project root containing a file system root.
Found input source directory: "[...]/my-project"
[...]/my-project/large_ignored_dir ignored: matches the --exclude regular expression
would reformat [...]/my-project/z.py

Oh no! 💥 💔 💥
1 file would be reformatted.

```
That seems to be the correct behavior.

But black is supposed to automatically ignore directories listed in `.gitignore` ([docs](https://black.readthedocs.io/en/stable/usage_and_configuration/file_collection_and_discovery.html#gitignore)).  So if I add a `.gitignore` file containing this line:
```
large_ignored_dir/
```
Then I should be able to run `black --check -v .` and see the same result as above.  But instead I see:
```
Identified `/` as project root containing a file system root.
Found input source directory: "[...]/my-project"
[...]/my-project/large_ignored_dir/inner ignored: matches a .gitignore file content
[...]/my-project/large_ignored_dir/a.py ignored: matches a .gitignore file content
[...]/my-project/large_ignored_dir/inner2 ignored: matches a .gitignore file content
[...]/my-project/large_ignored_dir/inner3 ignored: matches a .gitignore file content
would reformat [...]/my-project/z.py

Oh no! 💥 💔 💥
1 file would be reformatted.

```
Even though the final result is the same (only `z.py` is reformatted), in this second case black is wasting time checking files that it could have skipped.  Instead of ignoring all of `large_ignored_dir` it is descending one level inside it and then deciding to skip every file and subdirectory.

This has performance consequences with directories like `node_modules` and `__pycache__`.

**Environment**

- Black's version: 24.4.2
- OS and Python version: macOS Sonoma 14.5, M1 chip, Python 3.12

JelleZijlstra:
PR welcome to improve this.

devdanzin:
Interestingly, if the content of `.gitignore` is set to `large_ignored_dir` without a trailing slash, it works as expected:
```
Found input source directory: "[...]\my-project"
[...]

PR #4415, opened 2024-07-25
fix: Enhace black efficiently to  skip directories listed in .gitignore

Mr-Sunglasses:
### Description

fix: #4405  Enhace black efficiently to  skip directories listed in .gitignore

### Checklist - did you ...

### DIFF
--- a/src/black/files.py
+++ b/src/black/files.py
@@ -309,6 +309,8 @@ def _path_is_ignored(
     for gitignore_path, pattern in gitignore_dict.items():
         try:
             relative_path = path.relative_to(gitignore_path).as_posix()
+            if path.is_dir():
+                relative_path = relative_path + "/"
         except ValueError:
             break
         if pattern.match_file(relative_path):
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,8 @@
 
 <!-- Changes that improve Black's performance. -->
 
+- Improve performance when a large directory is listed in `.gitignore` (#4415)
+
 ### Output
 
 <!-- Changes to Black's terminal output and error messages -->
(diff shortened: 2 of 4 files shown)


src/black/files.py
```diff
@@ -309,6 +309,8 @@ def _path_is_ignored(
     for gitignore_path, pattern in gitignore_dict.items():
         try:
             relative_path = path.relative_to(gitignore_path).as_posix()
+            if path.is_dir():
```
JelleZijlstra, line 312: Any way we could test that this works as expected?

Mr-Sunglasses: I've add the test for it

Mr-Sunglasses:
@JelleZijlstra I've added the test for this and now black is almost 2x faster when it have to ignore directories from `.gitignore`

A benchmark I've tested, with `node_modules` as ignored.

Screenshot 2024-07-27 at 4 13 07 AM

Mr-Sunglasses:
> @JelleZijlstra I've added the test for this and now black is almost 2x faster when it have to ignore directories from `.gitignore`
> 
> A benchmark I've tested, with `node_modules` as ignored.
> 
> Screenshot 2024-07-27 at 4 13 07 AM

A question, should we add a changelog entry as it's a simple but significant change?

Mr-Sunglasses:
Should we add a changelog entry as it's a simple but significant change?

JelleZijlstra (approved):
Thank you! I pushed a news entry.

hauntsaninja:
Note this probably hurts perf in the normal case due to adding a bunch of new stats
This is more correct, but could be nice to rework things to use os.scandir and away from pathlib.Path

PR #4422, opened 2024-07-31
fix: respect braces better in f-string parsing

### LINKED ISSUE #4421: Crash on f-string with `\{`
Black fails to format a valid Python f-string containing `\{`. I minimized this from a file in Pygments, `pygments/lexers/int_fiction.py`.
```
% cat rf.py 
rf'{a}\{{[\}}'
% black --check rf.py
error: cannot format rf.py: Cannot parse: 1:10: rf'{a}\{{[\}}'

Oh no! 💥 💔 💥
% python rf.py 
  File "/Users/jelle/py/black/rf.py", line 1, in <module>
    rf'{a}\{{[\}}'
        ^
NameError: name 'a' is not defined
% python -V
Python 3.12.4
% black --version
black, 24.4.3.dev27+g7fa1faf (compiled: no)
Python (CPython) 3.12.4
```

cc @tusharsadhwani for f-strings.

tusharsadhwani:
### Description

Tiny inconsistency between two regexes, one had a part changed to `\\[^{]` whereas the other was left to be `\\.`, causing a crash in parsing in some cases.

The regex was a bit unreasonably complex though, it doesn't need to have `\\.` or `\\[^{]` in it anyway. This simplifies the regex and makes it more compatible.

Resolves #4421


JelleZijlstra:
Thanks! Could you add a changelog entry since this fixes crashes on some real-world code?

Also might be worth adding a few more test cases; I'll take a look tonight and maybe push some too.

tusharsadhwani:
This fixes the two crashing cases, but doesn't fix the incorrect parses.

Still looking into that one.

tusharsadhwani:
`rf"\{"a"}"` is not fixed yet, `rf"\{2}"` still parses wrong.

JelleZijlstra:
I have a fix for the remaining cases locally, will put up a PR once #4423 is done.

tusharsadhwani:
@JelleZijlstra I just made a fix, see if you can find any bugs with this one?

JelleZijlstra (approved):

PR #4423, opened 2024-08-01
Fix incorrect linenos on fstring tokens with escaped newlines

JelleZijlstra:
I don't think this can affect Black itself much (maybe for formatting
ranges), but I ran into this with https://github.com/JelleZijlstra/lib2toast

cc @tusharsadhwani

### DIFF
--- a/tests/test_tokenize.py
+++ b/tests/test_tokenize.py
@@ -60,7 +60,7 @@ def test_fstring() -> None:
         'f"x"',
         [
             Token("FSTRING_START", 'f"', (1, 0), (1, 2)),
-            Token("FSTRING_MIDDLE", 'x', (1, 2), (1, 3)),
+            Token("FSTRING_MIDDLE", "x", (1, 2), (1, 3)),
             Token("FSTRING_END", '"', (1, 3), (1, 4)),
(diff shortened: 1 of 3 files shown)


tests/test_tokenize.py
```diff
@@ -0,0 +1,120 @@
...
+            Token(type="ENDMARKER", string="", start=(3, 0), end=(3, 0)),
+        ],
+    )
+
+
+# Run "echo some code | python tests/test_tokenize.py" to generate test cases.
+if __name__ == "__main__":
+    code = sys.stdin.read()
+    tokens = get_tokens(code)
+    text = f"assert_tokenizes({code!r}, {tokens!r})"
```
tusharsadhwani, line 118: consider using the builtin `tokenize` module for generating tests? that way you don't have to verify the locations before adding the test.

tusharsadhwani: I'm not sure how compatible exactly are the tokens between Python and blib2to3 though.

tusharsadhwani: Yeah, for now this is good

tusharsadhwani (approved):

PR #4436, opened 2024-08-14
CI: Add Python 3.13 jobs to test matrices

EwoutH:
Enable Python 3.13 testing in the GitHub Action workflows by adding 3.13 jobs to the test matrices.

Python 3.13 is now in [release candidate](https://www.python.org/downloads/release/python-3130rc1/) phase and ABI stable.


JelleZijlstra (approved):
Thanks!

EwoutH:
Does an internal tooling change like this need a changelog entry?

JelleZijlstra:
I think it does; it should say that we now officially support Python 3.13. Could you add that and also add 3.13 to the trove classifiers in `pyproject.toml`?

EwoutH:
Something like this?

EwoutH:
https://github.com/psf/black/pull/4427 was needed for full Python 3.13 support, so I included that in this PR.

EwoutH:
I think I now have everything for full Python 3.13 support and testing. If I have missed anything, let me know!

JelleZijlstra:
Looks like the mypyc wheel build fails on 3.13. There are some open issues on the mypy side for adding 3.13 support. Possibly we could release only a pure-Python wheel on 3.13 for now.

hauntsaninja:
We already ship a pure Python wheel, so 3.13 users will just pick it up

EwoutH:
- mypy tracking issue: https://github.com/python/mypy/issues/17264
- mypyc tracking issue: https://github.com/mypyc/mypyc/issues/1056

Looks like it might be a while, shall I update this PR to not build the mypyc 3.13 wheel yet, and that we merge that in a separate PR?

JelleZijlstra:
Yes, that sounds good.

EwoutH:
Updated!

EwoutH:
Is there anything I still need to do?

PR #4440, opened 2024-08-22
fix: fix PEP 646 support of tuple unpacking

terencehonles:
### Description

This change fixes unpacking a tuple or generic type when *args is a type variable tuple.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Fix type annotation spacing between * and more complex type variable tuple (i.e. `def
+  fn(*args: *tuple[*Ts, T]) -> None: pass`) (#4440)
+
 ### Configuration
 
 <!-- Changes to how Black can be configured -->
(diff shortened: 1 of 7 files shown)


JelleZijlstra (changes requested):
Thanks, this is a good catch. However, as it is a change in style, it should be applied only to Black's preview style.

terencehonles:
> Thanks, this is a good catch. However, as it is a change in style, it should be applied only to Black's preview style.

Guidance on that probably would have been useful, but I believe I may have figured it out

JelleZijlstra (approved):
Thanks!

CHANGES.md
```diff
@@ -14,6 +14,9 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Fix type annotation spacing between * and more complex type variable tuple (i.e. `def
+  fn(*args: *tuple[*Ts, T]) -> None: pass`)
```
```suggestion
  fn(*args: *tuple[*Ts, T]) -> None: pass`) (#4440)
```

JelleZijlstra:
There's a CI failure that looks unrelated to your change, probably due to a typing change in a dependency. I can look into it over the next few days.

CHANGES.md
terencehonles: done

terencehonles:
> There's a CI failure that looks unrelated to your change, probably due to a typing change in a dependency. I can look into it over the next few days.

I fixed the last lint error by updating the schema

JelleZijlstra:
Thank you!

PR #4444, opened 2024-09-03
Fix 4399 `ASTSafetyError`  on comment in return type annotation

### LINKED ISSUE #4399: Black fails with `ASTSafetyError` on comment line in return type annotation
Hello, I recently hit a bug while working on adding type error suppressions to a codebase I'm working on.

**Describe the bug**

Black appears to add a `Tuple` to the AST when a comment is present in a multi-line return type annotation. This results in an `ASTSafetyError` and fails to lint the file.

**To Reproduce**

Run the following example file in [the online editor](https://black.vercel.app/?version=main)
```python
import typing as t

from my_library import Abstract
from my_other_library import api

_T = t.TypeVar("_T")

class MyClass(Generic[_T]):
    ...

def create_my_nested_class() -> (
    # some linter directive that need to be here
    MyClass
):
    class MyNestedClass(MyClass):
        pass
    return MyNestedClass
```

The resulting error is:
```
INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful:

--- src
+++ dst
@@ -155,17 +155,23 @@
         )  # /Return
         decorator_list=
         name=
         'create_my_nested_class',  # str
         returns=
-        Name(
+        Tuple(
             ctx=
             Load(
             )  # /Load
-            id=
-            'MyClass',  # str
-        )  # /Name
+            elts=
+            Name(
+                ctx=
+                Load(
+                )  # /Load
+                id=
+                'MyClass',  # str
+            )  # /Name
+        )  # /Tuple
         type_comment=
         None,  # NoneType
     )  # /FunctionDef
     type_ignores=
 )  # /Module
\ No newline at end of file
```

Compiling with Python shows no issues
```
$ python3.10 -m py_compile example.py
$ echo $?
0
```

**Expected behavior**

No `ASTSafetyError` thrown, and file is successfully formatted with no functional changes.

**Environment**

Locally:
- black.linux_x86_64, 24.4.2 (compiled: yes)
- Python (CPython) 3.10.9
- OS is Centos

Also reproduces on online editor above

dankrzeminski32:
### Description

Resolves #4399 

Added an extra check before adding commas to the elements in the body of a bracket split.

The check ensures that there is more than just one non-comment item in the list of return elements. It only performs this check for a bracket split in return annotations as it's non-problematic for parameter annotations. 

The extra comment was causing the `ASTSafetyError` (it led to the return type being interpreted as a tuple). 

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@
 
 <!-- Changes that affect Black's stable style -->
 
+Fix `ASTSafetyError` crash when return type contains one element and comments spanning
+multiple lines.
+
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@
 <!-- Changes that affect Black's stable style -->
 
 Fix `ASTSafetyError` crash when return type contains one element and comments spanning
-multiple lines.
+multiple lines. (#4444)
 
 ### Preview style
 
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -1116,6 +1116,14 @@ def bracket_split_build_line(
                 )
                 # Don't add one inside parenthesized return annotations
                 and get_annotation_type(leaves[0]) != "return"
+                # Don't add comment for single element return type with comments
+                and not (
+                    sum(leaf.type is not STANDALONE_COMMENT for leaf in leaves) < 2
+                    and any(
+                        leaf.parent is not None and leaf.parent.type != syms.parameters
+                        for leaf in leaves
+                    )
+                )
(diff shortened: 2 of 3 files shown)


hauntsaninja (commented):
Thank you for the PR!

I have a slightly different fix in mind. We should try and find a leaf that actually has the appropriate parent set. This will fix a similar crash in another case, see:
```
def foo() -> (
    # comment inside parenthesised new union return type
    int | str | bytes
):
    ...
```

I opened a PR with what I had in mind at https://github.com/psf/black/pull/4453. I'd be grateful if you had time to take a look and see if it makes sense to you!

PR #4452, opened 2024-09-08
Drop Python 3.8 support

hauntsaninja:
These are manual changes, we can run pyupgrade equivalents in a later PR

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -549,6 +549,7 @@ def main(  # noqa: C901
     """The uncompromising code formatter."""
     ctx.ensure_object(dict)
 
+    assert sys.version_info >= (3, 9), "Black requires Python 3.9+"
     if sys.version_info[:3] == (3, 12, 5):
         out(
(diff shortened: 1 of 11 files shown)


JelleZijlstra (approved):

src/black/__init__.py
```diff
@@ -549,6 +549,7 @@ def main(  # noqa: C901
     """The uncompromising code formatter."""
     ctx.ensure_object(dict)
 
+    assert sys.version_info >= (3, 9), "Black requires Python 3.9+"
```
jtmedley, line 552: Why is this added? This wasn't needed in the `__init__.py` file for previous minimum versions. Python version requirements are handled by `requires-python` in [pyproject.toml](https://github.com/psf/black/blob/main/pyproject.toml), and that was already correctly changed in this commit.

hauntsaninja: It's harmless enough, Black is definitely broken on 3.8, I have dev environments lying around with random Python versions. Is it causing you some issue?

PR #4457, opened 2024-09-12
Report Black version on internal error

hauntsaninja:
Pretty common for people or IDEs to confuse what Black version is run


JelleZijlstra (approved):
Looks good but I think you have to reformat the code.

PR #4460, opened 2024-09-16
Fix docs CI: use venv for uv to fix 'failed to create directory'

hugovk:
### Checklist - did you ...

- [n/a] Add an entry in `CHANGES.md` if necessary?
- [n/a] Add new / update outdated documentation?


hauntsaninja (approved):
Thanks for the fix! I'd also filed https://github.com/astral-sh/uv/issues/7417 upstream earlier :-)

PR #4462, opened 2024-09-18
Fix skipping Jupyter cells with unknown %% magic

AleksMat:
### Description

This CL fixes an inconsistency of when to skip formatting of Jupyter cells with unknown magic methods.

Black correctly skips formatting in this case:
```Python
%%unknown_custom_magic
<code in another language that black shouldn't format>
```
However if the cell started with some empty lines black still tried to format it and failed:
```


%%unknown_custom_magic
<code in another language that black shouldn't format>
```
Also if the cell started with a comment black tried to format it:
```Python
# Some comment
%%unknown_custom_magic
<code in another language that black shouldn't format>
```
(This example isn't possible in normal Jupyter notebooks but it works in Colab.)

This PR ensures that black will skip formatting cells with a custom magic function which is preceded by any number of empty lines and lines with comments.

Additional notes:
- I moved `validate_cell` function from `__init__.py` to `handle_ipynb_magics.py` module because it belongs there. (It also makes my maintenance of [Pyink fork](https://github.com/google/pyink) a bit easier if there are less things in `__init__.py` module.) But if there is a reason to keep it in `__init__.py` I can move it back.
- Function `_get_code_start` could also be implemented like this:

  ```Python
  def _get_code_start(src: str) -> str:
      match = re.search(r"^[\s\n]*([^#\s\n].*)$", src, flags=re.M)
      return match.group(1) if match else ""
  ```
  It might be a bit more efficient but it is also less readable. Let me know if you would prefer this implementation. 

### Checklist - did you ...

- [ ] ~~Add new / update outdated documentation?~~

### DIFF
--- a/src/black/handle_ipynb_magics.py
+++ b/src/black/handle_ipynb_magics.py
@@ -3,6 +3,7 @@
 import ast
 import collections
 import dataclasses
+import re
 import secrets
 import sys
 from functools import lru_cache
@@ -14,6 +15,7 @@ if sys.version_info >= (3, 10):
 else:
     from typing_extensions import TypeGuard
 
+from black.mode import Mode
 from black.output import out
 from black.report import NothingChanged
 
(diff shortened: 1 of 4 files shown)


src/black/handle_ipynb_magics.py
```diff
@@ -276,6 +306,21 @@ def unmask_cell(src: str, replacements: list[Replacement]) -> str:
...
 
+def _get_code_start(src: str) -> str:
+    """Provides the first line where the code starts.
+
+    Iterates over lines of code until it finds the first line that doesn't
+    contain only empty spaces and comments. It removes any empty spaces at the
+    start of the line and returns it. If such line doesn't exist, it returns an
+    empty string.
+    """
+    for match in re.finditer(".+", src):
```
JelleZijlstra, line 317:
```suggestion
    for match in src.splitlines():
```
Any reason not to use this?

AleksMat: There might be a slight performance difference because `re.finditer` will stop splitting string once the first line with code is found while `src.splitlines()` will always split the entire string by lines. But I haven't benchmarked anything and this probably won't affect much the performance of the entire formatting process.

I'm also ok with switching to `src.splitlines()` if that is the preferred way.

JelleZijlstra: Thanks, I guess this is fine.

PR #4464, opened 2024-09-24
Rename plugin/black.vim to ftplugin/python.vim

Freed-Wu:
Black is only for python.
So sourcing the plugin when vim user edit python file will save the most people's time.

### Checklist - did you ...


Freed-Wu:
BTW, source this vim plugin is 5th time-cost event:

```
event                  time percent plot
rocks.lua            177.24   27.70 ██████████████████████████
init.vim             143.03   22.36 █████████████████████
VimEnter autocommand 124.52   19.46 ██████████████████▎
opening buffers       65.87   10.30 █████████▋
black.vim             17.48    2.73 ██▋
```

Freed-Wu:
What is the status?

PR #4479, opened 2024-10-11
Store license identifier inside the License-Expression metadata field

cdce8p:
### Description
The `hatch` documentation recommends to use just `license = "MIT"` instead of `license = { text = "MIT" }`.
https://hatch.pypa.io/1.9/config/metadata/#spdx-expression

With that `hatchling` will write a compatible `License-Expression` to the package metadata.
```
...
License-Expression: MIT
...
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Change Log
 
-# Unreleased
+## Unreleased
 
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):
Thanks!

PR #4484, opened 2024-10-14
Fix formatting cells with magic methods and starting or trailing empty lines

AleksMat:
### Description

This PR fixes an issue where Jupyter cells containing any magic method and at least 1 empty line at the start or at the end of the cell don't get formatted at all. 

This is an example of a cell that doesn't get formatted:
```


%time
a=1


```

But if there are no magic methods then formatting works. The following cell:
```


#No magic methods
a=1


```
gets correctly reformatted into:
```
# No magic methods
a = 1
```

It turns out the problem is in `black.handle_ipynb_magics.mask_cell` function:
1. The call to `TransformerManager.transform_cell` returns a result without empty lines at the beginning of the cell. Because this part is implemented in `ipython` I didn't investigate why this is the case.
2. The call to `replace_magics` returns a result without empty lines at the end of the cell. This is because the implementation of `replace_magics` assumes that 
   ```Python
   "\n".join(src.splitlines())
    ```
   will always produce `src` again. This is not the case, e.g. `"\n".join("\n\n".splitlines())` returns `"\n"`.

3. Because of 1. and 2. the check `len(transformed.splitlines()) != len(src.splitlines())` is triggered and `NothingChanged` is raised.

#### Solution

1. A comment about `TransformerManager.transform_cell`  removing empty lines. Future improvements in `ipython` are possible.
2. Minor update in `replace_magics` so that it won't be removing empty lines at the end anymore.
3. The number of lines in `transformed` and `src` should be compared without starting and ending empty lines. Hence `len(transformed.strip().splitlines()) != len(src.strip().splitlines())`

I also added a few unit tests that test such cases.

### Checklist - did you ...

- [ ] ~Add new / update outdated documentation?~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@
 
 <!-- Changes that affect Black's stable style -->
 
-- Fix formatting cells with magic methods and starting or trailing empty lines (#4484)
+- Fix formatting cells in IPython notebooks with magic methods and starting or trailing empty lines (#4484)
 
 ### Preview style
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
-- Fix formatting cells in IPython notebooks with magic methods and starting or trailing empty lines (#4484)
+- Fix formatting cells in IPython notebooks with magic methods and starting or trailing
+  empty lines (#4484)
 
 ### Preview style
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -10,6 +10,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Fix formatting cells with magic methods and starting or trailing empty lines (#4484)
```
```suggestion
- Fix formatting cells in IPython notebooks with magic methods and starting or trailing empty lines (#4484)
```
Otherwise readers who only use Black for formatting regular Python code may be confused.

JelleZijlstra (commented):
Thanks!

PR #4503, opened 2024-10-25
note required python version for use_pyproject: true

Mattwmaster58:
docs don't work out-of-the-box in current state

### Description

docs

### Checklist - did you ...


hauntsaninja (approved):
Thanks!

Needs to be targeted to main, I can do that

PR #4512, opened 2024-11-14
Fix issue with newer upload-artifact in PyPI action

hauntsaninja:
Github is breaking older upload-artifact in a few weeks


webknjaz:
@hauntsaninja any reason you didn't migrate to Trusted Publishing? It can produce digital attestations out of the box now: https://blog.pypi.org/posts/2024-11-14-pypi-now-supports-digital-attestations/.

JelleZijlstra:
We should move to Trusted Publishing at some point, but that's a bit more work, so we'll do it when we do it I suppose. I'd probably accept a PR that does it for us.

PR #4529, opened 2024-12-09
Add 2248 tests

### LINKED ISSUE #2248: Formatting of conditional expressions in named function arguments

MeGaGiGaGon:
It looks like #2248 was fixed at some point, probably as part of the 2024 release. This adds an anti-regression test for, and closes #2248.

### DIFF
--- a/tests/data/cases/anti_regression_2248.py
+++ b/tests/data/cases/anti_regression_2248.py
@@ -1,0 +1,19 @@
+aaa = my_function(
+    foo="test, this is a sample value",
+    bar=some_long_value_name_foo_bar_baz
+    if some_boolean_variable
+    else some_fallback_value_foo_bar_baz,
+    baz="hello, this is a another value",
+)
+
+# output
+
+aaa = my_function(
+    foo="test, this is a sample value",
+    bar=(
+        some_long_value_name_foo_bar_baz
+        if some_boolean_variable
+        else some_fallback_value_foo_bar_baz
+    ),
+    baz="hello, this is a another value",
+)


JelleZijlstra:
I think this one is already tested exhaustively in https://github.com/psf/black/blob/main/tests/data/cases/conditional_expression.py .

MeGaGiGaGon:
Yep, looks like github search doesn't like multiline code. Sorry for the churn

JelleZijlstra:
For what it's worth I found it locally with `git grep '^ *else '`.

PR #4530, opened 2024-12-09
Add 1765 tests

### LINKED ISSUE #1765: black breaks the type annotation into multiple lines when it's possible to keep it on one line
**Describe the bug**

black breaks the type annotation into multiple lines, while existing formatting keep it on one line by using parentheses around expression

**To Reproduce**:

old formatting:
```
class Foo:
    def foo(self):
        if True:
            content_ids: Mapping[str, Optional[ContentId]] = (
                self.publisher_content_store.store_config_contents(files)
            )
```

after black:
```
class Foo:
    def foo(self):
        if True:
            content_ids: Mapping[
                str, Optional[ContentId]
            ] = self.publisher_content_store.store_config_contents(files)
```

playground link: https://black.now.sh/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AEZAK5dAD2IimZxl1N_WlXnON2nzNX9HFvHx6cBSdVHWyzJO6O50DK3ha9hxTlR0JwlWYXCXbF4CjYJWeEpyQIXtrrOkKmzjEkT25z6cJ4YzWm7pHh_LmznKdvIU8f2vDLKjoyxv-ZPlxXoOnqhPz8sGlioAQx7_azDc47CPs6gbJw5n5j0LZ-lBGbWsnS8ldMWJXrmleKBm2qB_EmzBNy6IQNsCElvvBmM-J3BIWV55zK6IAAAAAMi_lQVqdYNAAHKAZoCAAARFT-fscRn-wIAAAAABFla

**Expected behavior**
black keeps existing formatting

**Environment:**

- Version: both stable and master
- OS and Python version: Linux/Python 3.7.5+

Yes.

MeGaGiGaGon:
It looks like #1765 was fixed at some point. This adds an anti-regression test for, and closes #1765.


tests/data/cases/type_params.py
```diff
@@ -13,6 +13,13 @@ def it_gets_worse[WhatIsTheLongestTypeVarNameYouCanThinkOfEnoughToMakeBlackSplit
 
 def magic[Trailing, Comma,](): pass
 
+class Foo:
```
JelleZijlstra, line 16: This file is meant for PEP 695 syntax, which your new test doesn't use. Therefore, it's also run in 3.12+ mode. I couldn't find a better file to put it, so I'd suggest adding a new test file, maybe `annotation.py`.

MeGaGiGaGon: Moved to `annotations.py`

tests/data/cases/annotations.py
```diff
@@ -0,0 +1,17 @@
+# anti 1765 regression
```
JelleZijlstra, line 1:
```suggestion
# regression test for #1765
```

tests/data/cases/annotations.py
```diff
@@ -0,0 +1,17 @@
...
+class Foo:
+    def foo(self):
+        if True:
+            content_ids: Mapping[
+                str, Optional[ContentId]
+            ] = self.publisher_content_store.store_config_contents(files)
+
+# output
+
+# anti 1765 regression
```
cobaltt7, line 11:
```suggestion
# regression test for https://github.com/psf/black/issues/1765
```

PR #4531, opened 2024-12-09
Add `*.py diff=python` to `.gitattributes`

MeGaGiGaGon:
This adds the line `*.py text diff=python` to `.gitattributes`. I'm not 100% certain on all the effects this has, but it makes function name based finding git commands options like `git log -L :<function_name>:<file_path>` or `git blame -L :<function_name> <file_path>` work out of the box, which is a nice convenience.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 .git_archival.txt  export-subst
+*.py text diff=python
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 .git_archival.txt  export-subst
-*.py text diff=python
+*.py text=auto diff=python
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
This appears to break test, something with CRLF last names.

```
diff --git a/tests/data/cases/preview_long_strings__east_asian_width.py b/tests/data/cases/preview_long_strings__east_asian_width.py
```

MeGaGiGaGon:
As far as I can tell, the issue was because windows. I don't understand why anything broke, since github should use `core.autocrlf=true` by default, which should have the same behavior as `text`, but it doesn't. Removing the `text` still lets the diff function names work, but fixes the line ending nonsense to still be the `core.autocrlf` behavior.

Issue #4520, opened 2024-11-26
Freeze/bad performance on escaped quotation in f-string


scherrsasrf:
**Describe the bug**
black hangs (or takes an extremely long time?) to process a file with a multiline f-string containing a bunch of escaped quotation marks. 

**To Reproduce**
Run black against this file:
```python
def testing():
    somevar = "some value"
    return f"""
        {somevar} the following line causes the issue:
         \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" 
    """
```
Will just hang without any output.

**Expected behavior**
Should run to completion within at most 1 second.

**Environment**
- Black's version: 24.10.0 (was working with 24.4.2, probably introduced in 24.8.0)
- OS and Python version: macOS/Python 3.13.0

**Additional context**
Extracting the quotes into a variable seems to circumvent the issue:
```python
def testing():
    somevar = "some value"
    quotes = """\" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \" \""""
    return f"""
        {somevar} the following line causes the issue:
         {quotes} 
    """
```

MeGaGiGaGon:
The hang happens at [this line](https://github.com/psf/black/blob/17efac45f9e15750b1df7adb6432a8970e9054e0/src/blib2to3/pgen2/tokenize.py#L615) `endmatch = endprog.match(line)` 

At that point `endprog` is this regex `((?:\\\\N{|{{|\\\\"|"(?!"")|[^"{])*(?<!\\\\N){(?!{)|(?:\\\\.|"(?!"")|[^"\\\\])*""")` which has catastrophic backtracking.

[...]

Issue #2318, opened 2021-06-08
Switch to a new parser


JelleZijlstra:
Currently, Black uses a vendored version of [lib2to3](https://docs.python.org/3/library/2to3.html#module-lib2to3) for parsing. This works well for parsing Python 2 and early Python 3, but Python has now moved on to a PEG-based parser ([PEP 617](https://www.python.org/dev/peps/pep-0617/)), and lib2to3 is no longer being maintained.

So we need a new parser. There are a few existing options that we could leverage (Parso, LibCST), but it's going to be a lot of work to do the migration. WE're doing some early brainstorming in [a Google doc](https://docs.google.com/document/d/1QPqm01E7MIRi_l4jrgCGOFVThP_HINnkm4hmVZC_CSg/edit?fbclid=IwAR03VBxQLhXkrZiO7BDUQerkeQU7MOy51BHtJ9xqRMHonBF7_dF1Tef53fs#heading=h.qokppq9m1uzy). This issue exists so that we have a public record that we know this is a problem.

~~Concrete pieces of syntax that are blocked by this new grammar include parenthesized context managers and the match statement in Python 3.10.~~ (#2242 through #2586, #2667, #2758)

kamahen:
The main bug is: https://bugs.python.org/issue40360 (and also https://bugs.python.org/issue36541).

I think that there's a fairly straightforward way of wrapping the new Python parser to give the necessary functionality that "Black" (and other source-level tools) need. However, it's a non-trivial amount of work, and I'm loathe to do it unless I'm sure it'll be used and that nobody else is doing the work. (There appears to be one existing wrapper, namely leoAst.py; I've looked at it a bit but it seems much more complicated than necessary and therefore could be both difficult to use and a maintenance issue.)

Some other discussion at https://github.com/google/yapf/issues/825#issuecomment-868805396 , https://github.com/google/yapf/issues/894#issuecomment-799867767 and elsewhere.

ianliu:
Has [treesitter](https://tree-sitter.github.io/tree-sitter/) been considered? It already implements a parser for python here: https://github.com/tree-sitter/tree-sitter-python and I think it allows to build formaters upon it.

jakkdl:
LibCST now supports (according to readme) 3.0->3.11, though it does say
[...]

PR #4536, opened 2024-12-22
Replace the blib2to3 tokenizer with pytokens

tusharsadhwani:
### Description

Replaces black's tokenizer with a from-scratch rewrite.

Should help resolve problems like #4520 and others.

### What's left

- `ERRORTOKEN` handling is not very proper just yet, it'll require some tweaks.
- Edge case tests: Currently 366/381 tests are passing.

### DIFF
--- a/src/blib2to3/pgen2/tokenize.py
+++ b/src/blib2to3/pgen2/tokenize.py
@@ -45,13 +45,11 @@ from blib2to3.pgen2.token import (
     FSTRING_MIDDLE,
     FSTRING_START,
     INDENT,
-    LBRACE,
     NAME,
     NEWLINE,
     NL,
     NUMBER,
     OP,
-    RBRACE,
     STRING,
     tok_name,
 )
@@ -61,6 +59,11 @@ __credits__ = "GvR, ESR, Tim Peters, Thomas Wouters, Fred Drake, Skip Montanaro"
 
 import re
 from codecs import BOM_UTF8, lookup
+import token
+
+import pytokens
+from pytokens import TokenType
+
 
 from . import token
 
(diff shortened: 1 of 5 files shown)


tusharsadhwani:
@JelleZijlstra with this, the test suite is fully passing. Primer is failing (mostly just because some file in `hypothesis` failed to parse), I'll be working on that, but the code should be good for a first review.

MeGaGiGaGon:
I don't think we currently have any tests for this, but I just linked the above two issues here because they are the same bug in the parser where `\r`s cause issues. The most minimal reproduction is `{\r}`, and since this is a parser rewrite hopefully it can be solved. Note that this is currently only observable by directly calling internal methods due to how black reads input.

tusharsadhwani:
Thanks for linking this, I'll make sure these parse identically to how CPython does it.

tusharsadhwani:
Okay, primer is fixed, and all tests are green.

JelleZijlstra:
Might this fuzzer failure indicate a bug?

```
Falsifying example: test_idempotent_any_syntatically_valid_python(
    src_contents='\n#\r#',
    mode=Mode(target_versions=set(),
     line_length=88,
     string_normalization=False,
     is_pyi=False,
     is_ipynb=False,
     skip_source_first_line=False,
     magic_trailing_comma=False,
     python_cell_magics=set(),
     preview=False,
     unstable=False,
     enabled_features=set()),  # or any other generated value
)
```

src/blib2to3/pgen2/tokenize.py
```diff
@@ -45,1070 +43,218 @@
...
-    Each call to the function should return one line of input as a string.
 
-    The second parameter, tokeneater, must also be a callable object. It is
-    called once for each token, with five arguments, corresponding to the
-    tuples generated by generate_tokens().
+def transform_whitespace(
+    token: pytokens.Token, source: str, prev_token: Optional[pytokens.Token]
+) -> pytokens.Token:
+    r"""
+    Black treats `\\\n` at the end of a line as a 'NL' token, while it
```
JelleZijlstra, line 105: That doesn't sound particularly intentional, I'd be open to changing Black to remove this divergence.

tusharsadhwani:
> Might this fuzzer failure indicate a bug?

Yeah, but I'm pretty sure it is a bug in CPython. For now we can work it out in the tokenizer though. I'll add a flag in `pytokens` to fix it.

src/blib2to3/pgen2/tokenize.py
tusharsadhwani: Feel free to! I can give you a test case with the expected behaviour.

tusharsadhwani: So, this is enough as a test case actually:

```python
a \
  b
```

But, the reason this probably exists is to support formatting [this file](https://github.com/psf/black/blob/main/tests/data/cases/backslash_before_indent.py):

```python
class Plotter:
\
    pass

class AnotherCase:
  \
    """Some
    \
    Docstring
    """
```

MeGaGiGaGon:
I found another issue that this should close, https://github.com/psf/black/issues/2318

JelleZijlstra:
@MeGaGiGaGon no, that's about the parser, not the tokenizer. It would still be good long term to switch to a different parser, because the current parser only supports Python's recent grammar changes through a series of hacks.

hauntsaninja:
This seems awesome! Could be worth doing a quick benchmark though with a mypyc build (I guess we could potentially mypyc pytokens if there is a perf difference). I think to do that you have to set these env vars https://github.com/psf/black/blob/main/pyproject.toml#L181 (but honestly not too sure how hatch-mypyc is setup)

tusharsadhwani:
@JelleZijlstra edge cases have been taken care of.

@hauntsaninja I just tested `mypyc` on `pytokens` locally, and it got 2x faster at tokenizing the `black/profiling` folder.

src/blib2to3/pgen2/tokenize.py
tusharsadhwani: I tried removing this edge case, but it just seems to break the backslash before indent test case. Let's just keep it as-is for now and try to fix it later?

cobaltt7:
Resolves #4588 too

tusharsadhwani:
@JelleZijlstra can you confirm if the PR is good to review, or are changes needed?

JelleZijlstra (approved):
Let's do it!

tusharsadhwani:
@JelleZijlstra I don't have merge rights just yet :P

JelleZijlstra:
Sorry I had to resolve a merge conflict and then wait for CI

PR #4537, opened 2024-12-23
Fix type error

cobaltt7:
### Description

Fixes a type error introduced in https://github.com/python/typeshed/pull/13141
As mentioned in that PR, it is a legitimate bug, however `self.filename` is never not a string at runtime
Regardless, the type error is currently causing CI to fail and needs to be resolved

### Checklist - did you ...

- [N/A] Add an entry in `CHANGES.md` if necessary?
- [N/A] Add new / update outdated documentation?

### DIFF
--- a/src/blib2to3/pgen2/pgen.py
+++ b/src/blib2to3/pgen2/pgen.py
@@ -357,7 +357,9 @@ class ParserGenerator:
                 msg = msg % args
             except Exception:
                 msg = " ".join([msg] + list(map(str, args)))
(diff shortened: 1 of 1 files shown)


cobaltt7:
Diff-shades will continue to fail until after this is merged, since it needs to build the main branch as well as the PR branch, and the main branch can't be built until after this is merged

hauntsaninja (approved):

hauntsaninja:
Thank you!

PR #4538, opened 2024-12-23
Fix 3678 "problem parsing `with` statements"

### LINKED ISSUE #3678: problem parsing with statements
Some with statements can not be parsed by black.
They are unlikely to appear in real code, but they are valid python and can be parsed by cpython.

**Examples**

```python
with (name_2 for name_0 in name_4):
    pass
```

```python
with (a,*b):
    pass
```

None of these examples have a valid `__enter__` or `__exit__` and should fail at runtime.

black gives the following error:
```
error: cannot format example4.py: Cannot parse: 1:8: with a, *b:

Oh no! 💥 💔 💥
1 file failed to reformat.

```

**Expected behavior**

black should be able to parse and format this statements.

**Environment**

python -m black, 23.3.1.dev10+geb32729 (compiled: no)
Python (CPython) 3.11.3

MeGaGiGaGon:
### Description

This fixes #3678. The fix is done by adding two new functions to `node.py`, `is_tuple_containing_star` and `is_generator`, which are then added to the checks in `maybe_make_parens_invisible_in_atom` inside `linegen.py`, which is used by the function `remove_with_parens` that Jelle suggested modifying in a comment on #3678

Both functions are based on the existing `is_tuple_containing_walrus`. That makes me fairly confident in `is_tuple_containing_star`, since it does the same thing as `is_tuple_containing_walrus`, just with `syms.namedexpr_test` swapped for `syms.star_expr`. I'm less confident in `is_generator` since `for a in b` being called `syms.old_comp_for` seems weird and I don't totally understand `syms`/`_python_symbols`, but it seems to work.

Side note, `_python_symbols` is really hard to work with since `node.type` prints as an integer and the `_python_symbol` variable <--> integer relationship is done via `setattr`, so I ended up having to do `[print(x, eval(f"syms.{x}") for x in dir(syms) if x[0:2] != "__"]` to figure out what type generators get.

As usual, both names subject to bikeshedding. I chose `is_tuple_containing_star` because it compares with `syms.star_expr`, but  `is_tuple_containing_unpacking` could also work. I chose `is_generator` since the issue is with a generator, though that could be something else since it doesn't match the `syms.old_comp_for` comparison.

The tests contain both a top level case and one where the issue is nested, since the nested case caused issues for me while making the fix.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,9 @@
 - Fix formatting cells in IPython notebooks with magic methods and starting or trailing
   empty lines (#4484)
 
+- Fix crash when formatting `with` statements containing tuple genetators/unpacking
+  (#4538)
+
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@
 - Fix formatting cells in IPython notebooks with magic methods and starting or trailing
   empty lines (#4484)
 
-- Fix crash when formatting `with` statements containing tuple genetators/unpacking
+- Fix crash when formatting `with` statements containing tuple generators/unpacking
   (#4538)
 
 ### Preview style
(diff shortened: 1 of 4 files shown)


JelleZijlstra (approved):
Thanks!

Issue #2492, opened 2021-09-11
Document PyInstaller binaries


JelleZijlstra:
The documentation doesn't say anything about our PyInstaller binaries (or if it does I can't find it). Let's fix that.

joybh98:
Is this something that is still needed, if it is i would like to take it

cooperlees:
I only see https://black.readthedocs.io/en/stable/contributing/release_process.html?highlight=pyinstaller#publish-executables

Searching shows that too: https://black.readthedocs.io/en/stable/search.html?q=pyinstaller&check_keywords=yes&area=default

I guess @JelleZijlstra would like something about it as an available way to download black.

Maybe making a dedicated page for "Ways to install black" and link to it from the "installation" part of the "Getting Started" might be a good idea here.

Gettting started page: https://github.com/psf/black/blob/main/docs/getting_started.md
- Side note: I don't know why we recommend installing with jupyter by default - That seems wrong
  - I think we should split that out on the new "Ways to install black" and explain each [extra install](https://github.com/psf/black/blob/main/setup.py#L112-L117)

Today we have:
- PyPI (already kind of documented)
  - https://pypi.org/project/black/
  - We could maybe explain how we compile black and offer native python version too
- pyinstaller binaries via GitHub
  - https://github.com/psf/black/releases/tag/22.8.0
- Docker
  -  https://hub.docker.com/r/pyfound/black
- Pulling latest from GitHub itself (already documented on Getting Started)
  - i.e. `pip install git+https://github.com/psf/black`

I realize this is a lot of scope creep, but maybe parts of this could be useful to take on. What are others thoughts?

PR #4545, opened 2025-01-05
2492: Adding Information about PyInstaller Binaries

RohitGavale:
Added information about PyInstaller Binaries. This is my first contribution for learning purpose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,50 @@ If you can't wait for the latest _hotness_ and want to install from GitHub, use:
 
 `pip install git+https://github.com/psf/black`
 
+### PyInstaller binaries
+
+PyInstaller is a tool that packages Python programs into standalone executables, under Windows, Linux, and macOS. 
+This allows users to run Python applications without needing to install Python or any dependencies on their system. 
+
+- How PyInstaller Works:
+
+Compilation into Executables:
+PyInstaller analyzes your Python script and identifies all necessary dependencies, such as libraries, Python modules, and resources (e.g., images, configuration files).
+It then compiles your script and the identified dependencies into a single executable or a directory containing the executable and the necessary shared libraries.
+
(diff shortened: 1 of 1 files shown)


MeGaGiGaGon:
Thank you for your contribution. I've looked it over and have a couple comments:

You can create a link to another issue/pr by doing #<number>, ie #2492. If you put something along the lines of "Fixes #2492" in your opening PR description, it will link the issue so it is automatically closed on merge. This also works if you edit the message after creation.

I read your changes and #2492, and sadly I don't think your current changes match what #2492 was asking for. According to the [one place PyInstaller is mentioned in the docs](https://black.readthedocs.io/en/stable/contributing/release_process.html#publish-executables) the binaries are so that people can run black without having to install python. #2492 looks to be wanting more places in the documentation that say "Hey, we have PyInstaller binaries you can use if you don't have/want python". The best places to add those statements would probably be the [installation section of the getting started page](https://github.com/psf/black/blob/main/docs/getting_started.md#installation) and [the installation section of the readme](https://github.com/psf/black?tab=readme-ov-file#installation)

As [cooperlees](https://github.com/cooperlees) said in #2492, if you want to be more ambitious you could also make a dedicated docs section for all the different ways to install black, at least some of which were listed [by cooperlees in #2492](https://github.com/psf/black/issues/2492#issuecomment-1242368679). I'm not sure if that list remains accurate since it has been two years.

cooperlees:
> I'm not sure if that list remains accurate since it has been two years.

They are all still valid from my quick re-read.
- Naturally we probably want to use latest release tho if there is some way to link to that rather than - I think https://github.com/psf/black/releases/latest works ...

Himanshu5206:
Is the issue still open or closed, I want to contribute to it

Issue #4535, opened 2024-12-21
`# fmt: skip` is not being respected with one-liner functions


polirritmico:
**Describe the bug**

Hi, `# fmt: skip` doesn't skip the formatting in this code:

```python
def foo(): return "mock"  # fmt: skip
```

```sh
$ black foo.py; cat foo.py
```

Output:

```python
def foo():
    return "mock"  # fmt: skip
```

Similar case in the [playground](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AEpAJNdAD2IimZxl1N_WlXnON2nzN36C8EDS2OdozrF3A8N-yRxMzveJgvGD0d8QKTaox5vY_DvNkC0MxTYAIJP6wYvhPX_6e8_eA8axqa4Y00iUpSPgGJNJ0BXC0uC1k1IlIG9EiZ3ywxEX5_iWE6tChr3tTsKnUXni9YhX2RsZpxTMsl38XdOSWMqyjqIDQG6NIDSnK7WAAAAih4iHeqkrkAAAa8BqgIAAFlzLTexxGf7AgAAAAAEWVo=).

**Expected behavior**

From the docs I understand that Black should not reformat lines that contain `# fmt: skip` (The basics -> Usage -> Ignoring sections).

**Environment**

- Black's version: `24.10.0 (compiled: yes)`, `24.10.1.dev22+g6000d37 (compiled: no)`
- OS and Python version: `Gentoo Linux 2.17`, `Python 3.12.8`

Thanks!

Pedro-Muller29:
I investigated this issue and discovered the following:

1. **Behavior in Black Past Versions**:  
   This has never worked in Black (at least not in versions from the past two years; I didn’t check further back).

2. **Tree Comparison**:  
   The `lib2to3` tree generated is identical for the following two source codes:

   **Source 1**:
   ```python
   def foo(): return "mock"  # fmt: skip
   ```
   **Source 2**:
    ```python
   def foo():
       return "mock"  # fmt: skip
   ```
   **Generated Tree**
   ```
   [
        Node(funcdef, [
            Leaf(NAME, 'def'),
            Leaf(NAME, 'foo'),
            Node(parameters, [
                Leaf(LPAR, '('),
                Leaf(RPAR, ')')
            ]),
            Leaf(COLON, ':'),
            Node(suite, [
                Leaf(NEWLINE, '\n'),
                Leaf(INDENT, ''),
[...]

Issue #3682, opened 2023-05-07
`# fmt: skip` ignored in `if` conditionals


metov:
**Describe the bug**

`# fmt: skip` does not work in `if` conditionals.

**To Reproduce**
I expect the following code to not be reformatted, due to the skip directive. It is.
```
if True: print("yay") # fmt: skip
```

https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AC8AG9dAD2IimZxl1N_Wlws4TBexXdsrAK05TP_h68mLf6AjhWJHCRgXa5Rwd9gFE21oq9pVC8qdR56y-a_GL_VJamRio8oas6Eub07I8wP5I3sv2WdLkei-6Q30-uDRVSurPlC7LiqqMoMjKtW0_uwtGuaAAAAiskgpy3XxxYAAYsBvQEAAAblmGSxxGf7AgAAAAAEWVo%3D

**Expected behavior**

Since the line contains `# fmt: skip`, black should leave it alone.

**Environment**

Problem happens on Black Playground, clearly it is not env specific.

- Black's version: 23.3.0
- OS and Python version: Arch Linux, Python 3.10.9

metov:
This may be related to https://github.com/psf/black/issues/3364. I'm not really familiar with how exactly black works, but I assume the skip directive applies to a single "syntactic element", and the condition is split off from the statement into a separate element before the skip is applied. Clearly it should be the other way around.

This is only a hunch though. Please do not close either of these as duplicate until there is actual evidence that these are indeed the same bug.

treykeown:
I believe the underlying cause for my issue is related. `#fmt: skip` doesn't work for single-line class definitions like this:
```
class Foo: ...  # fmt: skip
```
It's always reformatted for me to
```
class Foo:
    ...  # fmt: skip
```

tim-timman:
Just want to add another situation this bugs out, but I've verified that #3978 fixes this (so please get that merged).

For example, take this code:

```python
# {} () [] / % & | and or ... - + .  etc as part of the "else" clause presents the bug
(
    ""
    if True else  # fmt: skip
    []
    # NOTE: this entire comment is completely GONE!
    # Doesn't matter how many rows
)
```

[...]

PR #4552, opened 2025-01-18
Fix bug where # fmt: skip is not being respected with one-liner functions

Pedro-Muller29:
### Description

Fixes https://github.com/psf/black/issues/3682
Fixes https://github.com/psf/black/issues/4535

The fix involves modifying the algorithm that determines which nodes to ignore in the `_generate_ignored_nodes_from_fmt_skip` function. Specifically, the adjustment applies to cases where the `# fmt: skip` comment is **not** immediately after the colon in an `if`, `while`, `def`, `class`, etc.

#### Previous Algorithm
The previous algorithm for selecting ignored nodes worked as follows:
1. Add the *previous sibling* of the current node to the list of ignored nodes.
2. Set the current node to its previous sibling.
3. Stop if the current node has no previous sibling or its previous sibling contains a `\n` in its prefix.

This approach worked well when the node with the comment existed at a tree level that covered the entire line. However, it failed in cases where this assumption was invalid. For example, consider the tree produced for the code in #4535:

```
Node(funcdef)
├── Leaf(NAME, 'def')   ─┐
├── Leaf(NAME, 'foo')    │
├── Node(parameters)     │
 │   ├── Leaf(LPAR, '(') │
 │   └── Leaf(RPAR, ')') ┣── This level is never reachable, so they won't be ignored by # fmt: skip.
├── Leaf(COLON, ':')     │
└── Node(simple_stmt)   ─┘
    ├── Leaf(STANDALONE_COMMENT, 'return "mock"  # fmt: skip') <-- The algorithm begins here.
    └── Leaf(NEWLINE, '\n')
Leaf(ENDMARKER, '')
```

As shown, only the content of the `return "mock"  # fmt: skip` would be ignored.  
This can be tested with the following input (note the single quotes):
```python
def foo(): return 'mock' # fmt: skip
```
The resulting output would be:
```python
def foo():
    return 'mock' # fmt: skip
```

Here, the single quotes are protected by the skip statement.

### New Algorithm

The new algorithm traverses the tree differently, ensuring it can visit all leaves in the order they appear in the source code (but in reverse). For the example tree:

1. The traversal moves from the comment to the colon (`:`), then to the `RPAR`, `LPAR`, and so on.
2. It stops when encountering a `NEWLINE` or `INDENT` node.
3. All visited leaves are ignored by `# fmt: skip`, expect the `NEWLINE` or `INDENT` node.

This approach ensures all relevant nodes are included in the ignored list.

### Checklist - did you ...

### DIFF
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -270,7 +270,7 @@ def generate_ignored_nodes(
     Stops at the end of the block.
     """
     if _contains_fmt_skip_comment(comment.value, mode):
-        yield from _generate_ignored_nodes_from_fmt_skip(leaf, comment)
+        yield from _generate_ignored_nodes_from_fmt_skip(leaf, comment, mode)
         return
     container: Optional[LN] = container_of(leaf)
     while container is not None and container.type != token.ENDMARKER:
@@ -309,11 +309,12 @@ def generate_ignored_nodes(
 
 
 def _generate_ignored_nodes_from_fmt_skip(
-    leaf: Leaf, comment: ProtoComment
+    leaf: Leaf, comment: ProtoComment, mode: Mode
 ) -> Iterator[LN]:
     """Generate all leaves that should be ignored by the `# fmt: skip` from `leaf`."""
     prev_sibling = leaf.prev_sibling
     parent = leaf.parent
+    ignored_nodes: list[LN] = []
     # Need to properly format the leaf prefix to compare it to comment.value,
     # which is also formatted
     comments = list_comments(leaf.prefix, is_endmarker=False)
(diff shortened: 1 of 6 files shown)


Pedro-Muller29:
I’d also like to mention that I included this in the Stable Style, as I considered it more of a bug fix than a new feature.  

However, I’m not yet fully familiar with the criteria for what qualifies as Stable versus Preview Style, and I couldn’t find much information about this in the documentation. Any guidance would be greatly appreciated.

tests/data/cases/fmtskip10.py
```diff
@@ -0,0 +1 @@
+def foo(): return "mock"  # fmt: skip
```
MeGaGiGaGon, line 1:
```suggestion
def foo(): return "mock"  # fmt: skip
if True: print("yay")  # fmt: skip
for i in range(10): print(i)  # fmt: skip
```
Looks to also fix #3682 so this adds two examples from that as test cases. The other issues not fixed by this PR in #3682 look to have been fixed by something in the 03-15-2024 release, probably #4248, so it should be good to add to the close-on-merge list for this PR along with #4535
(To do that, edit the opening description to include
```
Fixes #3682
Fixes #4535
```
)

cobaltt7:
> However, I’m not yet fully familiar with the criteria for what qualifies as Stable versus Preview Style, and I couldn’t find much information about this in the documentation. Any guidance would be greatly appreciated.

Typically speaking, everything should go in the preview style, the only exceptions being bug fixes for crashes and code changes that don't change code that has already been formatted by Black. This fits under the second case, so it does not need to go through the preview style if it doesn't make sense for it to.

tests/data/cases/fmtskip10.py
Pedro-Muller29: Thanks for the review.

I added the suggested tests + some extra ones.
I also mentioned #3682 on the PR description as suggested.

Pedro-Muller29:
> > However, I’m not yet fully familiar with the criteria for what qualifies as Stable versus Preview Style, and I couldn’t find much information about this in the documentation. Any guidance would be greatly appreciated.
> 
> Typically speaking, everything should go in the preview style, the only exceptions being bug fixes for crashes and code changes that don't change code that has already been formatted by Black. This fits under the second case, so it does not need to go through the preview style if it doesn't make sense for it to.

Thank you for the clarification!

I’ve already updated my PR to use the preview style. During this process, the test I added naturally started failing, since it was running on stable mode. I attempted to give the --preview flag to the test case, but it didn’t resolve the issue.

The test only worked when I gave it the `--unstable` flag, which allowed the code to enter the branch if `Preview.fix_fmt_skip_in_one_liners in mode`. I noticed that other PRs introducing features to the preview mode (e.g., #4498) also used the `--unstable` flag in their tests.

Is this the correct approach? If so, could you clarify the intended purpose of the `--preview` flag in test cases?

JelleZijlstra:
The idea of the stable style is that if you have a codebase formatted with Black (say) 25.1, and you upgrade to Black 25.2, nothing should change. This is useful so that users can upgrade Black and see bugfixes without having to commit reformatting. If this change only affects how Black formats code that wasn't previously formatted by Black.

Putting `--preview` in a test should be enough. `--unstable` also enables features in the UNSTABLE_FEATURES set, but shouldn't affect other preview features.

src/black/comments.py
```diff
@@ -313,31 +313,74 @@ def generate_ignored_nodes(
...
         return
     if prev_sibling is not None:
         leaf.prefix = ""
-        siblings = [prev_sibling]
-        while "\n" not in prev_sibling.prefix and prev_sibling.prev_sibling is not None:
-            prev_sibling = prev_sibling.prev_sibling
-            siblings.insert(0, prev_sibling)
-        yield from siblings
+
+        if Preview.multiline_string_handling not in mode:
```
JelleZijlstra, line 330: Wrong preview feature?

cobaltt7:
Can you run `tox -e generate_schema`? This will regenerate the schema for the configuration, which is changed by adding new preview features. That will fix CI failing.

Pedro-Muller29:
> Can you run `tox -e generate_schema`? This will regenerate the schema for the configuration, which is changed by adding new preview features. That will fix CI failing.

Thanks for the response, I wasn't sure how to proceed. I searched in the docs but I couldn't find anything that related to the CI error. Is there any other places that contains this info? (I searched on the ReadTheDocs page in the development section).

Thanks in advance.

Pedro-Muller29:
Now the CI seems to be failing because of the deprecation of the `upload-artifact v3` action in the `.github/workflows/diff_shades.yml` workflow, as can be seen [here](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/).

MeGaGiGaGon:
@Pedro-Muller29 I just fixed that in #4576, @JelleZijlstra could you re-run those two?

JelleZijlstra (approved):

PR #4555, opened 2025-01-21
Fix docs CI

MeGaGiGaGon:
### Description

Docs CI is now failing thanks to 
> The `sphinx.configuration` key is missing. This key is now required, see our [blog post](https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/) for more information.

It looks like black does the same thing as the `.readthedocs.yaml` example from that post, so I just copied it.

### Checklist - did you ...

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -16,3 +16,6 @@ python:
       path: .
       extra_requirements:
         - d
+
+sphinx:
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):
Thanks for fixing!

PR #4377, opened 2024-06-06
fix: Don't remove parenthesis around long dictionary values

cobaltt7:
### Description

Resolves #4158 (`string_processing` and `wrap_long_dict_values_in_parens`)
Resolves #3452 (`wrap_long_dict_values_in_parens`)
Slightly refactors `_maybe_split_omitting_optional_parens()` by moving some conditions for omitting optional parens into `_prefer_split_rhs_oop_over_rhs()` for simpler code logic
Fixes some typos in comments and error messages in `trans.py` that I found while debugging

I believe `wrap_long_dict_values_in_parens` is now ready to move from the unstable to preview style. Let me know if this is desirable in this PR.

### Checklist - did you ...

- [y] Add an entry in `CHANGES.md` if necessary?
- [y] Add / update tests if necessary?
- [-] Add new / update outdated documentation?

Issue #4158, opened 2024-01-20
wrap_long_dict_values_in_parens can remove parentheses


JelleZijlstra:
Given this input:

```python
x = {
    "xx_xxxxx_xxxxxxxxxx_xxxxxxxxx_xx": (
        "xx:xxxxxxxxxxxxxxxxx_xxxxx_xxxxxxx_xxxxxxxxxxx{xx}xxx_xxxxx_xxxxxxxxx_xxxxxxxxxxxx_xxxx"
    ),
}
```

The current draft new stable style outputs:

```python
x = {
    "xx_xxxxx_xxxxxxxxxx_xxxxxxxxx_xx": "xx:xxxxxxxxxxxxxxxxx_xxxxx_xxxxxxx_xxxxxxxxxxx{xx}xxx_xxxxx_xxxxxxxxx_xxxxxxxxxxxx_xxxx",
}
```

This is because the string is long enough that it doesn't fit within the line length even if it's wrapped in parentheses.

However, the new style feels worse: it's over the line length either way, but it's a lot *more* over the line length in the new formatting.

The new style is consistent with our formatting in a few other cases, however. For example, we also remove the parentheses in:

```python
xx_xxxxx_xxxxxxxxxx_xxxxxxxxx_xx = (
    xx_xxxxxxxxxxxxxxxxx_xxxxx_xxxxxxx_xxxxxxxxxxx_xx_xxx_xxxxx_xxxxxxxxx_xxxxxxxxxxxxxxxx
)
```

Is this bad enough to justify removing `wrap_long_dict_values_in_parens` from the new stable style? I'm not sure.

hauntsaninja:
Re: removing from 2024 stable style, there's also https://github.com/psf/black/issues/3452 (we've resolved #3758 and are likely to resolve #4129)

RedGuy12:
Interestingly enough, it looks like `string_processing` and `wrap_long_dict_values_in_parens` both individually trigger this behavior

Issue #3452, opened 2022-12-19
Suboptimal line split choices with `wrap_long_dict_values_in_parens`


tolomea:
**Describe the style change**

I'm looking at differences between 22.8.0 and 23.1a1 on our code base, most of it looks really good ❤️ as always thanks for the great work

however there is one pattern of change that doesn't seem ideal, it's possible this has been flagged already, I wasn't sure what to look for.

22.8.0
``` python
            common.models.DateTimeField: (
                datetime(2020, 1, 31, tzinfo=utc) + timedelta(days=i)
            ),
```

23.1a1
``` python
            common.models.DateTimeField: datetime(2020, 1, 31, tzinfo=utc) + timedelta(
                days=i
            ),
```

when a line needs breaking there must be some logic to decide where to break it, I guess it will find multiple candidates and rank them somehow, it looks like in this case it's started favouring the later break, whereas before it was favouring the structurally more significant one

tolomea:
I can confirm this doesn't happen in 22.12.0, so it's new in the 23 line

yilei:
This is caused by https://github.com/psf/black/pull/3440, agree it made this particular case worse. I can take a look once I have some free time.

JelleZijlstra:
Are you running with `--preview`? #3440 isn't slated for the 2023 stable style, so it's less urgent to fix.

Thanks @tolomea for reporting and @yilei for diagnosing.

tolomea:
re: --preview, yes we are, sorry I forgot to mention that, we like black so much we always want more

ichard26:
Just for future reference, please make it very clear which style your feedback is for. As you would imagine we care a lot more about feedback for the 2023 stable style and we'll assume any feedback on 23.1a1 to be about it unless noted.

I'll also make this more clear in the main thread and my blog post to avoid confusion :)

hauntsaninja:
[...]

PR #4561, opened 2025-01-26
Move `wrap_long_dict_values_in_parens` to the preview style

cobaltt7:
### Description

Moves `wrap_long_dict_values_in_parens` from the unstable to the preview style. Follows-up #4377, which closed both open issues about this feature (#3452 and #4158).

### Checklist - did you ...


docs/the_black_code_style/future_style.md
```diff
@@ -122,33 +121,6 @@ exceed the line length limit. Line continuation backslashes are converted into
 parenthesized strings. Unnecessary parentheses are stripped. The stability and status of
 this feature istracked in [this issue](https://github.com/psf/black/issues/2188).
 
-(labels/wrap-long-dict-values)=
```
JelleZijlstra, line 125: I don't think this should be removed. We happened to have these longer descriptions only for unstable features, but I think that was just by chance.

PR #4565, opened 2025-01-29
Fix hatch build

JelleZijlstra:


pyproject.toml
```diff
@@ -72,7 +72,8 @@ dependencies = [
   "tomli>=1.1.0; python_version < '3.11'",
   "typing_extensions>=4.0.1; python_version < '3.11'",
 ]
-dynamic = ["readme", "version"]
+version = "25.1.0"
```
JelleZijlstra, line 75: This should make it so we build wheels with the right tags. Once CI generates those, I can upload them to PyPI.

PR #4567, opened 2025-01-29
Fix mis-synced version check in black.vim

mgorny:
### Description

The message has been updated to indicate Python 3.9+, but the check still compared to 3.8.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,7 +20,7 @@
 
 ### Packaging
 
-<!-- Changes to how Black is packaged, such as dependency requirements -->
+- Fix the version check in the vim file to reject Python 3.8 (#4567)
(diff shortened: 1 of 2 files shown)


cobaltt7:
Thanks for the PR! Could you add a note describing the change to `CHANGES.md`?

mgorny:
Well, I don't mind doing that but it feels inconsistent, giving I'm only fixing #4473, and that change was done without a note.

JelleZijlstra:
Yes, please add a news entry since this is a user-visible change. I should have done that in the last PR but mistakenly thought all changes there were internal.

hauntsaninja (approved):

PR #4568, opened 2025-01-29
Re-add packaging CHANGES.md comment

MeGaGiGaGon:
### Description

Looks like #4567 replaced the comment with their entry, which bugs me, so this adds it back.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,8 @@
 
 ### Packaging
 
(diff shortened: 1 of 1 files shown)


hauntsaninja (commented):
While we're here, do you want to move the changelog entry to "Integrations" actually?

hauntsaninja (approved):

PR #4572, opened 2025-02-01
Fixed is_type_comment() to handle extra spaces in type comments (#2097)

Daiz-19:
### Description

This PR fixes the is_type_comment() function to correctly handle type comments with excessive whitespace. It normalizes spacing after # and :, ensuring that type comments are properly detected, even if they have extra spaces. This improves comment detection and prevents issues with misformatted type annotations.

### Checklist - did you ...

### DIFF
--- a/src/black/nodes.py
+++ b/src/black/nodes.py
@@ -1,7 +1,7 @@
 """
 blib2to3 Node/Leaf transformation-related utility functions.
 """
-
+import re
 import sys
 from collections.abc import Iterator
 from typing import Final, Generic, Literal, Optional, TypeVar, Union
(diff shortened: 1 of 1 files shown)


MeGaGiGaGon (changes requested):
Hello, thank you for your contribution to Black! I have a few things that need addressed before this is ready to be merged.
- Make sure you should read https://black.readthedocs.io/en/latest/contributing/the_basics.html . The next couple items of feedback all come from there, and it goes much more in-depth on exactly how and why to do things.
  - Since this is a user-facing change, you need to add a changelog entry
  - I'll admit I'm not familiar with this element of Black's formatting, so I couldn't figure out what this change actually does. Please add some new test cases to demonstrate that/make sure this doesn't get regressed in the future.
- Since you are fixing the issue, you can add it to the auto-close list by editing your opening description to include `Fixes #2097`
- In #2097 Jelle also said black should clean up the formatting, and it seems most people agreed with that, so it would also be nice if that was included. It looks like your current changes don't modify the comments.
  - This might require a change in approach though, since your current code would be overly-aggressive for reformatting if a comment isn't actually a type comment

PR #4576, opened 2025-02-06
Fix diff shades CI

MeGaGiGaGon:
### Description

Looks like upload-artifact@v3 is [now deprecated](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). Reading through the [migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) it looks like Black doesn't do any of the deprecated/changed behaviors, so this PR just bumps the number from 3 to 4.

### Checklist - did you ...

  - Not user facing, no changelog needed
  - Not nessacary, changes are to CI
  - Not nessacary, no behavior changed (hopefully)

### DIFF
--- a/.github/workflows/diff_shades.yml
+++ b/.github/workflows/diff_shades.yml
@@ -110,19 +110,19 @@ jobs:
           ${{ matrix.baseline-analysis }} ${{ matrix.target-analysis }}
 
       - name: Upload diff report
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.mode }}-diff.html
           path: diff.html
 
       - name: Upload baseline analysis
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.baseline-analysis }}
           path: ${{ matrix.baseline-analysis }}
 
       - name: Upload target analysis
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.target-analysis }}
           path: ${{ matrix.target-analysis }}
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Thanks! There's something wrong with the wheel build; seems unrelated but if you have time to look into it, hopefully not too bad to fix.

MeGaGiGaGon:
Looks like that did work to fix the diff_shades, but now the `mypyc` wheel build is failing. I think it might be spurious, or at least not my fualt, since the error is `Cannot download, all mirrors were already tried without success` from `sh -c 'yum install -y clang gcc'`, and I don't see why updating the upload-artifact version would affect it.

MeGaGiGaGon:
I tried looking at a previous run to see if it reports what url succeeded, but no dice. When I try to go to some of those URLs myself I also get 404s, so no clue what's happening, and extra not sure what to do since I'm on windows and very unfamiliar with anything linux.

MeGaGiGaGon:
@JelleZijlstra can you retry the `mypyc` build? I manually checked all the links it lists for the mirrors, and two of them work for downloading the file, so I think it might succeed if it re-runs. The two working links:
- http://mirror.facebook.net/almalinux/8.10/AppStream/x86_64/os/repodata/c35a17e12f1f9331776241e05fd568c487a7b4c8d0e8dc77527c313238408ae1-primary.xml.gz
- http://mirror.facebook.net/almalinux/8.10/AppStream/x86_64/os/repodata/31d0e735277fdcbc5e8cbb57177a827c3e9aa139e97a603980f6317165fdb4c9-modules.yaml.gz

JelleZijlstra:
Thanks, retried!

MeGaGiGaGon:
@JelleZijlstra it succeeded, so looks like it was just that one still working endpoint having a networking hiccup. I think this should be good for merge now that that's resolved.

PR #4577, opened 2025-02-07
mix_stderr parameter was removed from click 8.2.0

MeggyCal:
From click 8.2.0 changelog:

```
Keep stdout and stderr streams independent in CliRunner. Always collect stderr output and never raise an exception. Add a new output stream to simulate what the user sees in its terminal. Removes the mix_stderr parameter in CliRunner. [:issue:`2522`](https://github.com/pallets/click/blob/main/CHANGES.rst#id27) [:pr:`2523`](https://github.com/pallets/click/blob/main/CHANGES.rst#id29)
```

so `mix_stderr=False` throws an error and is redundant.

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -116,7 +116,7 @@ class BlackRunner(CliRunner):
     """Make sure STDOUT and STDERR are kept separate when testing Black via its CLI."""
 
     def __init__(self) -> None:
-        if Version(imp_version('click')) >= Version('8.2.0'):
+        if Version(imp_version("click")) >= Version("8.2.0"):
             super().__init__()
         else:
             super().__init__(mix_stderr=False)
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -26,8 +26,8 @@ import click
 import pytest
 from click import unstyle
 from click.testing import CliRunner
-from pathspec import PathSpec
 from packaging.version import Version
+from pathspec import PathSpec
 
 import black
 import black.files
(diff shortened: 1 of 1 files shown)


cobaltt7:
Also, wouldn't this change make the entire class redundant?

JelleZijlstra:
Yes, seems like we may have to make this conditional on Click version.

MeggyCal:
Ok, I will try to make it conditional :slightly_smiling_face:

MeggyCal:
sorry, I couldn't think of a more pythonic way

tests/test_black.py
```diff
@@ -14,6 +14,7 @@
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager, redirect_stderr
```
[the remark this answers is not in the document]
zbkash: Only one

JelleZijlstra (approved):

zbkash (approved):

zbkash (approved):

tests/test_black.py
```diff
@@ -114,7 +116,10 @@ class BlackRunner(CliRunner):
     """Make sure STDOUT and STDERR are kept separate when testing Black via its CLI."""
 
     def __init__(self) -> None:
-        super().__init__(mix_stderr=False)
+        if Version(imp_version('click')) >= Version('8.2.0'):
```
MeGaGiGaGon, line 119:
```suggestion
        if Version(imp_version("click")) >= Version("8.2.0"):
```
CI is unhappy because the quotes get reformatted to double quotes, but I think it should be gtg after this.

MeggyCal: ok

PR #4602, opened 2025-03-03
Fix parser for TypeVar bounds

### LINKED ISSUE #4599: Fails to parse lambda in PEP 695 TypeVar bound
**Describe the bug**

The following code can not be parsed/formatted by black:

``` python
class name_5[name_4: lambda: something]:
     pass
```

```
❯ black -t py312 bug.py
error: cannot format bug.py: Cannot parse for target version Python 3.12: 1:21: class name_5[name_4: lambda: something]:

Oh no! 💥 💔 💥
1 file failed to reformat.
```

but it can be parsed by cpython.

```
python3.12 scripts/min_code.py
```

**Environment**

- Black's version: current main (00c0d6d91ae3dc31894fadd37968ec0bb01e0d0d)
- OS and Python version: Linux/Python 3.12

**Additional context**

The bug was found by pysource-codegen (see #3908)

JelleZijlstra:
Fixes #4599

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,9 @@
 
 ### Parser
 
+- Fix bug where certain unusual expressions (e.g., lambdas) were not accepted
+  in type parameter bounds and defaults. (#4602)
+
 <!-- Changes to the parser or to version autodetection -->
 
 ### Performance
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,10 +24,10 @@
 
 ### Parser
 
+<!-- Changes to the parser or to version autodetection -->
+
 - Fix bug where certain unusual expressions (e.g., lambdas) were not accepted
   in type parameter bounds and defaults. (#4602)
-
-<!-- Changes to the parser or to version autodetection -->
 
 ### Performance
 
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -24,6 +24,9 @@
 
 ### Parser
 
+- Fix bug where certain unusual expressions (e.g., lambdas) were not accepted
+  in type parameter bounds and defaults. (#4602)
+
 <!-- Changes to the parser or to version autodetection -->
```
MeGaGiGaGon, line 30:
```suggestion
<!-- Changes to the parser or to version autodetection -->

- Fix bug where certain unusual expressions (e.g., lambdas) were not accepted
  in type parameter bounds and defaults. (#4602)
```
Slight nit, I just prefer this order.
Also the name test seems a bit sus, but I suppose its referring to something else instead of literal tests.
Slight nit, I just prefer this order

JelleZijlstra: I think the name "test" is because this grammar node is used in the "test" part of an if or while "if <test>: <then>".

hauntsaninja (approved):

PR #4606, opened 2025-03-04
update github-action to look for black version in "dependency-groups"

tsvikas:
### Description

"dependency-groups" is the mechanism for storing package requirements in `pyproject.toml`, recommended for formatting tools (see https://packaging.python.org/en/latest/specifications/dependency-groups/ )

this change allow the black action to look also in those locations when determining the version of black to install

### Checklist - did you ...

- [not needed] Add / update tests if necessary?

will be happy for help in those, or at least to see if my change is welcome before i start :)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,6 +49,8 @@
 <!-- For example, Docker, GitHub Actions, pre-commit, editors -->
 
 - Fix the version check in the vim file to reject Python 3.8 (#4567)
+- Enhance GitHub Action `psf/black` to read Black version from an additional
+  section in pyproject.toml: `[project.dependency-groups]` (#4606)
 
 ### Documentation
 
--- a/action/main.py
+++ b/action/main.py
@@ -71,9 +71,9 @@ def read_version_specifier_from_pyproject() -> str:
         return f"=={version}"
 
     arrays = [
+        *pyproject.get("dependency-groups", {}).values(),
         pyproject.get("project", {}).get("dependencies"),
         *pyproject.get("project", {}).get("optional-dependencies", {}).values(),
-        *pyproject.get("dependency-groups", {}).values(),
     ]
     for array in arrays:
         version = find_black_version_in_array(array)
--- a/action/main.py
+++ b/action/main.py
@@ -73,6 +73,7 @@ def read_version_specifier_from_pyproject() -> str:
     arrays = [
         pyproject.get("project", {}).get("dependencies"),
         *pyproject.get("project", {}).get("optional-dependencies", {}).values(),
+        *pyproject.get("dependency-groups", {}).values(),
     ]
     for array in arrays:
         version = find_black_version_in_array(array)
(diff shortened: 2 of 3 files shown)


MeGaGiGaGon:
I added skip news based on vibes/limited understanding, I don’t completely understand what this PR does so might be wrong.

tsvikas:
Thanks. I'll elaborate.  

Black has a GitHub Actions integration:  
<https://black.readthedocs.io/en/stable/integrations/github_actions.html>  

In that action, users can specify the Black version to run.  
Alternatively, they can set `use_pyproject: true`, which makes the action use the version specified in the user's `pyproject.toml`.  
This is preferable as it provides a single source of truth for the version for both local and CI runs.  

Until now, using `use_pyproject: true` made the action check two sections in `pyproject.toml`:  

- `project.dependencies`  
- `project.optional-dependencies.*`  

This PR adds a third location:  

- `project.dependency-groups.*`  

According to [Python Packaging User Guide](https://packaging.python.org/en/latest/specifications/dependency-groups/),  
this is the recommended place to specify the Black version.  

- It might be useful to inform users about this change since it expands compatibility with different setups, so Adding it to the changelog could be a good idea.  

- The documentation doesn’t specify where the action looks, so no updates are needed

- I haven't looked at tests.

MeGaGiGaGon:
Ah, that makes sense, thanks for the in-depth explanation. Then I think a changelog entry would make sense, and I think it would also be nice if you could add a small section to the end of the github action docs that says Black searches those three locations, since that sounds like some nice additional clarity to have.

I don't think there are tests for the action, at least I couldn't find them, but I could be wrong.

I also read though the docs on dependency-groups again, and it looks like you could have multiple different versions of black specified? This is probably an additional scope creep for this PR, but I looked though the code and it seems like Black just returns the first one it finds, which feels suboptimal to me, especially with the docs saying "Dependency Group Includes may specify the same package multiple times". I think that behavior should at least be noted in the github action docs, or if a different resolution strategy makes sense to you changed, then noted.

tsvikas:
> I also read though the docs on dependency-groups again, and it looks like you could have multiple different versions of black specified? This is probably an additional scope creep for this PR, but I looked though the code and it seems like Black just returns the first one it finds, which feels suboptimal to me, especially with the docs saying "Dependency Group Includes may specify the same package multiple times". I think that behavior should at least be noted in the github action docs, or if a different resolution strategy makes sense to you changed, then noted.

I'm not sure what the most common or reasonable use case would be for specifying Black multiple times. Given that, I suggest leaving the behavior intentionally undefined for now.

MeGaGiGaGon:
Ah I missed that section of the docs, then lgtm

MeGaGiGaGon (approved):

tsvikas:
Hey @MeGaGiGaGon, since this PR has been approved and all checks have passed, is there anything else needed before merging? Let me know if I should ping someone else. Thanks!

JelleZijlstra (approved):

PR #4611, opened 2025-03-12
Upgrade PyPI upload workflow to use Trusted Publishing (#4589)

### LINKED ISSUE #4589: [task] Upgrade `.github/workflows/pypi_upload.yml` to use Trusted Publishing
any reason you didn't migrate to Trusted Publishing? It can produce digital attestations out of the box now: https://blog.pypi.org/posts/2024-11-14-pypi-now-supports-digital-attestations/.

_Originally posted by @webknjaz in https://github.com/psf/black/issues/4512#issuecomment-2483427381_
            

              We should move to Trusted Publishing at some point, but that's a bit more work, so we'll do it when we do it I suppose. I'd probably accept a PR that does it for us.

_Originally posted by @JelleZijlstra in https://github.com/psf/black/issues/4512#issuecomment-2485528139_

The full guide is @ https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/.

Example of CI/CD setup that uploads individual dists as GHA workflow artifacts in cibuildwheel (and pure-python fallback) jobs with unique artifact names, and then downloads them in the publishing job: https://github.com/aio-libs/yarl/blob/426b7ac/.github/workflows/ci-cd.yml#L112-L124 + https://github.com/aio-libs/yarl/blob/426b7ac/.github/workflows/reusable-build-wheel.yml#L93-L100 + https://github.com/aio-libs/yarl/blob/426b7ac/.github/workflows/ci-cd.yml#L553-L558.

Roughly, this is what needs to be done:
* add uploading artifacts into each job producing them
* ensure that the artifact names are unique per single workflow run
* add downloading all of them and merging contents of multiple artifacts into a single dir, into the PyPI upload job
* Change `twine upload` to invoking `pypi-publish`
* Add OIDC permission to the PyPI job
* Add an environment called `pypi` to the job and set up required reviewers for said environment in the repo settings
* Get somebody with the Owner-level permissions to configure trust on the PyPI side

ahmed5145:
Fixes #4589

This PR upgrades the PyPI publishing workflow to use Trusted Publishing instead of token-based authentication. This change:
- Improves security by using OpenID Connect (OIDC) instead of long-lived tokens
- Removes the need for maintaining PyPI tokens in GitHub secrets
- Uses the official PyPA publishing action

Changes made:
- Added `id-token: write` permission for OIDC authentication
- Changed environment name to 'release'
- Switched to `pypa/gh-action-pypi-publish` action
- Removed twine dependency

Required actions after merging:
The repository maintainers will need to:
1. Configure Trusted Publishing in PyPI for the black project:
   - Go to https://pypi.org/manage/account/publishing/
   - Configure with:
     - Owner: psf
     - Repository: black
     - Workflow name: Build and publish
     - Environment: release
2. Create a 'release' environment in the GitHub repository settings
3. Remove the existing PyPI token from GitHub secrets (after verifying the new setup works)

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -10,12 +10,16 @@ on:
 
 permissions:
   contents: read
+  id-token: write  # Required for PyPI trusted publishing
 
 jobs:
   main:
     name: sdist + pure wheel
     runs-on: ubuntu-latest
     if: github.event_name == 'release'
+    environment:
+      name: release
+      url: https://pypi.org/p/black
 
     steps:
       - uses: actions/checkout@v4
@@ -26,19 +30,19 @@ jobs:
           python-version: "3.13"
           allow-prereleases: true
 
-      - name: Install latest pip, build, twine
+      - name: Install latest pip, build
         run: |
           python -m pip install --upgrade --disable-pip-version-check pip
-          python -m pip install --upgrade build twine
+          python -m pip install --upgrade build
 
       - name: Build wheel and source distributions
         run: python -m build
 
       - if: github.event_name == 'release'
-        name: Upload to PyPI via Twine
-        env:
-          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
-        run: twine upload --verbose -u '__token__' dist/*
+        name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          verbose: true
 
   generate_wheels_matrix:
     name: generate wheels matrix
@@ -84,6 +88,10 @@ jobs:
     name: mypyc wheels ${{ matrix.only }}
     needs: generate_wheels_matrix
     runs-on: ${{ matrix.os }}
+    if: github.event_name == 'release'
+    environment:
+      name: release
+      url: https://pypi.org/p/black
     strategy:
       fail-fast: false
       matrix:
@@ -103,10 +111,11 @@ jobs:
           path: ./wheelhouse/*.whl
 
       - if: github.event_name == 'release'
-        name: Upload wheels to PyPI via Twine
-        env:
-          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
-        run: pipx run twine upload --verbose -u '__token__' wheelhouse/*.whl
+        name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: wheelhouse/
+          verbose: true
 
   update-stable-branch:
     name: Update stable branch


cooperlees:
Cool - TIL there is something more official I long ago cooked up at a GitHub conference when Actions were announced. Let's make CI happy and this feels the right way to go to me. I'll be copying for my other projects too :)

ahmed5145:
Happy to hear that, please feel free to use across. It's quite impressive that you were there during the announcement of Actions.

It seems that there is an issue with the CI workflow in the PR. The diff-shades-comment job is failing because it cannot find the .pr-comment.json artifact. This artifact is expected by the comment-details command in the diff_shades_gha_helper.py script.

I've reviewed the script and it seems like the artifact is not being generated or uploaded in the diff-shades workflow. 
Could you provide guidance on how this artifact should be generated? Are there any known issues with the current CI setup that might be causing this?
Or if I am getting this completely wrong.
Thank you for your assistance!

JelleZijlstra:
This seems fine but it will need some changes on the PyPI side; I'll do that when I have some time. Not sure if @cooperlees has the requisite permissions.

.github/workflows/pypi_upload.yml
```diff
@@ -26,19 +30,19 @@ jobs:
...
 
-      - name: Install latest pip, build, twine
+      - name: Install latest pip, build
         run: |
           python -m pip install --upgrade --disable-pip-version-check pip
-          python -m pip install --upgrade build twine
+          python -m pip install --upgrade build
 
       - name: Build wheel and source distributions
         run: python -m build
```
webknjaz, line 41: Please avoid building in the same job as publishing. Especially with the elevated OIDC privileges. The transitive build deps shouldn't have such level of access.

#4589 links to proper examples.

.github/workflows/pypi_upload.yml
```diff
@@ -10,12 +10,16 @@ on:
...
+  id-token: write # Required for PyPI trusted publishing
 
 jobs:
   main:
     name: sdist + pure wheel
     runs-on: ubuntu-latest
     if: github.event_name == 'release'
+    environment:
+      name: release
+      url: https://pypi.org/p/black
```
woodruffw, line 22: Noting for the curious: the `environment` itself doesn't grant any special permissions to this job -- what it does is define an environment on GitHub's side, which can then be used to *further* constrain the publishing workflow (e.g. by adding required manual signoffs whenever the release workflow is triggered).

The PyPI documentation has some details on that: <https://docs.pypi.org/trusted-publishers/security-model/#provider-specific-considerations>

.github/workflows/pypi_upload.yml
cooperlees: Good call - but should we do that in a dedicated PR? We already have this setup today ...

cooperlees:
image
Turns out I had access - So I added as prompted, but it seems it has to be the .yml file name of the workflow ...

Is this ready to be merged now if I resolve conflicts?

.github/workflows/pypi_upload.yml
woodruffw: Yeah, I think there's potentially a larger CI/CD cleanup PR (or PRs) that could be done here. I can help chip away at that with some PRs 🙂

woodruffw:
> So I added as prompted, but it seems it has to be the .yml file name of the workflow ...
Is this ready to be merged now if I resolve conflicts?

Yep, it should be the filename of the workflow that performs the publishing operation 🙂

cobaltt7:
I didn't add a new blank changelog after releasing 25.12 because the next release is 26.1, which is WIP in #4892. Ideally we want to keep that mostly just formatting changes, and to be conservative when considering other PRs.
However, this should be fine since it doesn't actually change Black itself, the new changelog just needs to be added still (`python scripts/release.py -a`)

.github/workflows/pypi_upload.yml
webknjaz: @cooperlees the old setup doesn't give elevated privileges to the transitive build deps, this is quite coupled.

.github/workflows/pypi_upload.yml
```diff
@@ -91,6 +95,10 @@ jobs:
     name: mypyc wheels ${{ matrix.only }}
     needs: generate_wheels_matrix
     runs-on: ${{ matrix.os }}
+    if: github.event_name == 'release'
```
cobaltt7, line 98: Why do we want to gate the entire job to releases-only? The step is already release-gated. We want to build wheels on PRs to test and make sure builds don't fail.

webknjaz: Because building must not be happening in a job with such high privileges (once Trusted Publishing is set up) — the jobs should actually be made separate.

cobaltt7: @webknjaz I believe that's an unrelated change. We already build and publish in the same job. The build deps now have the elevated permissions -- which yes, should be fixed -- but if I'm not mistaken, they have those permissions regardless of if the workflow runs on PR or release. So, ignoring that issue, this PR removed the functionality of testing wheel builds on PRs. Is there a reason for that?

PR #4630, opened 2025-03-20
Fix parsing of walrus operator in complex with statements

### LINKED ISSUE #4629: failed to format a file with a with-statement and a named-expression
**Describe the bug**

The following code can not be parsed/formatted by black:

``` python
with {} as name_3, (name_2 := name_4):

    def name_3[**name_1]():
        pass
```

black reported the following error:
```
> black -l 100 -C bug.py
error: cannot format bug.py: Cannot parse: 1:26: with {} as name_3, name_2 := name_4:

Oh no! 💥 💔 💥
1 file failed to reformat.

```

but it can be parsed by cpython:
``` python
from ast import parse
parse(
    'with {} as name_3, (name_2 := name_4):\n'
    '\n'
    '    def name_3[**name_1]():\n'
    '        pass\n'
)

```
result:
```

returncode: 0
```

**Environment**

- Black's version: current main (dbb14eac93637cca7f9b88685a56c0b381e42caa)
- OS and Python version: Linux/Python 3.12.6 (main, Sep  9 2024, 22:11:19) [Clang 18.1.8 ]

**Additional context**

The bug was found by pysource-codegen (see #3908)
The issue description above was created by a script. Let me know if you think it can be improved.
One think I noted is that the following code causes also a formatting failure:
``` python
with a, (name_2 := name_4):
        pass
```
But pysource-minimize did not found this minimal example for some reason.

tusharsadhwani:
Resolves #4629 

### Description

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,8 @@
 
 <!-- Changes that affect Black's stable style -->
 - Fix crash while formatting a long `del` statement containing tuples (#4628)
-- Fix crash while formatting names expressions in complex `with` statements (#4630)
+- Fix crash while formatting expressions with walrus operator in complex `with`
+  statements (#4630)
 
 ### Preview style
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 
 <!-- Changes that affect Black's stable style -->
 - Fix crash while formatting a long `del` statement containing tuples (#4628)
+- Fix crash while formatting names expressions in complex `with` statements (#4630)
 
 ### Preview style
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -10,6 +10,7 @@
 
 <!-- Changes that affect Black's stable style -->
 - Fix crash while formatting a long `del` statement containing tuples (#4628)
+- Fix crash while formatting names expressions in complex `with` statements (#4630)
```
```suggestion
- Fix crash while formatting expressions using the walrus operator in complex `with` statements (#4630)
```
I think you meant "named" instead of "names" expressions. That's the formal name but I think "walrus" will be more recognizable to most readers. (Might need reformatting for line length)

hauntsaninja (approved):
Thanks!

PR #4635, opened 2025-03-21
Handle `# fmt: skip` followed by a comment

### LINKED ISSUE #4631: Internal Error when formatting a file with comments containing fmt: skip
**Describe the bug**
Internal Error when reformatting file with comments at the end, where the first line has `  # fmt: skip` added to it. All additional comment lines are removed.

**To Reproduce**

For example, take this code:

```python
def foo():
    pass


# comment1  # fmt: skip
# comment2
```

```sh
$ black file.py
```

The resulting error is:

> cannot format file.py: INTERNAL ERROR: Black 25.1.1.dev21+gdbb14ea on Python (CPython) 3.12.8 produced different code on the second pass of the formatter.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_5w5nafph.log
```
Mode(target_versions=set(), line_length=88, string_normalization=True, is_pyi=False, is_ipynb=False, skip_source_first_line=False, magic_trailing_comma=True, python_cell_magics=set(), preview=False, unstable=False, enabled_features=set())
--- source
+++ first pass
@@ -1,6 +1,3 @@
-def foo():
-    pass
-
-
-# comment1  # fmt: skip
-# comment2
\ No newline at end of file
+def foo():
+    pass
+# comment1  # fmt: skip
--- first pass
+++ second pass
@@ -1,3 +1,5 @@
 def foo():
     pass
+
+
 # comment1  # fmt: skip

```

**Expected behavior**

It should leave the lines of comments untouched.

**Environment**

My environment:
- Black's version: 25.1.0
- OS and Python version: Windows / Python (CPython) 3.10.6

This also happens on the Black Web Playground:
- Black's version: 25.1.1.dev21+gdbb14ea
- OS and Python version: Black Web Playground / Python (CPython) 3.12.8

**Additional context**
Originally, this was done with commented-out code (that's why `  # fmt: skip` was in a comment), but it happens with any comment. The only thing that seems to matter is if the skip tag is in the first comment line of a block, and the block of comment lines is at the end of the file. If the skip tag is on the second line of comments, or if there is additional python code after the comment block, the format succeeds and works as expected.

tusharsadhwani:
### Description

Resolves #4631 

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 - Fix crash while formatting a long `del` statement containing tuples (#4628)
 - Fix crash while formatting expressions using the walrus operator in complex
   `with` statements (#4630)
+- Handle `# fmt: skip` followed by a comment at the end of file (#4635)
 
 ### Preview style
 
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -321,7 +321,7 @@ def _generate_ignored_nodes_from_fmt_skip(
     if not comments or comment.value != comments[0].value:
         return
     if prev_sibling is not None:
-        leaf.prefix = ""
+        leaf.prefix = leaf.prefix[comment.consumed :]
 
         if Preview.fix_fmt_skip_in_one_liners not in mode:
             siblings = [prev_sibling]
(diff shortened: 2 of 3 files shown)


Anupam9142:
Done

JelleZijlstra (approved):

Issue #1333, opened 2020-04-03
Triple quote strings as function args are formatted badly


Mattwmaster58:
At the very least, I think black should keep the argument call paren on the same line. Ideally the opening triple quote and opening paren would be on the same line as well.

**Examples in the current _Black_ style**:

```py
            await isolated_page.evaluate(
                '''() => {
                history.pushState({}, '', '/first.html');
                history.pushState({}, '', '/second.html');
            }'''
            )
```

**Desired style**
```py
            await isolated_page.evaluate('''() => {
                history.pushState({}, '', '/first.html');
                history.pushState({}, '', '/second.html');
            }''')
```

Not sure if this is related to #256

yryuvraj:
This was fixed by the `multiline_string_handling` style change from #1879, which was stabilized in the 2026 stable style (#4892 / Black 26.1.0).

With current Black, the example from this issue formats as desired:

```python
await isolated_page.evaluate("""() => {
    history.pushState({}, '', '/first.html');
    history.pushState({}, '', '/second.html');
}""")
```

(Quotes are normalized to `"""` per Black's usual string normalization.)

PR #4638, opened 2025-03-23
format_function_call(): handles triple-quoted string args

yryuvraj:
### Description
Handles Issue : #1333 

### Checklist - did you ...

- [ ] Add an entry in `CHANGES.md` if necessary? -- Not Necessary

### DIFF
--- a/src/black/strings.py
+++ b/src/black/strings.py
@@ -377,8 +377,7 @@ def str_width(line_str: str) -> int:
 
 def count_chars_in_width(line_str: str, max_width: int) -> int:
     """Count the number of characters in `line_str` that would fit in a
-    terminal or editor of `max_width` (which respects Unicode East Asian
-    Width).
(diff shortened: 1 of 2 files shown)


MeGaGiGaGon (changes requested):
Unfortunately this doesn’t look to fix the issue, the defined function is never used to actually format code. The proper way to test this is adding a new test case to a file in the `tests/data/cases` folder. The fix should probably happen somewhere in the `src/black/linegen.py` file, and it would involve manipulating the produced ast instead of using regex.

hauntsaninja:
You'll also want the test case code to be Python, not Javascript. Feel free to reopen the PR if you have something that actually fixes the issue, like MeGaGiGaGon says!

yryuvraj:
Sorry @hauntsaninja @MeGaGiGaGon , I'll try and fix it! Thanks for suggesting fixes.

PR #4645, opened 2025-04-07
Standardize type comments to form `# type: (value)` (#2097)

ranjodhsingh1729:
### Description
**Fixes**: [#2097 – `is_type_comment()` is wrong](https://github.com/psf/black/issues/2097)

**Summary of Changes**:
- Modified function signatures across several modules to pass the `mode` parameter to `is_type_comment` and `make_comment`.
- Incorporated parts of PR #4467 by @Pedro-Muller29, particularly the fix in `mode.py` addressing the `unhashable type: mode` issue.

**Acknowledgment**:
Credit to @Pedro-Muller29 for the initial work in PR #4467.

### Checklist - did you ...

### DIFF
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -11,9 +11,9 @@ from black.nodes import (
     WHITESPACE,
     container_of,
     first_leaf_of,
+    is_type_comment_string,
     make_simple_prefix,
     preceding_leaf,
-    is_type_comment_string,
     syms,
 )
 from blib2to3.pgen2 import token
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -13,6 +13,7 @@ from black.nodes import (
     first_leaf_of,
     make_simple_prefix,
     preceding_leaf,
+    is_type_comment_string,
     syms,
 )
 from blib2to3.pgen2 import token
@@ -50,7 +51,7 @@ class ProtoComment:
     leading_whitespace: str  # leading whitespace before the comment, if any
 
 
-def generate_comments(leaf: LN) -> Iterator[Leaf]:
+def generate_comments(leaf: LN, mode: Mode) -> Iterator[Leaf]:
     """Clean the prefix of the `leaf` and generate comments from it, if any.
 
     Comments in lib2to3 are shoved into the whitespace prefix.  This happens
@@ -70,7 +71,9 @@ def generate_comments(leaf: LN) -> Iterator[Leaf]:
     are emitted with a fake STANDALONE_COMMENT token identifier.
     """
     total_consumed = 0
-    for pc in list_comments(leaf.prefix, is_endmarker=leaf.type == token.ENDMARKER):
+    for pc in list_comments(
+        leaf.prefix, is_endmarker=leaf.type == token.ENDMARKER, mode=mode
+    ):
         total_consumed = pc.consumed
         prefix = make_simple_prefix(pc.newlines, pc.form_feed)
         yield Leaf(pc.type, pc.value, prefix=prefix)
@@ -78,7 +81,7 @@ def generate_comments(leaf: LN) -> Iterator[Leaf]:
 
 
 @lru_cache(maxsize=4096)
-def list_comments(prefix: str, *, is_endmarker: bool) -> list[ProtoComment]:
+def list_comments(prefix: str, *, is_endmarker: bool, mode: Mode) -> list[ProtoComment]:
     """Return a list of :class:`ProtoComment` objects parsed from the given `prefix`."""
     result: list[ProtoComment] = []
     if not prefix or "#" not in prefix:
@@ -109,7 +112,7 @@ def list_comments(prefix: str, *, is_endmarker: bool) -> list[ProtoComment]:
             comment_type = token.COMMENT  # simple trailing comment
         else:
             comment_type = STANDALONE_COMMENT
-        comment = make_comment(line)
+        comment = make_comment(line, mode=mode)
         result.append(
             ProtoComment(
                 type=comment_type,
@@ -140,7 +143,7 @@ def normalize_trailing_prefix(leaf: LN, total_consumed: int) -> None:
     leaf.prefix = ""
 
 
-def make_comment(content: str) -> str:
+def make_comment(content: str, mode: Mode) -> str:
     """Return a consistently formatted comment from the given `content` string.
 
     All comments (except for "##", "#!", "#:", '#'") should have a single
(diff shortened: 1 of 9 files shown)


tusharsadhwani:
@ranjodhsingh1729 any reasons for closing the PR? It's fairly large so I haven't found time to review it yet.

ranjodhsingh1729:
@tusharsadhwani My Bad. I’m new to contributing and assumed I’d made an error when I saw no activity.

src/black/comments.py
```diff
@@ -157,9 +160,18 @@ def make_comment(content: str) -> str:
...
         and content[0] == "\N{NO-BREAK SPACE}"
-        and not content.lstrip().startswith("type:")
+        and not is_type_comment_string("# " + content.lstrip(), mode=mode)
     ):
         content = " " + content[1:]  # Replace NBSP by a simple space
+    if (
+        Preview.standardize_type_comments in mode
+        and content
+        and "\N{NO-BREAK SPACE}" not in content
+        and is_type_comment_string("# " + content.lstrip(), mode=mode)
```
cobaltt7:
```suggestion
        and is_type_comment_string("# " + content, mode=mode)
```

Just for the sake of conciseness - `.lstrip()` is already run within `is_type_comment_string()`

I think it makes sense to keep it as-is on L163, since there we might still be on stable style and can't rely on `is_type_comment_string()` to strip it.

src/black/comments.py
```diff
@@ -157,9 +160,18 @@ def make_comment(content: str) -> str:
...
+    ):
+        type_part, value_part = content.strip().split(":", 1)
```
cobaltt7:
```suggestion
        type_part, value_part = content.split(":", 1)
```
The `.strip()` here is unnecessary as both parts will be stripped individually

src/black/mode.py
```diff
@@ -285,3 +286,18 @@ def get_cache_key(self) -> str:
             features_and_magics,
         ]
         return ".".join(parts)
+
+    def __hash__(self) -> int:
```
cobaltt7, line 324: Why is this necessary in the context of this PR? And as Jelle asked on #4467, could we instead make dataclass generate the hash method?

cobaltt7 (commented):
Sorry for the delay in review. Looks pretty good! Just a few small things

src/black/mode.py
ranjodhsingh1729: In [src/black/comments.py#L84](https://github.com/ranjodhsingh1729/black/blob/b7abcd43774514a50e14c609b6584d865c804742/src/black/comments.py#L84), adding mode as a parameter to an @lru_cache-decorated function caused it to be hashed — which is why Mode needed to be hashable.

I initially tried converting the sets in Mode to frozensets, but if a regular set was passed in, it still raised a TypeError because the instance wasn’t fully hashable. So to avoid a more intrusive change, I went with a custom `__hash__` implementation, similar to what was done in #4467.

While revisiting this today, I came across [this StackOverflow thread](https://stackoverflow.com/questions/51199031/python-3-dataclass-initialization), which gave me the idea of using `__post_init__` along with `object.__setattr__` to convert passed-in sets to frozensets before any hashing occurs. (I have tested this and it works)

If you're okay with this approach, I can make the necessary changes and also address the extra strip() calls you pointed out.

tusharsadhwani: It would be easier to remove `lru_cache` from `make_comment`, and either add manual coaching or create an internal function that doesn't take `mode` within that function and put the lru_cache on that instead.

I don't think that method cares about every aspect of `mode` when it's doing the caching.

ranjodhsingh1729: I'm good with either approach. That said, updating the `Mode` class does feel like a more future-proof solution—if we ever need to pass mode into a cached function later on, we likely wouldn’t need to change anything further.

That does make me curious though—why isn’t `mode` global? I’m sure there’s a good reason; it just hasn’t clicked for me yet.

cobaltt7: IMO adding `__hash__` is the best solution for being more future-proof and less intrusive.

ranjodhsingh1729: This is how `Mode` would be modified :) if we use dataclass to generate hash:-

```
-@dataclass
+@dataclass(frozen=True)
 class Mode:
-    target_versions: set[TargetVersion] = field(default_factory=set)
+    target_versions: frozenset[TargetVersion] = field(default_factory=frozenset)
     line_length: int = DEFAULT_LINE_LENGTH
     string_normalization: bool = True
     is_pyi: bool = False
     is_ipynb: bool = False
     skip_source_first_line: bool = False
     magic_trailing_comma: bool = True
-    python_cell_magics: set[str] = field(default_factory=set)
+    python_cell_magics: frozenset[str] = field(default_factory=frozenset)
     preview: bool = False
     unstable: bool = False
-    enabled_features: set[Preview] = field(default_factory=set)
+    enabled_features: frozenset[Preview] = field(default_factory=frozenset)
+
+    def __post_init__(self):
+        object.__setattr__(self, 'target_versions', frozenset(self.target_versions))
+        object.__setattr__(self, 'python_cell_magics', frozenset(self.python_cell_magics))
+        object.__setattr__(self, 'enabled_features', frozenset(self.enabled_features))
```

Rest of the class remains as is.

ranjodhsingh1729: @tusharsadhwani does this capture what you intended?

```
def list_comments(prefix: str, *, is_endmarker: bool, mode: Mode) -> list[ProtoComment]:
    @lru_cache(maxsize=4096)
    def __list_comments(prefix: str, *, is_endmarker: bool) -> list[ProtoComment]
        """Return a list of :class:`ProtoComment` objects parsed from the given `prefix`."""
        result: list[ProtoComment] = []
        if not prefix or "#" not in prefix:
            return result

        consumed = 0
        nlines = 0
        ignored_lines = 0
        form_feed = False
        for index, full_line in enumerate(re.split("\r?\n|\r", prefix)):
            consumed += len(full_line) + 1  # adding the length of the split '\n'
            match = re.match(r"^(\s*)(\S.*|)$", full_line)
            assert match
            whitespace, line = match.groups()
            if not line:
                nlines += 1
                if "\f" in full_line:
                    form_feed = True
            if not line.startswith("#"):
                # Escaped newlines outside of a comment are not really newlines at
                # all. We treat a single-line comment following an escaped newline
                # as a simple trailing comment.
                if line.endswith("\\"):
                    ignored_lines += 1
                continue

            if index == ignored_lines and not is_endmarker:
                comment_type = token.COMMENT  # simple trailing comment
            else:
                comment_type = STANDALONE_COMMENT
            comment = make_comment(line, mode=mode)
            result.append(
                ProtoComment(
                    type=comment_type,
                    value=comment,
                    newlines=nlines,
                    consumed=consumed,
                    form_feed=form_feed,
                    leading_whitespace=whitespace,
                )
            )
            form_feed = False
            nlines = 0
        return result

    return __list_comments(prefix, is_endmarker=is_endmarker)
```

cobaltt7: That also seems like a good approach - @tusharsadhwani what do you think?

ranjodhsingh1729: Hi @tusharsadhwani

I meant to ask this earlier but totally forgot 😅 —  this doesn't warrant a immediate response, just something that crossed my mind.

If we put `@lru_cache` on an inner function that doesn’t get `mode` passed explicitly and instead relies on it from the enclosing scope, wouldn’t that cause issues if `mode` needed to change in the future? I know that currently `mode` doesn’t change mid-execution, but if it ever did, wouldn’t the cached function just keep returning the same result, ignoring the updated `mode`?

It seems like a sensible assumption to me that `mode` wouldn’t need to change mid-execution, so would it maybe make more sense to just freeze it instead?

I may be missing something. What do you think?

cobaltt7: This is good as-is

PR #4652, opened 2025-04-22
Fix #2097: Recognize type comments with spaces after #

### LINKED ISSUE #2097: is_type_comment() is wrong

anthonychen000:
Fixes #2097

### Description

Currently, Black fails to recognize `# type:` and `# type: ignore` comments if there's extra whitespace after the `#` (e.g., `#    type:`), as described in issue #2097. This can lead to these comments being treated as regular comments, potentially causing incorrect formatting or instability.

This PR updates the comment recognition logic (`is_type_comment`, `is_type_ignore_comment`, and helper functions like `is_type_comment_string` / `is_type_ignore_comment_string`) to correctly handle this leading whitespace after the `#`. It allows comments like `#    type: value` and `#    type: ignore` to be correctly identified.

As discussed in the issue thread, this change focuses *only* on fixing the **recognition** part of the bug and does **not** implement comment **standardization** (i.e., it doesn't reformat `#    type:` to `# type:`).

Additionally, this PR includes a small fix for unrelated `mypy` errors (`[import-untyped, unreachable]`) encountered in `action/main.py` which were required to pass local pre-commit checks.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@
 - Handle `# fmt: skip` followed by a comment at the end of file (#4635)
 - Fix crash when a tuple appears in the `as` clause of a `with` statement
   (#4634)
+- Fix crash when tuple is used as a context manager inside a `with` statement (#4646)
+- Recognize type comments and ignore comments with leading whitespace after `#` (#2097)
 
 ### Preview style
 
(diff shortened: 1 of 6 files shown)


cobaltt7:
I'm closing this PR since #4645 seems more comprehensive. Additionally, this PR seems to have some unrelated changes included - please make sure to only include the relevant changes in the future.

PR #4658, opened 2025-04-26
Fix bug with standalone comments in lambda default arguments

### LINKED ISSUE #4640: Standalone comment within parentheses in lambda default arguments crashes black
**Describe the bug**

Given the following code snippet:
```python
help(lambda x=(
    # comment
    "bar",
): False)
```
black crashes within the reformatting code with the error message:
```
error: cannot format <string>: Opening paren not found in `leaves`
```

I reproduced this using the current head of black (`black, 25.1.1.dev24+g2c135ed (compiled: no)`) and on the online formatter. I can see this on both Linux and macOS and multiple versions of Python (3.12 and 3.13).

I bisected the repro, and it looks like it started breaking in #4016.

Some other things I observed:
- Using other bracket groups (`[]` and `{}`) do not show this issue
- The magic trailing comma in the argument is important
- If the comment comes at the _end_, it's fine. This does not crash:
```python
help(lambda x=(
    "bar",
    # comment
): False)
```
- Expanding on the above, comments in the middle of arguments also cause the issue:
```python
help(lambda x=(
    "extremely lengthy argument 1 extremely lengthy argument 1",
    # comment
    "extremely lengthy argument 2 extremely lengthy argument 2",
): False)
```

**To Reproduce**
```
$ black --diff -v -c '''help(lambda x=(
    # comment
    "bar",
): False)
'''
```

siavasha:
### Description

Fixes #4640

This PR fixes a bug where Black would crash when formatting code with standalone comments within parentheses in lambda default arguments.

The issue occurs in the `has_magic_trailing_comma` method in `src/black/lines.py`. When a standalone comment is present within parentheses in a lambda default argument, the method tries to find the opening bracket in the line's leaves, but it wouldn't be there, causing a LookupError. 

The fix adds a try-except block to catch this LookupError and return True, which causes the line to be split appropriately. This approach is consistent with how Black handles other similar cases where it needs to determine whether to add a trailing comma.

Example code that previously crashed:
```python
help(lambda x=(
     # comment
     "bar",
): False)
```

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
 - Fix crash when a tuple appears in the `as` clause of a `with` statement
   (#4634)
 - Fix crash when tuple is used as a context manager inside a `with` statement (#4646)
+- Fix crash when standalone comment is within parentheses in lambda default arguments (#4640)
 
 ### Preview style
 
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -370,9 +370,21 @@ class Line:
         if self.is_import:
             return True
 
-        if closing.opening_bracket is not None and not is_one_sequence_between(
-            closing.opening_bracket, closing, self.leaves
-        ):
+        # Check if there are any standalone comments in the line
+        # If so, we need to split the line, so return True
+        for leaf in self.leaves:
+            if leaf.type == STANDALONE_COMMENT:
+                return True
+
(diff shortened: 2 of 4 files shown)


cobaltt7:
This PR introduced a crash when running on https://github.com/quora/pyanalyze/blob/master/pyanalyze%2Ftest_format_strings.py

PR #4662, opened 2025-05-04
Update Prettier pre-commit configuration

cobaltt7:
### Description

Updated the Prettier configuration and re-ran Prettier.
The pre-commit hook we were using has been archived, so I switched to an active and decently popular (600+ repo) fork. We were previously on an alpha version, and the fork doesn't support alpha versions, so I did have to downgrade the version. However, no changes in v4 affected us, and v3 is more freuently updated.
I also modified the file type list to include Markdown and remove language this project does not use.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/diff_shades.yml
+++ b/.github/workflows/diff_shades.yml
@@ -49,7 +49,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include: ${{ fromJson(needs.configure.outputs.matrix ) }}
+        include: ${{ fromJson(needs.configure.outputs.matrix) }}
 
     steps:
       - name: Checkout this repository (full clone)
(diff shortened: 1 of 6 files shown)


.github/workflows/diff_shades.yml
```diff
@@ -48,7 +49,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include: ${{ fromJson(needs.configure.outputs.matrix )}}
+        include: ${{ fromJson(needs.configure.outputs.matrix ) }}
```
JelleZijlstra, line 52:
```suggestion
        include: ${{ fromJson(needs.configure.outputs.matrix) }}
```
Does prettier not update this part?

JelleZijlstra (approved):

.github/workflows/diff_shades.yml
cobaltt7: Prettier does not format Actions syntax and formats them the same as other .yml files - the space before the `}}` was something I noticed and added manually, but I didn't see the space before the `)`.

PR #4665, opened 2025-05-12
CI: Remove now-uneeded workarounds

cobaltt7:
### Description

The issues mentioned in the comments have all been fixed. Checking CI still works as expected.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/diff_shades.yml
+++ b/.github/workflows/diff_shades.yml
@@ -34,7 +34,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
         run: >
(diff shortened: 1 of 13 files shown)


JelleZijlstra (approved):

JelleZijlstra:
The mypy failure in CI is likely because Click just had a new release, we may have to update something. Not sure about the unit test failures.

hauntsaninja (approved):
Nice!

cobaltt7:
it seems like mypy uses different versions of Click stubs depending on if it's run in the action vs on pre-commit.ci

JelleZijlstra (approved):

PR #4591, opened 2025-02-27
additional fix for click 8.2.0

mr-c:
### Description

Additional fix for the upcoming click 8.2.0; this is a backwards compatible fix with previous releases of click. 

### Checklist - did you ...

- [x] Add an entry in `CHANGES.md` if necessary?
- [x] Add / update tests if necessary?
- [x] Add new / update outdated documentation?

PR #4666, opened 2025-05-12
Update test_code_option_safe to work with click 8.2.0

danigm:
This patch fixes the test test_code_option_safe that's failing with click 8.2.0 after https://github.com/psf/black/pull/4591


cobaltt7:
Not too familiar with this, but in #4591 we made the 8.2 change conditional on the version- does this need to be conditional too? Or can we just bump the version instead

JelleZijlstra:
Since this is in tests I think we can just require the latest version. In the runtime code we should take a bit more care to allow multiple versions of Click, to avoid disturbing users who want to use one version or the other.

cobaltt7 (approved):

PR #4671, opened 2025-05-29
Fix CI mypyc 1.16 failure

MeGaGiGaGon:
### Description

This should fix the new failure happening with mypyc 1.16. I don't understand what was changed, and I also don't understand why the original code is so complex, so this reverts it to hopefully equivalent more normal code, which hopefully won't trip up mypyc.

I did not test this locally since I have no clue how.

### Checklist - did you ...

  - N/A, fixing CI failure, should not result in any changes at runtime
  - N/A, fixing CI failure
  - N/A, no docs needed for internal change

### DIFF
--- a/src/blib2to3/pgen2/parse.py
+++ b/src/blib2to3/pgen2/parse.py
@@ -89,18 +89,12 @@ class Recorder:
             self.parser.is_backtracking = is_backtracking
 
     def add_token(self, tok_type: int, tok_val: str, raw: bool = False) -> None:
-        func: Callable[..., Any]
-        if raw:
-            func = self.parser._addtoken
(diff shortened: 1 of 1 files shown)


MeGaGiGaGon:
I think that did fix it since both the `Build and publish / mypyc wheels` checks passed, If I understand correctly I think the diff-shades compare to the older version of black, so it's expected they would still fail? I don't know if there is anything else that needs to be done to fix it, or if it would automatically work once this is merged since it would become the new target for the baseline.

hauntsaninja (approved):
The right hand side is better than the left hand side, which is what you want out of a diff :-)

PR #4673, opened 2025-05-30
Fix backslash cr nl bug

MeGaGiGaGon:
### Description

This fixes another one of the hypothesis errors. It took me a long time to track down, but it was again in this translation layer: 

Since `transform_whitespace` only checked for `\\\n`, any `\\\r\n` whitespace token would break everything. This is because since the offset adjustments never happened, on the token after the `\\\r\n` there would be a position mismatch, triggering the code to add a `\n` to the prefix of that token. This is fine for the first pass, but after getting converted to code this causes the second pass to crash.
Input Code:
```py
x=\<carriage return><newline>
1
```
After first pass:
```py
x=<newline>
1
```
Crashes on second pass due to invalid code.

I also included the `\\\r` check for completeness, since even though code with `\\\r` seemed to work before it sounded to me like a bug waiting to happen, and has not caused any test failures (at least for me when running the tests locally).

### Checklist - did you ...

  - I didn't add a test, still I'm not certain how to handle these newline specific tests, and hypothesis should hopefully have this covered, but if someone else could write a test that would be great.
  - N/A crash fix, no docs changed

### DIFF
--- a/src/blib2to3/pgen2/tokenize.py
+++ b/src/blib2to3/pgen2/tokenize.py
@@ -113,7 +113,17 @@ def transform_whitespace(
         and prev_token.type not in (TokenType.nl, TokenType.newline)
     ):
         token_str = source[token.start_index : token.end_index]
-        if token_str.startswith("\\\n"):
+        if token_str.startswith("\\\r\n"):
+            return pytokens.Token(
+                TokenType.nl,
+                token.start_index,
+                token.start_index + 3,
+                token.start_line,
+                token.start_col,
+                token.start_line,
+                token.start_col + 3,
+            )
+        elif token_str.startswith("\\\n") or token_str.startswith("\\\r"):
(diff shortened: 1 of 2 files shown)


src/blib2to3/pgen2/tokenize.py
```diff
@@ -113,7 +113,17 @@ def transform_whitespace(
         and prev_token.type not in (TokenType.nl, TokenType.newline)
     ):
         token_str = source[token.start_index : token.end_index]
-        if token_str.startswith("\\\n"):
+        if token_str.startswith("\\\r\n"):
```
JelleZijlstra, line 116: Could you add a unit test? Probably something like `black.format_str("\\\r\n")` and assert it behaves sanely.

MeGaGiGaGon: Added

PR #4674, opened 2025-05-30
Attempt to fix carriage return issues

MeGaGiGaGon:
### Description

This is my attempt at fixing the carriage return problems. In combination with https://github.com/tusharsadhwani/pytokens/pull/1 , this should fix the hypothesis errors around `\r`.

I'm making both PRs drafts since my changes were made with the consideration of a sledgehammer, and were just focused on getting the issue to go away, so I'm uncertain if I've accidentally broken other things. All the black tests pass, as well as a local hypothesis run gave no problems, but I might have messed something up.

<details>
<summary> pytokens changes </summary>

My general change is based around the fact that I don't understand how you could fix `issue_128233` without making `\r`s behave properly as newlines, so that's what this PR does. If `issue_128233_handling` is enabled, `\r`s are treated as newlines instead of whitespace.

With this change, I came across issues where a lone `\r` newline would cause a double advance to happen, so I made `is_newline` return an `int` with the length of chars that should be removed. This keeps the old behavior intact, since 0 is still falsey and 1/2 are truthy, but also allows the methods that depend on `\r\n` handling to work correctly through checking if the result is 2.

</details>

<details open>
<summary> black changes </summary>

While working on the other hypothesis issues, I noticed that `\r`s weren't accounted for properly in the driver, so this adds those checks, leaving the `\n` ones first to hopefully leave `\r\n` behavior unchanged.

</details>

I also have no clue how to handle the cross-repository things that would make the tests show as passing, since they don't have the full effects without the other repo's changes.

One other note, I wonder if it would be easier to side-step all these problems completely and normalize the newlines early like how CPython does, probably in `src/black/__init__.py:format_str`

### Checklist - did you ...

  - Still unsure how to do the return type sensitive tests
  - Documentation not affected

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
 - Handle `# fmt: skip` followed by a comment at the end of file (#4635)
 - Fix crash when a tuple appears in the `as` clause of a `with` statement (#4634)
 - Fix crash when tuple is used as a context manager inside a `with` statement (#4646)
+- Fix issues/crashes caused by lack of carriage return checks (#4674)
 
 ### Preview style
 
--- a/src/blib2to3/pgen2/driver.py
+++ b/src/blib2to3/pgen2/driver.py
@@ -141,7 +141,7 @@ class Driver:
             if type in (tokenize.COMMENT, tokenize.NL):
                 prefix += value
                 lineno, column = end
-                if value.endswith("\n"):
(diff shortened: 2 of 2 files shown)


MeGaGiGaGon:
I managed to fix this a while ago in Black by normalizing eagerly before getting to tokenization, so these two PRs are no longer needed.

PR #4680, opened 2025-06-03
Fix f string expr split after

### LINKED ISSUE #4677: string_processing: Substitutions are never put at the end of wrapped f-strings
**Describe the bug**

When a long f-string is wrapped / split due to exceeding the line limit, Black will never put a substitution at the end of a f-string, resulting in unexpected wrapping behaviour.

**To Reproduce**

Input:

```python
print(
    f"short {1} loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong"
)
```

Output with `--unstable`:

```python
print(
    "short"
    f" {1} loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong"
)
```

[Playground link](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ADsAGpdAD2IimZxl1N_WmOQqEcOUylYcEXhdHqxLmtzFN-Q9ZLubshh24diXSjzsT1bCHXc-OftOMCkbTwe-Fs9XycD4Yez_J_v5_GDq4nFpOTWHTBONjF-IdtnuObkNsXEIX5labthcyzjeMTbawAAAAA7CxnaJpZ0zAABhgHtAQAAevkLE7HEZ_sCAAAAAARZWg==)

**Expected behavior**

```python
print(
    f"short {1}"
    " loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong"
)
```

**Environment**

- Black's version: 314f8cf

**Additional context**

Adding any character after `{1}` results in the expected behaviour of the substitution being put in the first line.

MeGaGiGaGon:
### Description

Fixes #4677

~~My assumption on how the issue started is from a documentation issue on `iter_fexpr_spans` - the docs said "Spans are half-open ranges (left inclusive, right exclusive)", but it actually returns closed ranges. I assume the author of `_get_illegal_split_indices` then saw that, and added one to the range, making the ranges include the char after the f-string expr.~~

Update: Wrong, the spans are actually half-open, and I also got confused by it like the original author.

To fix this, I just removed that +1 on the ranges.

I looked at the other usages of `iter_fexpr_spans`, but all of them seem to properly use the ranges ~~as closed~~ as half-open.

I updated the tests that were changed by this, it was a small subsection that looks fine, and since no other code tests were effected that supports that everything else uses the spans ~~as closed~~ as half-open. I also fixed the docs for `iter_fexpr_spans` to prevent this in the future.

### Checklist - did you ...

### DIFF
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -1307,6 +1307,7 @@ class BaseStringSplitter(StringTransformer):
 def iter_fexpr_spans(s: str) -> Iterator[tuple[int, int]]:
     """
     Yields spans corresponding to expressions in a given f-string.
+    Spans are closed ranges (left and right inclusive).
     Assumes the input string is a valid f-string, but will not crash if the input
     string is invalid.
     """
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -1307,7 +1307,6 @@ class BaseStringSplitter(StringTransformer):
 def iter_fexpr_spans(s: str) -> Iterator[tuple[int, int]]:
     """
     Yields spans corresponding to expressions in a given f-string.
-    Spans are half-open ranges (left inclusive, right exclusive).
     Assumes the input string is a valid f-string, but will not crash if the input
     string is invalid.
     """
@@ -1755,7 +1754,7 @@ class StringSplitter(BaseStringSplitter, CustomSplitMapMixin):
         ]
         for it in iterators:
             for begin, end in it:
-                illegal_indices.update(range(begin, end + 1))
+                illegal_indices.update(range(begin, end))
         return illegal_indices
 
     def _get_break_idx(self, string: str, max_break_idx: int) -> Optional[int]:
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -1307,7 +1307,7 @@ class BaseStringSplitter(StringTransformer):
 def iter_fexpr_spans(s: str) -> Iterator[tuple[int, int]]:
     """
     Yields spans corresponding to expressions in a given f-string.
-    Spans are closed ranges (left and right inclusive).
+    Spans are half-open ranges (left inclusive, right exclusive).
     Assumes the input string is a valid f-string, but will not crash if the input
     string is invalid.
     """
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,8 @@
 
 - Fix a bug where one-liner functions/conditionals marked with `# fmt: skip` would still
   be formatted (#4552)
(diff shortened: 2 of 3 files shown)


cobaltt7 (approved):

src/black/trans.py
```diff
@@ -1307,7 +1307,7 @@ def _prefer_paren_wrap_match(LL: list[Leaf]) -> Optional[int]:
 def iter_fexpr_spans(s: str) -> Iterator[tuple[int, int]]:
     """
     Yields spans corresponding to expressions in a given f-string.
-    Spans are half-open ranges (left inclusive, right exclusive).
+    Spans are closed ranges (left and right inclusive).
```
hauntsaninja: this isn't right, the spans are half-open ranges. See e.g. https://github.com/psf/black/blob/8310a118ba6690fef55aaf8d67c967852b4c997e/tests/test_trans.py#L26

hauntsaninja (changes requested):
I think the fix is right, but the change to the comment is wrong

src/black/trans.py
hauntsaninja:
```
>>> """text {var} text"""[5]
'{'
>>> """text {var} text"""[10]
' '
>>> """text {var} text"""[5:10]
'{var}'
```

MeGaGiGaGon: Looks like I also got confused by the ranges 😓 fixed

ToTorTH:
> ### Description
> Fixes #4677
> 
> ~My assumption on how the issue started is from a documentation issue on `iter_fexpr_spans` - the docs said "Spans are half-open ranges (left inclusive, right exclusive)", but it actually returns closed ranges. I assume the author of `_get_illegal_split_indices` then saw that, and added one to the range, making the ranges include the char after the f-string expr.~
> 
> Update: Wrong, the spans are actually half-open, and I also got confused by it like the original author.
> 
> To fix this, I just removed that +1 on the ranges.
> 
> I looked at the other usages of `iter_fexpr_spans`, but all of them seem to properly use the ranges ~as closed~ as half-open.
> 
> I updated the tests that were changed by this, it was a small subsection that looks fine, and since no other code tests were effected that supports that everything else uses the spans ~as closed~ as half-open. I also fixed the docs for `iter_fexpr_spans` to prevent this in the future.
> 
> ### Checklist - did you ...
> * [x]  Add an entry in `CHANGES.md` if necessary?
> * [x]  Add / update tests if necessary?
> * [x]  Add new / update outdated documentation?

https://github.com/psf/black/pull/4680#issue-3115456524

PR #4685, opened 2025-06-06
Convert legacy string formatting to f-strings

MeGaGiGaGon:
### Description

I got annoyed with dealing with percent formatting, so I decided to convert them to f-strings.

I'm not 100% sure about the `pytree` change since it's a reduction in quality, but I did it anyways because consistency.

There are still a couple instances of percent formatting in places, but they deal with external things like logging or tests so I didn't touch them.

There's a few more things I'd like to modernize since older python runtimes have stopped support, but since it's a lot I'm doing one thing at a time.

### Checklist - did you ...

  - N/A, `skip news` because behavior should be unchanged.
  - N/A, no tests affected.
  - N/A, documentation not affected.

### DIFF
--- a/src/blib2to3/pgen2/driver.py
+++ b/src/blib2to3/pgen2/driver.py
@@ -149,7 +149,9 @@ class Driver:
                 type = grammar.opmap[value]
             if debug:
                 assert type is not None
-                self.logger.debug(f"{token.tok_name[type]} {value!r} ({prefix=!r})")
+                self.logger.debug(
+                    "%s %r (prefix=%r)", token.tok_name[type], value, prefix
+                )
             if type == token.INDENT:
                 indent_columns.append(len(value))
                 _prefix = prefix + value


src/blib2to3/pgen2/driver.py
```diff
@@ -149,9 +149,7 @@ def parse_tokens(self, tokens: Iterable[TokenInfo], debug: bool = False) -> NL:
                 type = grammar.opmap[value]
             if debug:
                 assert type is not None
-                self.logger.debug(
-                    "%s %r (prefix=%r)", token.tok_name[type], value, prefix
```
JelleZijlstra: This is better to keep as is, since the log probably won't be executed and the previous code avoids creating the string.

JelleZijlstra (approved):
Looks good apart from the one I commented on

src/blib2to3/pgen2/driver.py
MeGaGiGaGon: I did consider that, but since it’s behind the `if debug:` my thoughts were the log level would always be at debug if that was reached - is that not right?

JelleZijlstra: I think we always pass debug as True, look in `black/parsing.py` for the two calls to `parse_string`.

Which is also questionable, maybe we should change that and get some perf win.

MeGaGiGaGon: Hm, just a guess but since it uses the logger instead of print is that so that the log level can deal with it? But another question: is there even a way for the user to get that output? Since I've never been able to get it to output even with like `-v`

JelleZijlstra: Yeah I don't think so, you'd need to manually grab the blib2to3 logger and set the log level probably. I think this code is just useless.

Maybe in a separate PR you can make it so parsing.py stops passing True for this debug argument.

PR #4686, opened 2025-06-07
Update python image version and reduce size for gallery docker image

huisman:
### Description

Update python image to 3.13:
Python version 3.8 is EOL and the gallery.py script depends on type hinting functionality introduced in [version 3.9](https://docs.python.org/3.9/whatsnew/3.9.html#type-hinting-generics-in-standard-collections).

Reduce size by adding `--no-install-recommends` and removing `/var/lib/apt/lists/*`: 
This reduces the size of the resulting docker image by not installing of keeping unneeded files.
 
### Checklist - did you ...

I don't believe this change warrants an entry in `CHANGES.md` or an updated test or documentation and should get the magical label to silence the CHANGELOG entry check.

### DIFF
--- a/docs/contributing/release_process.md
+++ b/docs/contributing/release_process.md
@@ -46,8 +46,6 @@ To cut a release:
    - `release.py` will calculate this and log to stderr for you copy paste pleasure
 1. File a PR editing `CHANGES.md` and the docs to version the latest changes
    - Run `python3 scripts/release.py [--debug]` to generate most changes
-     - Sub headings in the template, if they have no bullet points need manual removal
-       _PR welcome to improve :D_
 1. If `release.py` fail manually edit; otherwise, yay, skip this step!
    1. Replace the `## Unreleased` header with the version number
    1. Remove any empty sections for the current release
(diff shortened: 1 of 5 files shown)


gallery/Dockerfile
```diff
@@ -1,11 +1,12 @@
-FROM python:3.8.2-slim
+FROM python:3.13-slim
```
cooperlees: I can go either way here - I wonder if we want to just follow python:3-slim here if it exists? I guess ti depends how recent it's kept - I would hope latest stable major - i.e. 3.13 at the moment ...

cooperlees:
O, let's add a change log for this please.

gallery/Dockerfile
huisman: python:3-slim exists and it is kept up-to-date, but I haven't found an explicit statement on whether it follows the latest stable minor or major version. 

I think in this case it might be worth it to use python:3-slim. It would have prevented the image from not working due to the type hinting. My guess is that it would prevent more issues than pinning a specific version would prevent(especially since this image does not appear to be used very often.)

huisman:
> O, let's add a change log for this please.

Added a line to the Changes.md

cooperlees (approved):
Thanks - Will merge should CI pass.

cobaltt7:
CI passes (barring 3.12.4 Fuzz which is failing on main)

PR #4688, opened 2025-06-10
Small cleanup and release.py improvement

MeGaGiGaGon:
### Description

This has the changes mentioned in #4685 for not having the driver always use debug `True`, since while running black normally there is no way to change the log level to make those debug log calls useful.

This also has a small improvement to the `release.py` file since while I was reading the docs I noticed a part of the docs that said pr welcome for this. I updated the comment detection logic using https://www.rexegg.com/regex-quantifiers.php#lazy_solution as future proofing since I noticed it would break if there were ever a 3 line comment in the format, but that change could be reverted. The empty subheading removal assumes the file is standard-ly formatted. 

### Checklist - did you ...

  - N/A, no user facing changes
  - N/A, no tests affected

### DIFF
--- a/docs/contributing/release_process.md
+++ b/docs/contributing/release_process.md
@@ -47,7 +47,6 @@ To cut a release:
 1. File a PR editing `CHANGES.md` and the docs to version the latest changes
    - Run `python3 scripts/release.py [--debug]` to generate most changes
      - Sub headings in the template, if they have no bullet points need manual removal
-       _PR welcome to improve :D_
 1. If `release.py` fail manually edit; otherwise, yay, skip this step!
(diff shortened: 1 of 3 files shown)


JelleZijlstra (approved):
All changes make sense but I'd have put the parsing.py changes in a separate PR since they seem unrelated.

MeGaGiGaGon:
Yep, sorry got lazy. Moving the debug changes to new pr since it's easier and I'm bad with git

MeGaGiGaGon:
Oops, I didn't know renaming the branch would close the PR. Oh well I'll just make 2 new PRs since I've made a big mess out of this

PR #4692, opened 2025-06-12
Small cleanup to parsing.py

MeGaGiGaGon:
### Description

This has the changes mentioned in #4685 for not having the driver always use debug `True`, since while running black normally there is no way to change the log level to make those debug log calls useful, and should give a very small speedup.

### Checklist - did you ...

  - N/A, no user facing changes made
  - N/A, no tests affected
  - N/A, no docs affected


MeGaGiGaGon:
Merging since Jelle already approved in #4688

PR #4694, opened 2025-06-13
Misc lints

MeGaGiGaGon:
### Description

I was bored so I decided to run all the ruff lints against `src/black`, and only fix the ones I liked, so here's the results. I tried my best to only pick ones that I thought made a clear improvement to the code without causing too much churn. All the reasoning is in the commit messages, but I also copied it here because github cuts it off.

```
SIM201 negate-equal-op seems reasonable

RUF010 explicit-f-string-type-conversion seems reasonable

RUF003 ambiguous-unicode-character-comment churn but reasonable

RUF001 ambiguous-unicode-character-string changing the literal NBSP to using \N{} is a lot more editor friendly, good change

PLR1736 unnecessary-list-index-lookup good change, both speed and clarity improvement

PLC2401 non-ascii-name churn but good, original name takes a long time to parse that it is the empty set and is very had to type, while set() is more explicit and easy to type
Reverted since there were PRs on this before per Jelle

PERF401 manual-list-comprehension code clarity improvement

FURB136 if-expr-min-max code clarity improvement
Reverted, was on the edge with including this and is maybe worse perf per Jelle

SIM300 yoda-conditions improves readability and consistency

D301 escape-sequence-in-docstring some docstrings were broken due to having unescaped \n s, so I fixed those, as well as converted some others for consistency
```

### Checklist - did you ...

  - N/A no behavior changes made
  - N/A no tests affected
  - N/A no docs affected

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -40,6 +40,7 @@ from black.nodes import (
     ensure_visible,
     fstring_to_string,
     get_annotation_type,
+    has_sibling_with_type,
     is_arith_like,
     is_async_stmt_or_funcdef,
     is_atom_with_invisible_parens,
@@ -1628,6 +1629,11 @@ def maybe_make_parens_invisible_in_atom(
         or is_empty_tuple(node)
         or is_one_tuple(node)
         or (is_tuple(node) and parent.type == syms.asexpr_test)
+        or (
+            is_tuple(node)
+            and parent.type == syms.with_stmt
+            and has_sibling_with_type(node, token.COMMA)
+        )
         or (is_yield(node) and parent.type != syms.expr_stmt)
         or (
             # This condition tries to prevent removing non-optional brackets
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -40,7 +40,6 @@ from black.nodes import (
     ensure_visible,
     fstring_to_string,
     get_annotation_type,
-    has_sibling_with_type,
     is_arith_like,
     is_async_stmt_or_funcdef,
     is_atom_with_invisible_parens,
@@ -573,6 +572,7 @@ class LineGenerator(Visitor[Line]):
         self.current_line = Line(mode=self.mode)
 
         v = self.visit_stmt
+        Ø: set[str] = set()
         self.visit_assert_stmt = partial(v, keywords={"assert"}, parens={"assert", ","})
         self.visit_if_stmt = partial(
             v, keywords={"if", "else", "elif"}, parens={"if", "elif"}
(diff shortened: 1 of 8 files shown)


tusharsadhwani:
@MeGaGiGaGon for longevity i think it's better if you could: 

- Instead of selectively fixing the raised lints, create a ruff config that lets anyone else run the lints in a way that this PR passes the lint step

- Add a CI step that runs ruff so we can maintain the code quality going forward

MeGaGiGaGon:
The main reason I didn't do that is since especially with lints, I feel like it should be a lot more thought out and discussed than my approach. Since what lints you use defines the style of the code, I feel like any major changes to it, like switching to a whole new linting framework, should be a whole team effort. 

I'm also very new to linting in general, and the main reason I use ruff is that it is the thing I found first, followed by it having a lot of lints so it's very effective with my "enable all then trim" strategy. I'm unsure what else is out there in the ecosystem, which goes back to the "this should probably be a bigger discussion".

And even assuming that ruff would be chosen as the linter, I was very conservative with my choices here. There are some lints I saw that I do like, but would cause a large amount of churn to enable for smaller benefits, which is why I only went with ones that are both fairly contained and seem like clear improvements.

For context, there were 90 different lints raised, with 1253 total errors. Out of those I chose 10 lints to fix, making up 16 of the errors.

This also leads into/has ties with a different discussion I feel is worth having, which is on linting blib2to3. Currently it is still excluded from coverage, and also has a lot more lints that could be actually worth fixing. At least for my changes, a lot of them have been inside blib2to3, so I've been wondering if it would be worthwhile to add it to the lints and clean up the code.

That also has ties to a discussion I'm not sure ever got finished, #2318 (Switch to a new parser), and might be worth re-discussing since I'm not sure how the parser landscape has changed since 2021.

Also I'm tired, so sorry for the long wall of text. I'll probably come back and edit this to make it more readable tomorrow morning.

src/black/linegen.py
```diff
@@ -573,31 +573,30 @@ def __post_init__(self) -> None:
         self.current_line = Line(mode=self.mode)
 
         v = self.visit_stmt
-        Ø: set[str] = set()
```
JelleZijlstra: Let's keep this, we've refused several PRs getting rid of it. Using Unicode identifiers helps force the ecosystem to support them.

src/black/strings.py
```diff
@@ -355,7 +355,7 @@ def char_width(char: str) -> int:
         elif codepoint > end_codepoint:
             lowest = idx + 1
         else:
-            return 0 if width < 0 else width
+            return max(width, 0)
```
JelleZijlstra: This is likely slower than the old code though I don't know whether it matters. I do believe we call this function a lot though.

src/black/linegen.py
MeGaGiGaGon: Oops, I didn't know. Sounds good in that case, reverted

src/black/strings.py
MeGaGiGaGon: Sounds good, this change was on the edge of inclusion anyways, so reverted.

PR #4710, opened 2025-07-18
Add normalizing `\r` style newlines

MeGaGiGaGon:
### Description

This is another way to fix the issues with `\r`, and should permanently get rid of them - currently `\r` isn't included in the normalization due to limitations with `io.BytesIO` and probably other things. This PR makes it so that they are included in the normalization to `\n`, and are a valid target for all newlines to normalize to. This should permanently fix all `\r`-related issues, since now there will be none given to the tokenizer / parser / formatter internals.

---

This PR is mostly to gauge interest on if this would be a good direction to go in. If yes, there is a good amount of work that needs done on removing `\r`-related code that would no longer be applicable.

This should also hopefully fix the fuzzing finally.

---

A more in-depth explanation of the changes made:

The key change is to `decode_bytes`. `io.BytesIO.readline` is hard-coded to ignore `\r` line ends. This means that the returned lines have to be manually double-checked for `\r`s before the newline `readline` found. The code I used for this could be more condensed with tricks, but I was having trouble conceptualizing and making sure all branches were covered, so I kept it expanded out. Note that in the `\r\n` case, we don't have to double-check for another `\r\n` on found `\r`s, since `readline` would have chopped there instead.

Changing that allowed for changes to trickle back up the calling tree - `_format_str_once` now has newlines normalized early, and also does the actual newline normalization. `format_stdin_to_stdout` also needed some changes to account for `\r`s. `blackd` previously handled some newline normalization, but since it's now part of `_format_str_once` I was able to remove that.

---

I'm not 100% sure whether this should go directly into stable, or should go into preview. I'd like it to go directly into stable for a couple of reasons:
- It would be really annoying to keep two different newline behaviors.
- I think breakage should generally be small, as most files, especially ones in a git repo, should already have only 1 newline type.
- The behavior of keeping newlines untouched on no changes is still present.

And most importantly, the docs were only partially correct before. Since the newline normalization code lived in `blackd`, newlines would only be normalized using the first encountered when running through `blackd`, but not through `black` directly. For example, running this code while having `uv` installed and in the path:
```py
import subprocess
print(repr(subprocess.run(["uvx", "black", "-c", "\r\ntest\ntest2"], capture_output=True)))
```
Gives
```
CompletedProcess(args=['uvx', 'black', '-c', '\r\ntest\ntest2'], returncode=0, stdout=b'test\ntest2\n', stderr=b'')
```
The stdout is `test\ntest2\n`, so the newlines were normalized to `\n` despite the first newline being `\r\n`. This PR makes it correctly output `test\r\ntest2\r\n`

### Checklist - did you ...

- [x ] Add an entry in `CHANGES.md` if necessary?
  - Updated tests with changed behavior.
  - Added new test that permutes over all simple newline combinations.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 - Remove support for pre-python 3.7 `await/async` as soft keywords/variable names
   (#4676)
 - Fix crash on parenthesized expression inside a type parameter bound (#4684)
+- Normalize newlines on format to first newline in file (#4710)
 
 ### Preview style
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1010,7 +1010,7 @@ def format_stdin_to_stdout(
     if content is None:
         src, encoding, newline = decode_bytes(sys.stdin.buffer.read())
     else:
-        src, encoding, newline = content, "utf-8", ""
+        src, encoding, newline = content, "utf-8", "\n"
 
     dst = src
     try:
@@ -1026,11 +1026,11 @@ def format_stdin_to_stdout(
         )
         if write_back == WriteBack.YES:
             # Make sure there's a newline after the content
-            if dst and dst[-1] != "\n":
-                dst += "\n"
+            if dst and dst[-1] != "\n" and dst[-1] != "\r":
+                dst += newline
             f.write(dst)
         elif write_back in (WriteBack.DIFF, WriteBack.COLOR_DIFF):
-            now = datetime.now(timezone.utc)
+            now = datetime.now(tz=timezone.utc)
             src_name = f"STDIN\t{then}"
             dst_name = f"STDOUT\t{now}"
             d = diff(src, dst, src_name, dst_name)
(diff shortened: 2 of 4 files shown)


cobaltt7:
IMO this is the best solution. `\r` on its own is not a standard line ending on any modern OS, and as a formatter, it certainly makes sense for Black for standardize them

MeGaGiGaGon:
Oops, let me edit my opening comment since I wasn't very clear (I finished it very late last night)

JelleZijlstra:
The logic sounds good, but I think this should be limited to the preview style. We should guarantee that Black-formatted code stays the same in the stable style across the whole year, and this change would break that guarantee. It would break it only in a very unlikely situation, but I think it's better for user trust to stay strict about the guarantee.

MeGaGiGaGon:
It looks like making this preview doesn't fix the fuzz errors, which was the whole point :( It also makes the diff/code much nosier, :( x2

CHANGES.md
```diff
@@ -37,6 +37,8 @@
 - Wrap the `in` clause of comprehensions across lines if necessary (#4699)
 - Remove parentheses around multiple exception types in `except` and `except*` without
   `as`. (#4720)
+- Add `\r` style newlines to the potenial newlines to normalize file newlines both from
```
JelleZijlstra:
```suggestion
- Add `\r` style newlines to the potential newlines to normalize file newlines both from
```

docs/the_black_code_style/future_style.md
```diff
@@ -33,6 +33,8 @@ Currently, the following features are included in the preview style:
   across lines if it would otherwise exceed the maximum line length.
 - `remove_parens_around_except_types`: Remove parentheses around multiple exception
   types in `except` and `except*` without `as`. See PEP 758 for details.
+- `normalize_cr_newlines`: Add `\r` style newlines to the potenial newlines to normalize
```
JelleZijlstra:
```suggestion
- `normalize_cr_newlines`: Add `\r` style newlines to the potential newlines to normalize
```

PR #4714, opened 2025-07-23
Add FAQ section on Windows emojis not working

### LINKED ISSUE #3156: No Emojis on PowerShell

MeGaGiGaGon:
### Description

This PR adds an FAQ section on Windows emojis. The TDLR is "Black can't fix this, use [Windows Terminal](https://github.com/microsoft/terminal)".

This closes #3156, as it will now be documented that the issues with emojis are not Black's problem.

One piece of relevant reading is #3374, specifically [this comment by Jelle](https://github.com/psf/black/pull/3374#issuecomment-1369413290):
> Coming back to this, I'm not enthusiastic about including this in Black:
> 
> * We're monkeypatching the standard library, always a risky thing to do, and we're doing it only in a particular set of obscure conditions (when pytest is absent and we're on a particular version of Windows). That feels like a future maintainability nightmare.
> * Printing unicode isn't a core part of what Black is about. We're using Click to create Unicode-compatible output, so I feel like it should be Click's job to fix Unicode issues here. If Click doesn't think it's worth fixing, maybe that's a hint that we shouldn't either.
> * The bug this fixes isn't particularly bad. Emojis get printed as "\u1234" sequences, which is ugly but doesn't make Black unusable.

I agree with Jelle on this, emojis being printed as Unicode escapes or the missing character glyph certainly isn't ideal, but aside from looking bad still leaves Black fully functional, especially when combined with Windows Terminal existing to fix it.

I always use Windows Terminal, and I actually forgot this was an issue, since it handles all the cases fine, even the case of `echo "a+1" |  black --check - | more` that plain PowerShell fails on.

### Checklist - did you ...

  - Though not strictly nessacary, I added one anyways since this is giving a "won't fix" stance on an issue.
  - N/A, no functionality affected.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -137,3 +137,13 @@ wheels (including the pure Python wheel), so this command will use the [sdist].
 [mypyc]: https://mypyc.readthedocs.io/en/latest/
 [sdist]:
   https://packaging.python.org/en/latest/glossary/#term-Source-Distribution-or-sdist
+
+## Why are emojis not displaying correctly on Windows?
+
+When using Windows, the emojis in _Black_'s output may not display correctly. This is
+not fixable from _Black_'s end.
+
+Instead, run your chosen command line/shell through [Windows Terminal], which will
+properly handle rendering the emojis.
+
+[Windows Terminal]: https://github.com/microsoft/terminal
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -75,6 +75,8 @@
 <!-- Major changes to documentation and policies. Small docs changes
      don't need a changelog entry. -->
 
+- Add FAQ entry for windows emojis not displaying (#4714)
+
 ## 25.1.0
 
 ### Highlights


cobaltt7:
#3156 also says that setting `PYTHONUITF8=1` works, should that be mentioned in the docs as well?

docs/faq.md
```diff
@@ -137,3 +137,13 @@ wheels (including the pure Python wheel), so this command will use the [sdist].
 [mypyc]: https://mypyc.readthedocs.io/en/latest/
 [sdist]:
   https://packaging.python.org/en/latest/glossary/#term-Source-Distribution-or-sdist
+
+## Why are emojis not displaying correctly on Windows?
```
JelleZijlstra: I think usually "emoji" is both singular and plural
```suggestion
## Why are emoji not displaying correctly on Windows?
```
(Similar changes in a few other places)

docs/faq.md
```diff
@@ -137,3 +137,13 @@ wheels (including the pure Python wheel), so this command will use the [sdist].
...
+
+When using Windows, the emojis in _Black_'s output may not display correctly. This is
```
JelleZijlstra: The second line is technically a lie; we could fix it but choose not to.

docs/faq.md
MeGaGiGaGon: Updated

docs/faq.md
MeGaGiGaGon: It's a half-lie, since it might be fixable in PowerShell with the changes https://github.com/psf/black/pull/3374 was making, but definitely not in command prompt which doesn't have the font files for it. That's why I went for a blanket unfixable, but if you have a better idea for explaining it concisely we can go with that.

MeGaGiGaGon:
> #3156 also says that setting `PYTHONUITF8=1` works, should that be mentioned in the docs as well?

I chose not to for keeping the section simple, since that only fixes it on powershell, and not on cmd, where the font doesn't support emoji at all. We could do a more complex section, but for the same reason #3374 was closed I tried to keep it as simple as possible since dealing with emoji windows jank shouldn't be black's problem.

docs/faq.md
JelleZijlstra: Makes sense, didn't realize the fix only works for some contexts.

JelleZijlstra (approved):

PR #4726, opened 2025-08-04
Fix function `supports_feature` to return False when `target_versions`  is empty. [skip news]

ranjodhsingh1729:
### Description

```
def supports_feature(target_versions: set[TargetVersion], feature: Feature) -> bool:
    return all(feature in VERSION_TO_FEATURES[version] for version in target_versions)
```

Here when `target_versions` is empty, expression gets reduced to `all([])` which evaluates to `True` which is not the correct behaviour.

### Checklist - did you ...

      stability policy?

### DIFF
--- a/src/black/mode.py
+++ b/src/black/mode.py
@@ -190,9 +190,10 @@ VERSION_TO_FEATURES: dict[TargetVersion, set[Feature]] = {
 
 
 def supports_feature(target_versions: set[TargetVersion], feature: Feature) -> bool:
-    return bool(target_versions) and all(
-        feature in VERSION_TO_FEATURES[version] for version in target_versions
-    )
+    if not target_versions:
+        raise ValueError("target_versions must not be empty")
+
+    return all(feature in VERSION_TO_FEATURES[version] for version in target_versions)
 
 
 class Preview(Enum):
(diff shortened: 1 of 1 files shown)


tusharsadhwani:
Can you expand on a use case where this bites a user? Do we actually end up passing an empty list to `supports_feature` in some case?

ranjodhsingh1729:
> Can you expand on a use case where this bites a user? Do we actually end up passing an empty list to supports_feature in some case?

No, this doesn't affect functionality in any way.
There are only two places where this function is currently used:

1. In `get_grammars`, which simply returns all grammars if `target_versions` is empty - no call to `supports_feature` is made.
2. In `_format_str_once`, where `supports_feature` is called with `versions` obtained from `detect_target_versions`. `detect_target_versions` cannot possibly return an empty set (if it does that would indicate some issue elsewhere), so again, no issue here.

The misleading behavior of this function ended up wasting quite a bit of my time while working on #4720, so I figured it’s worth fixing to save others the trouble down the line.

tusharsadhwani:
If the function is never called with an empty value, and is never intended to be called with an empty value, let's throw a meaningful error instead of returning `False`.

P.S. The CI failure does seem unrelated.

tusharsadhwani (approved):

PR #4739, opened 2025-08-27
Optimize Line string method

av-runner:
### Description

This PR introduces an optimization for the `Line.__str__` method, improving runtime efficiency when rendering lines.

The original implementation repeatedly concatenates strings using `+=`, which is inefficient in Python due to the creation of intermediate string objects.
The refactored version significantly speeds up string creation by replacing repeated concatenation (`res += ...`) with the much more efficient `"".join()` method. 
This approach avoids creating a new string object on every iteration, drastically reducing memory allocations and overhead. 
Additionally, it uses a more performant generator expression to flatten the comments list, further contributing to the speedup.

Benchmarks indicate that this change provides a measurable speedup over the original version.
The code snippet below was compiled with mypy `python -m mypyc test.py` and then ran with Python 3.13.2 `python -c "import test; test.run_bench()"`.

### Checklist - did you ...

* [x] Implement any code style changes under the `--preview` style, following the
  stability policy?
* [x] Add an entry in `CHANGES.md` if necessary?
* [x] Add / update tests if necessary?
* [x] Add new / update outdated documentation?

### Performance Comparison

* **Improvement:** \~35-40 % faster than the original
* **Set-up:** Tested on MacBook Air M2 16 GB RAM.

Example results:

```
Original: 0.6753s
Optimized: 0.4052s
Improvement: 0.40 %
Speedup: 1.67
```

### Code Comparison (for validation)

You can reproduce the benchmark with the following snippet:

```python
from dataclasses import dataclass, field
from typing import Optional, Dict, List, Any, TypeVar
import timeit
import itertools

class Mode:
    pass

LeafID = int

@dataclass
class Leaf:
    value: str
    prefix: str = ""

    def __str__(self) -> str:
        return self.prefix + self.value

@dataclass
class BracketTracker:
    # Minimal stub for benchmark
    pass

@dataclass
class Line:

    mode: Mode = field(repr=False)
    depth: int = 0
    leaves: List[Leaf] = field(default_factory=list)
    comments: Dict[LeafID, List[Leaf]] = field(default_factory=dict)
    bracket_tracker: BracketTracker = field(default_factory=BracketTracker)
    inside_brackets: bool = False
    should_split_rhs: bool = False
    magic_trailing_comma: Optional[Leaf] = None

    def __bool__(self) -> bool:
        return bool(self.leaves or self.comments)

    # -------- Original version --------
    def str_original(self) -> str:
        """Render the line."""
        if not self:
            return "\n"

        indent = "    " * self.depth
        leaves = iter(self.leaves)
        first = next(leaves)
        res = f"{first.prefix}{indent}{first.value}"
        for leaf in leaves:
            res += str(leaf)
        for comment in itertools.chain.from_iterable(self.comments.values()):
            res += str(comment)

        return res + "\n"

    # -------- Optimized version --------
    def str_optimized(self) -> str:
        """Render the line."""
        if not self:
            return "\n"

        indent = "    " * self.depth
        leaves = iter(self.leaves)
        first = next(leaves)
        res = f"{first.prefix}{indent}{first.value}"
        res += ''.join(str(leaf) for leaf in leaves)
        comments_iter = itertools.chain.from_iterable(self.comments.values())
        comments = [str(comment) for comment in comments_iter]
        res += ''.join(comments)

        return res + "\n"


def make_test_line(n_leaves: int, n_comments: int) -> Line:
    leaves = [Leaf(str(i), prefix="_") for i in range(n_leaves)]
    comments = {0: [Leaf(f"c{j}") for j in range(n_comments)]}
    return Line(mode=Mode(), depth=2, leaves=leaves, comments=comments)


def run_bench():
    line = make_test_line(1000, 200)

    t_orig = timeit.timeit(lambda: line.str_original(), number=10000)
    t_opt = timeit.timeit(lambda: line.str_optimized(), number=10000)

    print(f"Original: {t_orig:.4f}s")
    print(f"Optimized: {t_opt:.4f}s")
    print(f"Improvement: {(t_orig - t_opt) / t_orig:.2f} %")
    print(f"Speedup: {t_orig / t_opt:.2f}")


if __name__ == "__main__":
    run_bench()

```

### DIFF
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -484,10 +484,10 @@ class Line:
         leaves = iter(self.leaves)
         first = next(leaves)
         res = f"{first.prefix}{indent}{first.value}"
-        res += ''.join(str(leaf) for leaf in leaves)
+        res += "".join(str(leaf) for leaf in leaves)
         comments_iter = itertools.chain.from_iterable(self.comments.values())
         comments = [str(comment) for comment in comments_iter]
-        res += ''.join(comments)
+        res += "".join(comments)
 
         return res + "\n"
 
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -484,10 +484,10 @@ class Line:
         leaves = iter(self.leaves)
         first = next(leaves)
         res = f"{first.prefix}{indent}{first.value}"
-        for leaf in leaves:
-            res += str(leaf)
-        for comment in itertools.chain.from_iterable(self.comments.values()):
-            res += str(comment)
+        res += ''.join(str(leaf) for leaf in leaves)
+        comments_iter = itertools.chain.from_iterable(self.comments.values())
+        comments = [str(comment) for comment in comments_iter]
+        res += ''.join(comments)
 
         return res + "\n"
 


JelleZijlstra:
Please show benchmark results for an end-to-end run of Black instead. Black is compiled with mypyc, which may have different performance characteristics.

av-runner:
Ok, I will do that. The results from the snippet above are from it being compiled with mypyc. Do you have anything specific in mind you would like to see?

JelleZijlstra:
Oh, your first message didn't mention that you ran the benchmark under mypyc. I can reproduce the speedup if I compile the file with mypyc, about 2x on my system.

Seems fine to merge then, thanks!

av-runner:
Great! The workflows seem to fail, is that normal, or should I look into it?

src/black/lines.py
```diff
@@ -484,10 +484,10 @@ def __str__(self) -> str:
...
         first = next(leaves)
         res = f"{first.prefix}{indent}{first.value}"
-        for leaf in leaves:
-            res += str(leaf)
-        for comment in itertools.chain.from_iterable(self.comments.values()):
-            res += str(comment)
+        res += ''.join(str(leaf) for leaf in leaves)
+        comments_iter = itertools.chain.from_iterable(self.comments.values())
+        comments = [str(comment) for comment in comments_iter]
+        res += ''.join(comments)
```
cobaltt7:
```suggestion
        res += "".join(str(leaf) for leaf in leaves)
        comments_iter = itertools.chain.from_iterable(self.comments.values())
        comments = [str(comment) for comment in comments_iter]
        res += "".join(comments)
```
The Black style uses double quotes, which is the main source of CI failing
(there's also a longstanding issue #4674 but that can be ignored)

PR #4741, opened 2025-08-31
ci: Run diff-shades on unstable instead of preview

cobaltt7:
### Description

This PR changes diff-shades to run on the stable and unstable styles as opposed to the stable and preview styles.

I opted to replace the preview job with the unstable job instead of adding a new job because all preview changes also exist in unstable so running both feels redundant, and it takes a long time to run so it wouldn't be worth it to have both jobs.

I wasn't sure if the "preview-changes" mode was meant to be interpreted as "Changes in the preview style" or "Preview new changes". I took it to mean the former and renamed it to `unstable-style-changes` for clarity, but I can change that if desired

### Checklist - did you ...

      stability policy?

### DIFF
--- a/.github/workflows/diff_shades.yml
+++ b/.github/workflows/diff_shades.yml
@@ -129,14 +129,14 @@ jobs:
           path: ${{ matrix.target-analysis }}
 
       - name: Generate summary file (PR only)
-        if: github.event_name == 'pull_request' && matrix.mode == 'preview-changes'
+        if: github.event_name == 'pull_request' && matrix.mode == 'preview-new-changes'
         run: >
           python helper.py comment-body ${{ matrix.baseline-analysis }}
           ${{ matrix.target-analysis }} ${{ matrix.baseline-sha }}
           ${{ matrix.target-sha }} ${{ github.event.pull_request.number }}
 
       - name: Upload summary file (PR only)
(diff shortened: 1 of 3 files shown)


cobaltt7:
Seems like it's hanging and eventually timing out. Probably a bug with string_processing

cobaltt7:
I re-ran the workflow a few times, as well as ran it locally, and couldn't reproduce the issue, so it must've been a one-off failure

ichard26:
> I wasn't sure if the "preview-changes" mode was meant to be interpreted as "Changes in the preview style" or "Preview new changes". I took it to mean the former and renamed it to unstable-style-changes for clarity, but I can change that if desired

As the person who devised diff-shades, I'll say that the intention with the `preview-changes` job was to show the impact of a given PR on Black's code style. That's why the baseline revision is the latest commit on main. In addition, it was merely meant to help reviewers determine whether a style change looked good or not using real world codebases. It wasn't meant to show exactly how downstream projects would be affected. In practice, I doubt many projects run Black with the preview style enabled. 

I'm honestly not sure how your two interpretations are meaningfully different, but I hope my extra context is useful nonetheless.

ichard26 (commented):
I've been an inactive maintainer for a long time now, but I don't have any concerns. As long as you've tested this and it has worked OK, I'd say you're good to press the :green_circle: button.

PR #4745, opened 2025-09-11
Fix mypy type issue

cobaltt7:
### Description

Fix type issue introduced in https://github.com/python/mypy/pull/19573

### Checklist - did you ...

      stability policy?

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -50,7 +50,7 @@ jobs:
       # Keep cibuildwheel version in sync with below
       - name: Install cibuildwheel and pypyp
         run: |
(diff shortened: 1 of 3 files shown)


cobaltt7:
diff-shades fail is expected because it can't build baseline revision. Don't know about wheels tests

JelleZijlstra:
I saw some errors about `__new__` in the test output. Can you try removing the `__new__` methods in `pytree.py`? Might also be able to fix them by using `super()`, but I don't think the methods provide much value anyway.

cobaltt7:
That didn't fix most of the issues. It looks like mypy 1.18 broke a lot, and pinning to 1.17 for now worked

PR #4750, opened 2025-09-18
Fix schema generation error caused by new click version

MeGaGiGaGon:
### Description

Not sure on the policy for updating versions, so to cope with the new click 8.3.0 version I used some code from #4577 for version conditional behavior. When >= 8.3.0, unset defaults are now the special UNSET sentinal instead of None. Hopefully nothing breaks on the click._utils usage since it's an _ module, might have to find a different solution if it does.

### Checklist - did you ...

- [N/A] Implement any code style changes under the `--preview` style, following the
      stability policy?
- [N/A] Add an entry in `CHANGES.md` if necessary?
- [N/A] Add new / update outdated documentation?

### DIFF
--- a/scripts/generate_schema.py
+++ b/scripts/generate_schema.py
@@ -1,7 +1,9 @@
 import json
 from typing import IO, Any
+from importlib.metadata import version as imp_version
 
 import click
+from packaging.version import Version
 
 import black
 
@@ -38,6 +40,9 @@ def generate_schema_from_click(
         result[name]["description"] = param.help
 
         if param.default is not None and not param.multiple:
+            if Version(imp_version("click")) >= Version("8.3.0"):
+                if param.default is click._utils.UNSET:
+                    continue
             result[name]["default"] = param.default
 
     return result


MeGaGiGaGon:
Looks like this does have mypy issues despite working to fix the issue

scripts/generate_schema.py
```diff
@@ -38,6 +40,9 @@ def generate_schema_from_click(
         result[name]["description"] = param.help
 
         if param.default is not None and not param.multiple:
+            if Version(imp_version("click")) >= Version("8.3.0"):
```
JelleZijlstra, line 43: Could also do something like `hasattr(click, "_utils") and hasattr(click._utils, "UNSET")`, though I guess it will still need a mypy type ignore. The current code is fine with me too.

Normally I'd be concerned about breaking future users if click removes or moves the private object later, but this is just an internal script; if it breaks later we can just fix it and users won't be affected.

MeGaGiGaGon: I searched around for a while but I don't see a better way of doing it that doesn't involve a mypy ignore, so :shrug:.
I like the current code since it says that whatever the problem/change is, it starts in click 8.3.0.

JelleZijlstra (approved):

Issue #4208, opened 2024-02-05
Future of `string_processing`


JelleZijlstra:
We merged string processing in May 2020 (#1132), almost four years ago. A few months later, I put the feature behind a flag because there were too many related bugs (#1609). Four years later, we're still in basically the same state: most of the open issues in https://github.com/psf/black/labels/C%3A%20preview%20style are about string processing.

I'm also increasingly convinced that string_processing's core feature, splitting up long strings, would be exceedingly controversial if we ever did turn it on by default. It is just too hard for a tool to figure out how to split a string in a way that makes sense to readers of the code. However, there are other things that string_processing does, and some of those are valuable and could perhaps be split off.

So my proposal is:
- Give up on splitting strings automatically; Black will simply not do that
- Split out any other useful features from string-processing into separate preview features that we can reasonably hope to enable in next year's stable style.

If you're interested in getting any part of string_processing into stable Black, please contribute a PR moving it to a separate preview feature.

I ran string_processing on my work codebase and noticed a few specific changes that could be split out. I'll list them in a separate comment.

JelleZijlstra:
Here are some examples of changes that string-processing makes in my codebase that I'd consider accepting in a standalone preview feature:

* Add parentheses around a string so it fits in the line limit
```
         xxxxxx = xx.xxxxxxx.XXXXXxxxxXxxXxfxxxxxx(
-            xxxxx_xxxxxxx_x3_xxxx_xxxxxxxx=f"x3://{XXX_X3_XXXXXX}/{XXXX_XXXXXXX_X3_XXXX_XXXXXXXX}",
+            xxxxx_xxxxxxx_x3_xxxx_xxxxxxxx=(
+                f"x3://{XXX_X3_XXXXXX}/{XXXX_XXXXXXX_X3_XXXX_XXXXXXXX}"
+            ),
             xxxxx_xxx=[
```
* Merge implicitly concatenated strings if the result fits in the line limit
```
     xxx_xxxxxxx_xxxxxxxx = (
-        "XXX(XX(xx XXXXXXX '{xxxxx_xx}' XXX '{xxx_xx}', {xxx}, 0)) "
-        "XX {xxx}_x{xxxxxx}"
+        "XXX(XX(xx XXXXXXX '{xxxxx_xx}' XXX '{xxx_xx}', {xxx}, 0)) XX {xxx}_x{xxxxxx}"
     )
```
* Combination of both, looks much nicer
```
     name_to_base_expression = {
[...]

PR #4760, opened 2025-09-20
Preview multiline string handling

cobaltt7:
### Description

In #4657, I fixed most issues with `multiline_string_handling`. Now that the fixes have been released, I think it makes sense to move it back to `--preview` for the next release, and hopefully into stable for v26.

As detailed in https://github.com/psf/black/issues/4159#issuecomment-2661780172, there is no good way to auto-fix the now unaligned indentation in multiline strings, and hopefully moving this feature to `--preview` will help us gather thoughts from users.

This may not belong in this PR, but I also modified the Future Style docs to have separate sections for the preview & unstable styles, as there are now multiple preview features with longer explanations. (I can make that change in a separate PR if wanted)
I also corrected the docs inconsistency I mentioned in https://github.com/psf/black/issues/4208#issuecomment-3315174458.

### Checklist - did you ...

      stability policy?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Move `multiline_string_handling` from `--unstable` to `--preview` (#4760)
+
 ### Configuration
 
 <!-- Changes to how Black can be configured -->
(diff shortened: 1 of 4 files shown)


cobaltt7:
The diff-shades report does seem to change a lot, mostly for the better, but there were also some issues in #4159 that can't really be fixed easily, so feedback is welcome! My hope is that moving it to preview can help people be aware that we're considering this change.

PR #4767, opened 2025-09-26
ci: try running diff-shades runs concurrently

cobaltt7:
### Checklist - did you ...

      stability policy?


cobaltt7:
This PR didn't help with speed on the `unstable` style much (which I've reverted for now), but it's still a speed improvement overall. There *is* a race condition if one of the projects commits at the same time the workflow is run, there's a very small chance that `base` could get one commit and `target` could get the other, but I don't think it's a big deal. It'd be fixed just by running the workflow again, and fixing that for sure would require changes to diff-shades to clone projects in a separate step.
I also fixed the PR comment not showing up - it still doesn't show up on this PR because the comment workflow is always run from `main`, not `pull_request_target`.

PR #4773, opened 2025-10-04
Add arm architecture to published executable

### LINKED ISSUE #4418: Offer more architectures in GitHub releases

I'd like to be able to use the GitHub release version of black on arm64 Linux hosts, but I can't.

Produce an arm64 Linux GitHub release of black. Ideally, this would include changing the existing linux release to include the appropriate architecture in the release file name (presumably "x86-64").

There's not really an alternative here. I want to use the binary release because it's quick and easy to get running on a machine in CI. I don't have to worry about having Python installed, which version of Python is installed, etc.

sgaist:
### Description

Add Ubuntu-22.04 for arm as builder.

This will allow people using arm based machines to have access to the black binary in the same fashion as for x86_64.

Fixes #4418

### Checklist - did you ...

- [X] Implement any code style changes under the `--preview` style, following the
      stability policy?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,7 +25,7 @@
 
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
-- Now GitHub actions build arm binaries (#4773)
+- Releases now include arm64 Linux binaries (#4773)
 
 ### Parser
 


CHANGES.md
```diff
@@ -25,6 +25,8 @@
 
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
+- Now GitHub actions build arm binaries (#4773)
```
JelleZijlstra:
```suggestion
- Releases now include arm64 Linux binaries (#4773)
```

PR #4777, opened 2025-10-05
fix(3.12_generics_syntax): Fixed formatting of 3.12 generics syntax

### LINKED ISSUE #4740: Bug: fails to properly format constrained generics with new 3.12 syntax
**Describe the bug**
when formatting the following snippet
```python
def func1[T: (int, str)](a,): ...
```
we get
```python
def func1[T: (
    int,
    str,
)](a,): ...
```
we would expect:
```python
def func[T: (int, str)](
    a,
): ...
```

**black version**
```
black --version
black, 25.1.0 (compiled: yes)
Python (CPython) 3.12.7
```

Exgene:
Fix type parameter expansion when function parameters have trailing comma

Fixes #4740

## Problem
When a function with type parameters has a trailing comma in its parameter list, Black incorrectly expanded the type parameter brackets instead of the function parameter list:

**Before:**
```python
def func1[T: (int, str)](a,): ...

# Formatted to (incorrect):
def func1[
    T: (int, str)
](a,): ...
```

**After:**
```python
def func1[T: (int, str)](a,): ...

# Formats to (correct):
def func1[T: (int, str)](
    a,
): ...
```

## Root Cause
The issue was in `left_hand_split()` which processes both `LSQB` (square brackets for type parameters) and `LPAR` (parentheses for function parameters) as potential split points. When a trailing comma forced a split, it would sometimes choose the type parameter brackets instead of the function parameter parentheses.

## Solution
Modified `left_hand_split()` to avoid selecting parentheses `(` as split points when they occur inside type parameter brackets `[...]`. This ensures type parameters stay on the head line and the split happens at the function parameter list where the trailing comma actually exists.

Closes #4740

I need some reference on how these test cases work so if you could help me with that it would be amazing and ill add the relevant test cases for it

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Fix bug where python 3.12 generics syntax split line happens weirdly (#4740)
+
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,10 +14,9 @@
 
 <!-- Changes that affect Black's preview style -->
 
-- Fix bug where python 3.12 generics syntax split line happens weirdly (#4740)
-
 - Fix bug where module docstrings would be treated as normal strings if preceeded by
   comments (#4764)
+- Fix bug where python 3.12 generics syntax split line happens weirdly (#4777)
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@
 <!-- Changes that affect Black's preview style -->
 
 - Fix bug where module docstrings would be treated as normal strings if preceeded by
-  comments (#4764)
+  comments (#4777)
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,14 +10,14 @@
 
 <!-- Changes that affect Black's stable style -->
 
-- Fix bug where python 3.12 generics syntax split line happens weirdly (#4740)
-
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
 
+- Fix bug where python 3.12 generics syntax split line happens weirdly (#4740)
+
 - Fix bug where module docstrings would be treated as normal strings if preceeded by
-  comments (#4777)
+  comments (#4764)
 
 ### Configuration
 
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -834,7 +834,9 @@ def left_hand_split(
                 current_leaves = tail_leaves if body_leaves else head_leaves
             current_leaves.append(leaf)
             if current_leaves is head_leaves:
-                if leaf.type == leaf_type:
+                if leaf.type == leaf_type and not (
+                    leaf_type == token.LPAR and depth > 0
+                ):
                     matching_bracket = leaf
                     current_leaves = body_leaves
         if matching_bracket and tail_leaves:
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -835,8 +835,8 @@ def left_hand_split(
             current_leaves.append(leaf)
             if current_leaves is head_leaves:
                 if leaf.type == leaf_type and (
-                    Preview.fix_type_expansion_split
-                    and not (leaf_type == token.LPAR and depth > 0)
+                    Preview.fix_type_expansion_split not in mode
+                    or not (leaf_type == token.LPAR and depth > 0)
                 ):
                     matching_bracket = leaf
                     current_leaves = body_leaves
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -834,8 +834,9 @@ def left_hand_split(
                 current_leaves = tail_leaves if body_leaves else head_leaves
             current_leaves.append(leaf)
             if current_leaves is head_leaves:
-                if leaf.type == leaf_type and not (
-                    leaf_type == token.LPAR and depth > 0
+                if leaf.type == leaf_type and (
+                    Preview.fix_type_expansion_split
+                    and not (leaf_type == token.LPAR and depth > 0)
                 ):
                     matching_bracket = leaf
                     current_leaves = body_leaves
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -37,6 +37,7 @@ Currently, the following features are included in the preview style:
   normalize file newlines both from and to.
 - `fix_module_docstring_detection`: Fix module docstrings being treated as normal
   strings if preceeded by comments.
+- `fix_type_expansion_split`: Fix type expansions split in generic functions.
 
 (labels/unstable-features)=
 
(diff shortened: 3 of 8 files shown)


tusharsadhwani:
Use this contributing basics section to setup the project: https://black.readthedocs.io/en/latest/contributing/the_basics.html#technicalities

and then run `pytest`.

For adding the new tests, follow the "Testing" section on the same page.

tusharsadhwani:
Also please add a simple explanation entry in the changelog, take a peek at merged PRs to get an idea.

Exgene:
> Use this contributing basics section to setup the project: https://black.readthedocs.io/en/latest/contributing/the_basics.html#technicalities
> 
> and then run `pytest`.
> 
> For adding the new tests, follow the "Testing" section on the same page.

Oh no i have done the testing setup and ran the tests no regressions seem to be found, I just didn't understand how the test cases quite worked so i wanted references on that to add for this particular case

Exgene:
> Also please add a simple explanation entry in the changelog, take a peek at merged PRs to get an idea.

Yeah sure on it, Thanks

tusharsadhwani:
From the doc: 

> Normally, tests should be created as files in the `tests/data/cases` directory. These files consist of up to three parts:
> [...]

Create a new file to create a new test. You can start with just the unformatted and formatted code snippet from your PR description, and then expand from there to cover more cases.

Exgene:
> From the doc:
> 
> > Normally, tests should be created as files in the `tests/data/cases` directory. These files consist of up to three parts:
> > [...]
> 
> Create a new file to create a new test. You can start with just the unformatted and formatted code snippet from your PR description, and then expand from there to cover more cases.

Yeah i got it, Didn't really see the section which explains how it works, Also i don't think this minor change requires an entry in the changes.md?

tusharsadhwani:
If that docs section felt like it's lacking context you can edit and improve it in another PR, happy to review.

Any change that has a user visible effect is a significant enough change to warrant a changelog entry IMO.

Exgene:
> If that docs section felt like it's lacking context you can edit and improve it in another PR, happy to review.
> 
> Any change that has a user visible effect is a significant enough change to warrant a changelog entry IMO.

Yes added basic test cases and updated the changelog to reflect the issue to be fixed should be okay now, Thanks!

tusharsadhwani:
Make sure not to format the test file, you seem to have changed the test file more than you intended to.

Also you should add a few more complex test cases that are adjacent to this bug, to ensure we handle those well too and don't regress in the future. Single or many item tuples for example, and maybe nested tuples.

Exgene:
> Make sure not to format the test file, you seem to have changed the test file more than you intended to.
> 
> Also you should add a few more complex test cases that are adjacent to this bug, to ensure we handle those well too and don't regress in the future. Single or many item tuples for example, and maybe nested tuples.

Oh right im dumb i ran the formatter on the test files, Will revert and add more test cases and try them

Exgene:
Updated with multiple tests as well and made the required changes, Let me know if any other changes are required

MeGaGiGaGon:
Thanks for the PR, I'm surprised it was that simple of a change. One more thing before this is good to go, since my previous fix for this already fixed the crash, and this code bug does not cause a crash (as far as I can tell), this will have to go in preview to prevent changing the stable style. #4764 is an example of how to do that, but the basic process is:
1. Add a new enum variant to `Preview` in `mode.py`
2. Use `if Preview.<variant name> in self.mode` to gate the new behavior behind preview
3. Move tests to a new preview specific file starting with `# flags: --preview`
4. Update `future_style.md` with a short description of the new mode changes
5. Update `black.schema.json` with the new preview name

Sorry that it's a bit of a headache to do, but we've recently been a bit loose with the stability policy, and I've been trying to make sure we follow it more closely.

For a bit more info, here's a link to the [stability policy](https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy).

Exgene:
> Thanks for the PR, I'm surprised it was that simple of a change. One more thing before this is good to go, since my previous fix for this already fixed the crash, and this code bug does not cause a crash (as far as I can tell), this will have to go in preview to prevent changing the stable style. #4764 is an example of how to do that, but the basic process is:
> 1. Add a new enum variant to `Preview` in `mode.py`
> 2. Use `if Preview.<variant name> in self.mode` to gate the new behavior behind preview
> 3. Move tests to a new preview specific file starting with `# flags: --preview`
> 4. Update `future_style.md` with a short description of the new mode changes
> 5. Update `black.schema.json` with the new preview name
> 
> Sorry that it's a bit of a headache to do, but we've recently been a bit loose with the stability policy, and I've been trying to make sure we follow it more closely.
> 
> For a bit more info, here's a link to the [stability policy](https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy).

Yeah no problem thats honestly fair, Ill do that thanks!

Exgene:
> Thanks for the PR, I'm surprised it was that simple of a change. One more thing before this is good to go, since my previous fix for this already fixed the crash, and this code bug does not cause a crash (as far as I can tell), this will have to go in preview to prevent changing the stable style. #4764 is an example of how to do that, but the basic process is:
> 
>     1. Add a new enum variant to `Preview` in `mode.py`
> 
>     2. Use `if Preview.<variant name> in self.mode` to gate the new behavior behind preview
> 
>     3. Move tests to a new preview specific file starting with `# flags: --preview`
> 
>     4. Update `future_style.md` with a short description of the new mode changes
> 
>     5. Update `black.schema.json` with the new preview name
> 
> 
> Sorry that it's a bit of a headache to do, but we've recently been a bit loose with the stability policy, and I've been trying to make sure we follow it more closely.
> 
> For a bit more info, here's a link to the [stability policy](https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy).

How do i gate this cause i can't really access the self.mode in `linegen.py` where the only change is outside the class i think?

Exgene:
@MeGaGiGaGon Let me know if this works

src/black/linegen.py
```diff
@@ -834,7 +834,10 @@ def left_hand_split(
                 current_leaves = tail_leaves if body_leaves else head_leaves
             current_leaves.append(leaf)
             if current_leaves is head_leaves:
-                if leaf.type == leaf_type:
+                if leaf.type == leaf_type and (
+                    Preview.fix_type_expansion_split
+                    and not (leaf_type == token.LPAR and depth > 0)
```
MeGaGiGaGon:
```suggestion
                    Preview.fix_type_expansion_split not in mode
                    or not (leaf_type == token.LPAR and depth > 0)
```
That should be it, looks like the function already takes `mode` as a parameter, so we can just use it.

CHANGES.md
```diff
@@ -10,12 +10,14 @@
...
+
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
 
 - Fix bug where module docstrings would be treated as normal strings if preceeded by
-  comments (#4764)
+  comments (#4777)
 
 ### Configuration
```
MeGaGiGaGon, line 22: The change entry for this PR should be moved to preview style, and the PR number for the other entry left unchanged

MeGaGiGaGon (changes requested):

Exgene:
@MeGaGiGaGon Yes done

Exgene:
Oh right it won't working on python version below 3.12 right i should update that for the test cases

MeGaGiGaGon (approved):

PR #4784, opened 2025-10-06
Try multiprocessing parallelism improvement

### LINKED ISSUE #4771: potential multiprocessing performance improvement
I swear this isn't just an advertisement for my youtube video... but along the same lines of my youtube video [multiprocessing: slow things first](https://youtu.be/h4vmU2DJoQg) I believe there's some optimization potential to the scheduling of files for black

I initially thought about this a bit when receiving a report from [freecad](https://github.com/freecad/freecad) about slow execution of black in pre-commit.ci -- (sorta unrelated to this) after looking into it I noticed two significant improvements to their setup: (1) use the pre-compiled black (next pre-commit.ci autoupdate will do this fleetwide automatically!) and (2) there were a bunch of very large files which black was spending most of its time on 

the second is interesting to me and where I reminded myself of potential optimizations!  the tl;dr of the optimization is to attempt "slower" things first such that a scheduler can appropriately binpack smaller things later (imagine the worst case scenario where the last thing you schedule is the slowest one and you've significantly extended your execution time while having mostly idle cores)

I decided to try an experiment with black and see if applying a similar optimization could improve performance.  my numbers below are on a VM with 12 cores against [this version of freecad](https://github.com/freecad/freecad/commit/bae76a4dc0a0cca79a55eec5b093258e34b95ba1) with this diff applied:

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 101f719e..8e277c8d 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,7 +44,6 @@ exclude: |
         src/Gui/3Dconnexion/navlib|
         src/Gui/QSint|
         src/Gui/Quarter|
-        src/Mod/Fem/femexamples|
         src/Mod/Import/App/SCL|
         src/Mod/Import/App/SCL_output|
         src/Mod/Mesh/App/TestData|
@@ -70,3 +69,13 @@ repos:
     rev: 719856d56a62953b8d2839fb9e851f25c3cfeef8  # frozen: v21.1.2
     hooks:
         -   id: clang-format
+-   repo: local
+    hooks:
+    - id: black-local
+      name: black
+      description: "Black: The uncompromising Python code formatter"
+      entry: black --line-length 100
+      language: system
+      minimum_pre_commit_version: 2.9.2
+      require_serial: true
+      types_or: [python, pyi]
```

and I wrote a little shell script to run this:

```bash
#!/usr/bin/env bash
set -euxo pipefail

pip uninstall -yqq black
pip install -qq /tmp/black --no-deps

rm -rf ~/.cache/black
pre-commit run black-local --all-files --verbose
```

(note this uses non-precompiled black -- I suspect the numbers are similar but maybe a little less pronounced on a precompiled black)

(note also I haven't intentionally chosen a terrible set of files -- if I wanted a worst case scenario I'd probably delete every file after the ones in `src/Mod/Fem/femexamples` such that it schedules those last)

first a little baseline -- I ran this a few times and this particular run seemed representative (again, not very scientific performance benchmark but I was mostly just exploring the idea)

```console
$ bash t.sh 
+ pip uninstall -yqq black
+ pip install -qq /tmp/black --no-deps
+ rm -rf /home/asottile/.cache/black
+ pre-commit run black-local --all-files --verbose
black....................................................................Passed
- hook id: black-local
- duration: 32.82s

All done! ✨ 🍰 ✨
1002 files left unchanged.

```

after applying this patch to black (which naively assumes that larger files will be slower -- I think this is a pretty good assumption):

```diff
diff --git a/src/black/concurrency.py b/src/black/concurrency.py
index f6a2b8a..fd8db5f 100644
--- a/src/black/concurrency.py
+++ b/src/black/concurrency.py
@@ -164,7 +164,7 @@ async def schedule_formatting(
                 executor, format_file_in_place, src, fast, mode, write_back, lock
             )
         ): src
-        for src in sorted(sources)
+        for src in sorted(sources, key=lambda f: (-f.stat().st_size, f))
     }
     pending = tasks.keys()
     try:
```

I get a pretty significant improvement! (again: anecdotal -- though a few runs were similar in timing)

```console
$ bash t.sh 
+ pip uninstall -yqq black
+ pip install -qq /tmp/black --no-deps
+ rm -rf /home/asottile/.cache/black
+ pre-commit run black-local --all-files --verbose
black....................................................................Passed
- hook id: black-local
- duration: 26.94s

All done! ✨ 🍰 ✨
1002 files left unchanged.

```

so maybe there's some potential here?

MeGaGiGaGon:
### Description

Trying the change from #4771 , I'd like to see if this helps reduce the diff-shades runtime, since the 22% speed increase from #4771 could be enough to make the action consistent again.

### Checklist - did you ...

      stability policy?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,8 @@
 
 <!-- Changes that improve Black's performance. -->
 
+- Format files from largest to smallest to improve benefits from concurrency (#4784)
+
 ### Output
 
 <!-- Changes to Black's terminal output and error messages -->
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -904,7 +904,7 @@ def reformat_one(
         else:
             cache = Cache.read(mode)
             if write_back not in (WriteBack.DIFF, WriteBack.COLOR_DIFF):
-                if not cache.is_changed(src):
+                if not cache.is_changed(src)[0]:
                     changed = Changed.CACHED
             if changed is not Changed.CACHED and format_file_in_place(
                 src, fast=fast, write_back=write_back, mode=mode, lines=lines
--- a/src/black/cache.py
+++ b/src/black/cache.py
@@ -8,7 +8,7 @@ import tempfile
 from collections.abc import Iterable
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import NamedTuple
+from typing import NamedTuple, Optional, Union
 
 from platformdirs import user_cache_dir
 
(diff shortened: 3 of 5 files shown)


JelleZijlstra (commented):
Seems fine to do, could also add a changelog entry since this likely will help perf in some cases.

I think Shantanu commented that we could reuse stat results from the cache logic, that might speed it up a bit more (avoiding syscalls).

ewest2350-ux:
why are they acting like a bitch like awe aint rebel before ??
thanks for the lesson

cobaltt7:
Looking at diff-shades, there is a noticeable difference in speed when running bigger files first. Stable style went from 6:51 to 4:23. However, preview style only went from 7:19 to 7:10.
Also, it may not be worth it to try reusing the `stat`s from the cache logic; they have to be passed between functions in weird ways. I got it working currently, but the cache tests need to be changed.
I'm not sure how reliable [the old diff-shades run](https://github.com/psf/black/actions/runs/18505365672/job/52732740064?pr=4784) from before `stat`s were reused is, but in that run, stable style increased from 6:13 to 6:23.

cobaltt7:
Closing as stale, new benchmarks should be run on current head compiled with mypyc. There's also probably better ways to implement it

PR #4805, opened 2025-10-22
t-string support

### LINKED ISSUE #4797: Python 3.14 support for black
**Describe the bug**

black doesn't support Python 3.14 syntax.

**To Reproduce**

For example, take this code:

```python
name = "world"
print(t"hello {name}")
```

```sh
$ black file.py --target-version py314
```

The resulting error is:

> error: cannot format t_string.py: Cannot parse for target version Python 3.14: 3:7: print(t"hello {name}")

**Expected behavior**

I expect black to be able to parse this.

**Environment**

- Black's version: 25.9.0
- OS and Python version: macOS 15.6.1, Python (CPython) 3.13.5

**Additional context**

None.

tusharsadhwani:
Resolves #4797 

What's left:

- [x] Formatting t-string quotes
- [x] Adding a new `T_STRINGS` Feature flag and gate it for `py314`
- [x] Fixing updating `python_symbols` types
- [x] Adding more tests
  - [x] for how t-string syntax differs from f-string syntax (so far it seems like all fstrings are valid t strings and vice versa)
  - [x] ensuring t-strings and f-strings can be nested in each other
- [x] Release `pytokens` and undo the `[tool.hatch.metadata]` change

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1248,6 +1248,7 @@ def _format_str_once(
         for feature in {
             Feature.PARENTHESIZED_CONTEXT_MANAGERS,
             Feature.UNPARENTHESIZED_EXCEPT_TYPES,
+            Feature.T_STRINGS,
         }
         if supports_feature(versions, feature)
     }
@@ -1364,6 +1365,8 @@ def get_features_used(  # noqa: C901
     for n in node.pre_order():
         if n.type == token.FSTRING_START:
             features.add(Feature.F_STRINGS)
+        elif n.type == token.TSTRING_START:
+            features.add(Feature.T_STRINGS)
         elif (
             n.type == token.RBRACE
             and n.parent is not None
--- a/src/black/mode.py
+++ b/src/black/mode.py
@@ -52,9 +52,10 @@ class Feature(Enum):
     DEBUG_F_STRINGS = 16
     PARENTHESIZED_CONTEXT_MANAGERS = 17
     TYPE_PARAMS = 18
-    FSTRING_PARSING = 19
+    # FSTRING_PARSING = 19  # unused
     TYPE_PARAM_DEFAULTS = 20
     UNPARENTHESIZED_EXCEPT_TYPES = 21
+    T_STRINGS = 22
     FORCE_OPTIONAL_PARENTHESES = 50
 
     # __future__ flags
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -53,7 +53,7 @@ repos:
           # v8.2 has breaking changes. We work around them at runtime, but we need the newer stubs.
           - packaging >= 22.0
           - platformdirs >= 2.1.0
-          - pytokens >= 0.1.10
+          - pytokens @ git+https://github.com/tusharsadhwani/pytokens@py314
```
cooperlees: We don't want to push that to PyPI first?

pyproject.toml
```diff
@@ -98,6 +98,9 @@ Changelog = "https://github.com/psf/black/blob/main/CHANGES.md"
 Repository = "https://github.com/psf/black"
 Issues = "https://github.com/psf/black/issues"
 
+[tool.hatch.metadata]
+allow-direct-references = true
```
cooperlees: nit: Do we want a comment as to why we are adding?

tusharsadhwani: PR is still WIP, once I've added all test cases and things pass I'll publish `pytokens` and then bump it here. This hatch change will also go away once we start using a canonical published version of `pytokens`

cooperlees: Ok cool, this is what I thought. Because the PR was not draft I thought it was ready ...

tests/data/cases/pep_750.py
```diff
@@ -0,0 +1,7 @@
+# flags: --minimum-version=3.14
```
matgrioni, line 1: Totally random person here with T-String issues in black :)

Consider adding test case for multi-line t-strings as well (as in t""" """), because I am using some tools that have an issue specifically with this construction.

tusharsadhwani: Will do! They do work just fine, I'm currently working on making t strings work inside f strings, but yes the test file needs a lot more cases which I plan on adding.

tusharsadhwani:
Correct, it's perfectly backwards compatible

cooperlees (approved):
Awesome. Thanks for adding in the extra tests and releasing `pytokens`

PR #4811, opened 2025-10-26
Fix #1245: Preserve comment formatting in fmt:off/on blocks

### LINKED ISSUE #1245: Comments between fmt:on/off are reformatted
This code:
```python
# fmt: off
 #nospace
 #  twospaces
# fmt: on
```
is changed to:
```python
# fmt: off
# nospace
#  twospaces
# fmt: on
```
I expected the lines betwen `# fmt: off` and `# fmt: on` to remain unchanged.

* Version: master (19.10b1.dev22+gbe49ac7)
* OS: Ubuntu Linux
* Python: 3.6.9

[Playground (master)](https://black.now.sh/?version=master&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ABvAFBdAD2IimZxl1N_Wg0-Ef3tYndKG4aPypUA1Ayrzf5zNZvLSGAENc30wvA2bEeMylIvqRdhWLTqYmXqf_IjzG5SN6r67wWKI0-xLCZn7fylTZeAABDeQELeGAvAAAFscLzgy7kftvN9AQAAAAAEWVo=)

elijahr:
### Description

  This fixes issue #1245 where comments between `# fmt: off` and `# fmt: on` directives were being reformatted even though they should be left untouched. Additionally, it fixes the inconsistent behavior where comments
  containing fmt directives (like `#fmt:skip`) were being normalized even though the directive's purpose is to prevent formatting.

  **Root Causes:**
  1. When ONLY comments exist between fmt directives (no code), `generate_ignored_nodes()` returns empty because comments don't create AST nodes - they exist as strings in node prefixes.

  2. Multiple consecutive fmt:off/on blocks in the same function were not all being preserved due to:
     - Identity check bug: After modifying the prefix, new comment objects were created, causing the identity check to fail
     - Whitespace check bug: STANDALONE_COMMENT nodes weren't treated as whitespace for standalone detection

  3. Comments containing fmt directives were being normalized by `make_comment()` even though the directive indicates formatting should be skipped.

  **Changes:**
  - Modified `normalize_fmt_off()` to detect and preserve comment-only blocks
  - Fixed logic to find fmt:off by value check instead of identity check
  - Treat STANDALONE_COMMENT as whitespace for standalone detection
  - Enhanced `make_comment()` to preserve exact formatting of comments containing any fmt directive (skip/off/on/yapf)
  - Added test coverage in `fmtonoff.py` and `fmtskip7.py`

  **This now correctly handles:**
  - File-level and function-level comment-only blocks
  - Multiple consecutive fmt:off/on blocks
  - Comments with unusual indentation/spacing
  - Comments containing fmt directives preserve exact formatting (e.g., `#fmt:skip` stays as-is instead of becoming `# fmt:skip`)

  **Note on style changes:**
  The second fix (preserving fmt directive comment formatting) does change Black's output, but we believe this is the correct and consistent behavior: if a user writes `#fmt:skip` or `#    fmt: skip` to skip formatting, it's inconsistent for
  Black to reformat the comment itself which is part of the line. This aligns with user intent.

  All 408 tests pass.

  ### Checklist - did you ...

    - **Partial** - The comment-only block fix is purely a bug fix. The fmt directive comment preservation could be considered a style change, but it's more about consistency: honoring the directive's intent to skip
  formatting. We've documented this clearly in CHANGES.md.

    - **Yes** - Added two entries in "Stable style" section, clearly noting the fmt directive formatting change

    - **Yes** - Comprehensive test coverage added to `fmtonoff.py`

    - **N/A** - No documentation changes needed; fixing behavior to be consistent with directive intent

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,10 +15,9 @@
 
 <!-- Changes that affect Black's stable style -->
 
-- Fix bug where comments between `# fmt: off` and `# fmt: on` were reformatted (#1245)
-- Comments containing fmt directives (`fmt:skip`, `fmt:off`, `fmt:on`) now preserve
-  their exact formatting instead of being normalized. For example, `#fmt:skip` is no
-  longer changed to `# fmt:skip` (#1245)
+- Fix bug where comments between `# fmt: off` and `# fmt: on` were reformatted (#4811)
+- Comments containing fmt directives now preserve their exact formatting instead of
+  being normalized (#4811)
 
 ### Preview style
 
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -604,8 +604,12 @@ def _contains_fmt_directive(
     comment_line: str, directives: set[str] = FMT_OFF | FMT_ON | FMT_SKIP
 ) -> bool:
     """
-    Checks if the given comment contains format directives alone or paired with other comments.
-    Defaults to checking all directives (skip, off, on, yapf), but can be narrowed to specific ones.
+    Checks if the given comment contains format directives, alone or paired with
+    other comments.
+
+    Defaults to checking all directives (skip, off, on, yapf), but can be
+    narrowed to specific ones.
+
     Matching styles:
       # foobar                    <-- single comment
       # foobar # foobar # foobar  <-- multiple comments


cobaltt7:
I made some refactors to make the code and diff cleaner, and resolved the CI issues, so as soon as it's green, this looks good to merge. Thanks for contributing!

Although this changes Black's output, it does so trivially. Plus, this actually doesn't break the stability policy, since pre-formatted code will not change. So it doesn't need to be gated under preview.

PR #4814, opened 2025-10-26
Add arm64 windows builds and tests

### LINKED ISSUE #4813: Add support for arm64 windows

I saw that arm64 support was recently added for linux, and since github actions now has [windows arm64 runner support](https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/) for public repos (`windows-11-arm`) would there be any appetite for supporting windows arm64 builds as well?

Support of native arm64 windows wheels and executable

Not sure if there is an alternative to the binary, but the pure python wheel could be an alternative to the native arm64 windows wheel.

finnagin:
### Description

     
Closes #4813 

This would add native arm64 wheel and binary builds as well as add add arm64 windows to the ci tests and the release tests.

### Checklist - did you ...

      stability policy?

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -98,8 +98,6 @@ jobs:
       - uses: pypa/cibuildwheel@v3.2.1
         with:
           only: ${{ matrix.only }}
-        env:
-          CIBW_TEST_REQUIRES_WINDOWS: "pytest^<9.0"
 
       - name: Upload wheels as workflow artifacts
         uses: actions/upload-artifact@v5
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,20 +1,13 @@
 # Change Log
 
-## Unreleased
-
-<!-- PR authors:
-     Please include the PR number in the changelog entry, not the issue number -->
+## 25.11.0
 
 ### Highlights
-
-<!-- Include any especially major or disruptive changes here -->
 
 - Enable base 3.14 support (#4804)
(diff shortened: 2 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -31,8 +31,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12.4", "3.13", "3.14", "pypy-3.9"]
-        os: [ubuntu-latest, macOS-latest, windows-latest]
+        python-version: ["3.9", "3.10", "3.11", "3.12.10", "3.13", "3.14", "pypy-3.9"]
```
finnagin, line 34: I bumped the 3.12 version from 3.12.4 to 3.12.10 here since the `setup-python` action fails to install 3.12.4 on arm64 windows and when I was looking back at the previous commit that set it to specify 3.12.4 it seemed to me it was to avoid an issue in 3.12.5 that was then fixed in 3.12.6.

.github/workflows/upload_binary.yml
```diff
@@ -38,7 +42,7 @@ jobs:
       - name: Set up latest Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.12.4"
+          python-version: "3.13"
```
finnagin, line 45: I changed this from 3.12.4 -> 3.13 so that the setup-python action does not fail when installing on arm64 windows. I could also change this to a later version of 3.12 like I did in `test.yml` if is that is preferable.

Issue #4826, opened 2025-11-09
tests are not compatiable with pytest 9.0


onerandomusername:
I'm not sure why its returned as an internal error when it happens in testing code, so that's also a bit odd. Might have to open a pytest issue as well.

Configuration was updated in pytest 9.0 (which now supports `[tool.pytest]` without the legacy ini_options in pyproject! :tada:) but that seems to break testing for black.

```py
py: commands[2]> pytest tests --run-optional no_jupyter --numprocesses auto --cov
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/qt/projects/psf/black/.tox/py/lib/python3.14/site-packages/_pytest/main.py", line 314, in wrap_session
INTERNALERROR>     config._do_configure()
INTERNALERROR>     ~~~~~~~~~~~~~~~~~~~~^^
INTERNALERROR>   File "/home/qt/projects/psf/black/.tox/py/lib/python3.14/site-packages/_pytest/config/__init__.py", line 1128, in _do_configure
INTERNALERROR>     self.hook.pytest_configure.call_historic(kwargs=dict(config=self))
INTERNALERROR>     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/qt/projects/psf/black/.tox/py/lib/python3.14/site-packages/pluggy/_hooks.py", line 534, in call_historic
INTERNALERROR>     res = self._hookexec(self.name, self._hookimpls.copy(), kwargs, False)
INTERNALERROR>   File "/home/qt/projects/psf/black/.tox/py/lib/python3.14/site-packages/pluggy/_manager.py", line 120, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>            ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/qt/projects/psf/black/.tox/py/lib/python3.14/site-packages/pluggy/_callers.py", line 167, in _multicall
INTERNALERROR>     raise exception
INTERNALERROR>   File "/home/qt/projects/psf/black/.tox/py/lib/python3.14/site-packages/pluggy/_callers.py", line 121, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
[...]

PR #4827, opened 2025-11-09
build: restrict to pytest 9.0 due to breakage in custom pytest_configure

onerandomusername:
Hotfix for getting testing back to a stable configuration without changing a bunch of the test internals right before a release :)

see GH-4826 for more details.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -162,7 +162,7 @@ skip = ["*-manylinux_i686", "*-musllinux_*", "*-win32", "pp*"]
 # This is the bare minimum needed to run the test suite. Pulling in the full
 # test_requirements.txt would download a bunch of other packages not necessary
 # here and would slow down the testing step a fair bit.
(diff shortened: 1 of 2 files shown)


cooperlees (approved):
Many thanks! Also, thanks for opening the issue so we (or I might as copilot) will fix it :)

cooperlees:
Huh, it still selected 9.0 - We must need it somewhere else. tox? If we can, let's try and get it so we only have to define a pin in one place, but that can also wait if it needs to much work ... 
- Maybe pre install pytest in the action using our requirements file?

Might be inbuilt to the mypyc building?:
`pip install virtualenv -c /constraints.txt` ... I guess need to workout what generates that

hauntsaninja:
Yeah, see https://github.com/psf/black/blob/main/pyproject.toml#L162-L165

cooperlees (approved):

PR #4831, opened 2025-11-10
Use build[uv] as cibuildwheel frontend

hauntsaninja:
Maybe this will make the pip errors on Windows go away


hauntsaninja:
Reverting this in https://github.com/psf/black/pull/4832 for Windows reasons

PR #4833, opened 2025-11-10
ci: add label for running all builds on a pull request

onerandomusername:
We currently don't run all builds on a pull request and only run them on main. This changes it to run all builds on a pull request when the 'ci: build all' label is added.

This is currently implemented in a hotfix manner which will only work once a commit is pushed *after* the label is added.

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -53,7 +53,8 @@ jobs:
           pipx install cibuildwheel==3.2.1
           pipx install pypyp==1.3.0
       - name: generate matrix
-        if: github.event_name != 'pull_request'
+        if: |
+          github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'ci: build all wheels')
         run: |
           {
             cibuildwheel --print-build-identifiers --platform linux \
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi_upload.yml
```diff
@@ -53,7 +53,9 @@ jobs:
           pipx install cibuildwheel==3.2.1
           pipx install pypyp==1.3.0
       - name: generate matrix
-        if: github.event_name != 'pull_request'
+        if: |
+          github.event_name != 'pull_request' ||
+          github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'ci: build all wheels')
```
cobaltt7: shouldn't this be checking if it does have the label?
also, `github.event_name == 'pull_request'` will always be true at this point
```suggestion
          contains(github.event.pull_request.labels.*.name, 'ci: build all wheels')
```

onerandomusername: https://github.com/psf/black/pull/4833/commits/579792d12bdd6f510833bc48a2d950149ccff751

cobaltt7 (approved):

PR #4834, opened 2025-11-10
Revert "Use build[uv] as cibuildwheel frontend"

cobaltt7:
Reverts psf/black#4831
uv isn't installed on windows


cobaltt7:
oops, i see this was added to #4832

PR #4846, opened 2025-11-18
docs: clarify behavior of  label and commit requirement

amritamishra01:
This PR adds documentation explaining the behavior of the `ci: build all wheels` label.

### Why this change is needed
The label exists to trigger a full wheel build on a PR, but GitHub Actions does not automatically rerun CI when a label is added.  
This leads to confusion because contributors expect the wheel build to start immediately, but it only works after pushing a new commit.

### What this PR adds
- A new section in `CONTRIBUTING.md` describing:
  - the purpose of the `ci: build all wheels` label,
  - the requirement to push a commit after applying the label,
  - how to trigger the workflow using an empty commit,
  - why GitHub Actions is not configured to run wheel builds directly on label events (security/resource reasons).

### Security and design considerations
This PR intentionally avoids modifying CI triggers because building wheels on label-added events can introduce security issues (e.g., running untrusted fork code with elevated permissions) and heavy CI load.  
Adding documentation is the safest and maintainer-aligned solution.

### References
- Discussion in issue #4838
- Closes #4857

PR #4857, opened 2025-11-22
Add CI-related label documentation to issue triage guide

amritamishra01:
Add CI-related label documentation to issue triage guide

Description

This PR adds documentation for the ci: build all wheels label to the Issue Triage guide.

The new section explains:

what the label is used for

why GitHub Actions does not trigger CI automatically when the label is added

the requirement to push a new commit (even an empty one) to start the wheel build

security and resource reasons for not triggering builds on label events

that this label should only be applied by maintainers or triage users

This aligns the documentation with current project behavior and helps avoid confusion for contributors.

Checklist

 Documentation updated

 No user-facing changes

 No test changes needed

Closes #4846

### DIFF
--- a/docs/contributing/issue_triage.md
+++ b/docs/contributing/issue_triage.md
@@ -97,20 +97,24 @@ rigorous about it. Just follow your judgement on what labels make sense for the
 specific PR (if any even make sense).
 ```
 
-```md
+````md
 ## CI-related labels
 
 ### Label: `ci: build all wheels`
 
-This label is used by maintainers or triage users to request a full wheel build for a pull request.
+This label is used by maintainers or triage users to request a full wheel build for a
+pull request.
 
 #### Important behavior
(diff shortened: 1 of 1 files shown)


amritamishra01:
I moved the CI-related label documentation into the Issue Triage guide as suggested.
Please let me know if you'd prefer a different location or structure — happy to adjust.

cobaltt7:
You already had a PR open, you didn't need to create a new one?

Also, this change has the entire section wrapped in a code block, that surely wasn't intentional. Was this written by AI? It's okay to use AIs to assist, but it's still required for you to check what it does and make sure nothing was done incorrectly or unnecessarily.

Most labels here are documented with a short, concise description. This label should be documented the same way. Include quick explanations on why all wheels aren't always built, when you would want to force all wheels to build, and when the workflow is triggered - it should only be about 2-3 sentences total. It's unnecessary to say only maintainers should use it - only maintainers can add labels anyway.

PR #4865, opened 2025-11-24
Remove unnecessary parentheses from L.H.S of assignment statements

### LINKED ISSUE #4725: LHS parenethesis could be removed sometimes
**Describe the style change**

If you have `(...) = something`, the LHS parenthesis, I think, should not be kept.

```python
def a():
    return 1, 2, 3

(b) = a()[0]
(c, *_) = a()
```

**Desired style**

```python
def a():
    return 1, 2, 3

b = a()[0]
c, *_ = a()
```

Nikhil172913832:
### Description

Closes #4725

This PR removes unnecessary parentheses on the left-hand side of assignments. Black currently keeps parentheses around single targets and ordinary unpacking patterns, for example `(b) = a()[0]` or `(c, *_) = a()`, even though they don’t change the semantics and just add noise.

The update adjusts `normalize_invisible_parens` so these parentheses are only preserved when they actually matter. They’re still kept in cases involving:
- a magic trailing comma that indicates intentional tuple syntax
- atoms that span multiple lines
- unpacking patterns that rely on star expressions

Two small helpers were added to detect trailing-comma atoms and multi-line atoms. The assignment path now uses `allow_star_expr` so unpacking works as expected. Relevant tests are included.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/diff_shades.yml
+++ b/.github/workflows/diff_shades.yml
@@ -163,7 +163,7 @@ jobs:
   compare:
     name: compare / ${{ matrix.mode }}
     needs: ["configure", "analysis-base", "analysis-target"]
-    if: !cancelled()
+    if: ${{ !cancelled() }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false


tests/data/cases/remove_parens_from_lhs.py
```diff
@@ -0,0 +1,35 @@
...
+
+# Single variable with unnecessary parentheses
+b = a()[0]
+
+# Tuple unpacking with unnecessary parentheses
+c, *_ = a()
+
+# These should not be changed - parentheses are necessary
+(d,) = a()  # single-element tuple
+e = (1 + 2) * 3  # RHS has precedence needs
```
cobaltt7, line 17: Did you accidentally format the input?

cobaltt7 (changes requested):
Thanks for the contribution! Since this changes Black's code style, it will have to be added to the Preview style. See commit https://github.com/psf/black/commit/82030302ea3d70f233d418d9abd3645d543a400c for an example of how that would be done. That should also theoretically fix the diff-shades failures (edit: actually it won't, https://github.com/ichard26/diff-shades/pull/20).

Please also add an entry to the changelog and run black on the codebase (ie, with `python src/black .` in the repo's root directory)

tests/data/cases/remove_parens_from_lhs.py
Nikhil172913832: Yes, I think I might have done that, I have now replaced the formatted code with the unformatted version in the original issue.

Nikhil172913832:
Thanks for the review. I’ve moved the change into the Preview style following 8203030, added the changelog entry, and reformatted with black. I also updated the docs, regenerated the schema, and added a test case under --preview. All tests pass.

gallery/gallery.py
```diff
@@ -103,9 +103,7 @@ def download_and_extract(package: str, version: str | None, directory: Path) ->
     return directory / result_dir
 
 
-def get_package(
-    package: str, version: str | None, directory: Path
-) -> Path | None:
+def get_package(package: str, version: str | None, directory: Path) -> Path | None:
```
cobaltt7, line 106:
```suggestion
def get_package(
    package: str, version: str | None, directory: Path
) -> Path | None:
```

Apparently this file isn't included in CI's formatted check; formatted it in #4867

cobaltt7:
diff-shades changes look good. Thanks for the contribution!

PR #4866, opened 2025-11-25
Try narrowing down mypyc issues

ilevkivskyi:


ilevkivskyi:
Actually nvm, I will bisect locally (this was a lucky guess btw).

PR #4870, opened 2025-11-26
Create test-automagic_parser.py

arrMan-dev:
Unit testing to fix the check automagic parser for syntax like ast.parse or IPython to be skipped

### Checklist - did you ...

      stability policy?


MeGaGiGaGon:
Thank you for the PRs, however this is not needed. We already have magic handling implemented, which you can find here https://github.com/psf/black/blob/main/src/black/handle_ipynb_magics.py .

PR #4878, opened 2025-12-01
Pin Hatch to hopefully fix Docker builds

cobaltt7:
### Description

Appears to fix the issue on my fork, but I had to modify the workflow to get it to run at all, so we'll see.
(thanks @hauntsaninja for noticing the new version!)

### Checklist - did you ...

      stability policy?


cooperlees:
I can possibly look into this more, but I wonder if we can set this in the pyproject.toml or something more central so docker just gets the pin cleaned up etc. when ready etc. etc. rather than hacking the `Dockerfile` moving forward ...

JelleZijlstra:
Might make sense as a dependency group for build deps?

PR #4881, opened 2025-12-04
wrap_comprehension_in: Always wrap if there's delimiters

### LINKED ISSUE #4877: Black preview style removes parentheses that make code harder to read
```
         filtered_trait_instructions: list[str] = [
-            trait for trait in (traits_instructions or []) if trait
+            trait for trait in traits_instructions or [] if trait
         ]
```

There are ones involving `await` that are a little more arguable

cobaltt7:
### Description

Background info: On current stable, Black does not add nor remove parens around `in` clauses in comprehensions at all. `wrap_comprehension_in` standardized it to always add them if it would then fit in the line limit, and to remove them otherwise.

Resolves #4877

My approach was to always wrap the `in` clause if there's delimiters in it (not counting dots). This resulted in changes like this in Black's codebase, which I think are better?
image
image

I'm not sure if this change is best. It makes `wrap_comprehension_in` much more intrusive, at the very least. Will have to look at shades.

cc @hauntsaninja 

### Checklist - did you ...

- [y] Implement any code style changes under the `--preview` style, following the
      stability policy?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 - Fix `string_processing` crashing on unassigned long string literals with trailing
   commas (one-item tuples) (#4929)
 - Simplify implementation of the power operator "hugging" logic (#4918)
+- Always wrap `in` clauses of comprehensions if they have delimiters (#4881)
 
 ### Configuration
 
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -15,8 +15,8 @@ from black.brackets import (
     DOT_PRIORITY,
     STRING_PRIORITY,
     get_leaves_inside_matching_brackets,
-    max_delimiter_priority_in_atom,
     max_delimiter_priority,
+    max_delimiter_priority_in_atom,
 )
 from black.comments import (
     FMT_OFF,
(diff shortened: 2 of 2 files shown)


cobaltt7:
I agree that this change probably shouldn't be made for v26, and we should either ship `wrap_comprehension_in` as-is for v26, or not at all. I think we should see what feedback we get from the beta and decide then.
However, another point to consider: Currently, Black doesn't standardize parens here at all. If we ship something this year and something else next year, it causes similar churn twice. So I think ideally we should ship it all in v26 (probably unrealistic), or all in v27.

JelleZijlstra:
For what it's worth, in the two examples in the OP I feel the extra parentheses make the formatting worse. In the diff-shades output there are some changes I like, but I find myself disliking most of the ones where the in clause has an operator like `+` and previously fit on one line. The parentheses look better if the in clause contains a boolean op (and/or), if there's an `if` clause, or if the in clause already spans multiple lines and was previously split within a call.

My ideal would be that we add parentheses only if needed to split the clause over multiple lines, remove parentheses only if the clause consists of a single extremely simple node (e.g., a single name, literal, or call), and otherwise preserve the user's decision.

Given that what we have on main right now arguably makes the code worse in some cases (#4877) and this proposed change is also not always clearly better, I agree it's better to back out this change from the stable style for this year. Hopefully for next year we can then get a better version in place.

PR #4883, opened 2025-12-05
Fix `# fmt: skip` ignored in deeply nested expressions

### LINKED ISSUE #4730: `# fmt: skip` ignored inside multi-part if-clause
**Describe the bug**

`#fmt: skip` is ignored in the specific case demonstrated below.

_Edit_: Actually, thinking about it again I realised that (to me), the actual bug is that black changes the code at all. However, I'd already be happy if I could at least keep it from disfiguring my code via `fmt: skip`.

**To Reproduce**

Let black format this piece of code:

```python
class ClassWithALongName:
    Constant1 = 1
    Constant2 = 2
    Constant3 = 3


def test():
    if (
        "cond1" == "cond1"
        and "cond2" == "cond2"
        and 1 in (  # fmt: skip
            ClassWithALongName.Constant1,
            ClassWithALongName.Constant2,
            ClassWithALongName.Constant3,
        )
    ):
        return True
    return False
```

**Expected behavior**

Leave the line marked with  `# fmt: skip` unchanged.
Instead, it moves the `in` to the next line, which I find confusing and hard to read:

```python
def test():
    if (
        "cond1" == "cond1"
        and "cond2" == "cond2"
        and 1
        in (  # fmt: skip
            ClassWithALongName.Constant1,
            ClassWithALongName.Constant2,
            ClassWithALongName.Constant3,
        )
    ):
        return True
    return False
```

**Environment**

Arch Linux
black, 25.1.0 (compiled: no)
Python (CPython) 3.13.5

Nikhil172913832:
### Description

Closes #4730

Black was ignoring `# fmt: skip` in deeply nested expressions (e.g., multi-part if-clauses). This change updates the tree climbing logic to correctly find the start of the line and all relevant siblings, ensuring that `# fmt: skip` works as intended even in complex nested cases.

### Checklist - did you ...

      stability policy?

### DIFF
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -802,22 +802,6 @@ def _generate_ignored_nodes_from_fmt_skip(
             ignored_nodes = [node for node in ignored_nodes if node.parent != parent]
             ignored_nodes.append(parent)
 
-            # If the atom's parent is a binary operation (comparison, arithmetic, etc.)
-            # and we've also captured siblings of the atom from the same parent, then
-            # include the entire parent to avoid leaving an incomplete binary operation
-            if (
-                parent.parent is not None
-                and isinstance(parent.parent, Node)
-                and any(
-                    node.parent == parent.parent and node != parent
-                    for node in ignored_nodes
-                )
-            ):
-                ignored_nodes = [
-                    node for node in ignored_nodes if node.parent != parent.parent
-                ]
-                ignored_nodes.append(parent.parent)
-
         yield from ignored_nodes
     elif (
         parent is not None and parent.type == syms.suite and leaf.type == token.NEWLINE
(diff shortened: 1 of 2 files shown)


cobaltt7:
The diff-shades doesn't look right, it looks like some changes from `multiline_string_handling` have been reverted. Could you take a look at that? (click the "available in the logs") link

Also, as a heads up, we're going to be releasing 25.12.0 over the weekend, and we want to keep 26.1.0 limited to purely stablizing preview styles as much as possible, so this probably won't get merged until after those are released.

Nikhil172913832:
Thanks for the review. I’ve made the required changes. Happy to wait for the right window after the upcoming releases.

cobaltt7:
Since this looks good and we haven't released yet, I'll go ahead and merge this now. Thanks!

PR #4884, opened 2025-12-05
Fix #4653 If guard explosion in case statement

### LINKED ISSUE #4653: Trailing comma in `case` pattern causes `if` guard to explode
Black is exploding the `if` guard in `case` patterns which have trailing commas in them, even if the guard expression fits in one line. The trailing comma logic should apply only to the `case` pattern, not the `if` guard.

**Black https://github.com/psf/black/commit/d0ff3b**

[Playground link](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AHyAKFdAD2IimZxl1N_WlbvK5V9HwFY_XA6IUcMIo1tyYCCoFcD4ulUXLciN-oOsLyTg3jeQOCSB4n_hjtP0pinvPpG8c3P-YF68pR7Qz8LQZlIYqo1mz-CYSOrKyew5V3b7Uw9DgX8-VXf4YUwtZKb8XT0dXeojsS0HOyiiLZEAtICZZXtPswUvANGrr3xUn2oIcjaOQ6s8akHAR0JfXassLFxWsAAAAAAACZkTt4v_8QtAAG9AfMDAADW-vTOscRn-wIAAAAABFla)

## Options

`--line-length=88`
`--safe`

## Input

```python
def f(x):
    match x:
        # good refactor
        case [
            y
        ] if y == 123:
            pass

        case [
            y
        ] if True:
            pass

        case [
            y,
        ] if True:
            pass

        # bad refactor
        case [
            y,
        ] if y == 123:
            pass


```

## Output

```python
def f(x):
    match x:
        # good refactor
        case [y] if y == 123:
            pass

        case [y] if True:
            pass

        case [
            y,
        ] if True:
            pass

        # bad refactor
        case [
            y,
        ] if (
            y == 123
        ):
            pass
```

## Expected

This case:

```python
    match x:
        case [
            y,
        ] if y == 123:
            pass
```

Should not be refactored, as the `if` guard can fit on one line. The pattern above is forcing a refactor as it has a trailing comma, which should not happen as the pattern and the guard are separate expressions. Using `--skip-magic-trailing-comma` "fixes" this, but also causes trailing commas to be ignored globally, which is not desirable.

I didn't catch this when I wrote #4214 because this issue only crops up when the pattern has trailing commas, and the `if` guard is a more complex expression.

Since I wrote #4214 I can fix this one as well.

rxjacob:
### Description

Black is exploding the if guard in case patterns which have trailing commas in them, even if the guard expression fits in one line. The trailing comma logic should apply only to the case pattern, not the if guard.

### Checklist - did you ...

- [X] Implement any code style changes under the `--preview` style, following the
      stability policy?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ d = ["aiohttp>=3.10"]
 jupyter = ["ipython>=7.8.0", "tokenize-rt>=3.2.0"]
 
 [dependency-groups]
-build = ["hatch==1.15.1", "hatch-fancy-pypi-readme", "hatch-vcs"]
+build = ["hatch==1.15.1", "hatch-fancy-pypi-readme", "hatch-vcs", "virtualenv<21.0.0"]
 wheels = ["cibuildwheel==3.3.1", "pypyp==1.3.0"]
 binary = ["pyinstaller", "wheel>=0.45.1"]
 


cobaltt7:
This code doesn't build.

rxjacob:
> This code doesn't build.

Updated the PR with the fix.

cobaltt7:
Thanks! Can you gate this change to the preview style, since it changes previous formatting?

rxjacob:
> Thanks! Can you gate this change to the preview style, since it changes previous formatting?

Done.

cobaltt7:
Sorry, please add and check for a specific preview style, instead of the entire preview style as a whole. Here's a recent example: https://github.com/psf/black/pull/4865/changes. To test it, add the line `# flags: --preview` to the top of the test file.

Also, it looks like there's some unintentional formatting changes in one of the other tests

src/black/mode.py
```diff
@@ -229,13 +229,15 @@ class Preview(Enum):
...
+    fix_if_guard_explosion_in_case_statement = auto()
 
 
 UNSTABLE_FEATURES: set[Preview] = {
     # Many issues, see summary in https://github.com/psf/black/issues/4208
     Preview.string_processing,
     # See issue #4036 (crash), #4098, #4099 (proposed tweaks)
     Preview.hug_parens_with_braces_and_square_brackets,
+    Preview.fix_if_guard_explosion_in_case_statement,
 }
```
cobaltt7, line 240: This shouldn't be in the unstable style unless it has known issues (and we wouldn't merge anything that already has known issues!)

```suggestion
    Preview.hug_parens_with_braces_and_square_brackets,
}
```

PR #4893, opened 2025-12-08
Revert "Fix `# fmt: skip` ignored in deeply nested expressions"

cobaltt7:
Reverts psf/black#4883

Sorry @Nikhil172913832, I missed a part of this that causes issues once I move this out of preview:

`--no-preview-line-length-1` is meant to be a cop-out for known broken tests that were already broken when tests started being checked with `--preview --line-length=1`. It's not meant to be used for newly-added tests.

Running with `--line-length=1` is a good way to find stability issues. No test should crash when ran with it, because Black should never crash when run on valid code.

Please feel welcome to fix the issue and recreate your PR for it to be merged once v26.1 is released.


Nikhil172913832:
No problem @cobaltt7. I will fix it and create a new pr.

PR #4898, opened 2025-12-09
docs: update pre-commit exclude and force-exclude explanation

### LINKED ISSUE #3015: Document using pre-commit and how --force-exclude can help

Nidhijha-10:
This PR improves the documentation for how Black handles exclude patterns when
used with pre-commit. Since pre-commit passes file paths directly to Black,
Black's normal exclude rules do not apply unless --force-exclude is used.

The updated section explains:
- why pre-commit's own exclude should be preferred,
- how exclusions can be configured in pyproject.toml,
- when --force-exclude is helpful as a fallback.

  Closes #3015
Closes #4957

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -17,8 +17,8 @@ current development version. To confirm this, you have three options:
    - create a new virtualenv (make sure it's the same Python version);
    - clone this repository;
    - run `pip install -e .[d]`;
-   - run `pip install -r test_requirements.txt`
-   - make sure it's sane by running `python -m pytest`; and
+   - run `pip install --group tests`
+   - make sure it's sane by running `python -m pytest -n auto`; and
    - run `black` like you did last time.
 -->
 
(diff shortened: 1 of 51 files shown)


docs/integrations/source_version_control.md
```diff
@@ -51,3 +51,54 @@ The `black-jupyter` hook became available in version 21.8b0.
...
+[tool.black]
+force_exclude = '''
+/(
+    migrations
+  | build
+  | dist
+)/
+'''
+```
+
```
cobaltt7, line 92: The intro and section on pre-commit's `exclude` feel well written, thanks. However, this section sounds a bit redundant. Could you combine it into the last section?

Nidhijha-10: @cobaltt7 Thanks for the review! I’ve combined the pre-commit exclude explanation into the last section to avoid redundancy and pushed the update. Please let me know if this looks good.

cobaltt7: You merged the wrong sections.

Previously, you had
- the why
- pre-commit excludes
- using force-include in pyproject.toml
- using force-include in pre-commit config

You combined the "why" into the "pyproject.toml" section, which is wrong. Plesse keep the why, and merge the two "force-exclude" sections.

PR #4899, opened 2025-12-09
Fix #3955: Add parentheses around adjacent string literals (preview style)

### LINKED ISSUE #3955: (🐞) String processing: wrap concatenated strings in parens in all cases
# input:
```py
f(
    "lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor" 
    "incididunt ut labore et dolore magna aliqua Ut enim ad minim",
    "",
)
```
# Expected output:
```py
f(
    (
        "lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor" 
        "incididunt ut labore et dolore magna aliqua Ut enim ad minim"
    ),
    "",
)
```
# Actual:
No change

# Related:
- https://github.com/psf/black/issues/3292

Apparently this was fixed, but I'm not seeing it with:

Black version 23.10.0

Nidhijha-10:
This PR fixes Issue #3955. 
When there are multiple string literals written next to each other (implicit string concatenation), Black currently leaves them as-is. In preview mode, this causes unstable formatting.

I added a small check in `visit_atom()` that detects when an atom has more than one string leaf, and then wraps the whole atom in parentheses. This makes the output more readable and consistent.

Example:
x = "hello" "world"

becomes:
x = (
    "hello"
    "world"
)

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -581,7 +581,7 @@ class LineGenerator(Visitor[Line]):
             if len(string_children) > 1:
                 wrap_in_parentheses(node, node, visible=True)
                 return
-            
+
         yield from self.visit_default(node)
 
     def visit_fstring(self, node: Node) -> Iterator[Line]:


src/black/linegen.py
```diff
@@ -571,6 +571,16 @@ def visit_atom(self, node: Node) -> Iterator[Line]:
                     mode=self.mode,
                     features=self.features,
                 )
+        if self.mode.preview:
```
JelleZijlstra: We should check for a specific preview feature, not the general preview toggle.

Also, if this fixes a crash and doesn't change formatting for code we already support, this doesn't need to go through preview style.

Nidhijha-10: Thanks for the clarification! @JelleZijlstra 
You’re right, this change is intended to prevent a crash and does not alter formatting for code that was already supported. I’ll remove the general preview check and update the logic accordingly. Please let me know if that sounds correct.

cobaltt7:
Please add tests and a changelog.

Nidhijha-10:
Hi @cobaltt7 , I’ve added a dedicated test case for adjacent implicit string literals and included a changelog entry as requested. Please let me know if any additional changes are needed. Thanks!

tests/data/cases/adjacent_implicit_strings.py
```diff
@@ -0,0 +1,5 @@
+# input
+x = "hello" "world"
+
+# output
+x = ("hello" "world")
```
cobaltt7: This does not look like a desired change. Additionally, there are many CI failures.

cobaltt7:
This is still not done properly. Please feel free to make a new PR with proper tests and following the requirements in #3955:
> Only wrap the string if it shares a boundary with an expression that starts with a string literal
Don't wrap in cases of following keyword arguments

PR #4901, opened 2025-12-09
Pin all GitHub Actions references

woodruffw:
### Description

This hash-pins all GitHub Actions.

I used `pinact run -v` to perform and verify this; a separate tool like `gha-update` could be used to cross-check them for honesty!

Note: This only pins the actions; I haven't attempted any bumps. Dependabot will keep them updated, including updating the version comments.

For context: this is an important (but not final) step towards making GitHub Actions runs generally more hermetic/reproducible. One of the big issues with GitHub Actions is that it *encourages* mutable tag usage by default, meaning that an attacker who manages to take over a third-party action can pretty easily pivot onto lots of critical project by force-pushing over an existing version tag. Hash-pinning prevents that.

For more information, I've put some docs on hash pinning in zizmor's audit docs: https://docs.zizmor.sh/audits/#unpinned-uses

### Checklist - did you ...

      stability policy?

Leaving all of the above blank since I believe this is an internal-only CI change 🙂

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - name: Grep CHANGES.md for PR number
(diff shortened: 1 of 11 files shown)


woodruffw:
xref #4611 for some backing motivation here -- I'll do some more follow-up PRs for other findings from [zizmor](https://docs.zizmor.sh) 🙂

PR #4903, opened 2025-12-10
Improve fmt:skip handling in nested expressions with checks

### LINKED ISSUE #4730: `# fmt: skip` ignored inside multi-part if-clause

Nikhil172913832:
### Description

Closes #4730 

This PR addresses issue #4730, where `# fmt: skip` was being ignored or causing crashes in deeply nested expressions. The root cause is that Black's handling of `fmt: skip` converts AST nodes to string literals, which bypasses bracket tracking and can result in missing attributes like `bracket_depth`. This led to `AttributeError` exceptions, especially with extreme settings like `--line-length=1`.

**Key changes:**
- Added `hasattr(leaf, "bracket_depth")` checks in `src/black/lines.py` to prevent crashes when processing leaves generated by `fmt: skip`.
- Added a test (`tests/data/cases/fmtskip_nested_expression.py`) using the original code from the issue, without relying on the `--no-preview-line-length-1` flag, as requested by maintainers.
- Updated `CHANGES.md` to document the fix.

### Checklist - did you ...

      stability policy?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
-- Fix `# fmt: skip` being ignored in nested `if` expressions with parenthesized `in` clauses (#4903)
+- Fix `# fmt: skip` being ignored in nested `if` expressions with parenthesized `in`
+  clauses (#4903)
 - Fix crash when an f-string follows a `# fmt: off` comment inside brackets (#5097)
 - Add support for unpacking in comprehensions (PEP 798) and for lazy imports (PEP 810),
   both new syntactic features in Python 3.15 (#5048)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Fix `# fmt: skip` being ignored in nested `if` expressions with parenthesized `in` clauses (#4903)
 - Fix crash when an f-string follows a `# fmt: off` comment inside brackets (#5097)
 - Add support for unpacking in comprehensions (PEP 798) and for lazy imports (PEP 810),
   both new syntactic features in Python 3.15 (#5048)
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -7,6 +7,7 @@ from typing import Final, Union
 from black.mode import Mode
 from black.nodes import (
     CLOSING_BRACKETS,
+    OPENING_BRACKETS,
     STANDALONE_COMMENT,
     STATEMENT,
     WHITESPACE,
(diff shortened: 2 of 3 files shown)


Nikhil172913832:
Hi @cobaltt7, can you have a look at this PR when you get time.

tests/data/cases/fmtskip_nested_expression.py
```diff
@@ -0,0 +1,41 @@
...
+    Constant2 = 2
+    Constant3 = 3
+
+
+def test():
+    if (
+        "cond1" == "cond1"
+        and "cond2" == "cond2"
+        and 1
+        in (  # fmt: skip
```
cobaltt7: This line shouldn't be formatted due to the `fmt:skip` (that's the bug in the original issue)

Nikhil172913832: Sure, I will look into it

themavik (commented):
hasattr(leaf, bracket_depth) in is_line_short_enough avoids fmt:skip nested in (...) blowing up; fmtskip_nested_expression.py covers the formatter. nit: getattr(leaf, bracket_depth, None) might read clearer than continue/pass around comma tracking if you revisit this.

cobaltt7:
Hey @Nikhil172913832, are you still working on this PR?

Nikhil172913832:
@cobaltt7 , I’ve tried a few different approaches, but haven’t been able to get it working yet. It likely needs a deeper fix. I’ll continue investigating, but I’m open to any suggestions if you have ideas.

Nikhil172913832:
Hi @cobaltt7 , can you please review the pr.

cobaltt7:
Thanks!

PR #4904, opened 2025-12-11
Add CHANGES template post 25.12.0

cooperlees:
Add post release so I can land another PR

Seem we protected main so one has to do a PR for this now ...


cooperlees:
Is this the right way now?

PR #4905, opened 2025-12-11
Drop CI permissions, eliminate persisted credentials

woodruffw:
### Description

This follows #4901 with some more CI security improvements. Namely:

* Where possible, the workflow-level default permissions have been fully dropped and job-level permissions are used to limit permissions to the smallest unit of work possible.
* All `actions/checkout` usage now includes `persist-credentials: false` to avoid accidental credential persistence/leakage. Some exceptions to this that actually use the persisted credential (e.g. for `git push`) have `persist-credentials: true` instead with a comment, making explicit the implicit default.

See zizmor's docs for more info on both of these:

* https://docs.zizmor.sh/audits/#excessive-permissions
* https://docs.zizmor.sh/audits/#artipacked

### Checklist - did you ...

      stability policy?

Like with #4901, I believe none of the above apply since this is an internal-only change 🙂

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,7 +58,8 @@ jobs:
 
       - name: Unit tests
         if: "!startsWith(matrix.python-version, 'pypy')"
-        run: tox -e ci-py$(echo ${{ matrix.python-version }} | tr -d '.') -- -v --color=yes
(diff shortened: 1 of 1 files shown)


hauntsaninja:
Thank you!

PR #4906, opened 2025-12-11
Lingering CI/CD pins, add cooldowns, remove template injections

woodruffw:
### Description

This follows #4901 and #4905 with some more small CI/CD security improvements. It hash-pins some of the dependencies added with #4611, minimizes more workflow/job permissions, and eliminates a few template injections (which probably aren't exploitable in practice in this context, but are still good to remove IMO!)

Following this, there's only one finding left from zizmor (which will unfortunately be nontrivial to fix, since it involves a `workflow_run` trigger). I'm happy to try and take a look at that, but the "fix" might be a removal of functionality so I'll file an issue for consultation first.

Separately, I'm happy to send a PR enabling zizmor in your CI, either through pre-commit or [zizmor-action](https://github.com/zizmorcore/zizmor-action). Let me know if either of these would be helpful; there's plenty of PSF / PyPA / PyPI / etc. reference material I can share for others using it!

### Checklist - did you ...

      stability policy?

Like with the other PRs, I think none of the above apply 🙂

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,9 +8,13 @@ updates:
     schedule:
       interval: "weekly"
     labels: ["skip news", "C: dependencies"]
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: "pip"
     directory: "docs/"
     schedule:
       interval: "weekly"
     labels: ["skip news", "C: dependencies", "T: documentation"]
+    cooldown:
+      default-days: 7
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -68,4 +68,6 @@ jobs:
           tags: pyfound/black:latest_prerelease
 
       - name: Image digest
-        run: echo ${{ steps.docker_build.outputs.digest }}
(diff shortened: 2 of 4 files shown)


.github/dependabot.yml
```diff
@@ -8,9 +8,13 @@ updates:
     schedule:
       interval: "weekly"
     labels: ["skip news", "C: dependencies"]
+    cooldown:
+      default-days: 7
```
woodruffw, line 12: Note: this and below prevents Dependabot from sending update PRs for dependency updates that are less than 7 days old. You can tweak this number (or I can remove it entirely), but I'd recommend some degree of "cooldown" since it's currently effective against the kinds of package compromises that have become increasingly common: https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

JelleZijlstra:
Thanks so much! I'd accept a PR adding zizmor to CI.

As for the remaining workflow_run issue, looks like it's https://github.com/psf/black/blob/main/.github/workflows/diff_shades_comment.yml . We'd have to talk about how to handle it; maybe it's fine to restrict the capabilities of the workflow in some way.

PR #4910, opened 2025-12-12
Fix inline # fmt: skip handling to prevent unwanted line splitting.

deveshbervar:
### Summary -
This PR fixes an issue where an inline `# fmt: skip` comment inside a function call
did not fully prevent Black from splitting or reformatting that argument.

Example:

```python
join(
    packages,
    func.coalesce(...) == packages.c.id,  # fmt: skip
    isouter=True,
)

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -1948,6 +1948,7 @@ def generate_trailers_to_omit(line: Line, line_length: int) -> Iterator[set[Leaf
 
 from black.comments import _contains_fmt_directive
 
+
 def run_transformer(
     line: Line,
     transform: Transformer,
--- a/changelog/fmt-skip-inline.md
+++ b/changelog/fmt-skip-inline.md
@@ -1,4 +1,4 @@
 ### Fixed
 
-- Inline `# fmt: skip` comments inside function calls now correctly prevent
-  Black from splitting or reformatting the affected line.
+- Inline `# fmt: skip` comments inside function calls now correctly prevent Black from
+  splitting or reformatting the affected line.

### CODE REVIEW

File: src/black/linegen.py [hunk 1]
```diff
@@ -1997,9 +1954,81 @@ def run_transformer(
     *,
     line_str: str = "",
 ) -> list[Line]:
+    import re
+
+    # tolerant pattern for fmt: skip (spaces allowed, case-insensitive)
+    _FMT_SKIP_RE = re.compile(r"fmt\s*:\s*skip", re.IGNORECASE)
```
cobaltt7: Please use the `contains_fmt_directive` function

deveshbervar: Thanks for the review @cobaltt7!
I've removed the custom regex and updated the implementation to use
`_contains_fmt_directive` as suggested. A changelog entry has also been added.

Please let me know if you'd like any adjustments.

cobaltt7:
This PR contains unrelated changes, as well as some things that don't follow our repository structure (pull request template not used, changelog not properly added, no tests)
For these reasons I'm inclined to believe that this PR is AI-generated. While using AI in itself isn't forbidden, we do expect you to look over what the AI did, that you've reviewed the code it writes, and that submitted PRs are up to our code standards and in line with the project expectations and requirements. **Unfortunately, your contribution does not meet this bar.**

We don't have the resources to review AI-generated PRs, which are often of low quality. If you aren't already aware:

- AIs often [hallucinate](https://en.wikipedia.org/wiki/Hallucination_(artificial_intelligence)), i.e. lie and make stuff up.
- AI does understand subtle context. While its output may be factually correct, it is likely that it is irrelevant/inapplicable to the exact situation because of unknown requirements, missing knowledge, etc.
- AI generated texts are also generally hard and tiring to read. AI often generates walls of text that are full of fluff and redundant statements.

Again, you can use AIs as part of your work, but you are responsible for verifying their output and that they do not impact others' ability to contribute productively.

Issue #3083, opened 2022-05-22
Documentation: mention tools that can do doctests formatting


KelSolaar:
**Is your feature request related to a problem? Please describe.**

Black currently does not format Doctests in Docstrings. This issues was reported in https://github.com/psf/black/issues/745 which was superseeded by https://github.com/psf/black/issues/144.

As the latter was closed though, nothing was done about the Doctests.

**Describe the solution you'd like**

I would be keen for *black* to format Doctest.

For example, given the following Python Doctests code:

```python
    >>> XYZ = np.array([0.20654008, 0.12197225, 0.05136952])
    >>> LMS_to_LMS_p_callable = lambda x: x ** 0.43
    >>> matrix_XYZ_to_LMS = np.array(
    ...     [[0.4002, 0.7075, -0.0807],
    ...      [-0.2280, 1.1500, 0.0612],
    ...      [0.0000, 0.0000, 0.9184]])
    >>> matrix_LMS_p_to_Iab = np.array(
    ...     [[0.4000, 0.4000, 0.2000],
    ...      [4.4550, -4.8510, 0.3960],
    ...      [0.8056, 0.3572, -1.1628]])
    >>> XYZ_to_Iab(
    ...     XYZ, LMS_to_LMS_p_callable, matrix_XYZ_to_LMS, matrix_LMS_p_to_Iab)
```

*black* could produce the following output:

```python
    >>> XYZ = np.array([0.20654008, 0.12197225, 0.05136952])
    >>> LMS_to_LMS_p_callable = lambda x: x**0.43
    >>> matrix_XYZ_to_LMS = np.array(
    ...     [
    ...         [0.4002, 0.7075, -0.0807],
    ...         [-0.2280, 1.1500, 0.0612],
    ...         [0.0000, 0.0000, 0.9184],
    ...     ]
    ... )
    >>> matrix_LMS_p_to_Iab = np.array(
    ...     [
    ...         [0.4000, 0.4000, 0.2000],
    ...         [4.4550, -4.8510, 0.3960],
    ...         [0.8056, 0.3572, -1.1628],
    ...     ]
    ... )
    >>> XYZ_to_Iab(
    ...     XYZ, LMS_to_LMS_p_callable, matrix_XYZ_to_LMS, matrix_LMS_p_to_Iab
    ... )
```

**Describe alternatives you've considered**

I'm manually formatting the code.

**Additional context**

N/A

felix-hilden:
[...]

Issue #4757, opened 2025-09-19
black 25.9.0 throwing `BLK900 Failed to load file: decode_bytes() missing required argument 'mode' (pos 2)`


KalebCoberly:
**Describe the bug**

When upgrading to `black==25.9.0` (from 25.1.0), I get the following error for every file when running `flake8`:
```sh
BLK900 Failed to load file: decode_bytes() missing 1 required positional argument: 'mode'
```

**To Reproduce**

```sh
$ pip install black==25.9.0
$ flake8 src/
<for every file in src/> BLK900 Failed to load file: decode_bytes() missing 1 required positional argument: 'mode'
```

```sh
$ pip install black<25.9.0
<Installs 25.1.0>
$ flake8 src/
<no errors>
```

`flake8` and `flake8-black` (and `flake8-*`) installed in each scenario are the same, 7.3.0 and 0.3.6 respectively.

Running with a fresh install from `black` main (`pip install -e .[d]`) does not fix the issue. (All black tests passed.)

**Expected behavior**

No errors.

**Environment**

- Black's version: 25.9.0
- OS and Python version: MacOS Sequoia 15.5, Python 3.11-3.13

**Additional context**

Not happening on Linux build. Will try updating Mac next.

Just started yesterday when unpinned black dependencies started installing 25.9.0 instead of 25.1.0.

cobaltt7:
It sounds like `flake8-black` is using Black's public API, which is undocumented and unstable (see #779).
This is an issue that must be brought up with the `flake8-black` team and fixed there.

MeGaGiGaGon:
Yep. As part of #4710 I changed the signature of `decode_bytes` to require the mode (`black.Mode`) be passed in as a second argument.

KalebCoberly:
Ah, so the API under semvar is only the CLI?

MeGaGiGaGon:
Yes, see #779 for some discussion on it, but currently only the behavior of the CLI is guaranteed, and the functions inside black can change at any time.

PR #4916, opened 2025-12-24
Added documentation for doctest formatting tools

samuelikohn:
### Description

Addresses issue #3083. Adds a documentation page under integration for tools that can apply *Black* formatting to doctests. This is meant to inform users of 3rd party tools that cover functionality not covered by *Black*. New doc page includes a brief description of [blackdoc](https://pypi.org/project/blackdoc/) and [blacken-docs](https://pypi.org/project/blacken-docs/). The integrations index page was also updated to list this new page.

### Checklist - did you ...

      stability policy?

- Updated `CHANGES.md`
- Confirmed documentation pages build as expected

### DIFF
--- a/docs/integrations/doctest_formatting.md
+++ b/docs/integrations/doctest_formatting.md
@@ -7,6 +7,65 @@ docstrings or documentation files (e.g. doctests), will not be formatted by _Bla
 Listed below are tools that apply _Black_ formatting to code inside docstrings and
 documentation files.
 
+> Note: There are some observed inconsistencies between the below packages. Because of
+> this, we hesitate to make any recommendations. Any installed packages are installed at
+> your own risk. We also encourage anyone to contribute documentation for additional
+> packages that apply _Black_ formatting to doctests.
+
+## blacken-docs
+
+[`blacken-docs`](https://pypi.org/project/blacken-docs/) is primarly used to apply
+_Black_ formatting to code in documentation files (e.g. `.rst`, `.md`, `.tex`).
+
+`blacken-docs` supports the following:
+
+- Python code blocks in Markdown, reStructuredText, and LaTeX files. Similar to
+  `blackdoc`, normal _Black_ formatting is applied, so doctests inside Python code
+  blocks will not be formatted.
+
+  ````md
+  ```python
+  print("Hello world!")
+  ```
+  ````
+
+  ```rst
+  .. code-block:: python
+      print("Hello world!")
+  ```
+
+  ```latex
+  \begin{minted}{python}
+  print("Hello world!")
+  \end{minted}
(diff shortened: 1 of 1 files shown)


cobaltt7:
I'm a bit hesitant about advertising small third-party tools that integrate directly with Black. One issue that comes to mind (beyond the obvious security concerns) is #4757, where a similar small project started crashing after we changed our public API, which is not stable or covered under semver (see #779).

Do the maintainers of these packages know that we're considering linking to them on our docs?

samuelikohn:
> I'm a bit hesitant about advertising small third-party tools that integrate directly with Black. One issue that comes to mind (beyond the obvious security concerns) is #4757, where a similar small project started crashing after we changed our public API, which is not stable or covered under semver (see #779).
> 
> Do the maintainers of these packages know that we're considering linking to them on our docs?

That's a fair point. My interpretation of the discussion around #3083 was that this would be a nice-to-have. The wording can certainly be changed, but based on the issues you brought up it may be best to scrap the issue altogether. But afaik the maintainers haven't been contacted.

keewis:
> Do the maintainers of these packages know that we're considering linking to them on our docs?

Speaking as the maintainer of `blackdoc` I'm aware now, and would totally be fine with it.

I'm also aware that `black` doesn't have a stable public API, such that changes in `black` may end up causing errors downstream. However, that's nothing new, it has been that way ever since I created `blackdoc` almost 5 years ago, and would typically cause a brief period of being incompatible with the newest `black` until I've had the time to merge and release a fix (which thankfully doesn't happen very often, so I didn't have to release much more than once a year).

Also, since I have seen claims that `blackdoc` is not actively maintained: it _is_ maintained, but reactions may be delayed depending on how much free time I have (I'm maintaining it on my own), although I will try to prioritize anything urgent.

> beyond the obvious security concerns

can you say a bit more about what you're referring to with this? Any of the downstream packages are clearly separate packages, and by using them you'd have to trust both `black` and the downstream project's maintainers.

I would add a note to the docs above stating that these are not officially supported, and any issues should be raised there first.

cobaltt7:
My main concern was advertising other projects could seem like an endorsement, and I wasn't sure if Black would be in hot water if one of the projects it advertised was turned into some sort of malware. However, I recently realized that Black already advertises projects in this way (see [Using _Black_ with other tools](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html) and [Editor integration](https://black.readthedocs.io/en/stable/integrations/editors.html)) so it's probably fine. Additionally, Black users are probably more technical and more aware of the risks of installing programs compared to the users of other projects I've maintained in the past that I was thinking about when I made my previous comment.

PR #4918, opened 2025-12-25
Simplify power operator hugging implementation

adamsol:
### Description

This PR replaces the current power operator hugging implementation, which uses transformers and is quite complicated, with a simpler one, implemented directly in the `whitespace` function.

The original implementation was created in #2726 and later fixed in #2806, #2874, #3942, #4154. Currently it takes around 150 lines of code, while the new implementation takes around 30 lines.

A minor side-effect of the change is that the whitespace is removed also if the power operator is moved to a separate line, so we'll have `1<newline>**1` instead of `1<newline>** 1`. But this shouldn't occur in any normal code and wasn't even tested in the original PR.

Here's how this commit would look like (all deleted lines etc.) without preview mode: https://github.com/adamsol/black/commit/2444f5c6c6d3f78c340a27fe1a85c5942ce33a6b

The main rationale for this refactoring was to extract the condition for power operator hugging into one function, so that the logic could be shared with another change that I want to upload later (edit: #4925). While making it, I also found a small bug related to the power operator (in the `visit_factor` function), for which I'm planning to submit a separate PR.

Let me know if there could be some specific reason the original implementation did not use the `whitespace` function, or if I'm missing anything else (this is my first commit here).

### Checklist - did you ...

      stability policy?

### DIFF
--- a/src/black/nodes.py
+++ b/src/black/nodes.py
@@ -8,7 +8,7 @@ from typing import Final, Generic, Literal, TypeGuard, TypeVar, Union
 from mypy_extensions import mypyc_attr
 
 from black.cache import CACHE_DIR
-from black.mode import Mode
+from black.mode import Mode, Preview
 from black.strings import get_string_prefix, has_triple_quotes
 from blib2to3 import pygram
 from blib2to3.pgen2 import token


cobaltt7:
This change looks like an improvement to me, thanks! Will merge after the 2026 style is stabilized, because we want to reduce noise in that release.

Issue #3606, opened 2023-03-14
Parenthesizing a lambda's body if it contain line wraps


huzecong:
**Describe the style change**

It's possible to have line wraps in the body of a lambda if the lambda expression as a whole is parenthesized, or used in contexts such as function arguments. Currently black would give the wrapped lines the same indent as the line that `lambda` is on, which could cause confusion. I propose to have black automatically parenthesize the lambda's body so it could be further indented.

**Examples in the current _Black_ style**

```python
some_func = (
    lambda arg1, arg2: do_something_with_the_two_args(arg1, arg2)
    if arg2 is not None
    else do_something_with_only_one_arg(arg1)
)

some_function(
    fn=lambda arg1, arg2: do_something_with_the_two_args(arg1, arg2)
    if arg2 is not None
    else do_something_with_only_one_arg(arg1)
)
```

The current style is very confusing in my opinion, because it's hard to tell at first glance what the boundary of the lambda expression is. Since usually lambda expressions are single-line, one might be more likely to initially parse this as:
```python
    (lambda arg1, arg2: do_something_with_the_two_args(arg1, arg2))
    if (arg2 is not None)
    else (do_something_with_only_one_arg(arg1))
```

**Desired style**

```python
some_func = lambda arg1, arg2: (
    do_something_with_the_two_args(arg1, arg2)
    if arg2 is not None
    else do_something_with_only_one_arg(arg1)
)

some_function(
    fn=lambda arg1, arg2: (
        do_something_with_the_two_args(arg1, arg2)
        if arg2 is not None
        else do_something_with_only_one_arg(arg1)
    )
)
```
Parentheses are inserted inside the lambda's body, so it has an extra level of indentation that clearly marks the boundaries of the lambda.

**Additional context**

The function argument case might be similar to #2248 and #3510, both of which I also endorse.

Issue #3494, opened 2023-01-12
Enforce parentheses around star unpacking containers modified with operators


jakkdl:
**Describe the style change**
Star unpacking containers in combination with modifying the container with an operator has a somewhat unintuitive operation ordering, and I think should probably have enforced parentheses.

**Examples in the current _Black_ style**

```python
def foo(other_var: Any):
    my_set: set[str] = {"a", "b", "c"}
    my_list: list[int] = [0, 1]

    print("hello", *my_set | {d}, "world")
    bar(*my_list + [2])

    # TypeError if other_var is not a set, but could at a glance be interpreted as
    # "return other_var if my_set is empty"
    return *my_set | other_var
```

**Desired style**

```python
def foo(other_var: Any):
    my_set: set[str] = {"a", "b", "c"}
    my_list: list[int] = [0, 1]

    print("hello", *(my_set | {d}), "world")
    bar(*(my_list + [2]))

    # very obvious that other_var is added to my_set
    return *(my_set | other_var)
```

**Additional context**
Black will currently not remove the parentheses either, so if you manually add them they'll stay.
I recently wrote code with `foo(*container|{"value"})`, and was personally not aware of how python would prioritize the operators and was a bit surprised that Black didn't have any opinions.

I'm not completely sure how common unintuitive code with this is, or if it just leads to a ton of extra parentheses that are hard to read and users can manually add parens when it's needed, but I thought it was worth bringing up either way.

JelleZijlstra:
Seems reasonable to make this change, I think the parentheses add clarity here.

PR #4925, opened 2025-12-26
Parenthesize complex expressions passed as keyword arguments or parameter defaults

### LINKED ISSUE #3510: Formatting of long expressions in arguments
**Describe the style change**

#3440 (closing #620 and #808) improved handling of long expressions in dict values by putting them in parentheses, but the same values in arguments aren't always handled nicely.

```python
dict(
    key_key_key_key=a_long_long_long_long_long
    * b_long_long_long_long_long
    / 100.0
)

dict(
    key_key_key_key_key_key_key_key_key=lambda: some_very_long_function_name(
        x, y, z
    )
)
```

**Desired style**

```python
dict(
    key_key_key_key=(
        a_long_long_long_long_long * b_long_long_long_long_long / 100.0
    )
)

dict(
    key_key_key_key_key_key_key_key_key=(
        lambda: some_very_long_function_name(x, y, z)
    )
)
```

**Additional context**

The above problem surprisingly isn't present when the expression is the ternary operator. I found this related issue: #2248, and I think it could be closed now, as on the --preview on main it's fixed.

Here's a comparison of some structures where formatting is good and not good: 

https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AVtAZFdAD2IimZxl1N_Wg0-DjgFSwwpr4QAHGnZVCigXiEsaw3kDKWOCEa4OyVpau2mCKO2huSgEtYJFakCvJdOz0Tnz9A9VXZaY6yvVt0g0T7eVV1sdafqJHPQjFCF87XgeS5o61J0hfwUrJyv9vemJ3drIV67OMA4NPtoBxDB0jAnP0w3_FCw4CwHZNzHatn3Mmo8JDZbQgL9tZuHGhbY93lVPrMOOj7V13lJpYG7iN4EAeOxV-wIMLcnBSj9dt2P1447cRZOFBAPbIbLCz6KCdzbNRgUw7Z2AoNrHepi73I-BQJ-tVCrypINXFeYTLtWI5ijTFPZ_c48_7tnfGJvb9lSt20aVWHGlcDJzbeHWZA5x90aMTtBD_YqD4xSuSyF2kcLxHFGTemWtI7jxvkqSbsB1L1TVuEqG1ERUBH7wYVedEYUorPtewAk13xN8SdhklZ3xTKO_WsKJ2NSK4bvFkLnxJuMSyIlc0tGdWdyFUdJub9EJwuKoneabBy__ez6gsL3D1DbXrAxE-ETk4_7yaQi0S4AAAAAAK98cuj9XVqQAAGtA-4KAADgH0AascRn-wIAAAAABFla

Without preview enabled, cases A-J excluding E are ugly, with preview enabled only C, D, I and J are ugly (and these are the examples presented above).

adamsol:
### Description

In a nutshell, this PR turns `foo(bar=x + y)`, where the spacing looks confusing/illogical, into `foo(bar=(x + y))`.

The general idea is to make whitespace consistent with operator precedence - it is similar in spirit to #157 or #646.

See [this short discussion](https://discuss.python.org/t/pep-8-formatting-of-function-arguments-with-complex-expressions/104311) for more details. A few developers agreed that the current formatting is "a little awkward" or would be "worthwhile to fix".

The proposed formatting already exists in a few places in the Black repository:
* https://github.com/psf/black/blob/c3cc5a95d4f72e6ccc27ebae23344fce8cc70786/tests/data/cases/expression.py#L39
* https://github.com/psf/black/blob/c3cc5a95d4f72e6ccc27ebae23344fce8cc70786/tests/data/cases/expression.py#L181

As well as in other popular Python projects:
* [django](https://github.com/django/django/blob/7bf3ac3ee255bcfe329e3203c7a2555b1275d506/django/db/models/base.py#L1006)
* [numpy](https://github.com/numpy/numpy/blob/main/numpy/ma/tests/test_core.py#L3875)
* [pip](https://github.com/pypa/pip/blob/468f61214e2782b9e317e84187281982cfb5a8cc/src/pip/_internal/commands/configuration.py#L125)
* [rich](https://github.com/Textualize/rich/blob/53757bc234cf18977cade41a5b64f3abaccb0b85/rich/color.py#L392)

I'm interested to hear what you think about this proposal!

PS. This PR shares some logic (`is_simple_exponentiation`) with #4918. I will rebase one onto another if either is merged first.

### Checklist - did you ...

      stability policy?


cobaltt7:
I'm split on this. I agree in most cases it improves formatting, but it does affect a lot of code. In some cases it feels unnecessary to me (like when the only operator is `not` or `await`). Would love to hear other maintainers' opinions!

adamsol:
Thanks for the feedback.

When I ran it locally on the Django codebase, there were around 500 changes, which may seem like a lot, but it's only 0.1% of all ~500k lines of code.

As for `not`, I've seen some people already parenthesize it, like in the [previously linked example from django](https://github.com/django/django/blob/7bf3ac3ee255bcfe329e3203c7a2555b1275d506/django/db/models/base.py#L1006) or [another one from numpy](https://github.com/numpy/numpy/blob/d0de6a1a90bb7740cc410144795b203720c82028/numpy/lib/_shape_base_impl.py#L1145). However, the main point of this PR was to add parentheses around binary operations, so we could definitely exclude `not`/`await` if this is what other maintainers agree with, just like there's already an exclusion for lambdas (where parentheses also felt unnecessary to me).

adamsol:
I've stumbled upon an issue that this PR would fix: #3510 (minus the lambda part, though both versions are fine IMO). [Example from Django](https://github.com/django/django/blob/d61838761f17d7c934c7be288fadfa14f471b598/tests/constraints/models.py#L36):

```diff
    rebate = models.GeneratedField(
-       expression=Coalesce("price", 0)
-       - Coalesce("discounted_price", Coalesce("price", 0)),
+       expression=(
+           Coalesce("price", 0) - Coalesce("discounted_price", Coalesce("price", 0))
+       ),
        output_field=models.IntegerField(),
        db_persist=False,
    )
```

cobaltt7:
The lambda part is covered by #3606, so I've linked the issue to this PR

cobaltt7:
FYI, Fuzz failures are unrelated

adamsol:
Hey, I wanted to ask about the status of this PR. Is there something I should do so that it can move forward?

adamsol:
In #3095 and #3494 (probably duplicates), there was a similar idea of adding parentheses in confusing expressions like `*a + b`, and both issues were accepted as worth implementing back then. Will this PR get any more interest if I generalize it to handle also those unpacking cases (as the logic should be identical), or should we just close it?

PR #4928, opened 2025-12-29
Various CI and doc refactors

### LINKED ISSUE #4926: Split the Build and Publish workflow into more jobs
> Please avoid building in the same job as publishing. Especially with the elevated OIDC privileges. The transitive build deps shouldn't have such level of access.
> 
> #4589 links to proper examples.
> 
_Originally posted by webknjaz in https://github.com/psf/black/pull/4611#discussion_r2464114574_
            

Making an issue so this doesn't get lost.

cobaltt7:
### Description

Various unrelated refactors, most of which are too small for their own PR

Since most changes are largely unrelated, it may be easier to review this PR commit-by-commit, so I made sure to keep each change contained to a single commit and I've listed them here:

- 0457459: Rename workflows, jobs, and steps for consistency, clarity, and correctness
- 465f2b5: Set up dependency groups in `pyproject.toml`
  - See https://github.com/psf/black/pull/4878#issuecomment-3597393207
  - Use them most places (some places are migrated in future commits, they're used everywhere by the end of the PR)
  - Also bumps some action runner versions
  - And updates dependabot/pre-commit config to check for more updates
  - Remove second unnecessary mypy config now that the default minimum version is 3.10
- 2399c5d: Add and improve `paths` matchers on action triggers
- 00fb065: Split up build workflow jobs into build and publish
  - Fixes #4926
  - Also removed the spurious `if` checks (see https://github.com/psf/black/pull/4611#discussion_r2635275562)
  - Prevents the workflow from trying to update the `stable` branch on prereleases, which doesn't currently work anyway and just causes the workflow to "fail"
- 5ff2a0f: Correct wrong/outdated information in release & build docs
- 5b1b02e: Don't wrap prose in issue/PR templates
  - GitHub issue comments preserve all newlines
  - GitHub releases do as well, so this should maybe be expanded to include CHANGES.md too, but that would be a more intrusive change
- d2e7a67: Manually reformat places that don't get autoformatted
  - Most notably:
    - vim files (and Python heredocs)
    - Sphinx attestations
    - pyproject.toml
  - I used 4-space indents to be consistent with Black
  - Files formatted with Prettier are currently defaulting to 2-space indents, but I left that alone for now
- 01a1e4c: Modfiy codeblock languages
  - Standardize `/py(thon)?(3)?/` to `python`
  - Only use `console` when the ps prompt is included, and don't put comments where output should be
  - Add a few other missing languages
- d351d1b: Minor misc docs rewordings (and add myself to AUTHORS.md)
- 5977512: Run docker builds in parallel using native builders
  - Cuts build times from ~45 to ~6 minutes
- 3d7b406: Refactor diff-shades workflow
  - It still uses the `workflow_run` trigger unfortunately (see #4906); there's no way to comment on the PR without it
    - Having said that, I think the usage of this is "safe"
    - To my understanding, the worst someone could do is modify the comment to whatever they wanted
    - I don't think it'd be worth it to remove the comment functionality just to satisfy the linter
    - Also, a maintainer has to manually approve CI on PRs unless they've contributed before, so it'd be a fair amount of effort for someone to troll
  - Reword PR comment to be clearer and more detailed (example: https://github.com/cobaltt7/black/pull/7#issuecomment-3703166845)
  - Work around Prettier format issues
  - Update docs appropriately

*all CI changes were tested as best I could*

### Checklist - did you ...

-   [-] Implement any code style changes under the `--preview` style, following the stability policy?
-   [-] Add an entry in `CHANGES.md` if necessary?
-   [-] Add / update tests if necessary?
-   [y] Add new / update outdated documentation?

### DIFF
--- a/.github/workflows/diff_shades.yml
+++ b/.github/workflows/diff_shades.yml
@@ -3,10 +3,20 @@ name: diff-shades
 on:
   push:
     branches: [main]
-    paths: ["src/**", "pyproject.toml", ".github/workflows/*"]
+    paths:
+      - src/**
+      - pyproject.toml
+      - scripts/diff_shades_gha_helper.py
+      - .github/workflows/diff_shades.yml
+      - .github/workflows/diff_shades_comment.yml
 
   pull_request:
-    paths: ["src/**", "pyproject.toml", ".github/workflows/*"]
+    paths:
+      - src/**
+      - pyproject.toml
+      - scripts/diff_shades_gha_helper.py
+      - .github/workflows/diff_shades.yml
+      - .github/workflows/diff_shades_comment.yml
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
@@ -34,17 +44,14 @@ jobs:
       - uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
         with:
           python-version: "3.13"
-
-      - name: Install diff-shades and support dependencies
-        run: |
-          python -m pip install 'click>=8.1.7' packaging urllib3
-          python -m pip install https://github.com/ichard26/diff-shades/archive/stable.zip
+          pip-version: "25.3"
+          pip-install: --group diff-shades --group diff-shades-comment
 
       - name: Calculate run configuration & metadata
         id: set-config
         env:
           GITHUB_TOKEN: ${{ github.token }}
-        run: python scripts/diff_shades_gha_helper.py config ${{ github.event_name }}
+        run: python scripts/diff_shades_gha_helper.py config
 
   analysis-base:
     name: analysis / base / ${{ matrix.mode }}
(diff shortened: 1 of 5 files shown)


cobaltt7:
Oh, so that's why the workflow was restricted to `release`...I think I have to split up the jobs to avoid those "deployed to release" notices, which we needed to do anyway

cobaltt7:
All issues have been fixed, ready for review!

.github/ISSUE_TEMPLATE/style_issue.md
```diff
@@ -29,6 +29,7 @@ def f():
 ```python
 def f(
     ):
+    "Make sure this code is blackened"""
```
hauntsaninja: The "current Black style" needs to be Blackened, but the proposed desired style can be anything (also preexisting, but the triple quotes are mismatched :-) )

Maybe "This code should be formatted as per the current Black style" would be clearer

docs/contributing/the_basics.md
```diff
@@ -9,28 +9,28 @@ system.
 
 First clone the _Black_ repository:
 
-```console
+```sh
```
hauntsaninja: does this syntax highlight well, given that we include the PS1 prompt?

hauntsaninja (approved):
Thank you! Looks good from a skim! (And happy new year! :-) )

docs/contributing/the_basics.md
cobaltt7: The issue I was trying to fix was this:
image image

The comments aren't formatted using `console`. I changed it to `sh` because I didn't think the prompt ever was formatted so it wouldn't matter, but it *is* formatted when using `console` and not when using `sh`. The color is just very similar to the normal text color so I didn't notice. The difference is more noticeable in light mode:

imageimage

Instead of changing `console` to `sh` except when output is shown, I'll change it except when the prompt is included, and change this specific example to format the comments correctly.

PR #4957, opened 2026-01-13
docs: document --force-exclude for pre-commit workflows

### LINKED ISSUE #3015: Document using pre-commit and how --force-exclude can help

EricGrill:
## Summary

Adds documentation explaining how to properly exclude files when using Black with pre-commit.

This addresses a common point of confusion: Black's `--exclude` option doesn't work with pre-commit because pre-commit passes files directly via the command line rather than letting Black discover files recursively.

The new section covers:
- **Why `--exclude` doesn't work**: Pre-commit passes files via CLI, bypassing recursive discovery
- **Option 1 (Recommended)**: Using pre-commit's native `exclude` configuration
- **Option 2**: Using Black's `--force-exclude` with patterns in `pyproject.toml`

Closes #3015
Closes #4898

## Test plan

- Built docs locally to verify formatting renders correctly
- Verified YAML and TOML code examples are syntactically correct

### DIFF
--- a/docs/integrations/source_version_control.md
+++ b/docs/integrations/source_version_control.md
@@ -78,8 +78,8 @@ Black, avoiding unnecessary processing.
 ### Option 2: Use Black's force-exclude in pyproject.toml
 
 Black's `force-exclude` configuration option excludes files even when they are passed
-explicitly as command-line arguments (which is how pre-commit invokes Black). Simply
-add the pattern to your `pyproject.toml`:
+explicitly as command-line arguments (which is how pre-commit invokes Black). Simply add
+the pattern to your `pyproject.toml`:
 
 ```toml
 [tool.black]


cobaltt7:
This is not the correct usage of `--force-exclude`. Please avoid using AI to write documentation.


We don't have the resources to review AI-generated PRs, which are often of low quality. If you aren't already aware:

- AIs often [hallucinate](https://en.wikipedia.org/wiki/Hallucination_(artificial_intelligence)), i.e. lie and make stuff up.
- AI does understand subtle context. While its output may be factually correct, it is likely that it is irrelevant/inapplicable to the exact situation because of unknown requirements, missing knowledge, etc.
- AI generated texts are also generally hard and tiring to read. AI often generates walls of text that are full of fluff and redundant statements.

You can use AIs as part of your work, but you are responsible for verifying their output and that they do not impact others' ability to contribute productively. For documentation PRs, please avoid using AI entirely because accuracy is even more important. If it's an English fluency issue, I'd advice not making docs PRs, or at the very least, write them yourself and only use AI for proofreading, and then re-verify the accuracy.

EricGrill:
Fixed. Removed the incorrect `args: [--force-exclude]` example - I incorrectly showed it as a bare flag when it requires a pattern argument. 

The documentation now correctly explains that Black automatically reads the `force-exclude` configuration from `pyproject.toml`, so no CLI arguments are needed when using pre-commit.

PR #4964, opened 2026-01-19
Don't double-decode input, causing non-UTF-8 files to be corrupted

### LINKED ISSUE #4963: pygettext.py: Black 26.1.0 on Python (CPython) 3.14 produced different code on the second pass of the formatter
**Describe the bug**

Black crashes with the mentioned error on a very old script that we have included for ages

**To Reproduce**

See: https://github.com/sabnzbd/sabnzbd/blob/develop/tools/pygettext.py

Log as mentioned in the error:
[blk_ym54yfeb.log](https://github.com/user-attachments/files/24714471/blk_ym54yfeb.log)

This also happens when I use black.vercel.app. 
It crashes.

**Expected behavior**

No failure

**Environment**

Black 26.1.0
Windows and black.vercel.app. 

**Additional context**

cobaltt7:
### Description

Fixes #4963 

`decode_bytes` was already called in `format_file_in_place`/`format_stdin_to_stdout`, so
`src_contents` in `_format_str_once` had already been decoded.

### Checklist - did you ...

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -2096,7 +2096,11 @@ class BlackTestCase(BlackBaseTestCase):
                 b"# 2002-11-22 J\xfcrgen Hermann <jh@web.de>\n"
             )
             first_line.close()
-            self.assertFalse(ff(Path(first_line.name)))
+            self.assertFalse(
+                ff(Path(first_line.name)),
+                "Failed to properly detect encoding",
+            )
+
         with NamedTemporaryFile(delete=False) as second_line:
             second_line.write(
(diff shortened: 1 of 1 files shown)


MeGaGiGaGon:
We should use a `tests/test_black.py` test for this instead of a file test, like how I did in https://github.com/psf/black/commit/626b32fe2b5387656be6694da9a4b7a3148fb892, since anything with different encodings can be very brittle

cobaltt7:
Fuzz error is unrelated

PR #4970, opened 2026-01-23
Partially revert actions/checkout bump

cobaltt7:
actions/checkout versions beyond 5.0.1 check out a detached HEAD when running on releases. Confirmed this is still reproducible on 6.0.2

Refs: https://github.com/actions/checkout/issues/2280, #4969


cobaltt7:
I misremembered the bug, it's not that it checks out a detached HEAD, it's that it checks out the commit hash instead of the tag. This does actually appear to be fixed in 6.0.2, but if actions fail next release, this may be a reason

PR #4973, opened 2026-01-26
Fix #4945: parse multiline backslash correctly

### LINKED ISSUE #4945: Cannot parse multi-line code split using backslash
**Describe the bug**

When my code contains a multi-line expression that's concatenated using backslash (`\`) and no indentation, black cannot parse my code

**To Reproduce**

For example, take this code:

```python
if True:
    foo = 1+\
2
    print(foo)
```

```sh
$ black file.py --target-version py313
```

The resulting error is:

> error: cannot format file.py: Cannot parse for target version Python 3.13: 3:0: 2

**Expected behavior**

The file should be formatted to:
```
if True:
    foo = 1 + 2
    print(foo)
```

**Environment**

- Black's version: 25.12.0
- OS and Python version: Debian 13, Python (CPython) 3.14.2

**Additional context**

nileshsingal:
### Description

Fixes #4945.

The parser did not correctly handle multiline statements that use explicit
backslash (`\`) line continuations. In certain cases, continued lines were
treated as separate logical units instead of a single statement, leading to
incorrect tokenization and formatting behavior.

This issue could surface when multiple backslash-continued lines were combined
with indentation or nested constructs, even though the input was valid Python
syntax.

This PR fixes the issue by normalizing the input before parsing: explicit
backslash-newline continuations are removed prior to indentation handling.
CPython removes backslash-newline sequences during tokenization, but `blib2to3`
does not, which could previously result in invalid indentation states during
parsing.

By aligning the preprocessing behavior with CPython’s tokenizer, the parser
now handles multiline backslash continuations deterministically and in line
with Python’s continuation rules.

A regression test is added to cover the previously failing multiline backslash
cases and ensure the behavior remains stable going forward.

### Checklist - did you ...

  N/A – this is a parser bug fix, not a code style change

  Added

  Added case file: `tests/data/cases/backslash_continuation_no_indent.py`

  N/A – no documentation affected by this fix

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@
 
 <!-- Changes that affect Black's stable style -->
 
-- Don't double-decode input, causing non-UTF-8 files to be corrupted (#4964)
+- Don't double-decode input, causing non-UTF-8 files to be corrupted (#4973)
 
 ### Preview style
 


cobaltt7:
This is a bug in pytokens, there's already a PR up to fix it there - https://github.com/tusharsadhwani/pytokens/pull/19

PR #4978, opened 2026-02-01
A few more small CI tweaks

cobaltt7:
### Description

A small followup to #4928 with a few more things I noticed since then

Notable changes:

- Change Fuzz to run on schedule & on push instead of on PRs, and open an issue with failures
- Add a post release workflow that adds a new changelog, and move the update-stable job to it
- Pinned most dev/ci depedencies to a specific version, except for bigger projects with lots of updates
- For runtime dependencies, I changed most `>=` to `~=` to avoid bumping major versions
  I didn't change any ranges that already included multiple major versions
  No currently-released versions that were previously allowed are now disallowed
- Update cibuildhweel in GHA and pyproject together
- Use default dependabot update schedule
- Rename `skip news` to `ci: skip news` for consistency with other CI labels
  (changelog action will fail on this PR until it's updated on the GH side)

### Checklist - did you ...

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -5,7 +5,12 @@ import re
 import sys
 import tokenize
 import traceback
-from collections.abc import Collection, Generator, MutableMapping, Sequence
+from collections.abc import (
+    Collection,
+    Generator,
+    MutableMapping,
+    Sequence,
+)
 from contextlib import nullcontext
 from dataclasses import replace
 from datetime import datetime, timezone


pyproject.toml
```diff
@@ -53,54 +53,54 @@ classifiers = [
     "Topic :: Software Development :: Quality Assurance",
 ]
 dependencies = [
-    "click>=8.0.0",
-    "mypy_extensions>=0.4.3",
+    "click~=8.0",
```
JelleZijlstra: I'm strongly against these changes. In practice, major versions aren't an especially good indicator of when something will break. And when dealing with a large Python environment, unnecessary upper version constraints are often a pain to deal with because they restrict what packages can live in the same environment. I would only use upper bounds if we *know* that Black is incompatible with some version of a package.

hauntsaninja: see also https://iscinumpy.dev/post/bound-version-constraints/

cobaltt7: Thanks for the explanations @JelleZijlstra and @hauntsaninja! I wasn't aware that Python dependency resolution worked that way. I agree it makes sense not to pin, and fix issues as they come up. I've reverted most of the pins.

> see also [iscinumpy.dev/post/bound-version-constraints](https://iscinumpy.dev/post/bound-version-constraints/)

Thanks for this as well - very insightful read.

JelleZijlstra: Thanks! Worth noting that this applies most strongly to Black's own runtime dependencies, because those affect what users can do when they use Black. For things only installed in our own CI, pinning is less problematic, though I'd still try to minimize it.

PR #4992, opened 2026-02-15
enhance: enhancement of the syntex error reporting

### LINKED ISSUE #4820: Improve syntax error reporting for a better user experience
When black encounters a Python file with a syntax error, the current error message is often dense, confusing. This forces the developer to spend extra time deciphering what went wrong and where.

Image

Image

I propose enhancing the error reporting for SyntaxErrors to be more user-friendly, clear, and actionable.
Instead of the current one-line error, black could print a more detailed, multi-line message that pinpoints the error, much like Python's own interpreter.

**Proposed new error output:**

$ black test.py
Error: Cannot parse test.py

  black's parser found a syntax error on or near line 4.
  This is often caused by a missing ':' on the line before.

  File "test.py", line 3:
    def my_func()
                 ^
  SyntaxError: expected ':'

Please fix the syntax error before formatting.

 Another alternative is to use a separate linter, but it would be a significant user experience improvement if black could provide this simple feedback on its own.

This proposal is not about adding new formatting rules or configuration. It's about improving the core user experience and developer productivity. By providing clear, actionable error messages, black becomes an even more helpful tool.

This could likely be implemented by catching the SyntaxError (or parser-specific error) and extracting the line number, column, and text to format it into a friendlier "pretty print" error message before exiting.

MohamedSharfan:
### Description

My fix improves Black's syntex error reporting when parsing fails.

I have made a test file to see the error messege that reports
image
  
currently when Black encounters invalid python syntex, it reports something like this 
image

The changes i have changes to enhance the error message reporting

1. Reporting by lines and column so maintainer can easily find the error
2. Displaying the error line and using caret(^) pointing to show exactly where the error is 

image   

fixes #4820 
closes #5068

### Checklist - did you ...

### DIFF
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -85,9 +85,9 @@ def lib2to3_parse(
                 f"SyntaxError is in line {lineno}, column {column}:\n"
                 f"    {faulty_line}\n"
                 f"    {caret_line}"
-            )  
+            )
             errors[grammar.version] = InvalidInput(message)
-            
+
         except TokenError as te:
             # In edge cases these are raised; and typically don't have a "faulty_line".
             lineno, column = te.args[1]


cobaltt7:
The message should be changed here, inside lib2to3_parse: https://github.com/psf/black/blob/e015af4a681dd2ac847188b13a81dc4a573d4dcb/src/black/parsing.py#L84

Also, Copilot's suggestions are correct; please review them.

MohamedSharfan:
> The message should be changed here, inside lib2to3_parse:
> 
> https://github.com/psf/black/blob/e015af4a681dd2ac847188b13a81dc4a573d4dcb/src/black/parsing.py#L84
> 
> Also, Copilot's suggestions are correct; please review them.

ok sure i will review them and commit again

src/black/parsing.py
```diff
@@ -80,9 +80,13 @@ def lib2to3_parse(
                 faulty_line = lines[lineno - 1]
             except IndexError:
                 faulty_line = "<line number missing in source>"
-            errors[grammar.version] = InvalidInput(
-                f"Cannot parse{tv_str}: {lineno}:{column}: {faulty_line}"
+            caret_line = "^".rjust(column)
+            message = (
+                f"SyntaxError is in line {lineno}, column {column}:\n"
```
cobaltt7, line 85: Please use wording similar to how it was previously, this message is confusing (and spelled wrong)!

Please also review the test failures.

cobaltt7 (changes requested):

PR #4993, opened 2026-02-17
Fix crash on standalone comment in lambda default arguments

### LINKED ISSUE #4640: Standalone comment within parentheses in lambda default arguments crashes black

worksbyfriday:
## Summary
- Fixes #4640
- Closes #4658
- Black crashes with `"Opening paren not found in 'leaves'"` when a standalone comment appears inside parentheses in a lambda default argument with a trailing comma
- Root cause: after line splitting, the opening bracket ends up in a different `Line` object than the closing bracket, so `is_one_sequence_between()` can't find it via identity comparison in the current line's leaves

## Fix
Return `False` instead of raising `LookupError` when the opening bracket isn't found in the leaves list. This is the correct conservative answer — if we can't find both brackets in the same line, we can't determine the sequence between them, so we shouldn't assume it's a one-sequence.

## Reproduction
```python
# This crashes Black:
help(lambda x=(
    # comment
    "bar",
): False)
```

## Test plan
- [x] Verified both reproduction cases from #4640 now format correctly
- [x] All 393 existing tests pass (3 skipped, 4 subtests passed)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ d = ["aiohttp>=3.10"]
 jupyter = ["ipython>=7.8.0", "tokenize-rt>=3.2.0"]
 
 [dependency-groups]
-build = ["hatch==1.15.1", "hatch-fancy-pypi-readme", "hatch-vcs"]
+build = ["hatch==1.15.1", "hatch-fancy-pypi-readme", "hatch-vcs", "virtualenv<21.0.0"]
 wheels = ["cibuildwheel==3.3.1", "pypyp==1.3.0"]
 binary = ["pyinstaller", "wheel>=0.45.1"]
 


JelleZijlstra:
Please add a unit test for this behavior.

worksbyfriday:
The test was added in commit 2fdde64 — `tests/data/cases/comments_in_lambda_default.py` covers both reproduction cases from #4640. It's auto-discovered by `test_simple_format` and passes:

```
tests/test_format.py::test_simple_format[comments_in_lambda_default] PASSED
```

PR #4997, opened 2026-02-20
Mark first block before matching content as changed in line ranges mapping

bysiber:
In `_calculate_lines_mappings`, when the first `matching_blocks` entry doesn't start at offset 0, the content before it is different between original and modified. This block should be marked as `is_changed_block=True`, but it's currently marked `False`.

**What goes wrong**

The `adjusted_lines` function uses `is_changed_block` to decide how to remap line ranges:
- For unchanged blocks, it applies a linear offset
- For changed blocks, it expands to the full block extent

When the first changed block is mislabeled as unchanged, a linear offset is applied instead of block expansion. For example, if the original file has 5 lines changed at the top that become 3 lines after formatting, a `--line-ranges 1-5` would compute the new range as `(1, 5)` via offset instead of the correct `(1, 3)` via block expansion. The range extends past the actual modified content.

The `else` branch (for `i > 0`) correctly marks inter-block gaps as `is_changed_block=True` — this first-block case was just missed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@
 <!-- Changes that affect Black's stable style -->
 
 - Don't double-decode input, causing non-UTF-8 files to be corrupted (#4964)
+- Fix `--line-ranges` not marking the first block as changed when it contains
+  modifications before the first matching content (#4997)
 
 ### Preview style
 
--- a/src/black/ranges.py
+++ b/src/black/ranges.py
@@ -492,7 +492,7 @@ def _calculate_lines_mappings(
                         original_end=block.a,
                         modified_start=1,
                         modified_end=block.b,
-                        is_changed_block=False,
+                        is_changed_block=True,
                     )
                 )
         else:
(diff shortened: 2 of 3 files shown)


JelleZijlstra:
Thanks! Could you add a unit test for this behavior and a NEWS entry describing the user-visible change?

bysiber:
Added both — unit test in test_ranges.py (test_first_block_changed) and a CHANGES.md entry under Stable style. The test verifies that changes in the first block before any matching content are correctly included in the adjusted line ranges.

cobaltt7:
@bysiber Please review the test failures

themavik (commented):
is_changed_block=True for that first LineMapping fixes adjusted_lines when only the leading line differs; test_first_block_changed is the right minimal repro. nit: if a future case has an unchanged prefix that still maps through block.a, add another regression beside the x=10 edit.

cobaltt7:
Hey @bysiber, are you still updating this PR?

cobaltt7:
Closing as stale - please feel free to reopen/recreate if you or anyone else is interested in continuing this change.

PR #5007, opened 2026-02-26
fix: replace 15 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


cobaltt7:
These are almost all test files, and the other change introduces a syntax error

PR #5021, opened 2026-03-03
Improve formatting of `@overload`-decorated function groups in stub files, especially those with docstrings

AlexWaygood:
⚠️ I'm pretty unfamiliar with Black's codebase, so this patch was mostly authored by Claude. I verified that it has extensive test coverage, that it has the effect that I want, and that it's well documented, but I can't say much about whether it's making the change in the right way or not.

### Description

In `.pyi` stub files, Black currently has no special handling for groups of consecutive decorated functions that share the same name. This comes up quite often with `@overload` groups, and it results in especially unfortunate formatting if one of the `@overload`-decorated functions has a docstring in it. Overloads can end up with inconsistent blank lines — sometimes separated from each other, sometimes not — depending on the surrounding context and whether any overloads have docstrings:

```py
# Before (status quo)

@overload                                                                                                                                       
def foo(x: int) -> int:                                         
    """Docs."""

@overload
def foo(x: str) -> str: ...  # a blank line is enforced between the `foo` overloads due to the docstring
def bar(x): ...              # a blank line is removed between the last `foo` overload and `bar`

class MyClass:
    attr: int
    @overload                # no blank line after attr
    def method(self, x: int) -> int: ...
    @overload
    def method(self, x: str) -> str: ...
    attr2: str               # no blank line before attr2
```

This PR adds a new preview feature (`pyi_overload_group_blank_lines`) that enforces two rules for decorated function statements in `.pyi` files:

1. Before a decorated function: always insert a blank line, unless the immediately preceding statement is a decorated function with the same name (i.e. part of the same overload group), or it is the first statement in the block.
2. After a decorated function: always insert a blank line, unless the immediately following statement is a decorated function with the same name.

These rules apply regardless of whether the adjacent statement is a function definition, a class, or a simple statement like a variable annotation.

```py
# After (with `--preview`)

@overload
def foo(x: int) -> int:
    """Docs."""
@overload
def foo(x: str) -> str: ...  # no blank line in between the `foo` overloads

def bar(x): ...              # blank line after `foo` group

class MyClass:
    attr: int

    @overload                # blank line before group
    def method(self, x: int) -> int: ...
    @overload
    def method(self, x: str) -> str: ...

    attr2: str               # blank line after group
```

The implementation tracks the most recently seen decorated function's name and depth in `EmptyLineTracker`, and uses this state to determine whether to collapse or separate adjacent statements.

### Motivation

The motivation here is that ty codemods docstrings into its vendored copy of typeshed's stubs, and then reformats the stubs with Black after the codemod. We use Black to reformat the stubs because typeshed uses Black upstream, and we want to minimze the diff between our modified version of typeshed and the upstream version of typeshed. Black's formatting also has the nice property that it will never reformat a line with a `# type: ignore` comment, which means that it's easy to run typeshed's tests on our codemodded version of typeshed and verify that our codemodding didn't break anything in typeshed.

I ran Black from this branch on ty's vendored typeshed stubs, and you can see the result here -- I think it's a big improvement, personally! https://github.com/astral-sh/ruff/compare/main...AlexWaygood:ruff:reformat-stubs?expand=1.

Fixes https://github.com/astral-sh/ty/issues/2945

### Checklist - did you ...


AlexWaygood:
> **diff-shades** results comparing this PR ([9540cbc](https://github.com/psf/black/commit/9540cbc974f63395c1b70a40bda51ae24f355e0b)) to main ([457320a](https://github.com/psf/black/commit/457320aeb1b409a79ca2cbd04348d4ade4909dfd)):

This looks about as-expected! Most stubs in the wild don't have docstrings in them, so on most codebases this will result in blank lines being added before or after decorated functions and overload groups. In my opinion, that makes things more readable.

docs/the_black_code_style/future_style.md
```diff
@@ -25,6 +25,56 @@ Currently, the following features are included in the preview style:
...
   patterns which have trailing commas in them, even if the guard expression fits in one
   line
+- `pyi_overload_group_blank_lines`: In `.pyi` stub files, enforce consistent blank line
+  handling around decorated function groups. ([see below](labels/pyi-overload-group))
+
+(labels/pyi-overload-group)=
+
+### Consistent blank lines around decorated function groups in stub files
+
+In `.pyi` stub files, Black now enforces consistent blank lines around groups of
```
JelleZijlstra: The previous rule was that we only put blank lines after a docstring, right? Seems a bit weird then to say it was not "consistent".

AlexWaygood: Yes, that's fair, I'll revise this

docs/the_black_code_style/future_style.md
```diff
@@ -25,6 +25,57 @@ Currently, the following features are included in the preview style:
 - `fix_if_guard_explosion_in_case_statement`: fixed exploding of the if guard in case
   patterns which have trailing commas in them, even if the guard expression fits in one
   line
+- `pyi_overload_group_blank_lines`: In `.pyi` stub files, improve heuristics around when
+  blank lines should appear before, after and within decorated function groups.
+  ([see below](labels/pyi-overload-group))
+
+(labels/pyi-overload-group)=
+
+### Improved heuristics for blank lines before, after and within decorated function groups in stub files
```
cobaltt7: Could you move this section below `wrap-long-dict-values` so the extended descriptions are ordered the same way as in the list?

Also, I'd prefer a shorter heading, maybe "Grouping decorated functions in stub files"

tests/test_black.py
```diff
@@ -1147,7 +1147,9 @@ def test_multi_file_force_pyi(self) -> None:
     def test_pipe_force_pyi(self) -> None:
         source, expected = read_data("miscellaneous", "force_pyi")
         result = CliRunner().invoke(
-            black.main, ["-", "-q", "--pyi"], input=BytesIO(source.encode("utf-8"))
```
cobaltt7, line 1150: Is this change necessary? It doesn't look like the `force_pyi` folder is even modified in this PR

AlexWaygood: Yeah, the test failed without the change — will check the details about why that was tomorrow morning! Sorry, I meant to add a comment about the reason for this change but forgot

AlexWaygood:
> I agree that overloaded functions with the same name should be grouped together, but does it make sense to apply this to _all_ decorators? Otherwise, we end up with code like this - multiple undecorated functions grouped together, then groups of one decorated function each:

Jelle gave me similar feedback offline — since you both gave the same feedback independently, and this was the design aspect I was least sure of, I'll change this. Thanks!

tests/test_black.py
AlexWaygood: Ah, it looks like it's no longer required after the changes I just pushed in c40e0cb13428acead0998bef5f82f136ee09434e, so I can revert it again.

But IIRC it was an issue where the test was insufficiently isolated against black's own pyproject.toml file, which specifies `preview = true`. This test was picking up that pyproject.toml file, which meant that the asserted output of `black --pyi` no longer passed on an earlier version of this PR. That, in and of itself, would not be sufficient to motivate this change (we can just change the expected output for the test) -- but then other tests that _were_ isolated against black's own pyproject.toml file started failing.

So it does seem like there might be a test-isolation problem here for us to solve independently of this PR.

AlexWaygood: (reverted in f88aeb204aee072dce035d748c8a3e42bf6d347a)

src/black/lines.py
```diff
@@ -544,6 +544,132 @@ class EmptyLineTracker:
...
+                        return EmptyLineTracker._get_funcdef_name(subchild) == cur_name
+                return False
+            elif sibling.type in (
+                token.NEWLINE,
+                token.NL,
+                token.INDENT,
+                token.DEDENT,
+                token.COMMENT,
+                token.ENDMARKER,
+            ):
```
AlexWaygood: I'd normally use a set for this kind of containment check, but IIRC mypyc optimizes tuples much better than sets (though maybe that's outdated lore by now...? Not sure...), and I know black compiles with mypyc

AlexWaygood:
> I agree that overloaded functions with the same name should be grouped together, but does it make sense to apply this to _all_ decorators?

Okay, I addressed this point, and also fixed a whole bunch of edge cases thrown up by diff-shades. The heuristics hopefully now mean that the patch only changes formatting in `.pyi` files for cases involving `@overload` groups or properties with setters. I also updated the branch of Ruff where I applied this branch's formatting to ty's vendored typeshed stubs: https://github.com/astral-sh/ruff/compare/main...AlexWaygood:ruff:reformat-stubs?expand=1

AlexWaygood:
Changing the heuristic so that we only add blank lines around decorated functions if there's a sequence of >=2 with the same name _has_ added a fair amount of complexity to this patch. If this is now too complex to stomach, one way to simplify would be to remove some of the handling that I've added to take care of decorated-function groups where one of the decorated functions is inside an `if sys.version_info` block or similar and others are not.

<details>
<summary>Possible simplification</summary>

```diff
diff --git a/src/black/lines.py b/src/black/lines.py
index da54666..96a0ff1 100644
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -683,93 +683,6 @@ def _is_start_of_decorated_group(line: Line) -> bool:
             return False
         return EmptyLineTracker._decorated_node_starts_group(decorated_node)
 
-    @staticmethod
-    def _get_block_first_decorated_funcname(line: Line) -> str | None:
-        """Return the function name of the first decorated function in a block.
-
-        *line* must be a block-opening line (ending with ``:``) such as
-        ``if ...:``.  Returns ``None`` when the block doesn't start with a
-        decorated function.
-        """
-        if not line.leaves or line.leaves[-1].type != token.COLON:
-            return None
-        suite = line.leaves[-1].next_sibling
-        if suite is None or not isinstance(suite, Node) or suite.type != syms.suite:
-            return None
-        for child in suite.children:
-            if isinstance(child, Node) and child.type == syms.decorated:
-                for sub in child.children:
-                    if sub.type in (syms.funcdef, syms.async_funcdef):
-                        return EmptyLineTracker._get_funcdef_name(sub)
-                return None
-            if child.type in (token.NEWLINE, token.NL, token.INDENT, token.DEDENT):
-                continue
-            return None
-        return None
-
-    @staticmethod
-    def _block_is_part_of_overload_group(line: Line) -> bool:
-        """Check if a block-opening line contains a decorated function that is
-        part of a larger overload group — either because the ``if_stmt``'s next
-        sibling is a same-name decorated function, or because another branch of
-        the same ``if_stmt`` has one.
-        """
-        func_name = EmptyLineTracker._get_block_first_decorated_funcname(line)
-        if func_name is None:
-            return False
-
-        suite = line.leaves[-1].next_sibling
-        if suite is None or not isinstance(suite, Node):
-            return False
-        if_stmt = suite.parent
-        if if_stmt is None:
-            return False
-
-        # Check if the if_stmt's next sibling is a same-name decorated function.
-        sibling = if_stmt.next_sibling
-        while sibling is not None:
-            if sibling.type == syms.decorated:
-                for sub in sibling.children:
-                    if (
-                        sub.type in (syms.funcdef, syms.async_funcdef)
-                        and EmptyLineTracker._get_funcdef_name(sub) == func_name
-                    ):
-                        return True
-                break
-            elif sibling.type in (
-                token.NEWLINE,
-                token.NL,
-                token.INDENT,
-                token.DEDENT,
-                token.COMMENT,
-                token.ENDMARKER,
-            ):
-                sibling = sibling.next_sibling
-            else:
-                break
-
-        # Check other branches (elif/else) of the same if_stmt.
-        for child in if_stmt.children:
-            if (
-                isinstance(child, Node)
-                and child.type == syms.suite
-                and child is not suite
-            ):
-                for stmt in child.children:
-                    if not isinstance(stmt, Node):
-                        continue
-                    if stmt.type != syms.decorated:
-                        continue
-                    for sub in stmt.children:
-                        if (
-                            sub.type in (syms.funcdef, syms.async_funcdef)
-                            and EmptyLineTracker._get_funcdef_name(sub) == func_name
-                        ):
-                            return True
-                    break
-
-        return False
-
     def maybe_empty_lines(self, current_line: Line) -> LinesBlock:
         """Return the number of extra empty lines before and after the `current_line`.
 
@@ -835,15 +748,7 @@ def maybe_empty_lines(self, current_line: Line) -> LinesBlock:
                 and not current_line.is_comment
                 and (
                     self._pyi_previous_decorated_func is None
-                    or (
-                        current_line.depth <= self._pyi_previous_decorated_func.depth
-                        # Don't reset on else/elif — they continue an if/else
-                        # chain that may contain overloads at a deeper depth.
-                        and not (
-                            current_line.leaves
-                            and current_line.leaves[0].value in ("else", "elif")
-                        )
-                    )
+                    or current_line.depth <= self._pyi_previous_decorated_func.depth
                 )
             ):
                 # Only reset when we see a non-decorator line at the same or
@@ -920,32 +825,15 @@ def _maybe_empty_lines(self, current_line: Line) -> tuple[int, int]:
                     and self._pyi_previous_decorated_func.is_multi
                     and not current_line.is_comment
                     and self.previous_line.depth >= current_line.depth
-                    and not (
-                        current_line.leaves
-                        and current_line.leaves[0].value in ("else", "elif")
-                    )
                 ):
                     if self._is_in_current_group(current_line):
                         before = 0
-                    elif current_line.opens_block and (
-                        self._get_block_first_decorated_funcname(current_line)
-                        == self._pyi_previous_decorated_func.name
-                    ):
-                        before = 0
                     else:
                         before = 1
                 elif depth and not current_line.is_def and self.previous_line.is_def:
-                    if (
-                        Preview.pyi_overload_group_blank_lines in self.mode
-                        and current_line.opens_block
-                        and self.previous_line.depth <= current_line.depth
-                        and self._block_is_part_of_overload_group(current_line)
-                    ):
-                        before = 1
-                    else:
-                        # Empty lines between attributes and methods should
-                        # be preserved.
-                        before = 1 if user_had_newline else 0
+                    # Empty lines between attributes and methods should
+                    # be preserved.
+                    before = 1 if user_had_newline else 0
                 elif depth:
                     before = 0
                 else:
diff --git a/tests/data/cases/pyi_overload_groups.py b/tests/data/cases/pyi_overload_groups.py
index 2466852..1a461da 100644
--- a/tests/data/cases/pyi_overload_groups.py
+++ b/tests/data/cases/pyi_overload_groups.py
@@ -578,7 +578,6 @@ def conditional_overload(x: int) -> int:
 
 @overload
 def conditional_overload(x: str) -> str: ...
-
 def after_conditional_overload() -> None: ...
 def before_conditional_overload_with_gaps() -> None: ...
 
@@ -594,7 +593,6 @@ def conditional_overload_with_gaps(x: int) -> int:
 
 @overload
 def conditional_overload_with_gaps(x: str) -> str: ...
-
 def after_conditional_overload_with_gaps() -> None: ...
 
 if sys.version_info >= (3, 10):
@@ -663,7 +661,6 @@ def after_comment_group() -> None: ...
 
 class ClassWithConditionalOverloads:
     def method(self) -> None: ...
-
     if sys.version_info >= (3, 10):
         @overload
         def conditional(self, x: int) -> int: ...
@@ -679,6 +676,7 @@ class ClassWithOverloadThenConditional:
     def method(self, x: int) -> int: ...
     @overload
     def method(self, x: str) -> str: ...
+
     if sys.version_info >= (3, 10):
         @overload
         def method(self, x: bytes) -> bytes: ...
```

</details>

tests/data/cases/pyi_overload_groups.py
```diff
@@ -0,0 +1,697 @@
...
+
+else:
+    @overload
+    def if_else_other_name(x: int) -> int: ...
+    @overload
+    def if_else_other_name(x: str) -> str: ...
+
+@overload
+def commented(x: int) -> int: ...
+
```
JelleZijlstra: Why is there a blank line here and not between the commented methods a few lines down?

srittau:
This new formatting makes sense to me. Decorators are a "visual break", separating two overloaded functions from each other in a way. Especially the last overload looks more tightly coupled to the succeeding function than the previous overload. This will hopefully help to group overloads more closely.

tests/data/cases/pyi_overload_groups.py
AlexWaygood: hmm, good catch. that seems pretty inconsistent. will fix.

AlexWaygood:
I _think_ that Windows/ARM failure is a flake?

tests/data/cases/pyi_overload_groups.py
AlexWaygood: fixed

Akuli:
I like the new formatting too.

srittau:
FWIW, I've now closed that typeshed issue. Apart from Alex, Jelle, Akuli and myself have commented, which makes four typeshed maintainers. :)

AlexWaygood:
Thanks folks :D

Issue #5029, opened 2026-03-06
Security: Zip Slip / Tar Slip path traversal in gallery/gallery.py archive extraction


Vext-Labs:
This issue was filed in error and has been withdrawn. We apologize for the noise.

JelleZijlstra:
Maybe we should just delete `gallery.py`, I am not aware of anyone or anything using it.

Vext-Labs:
@JelleZijlstra — That's a pragmatic approach. If gallery.py isn't actively used, removing it eliminates the attack surface entirely. Deleting vulnerable code you don't need is often the best fix.

If there's interest in keeping it, the fix would be adding `filter='data'` to `tarfile.extractall()` (Python 3.12+) or manual path validation for older versions.

Vext-Labs:
Withdrawing this report. We apologize for the noise — this should have been filed through proper security disclosure channels, not as a public issue. Closing.

PR #5030, opened 2026-03-06
Remove gallery/

JelleZijlstra:
This hasn't seen wide use and has potential security issues (#5029).

### Checklist - did you ...


cobaltt7:
Should we also remove https://github.com/psf/black/blob/main/scripts%2Fmigrate-black.py? It's probably a similar situation

JelleZijlstra:
Possibly, but seems less urgent as it doesn't have the same security implications.

PR #5071, opened 2026-04-03
Upgrade mypy

hauntsaninja:
closes #5084


hauntsaninja:
Great, looks like that worked. Will just need to wait for pre-commit mirror to pick up the change

PR #5074, opened 2026-04-04
Fix blackd returning 500 for invalid f-strings

### LINKED ISSUE #3616: (🐞) blackd bad f-string fragment causes 500 response
**Describe the bug**

Blackd is inconsistent with black in regards to formatting invalid f-strings. While black returns a consistent error code of '123' for `,` and `f"{}"`, blackd returns an expected 400 for the former, but a surprising 500 for the later.

**To Reproduce**
```
👉 curl -XPOST "localhost:9090" -d "," -i
400: bad request, take it away

Cannot parse: 1:0: , 
👉 curl -XPOST "localhost:9090" -d 'f"{}"' -i
500: internal server error

cannot use --safe with this file; failed to parse source file AST: f-string: empty expression not allowed (<unknown>, line 1)
This could be caused by running Black with an older Python version that does not support new syntax used in your source file. 
```

**Environment**

- Black's version: 23.1.0
- OS and Python version:  windows / python 3.11.2

Bahtya:
## Problem

When `blackd` receives code with invalid f-strings (e.g. `f"{}"`), `ast.parse()` raises `ASTSafetyError`. This falls through to the generic `except Exception` handler, returning HTTP 500 (Internal Server Error).

It should return HTTP 400 (Bad Request), consistent with how `InvalidInput` is handled.

## Solution

Add a specific `except ASTSafetyError` handler before the generic `except Exception`, returning HTTP 400.

Fixes #3616

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -66,6 +66,7 @@ from black.output import color_diff, diff, dump_to_file, err, ipynb_diff, out
 from black.parsing import (  # noqa F401
     ASTSafetyError,
     InvalidInput,
+    SourceASTParseError,
     lib2to3_parse,
     parse_ast,
     stringify_ast,
@@ -1632,7 +1633,7 @@ def assert_equivalent(src: str, dst: str) -> None:
     try:
         src_ast = parse_ast(src)
     except Exception as exc:
-        raise ASTSafetyError(
+        raise SourceASTParseError(
             "cannot use --safe with this file; failed to parse source file AST: "
             f"{exc}\n"
             "This could be caused by running Black with an older Python version "
--- a/src/black/parsing.py
+++ b/src/black/parsing.py
@@ -106,6 +106,14 @@ class ASTSafetyError(Exception):
     """Raised when Black's generated code is not equivalent to the old AST."""
 
 
+class SourceASTParseError(Exception):
+    """Raised when the source file's AST cannot be parsed by ast.parse.
+
+    This indicates a user-input issue (e.g. syntax not supported by the
+    runtime Python) rather than a bug in Black itself.
+    """
+
+
 def _parse_single_version(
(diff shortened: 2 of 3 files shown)


JelleZijlstra:
This is not correct. ASTSafetyError can indicate a bug in Black, which should be a 500. The linked issue appears to be a case where Black's parser is more lenient than `ast.parse`. Perhaps we should split that condition off into its own exception and turn it into a 400.

Before you do this make sure you understand what each use of ASTSafetyError in Black is for and what it indicates.

Bahtya:
Thanks for the review @JelleZijlstra, and you're absolutely right. I took the time to study all the uses of `ASTSafetyError` in Black:

1. **Source AST parse failure** (`assert_equivalent`, line 1635) — user's code can't be parsed by `ast.parse()`, but Black's more lenient parser accepted it → **user input issue → 400**
2. **Dest AST parse failure** (`assert_equivalent`, line 1646) — Black produced invalid output → **Black bug → 500**
3. **Source ≠ Dest** (`assert_equivalent`, line 1656) — Black produced non-equivalent output → **Black bug → 500**
4. **Version mismatch** (`format_file_contents`, line 1091) — re-raises with context → depends on underlying cause

My original PR incorrectly treats ALL of these as 400, which is wrong for cases 2 and 3.

I agree with your suggestion to split the source-parse-failure case into its own exception. I'll prepare a new version that:

1. Introduces a new exception (e.g. `SourceASTParseError`) for case 1
2. Keeps `ASTSafetyError` for cases 2-3 (Black bugs)
3. Updates `blackd` to catch `SourceASTParseError` → 400, while `ASTSafetyError` continues to fall through to the generic 500 handler

Let me know if you'd prefer a different name for the new exception, or if you'd like a different approach entirely. Happy to implement whichever direction you prefer.

PR #5092, opened 2026-04-11
In stub files, enforce a blank line between a function and a decorated class definition

### LINKED ISSUE #4256: Black removes blank lines in between a function and a decorated class in a stub file
**Describe the bug**

Black reformats the following `.pyi` stub file:

```py
def foo(): ...

@decorator
class Bar: ...
```

to this:

```py
def foo(): ...
@decorator
class Bar: ...
```

If `Bar` does not have a decorator, black permits the empty line between the two definitions. I think the empty line improves readability here quite a bit, so I'd prefer it if black would allow (or even enforce it). Decorated classes haven't historically been very common in stub files (previously, I think the only major one was maybe `@final`?), but we'll get more of them now that we have `@deprecated`: see https://github.com/python/typeshed/pull/11488

AlexWaygood:
### Description

In `.pyi` stub files, Black already enforces a blank line before a (decorated or undecorated) class definition in many situations. However, it fails to do so if the decorated class definition is immediately preceded by a function.

For this example file, saved as `foo.pyi`:

```py
def a(): ...
class A: ...

@deco
class B: ...
@deco
class C: ...

def d(): ...

@deco
class D: ...

def e(): ...
@deco
class E: ...
```

Black applies this formatting on `main` with `--preview`

```diff
--- foo.pyi	2026-04-11 13:21:51.522132+00:00
+++ foo.pyi	2026-04-11 13:22:54.144533+00:00
@@ -1,15 +1,16 @@
 def a(): ...
+
 class A: ...
 
 @deco
 class B: ...
+
 @deco
 class C: ...
 
 def d(): ...
-
 @deco
 class D: ...
 
 def e(): ...
 @deco
```

On this PR branch, the formatting changes to:

```diff
--- foo.pyi	2026-04-11 13:21:51.522132+00:00
+++ foo.pyi	2026-04-11 13:23:18.497329+00:00
@@ -1,16 +1,19 @@
 def a(): ...
+
 class A: ...
 
 @deco
 class B: ...
+
 @deco
 class C: ...
 
 def d(): ...
 
 @deco
 class D: ...
 
 def e(): ...
+
 @deco
 class E: ...
```

which feels much more consistent.

The fix adds a `_decorator_decorates_class()` helper that walks the AST to check whether a decorator line decorates a `classdef` node, and uses it to insert a blank line in `_maybe_empty_lines_for_class_or_def` when appropriate.

Fixes https://github.com/psf/black/issues/4256. The same bug was fixed in the Ruff formatter in June, in https://github.com/astral-sh/ruff/pull/18888

### Checklist - did you ...


CHANGES.md
```diff
@@ -23,6 +23,8 @@
 - Improve heuristics around whether blank lines should appear before, within and after
   groups of same-name decorated functions (such as `@overload` groups) in `.pyi` stub
   files (#5021)
+- Fix blank lines being removed between a function and a decorated class in `.pyi` stub
+  files (#4256)
```
AlexWaygood:
```suggestion
  files (#5092)
```

AlexWaygood:
The diff-shades preview changes all look like clear improvements to me

docs/the_black_code_style/future_style.md
```diff
@@ -189,6 +192,34 @@ def foo(x: str) -> str: ...
...
 
+(labels/pyi-blank-line-before-decorated-class)=
+
+### Blank line before decorated classes in stub files
+
+In `.pyi` stub files, Black already enforces blank lines around class definitions.
+However, when a class has a decorator, Black previously failed to enforce this, removing
+the blank line between a function and the decorator:
+
+```python
```
MeGaGiGaGon: This may be clearer if diffs were used for the examples, ie for before:
```diff
 def foo(): ...
-
 @decorator
 class Bar: ...
```
and for after:
```diff
 def foo(): ...
+
 @decorator
 class Bar: ...
```
(not suggestion because github fails with nested blocks)

AlexWaygood: > (not suggestion because github fails with nested blocks)

ah, you can add a suggestion with a nested diff block by using a triple-fenced `diff` block inside a quadruple-fenced `suggestion` block! E.g.

````suggestion
here is my markdown prose

```python
a = 42
```

and here is some more markdown prose
````

Comment (AlexWaygood) [reply docs/the_black_code_style/future_style.md]: (I switched to using diff examples!)

**Review by MeGaGiGaGon — APPROVED** [review]

File: docs/the_black_code_style/future_style.md [hunk 3]
```diff
@@ -189,6 +192,44 @@ def foo(x: str) -> str: ...
...
+  def foo(): ...
+
+  @decorator
+  class Bar: ...
+
+  def baz(): ...
++
+  @decorator
+  class Spam: ...
+```
```
cobaltt7, line 231:
````suggestion
```diff
# Before

def foo(): ...
-
@decorator
class Bar: ...

def baz(): ...
@decorator
class Spam: ...
```

With this feature enabled, a blank line is enforced, consistent with how undecorated
classes are handled:

```diff
# After (with --preview)

def foo(): ...

@decorator
class Bar: ...

def baz(): ...
+
@decorator
class Spam: ...
```
````

cobaltt7 (approved):

docs/the_black_code_style/future_style.md
AlexWaygood: Hmm, this is a diff code block (following @MeGaGiGaGon's suggestion) — aren't unchanged lines in diffs usually indented two spaces, so that they're indented more than the `+` and `-` signs at the beginning of changed lines?

cobaltt7: Ah, right, that makes sense. I missed that, sorry

AlexWaygood: No worries

JelleZijlstra (approved):

PR #5095, opened 2026-04-12
Fix unnecessary parentheses in indexed assignment RHS

### LINKED ISSUE #4349: Unnecessary Parentheses Added to Expression in Indexed Assignment
**Describe the bug**

Black unnecessarily adds parentheses around the right-hand side expression in indexed assignments when breaking lines for long index names. This results in added visual clutter and does not conform to typical Python formatting standards where no additional parentheses would be expected. In particular, PyLint raises the [superfluous-parens](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/superfluous-parens.html) warning.

**To Reproduce**

Given the following Python script:

```python
import numpy as np

def main():
    dictionary_of_arrays = {
        "long_key_name_for_the_example": np.zeros((3, 1)),
        "short": np.zeros((3, 1)),
    }
    index_zero = 0
    very_long_index_name = 1
    dictionary_of_arrays["long_key_name_for_the_example"][very_long_index_name, index_zero] = 10 - 5
```

When formatted using the online Black formatter at <https://black.vercel.app/?version=main>, the assignment line is formatted as follows:

```python
dictionary_of_arrays["long_key_name_for_the_example"][
    very_long_index_name, index_zero
] = (10 - 5)
```

**Expected behavior**

The expected code should look like this:

```python
dictionary_of_arrays["long_key_name_for_the_example"][
    very_long_index_name, index_zero
] = 10 - 5
```

**Environment**

- Black's version: main (online formatter)
- OS and Python version: N/A (used the online formatter)

**Additional context**

None.

ParamChordiya:
## Summary

Fixes #4349.

When an assignment target contains brackets (e.g. indexed access like `x[key] = expr`), Black would wrap the right-hand side expression in unnecessary parentheses when the line was too long.

**Before:**
```python
dictionary_of_arrays["long_key_name_for_the_example"][
    very_long_index_name, index_zero
] = (10 - 5)
```

**After:**
```python
dictionary_of_arrays["long_key_name_for_the_example"][
    very_long_index_name, index_zero
] = 10 - 5
```

## Details

The root cause is in `can_omit_invisible_parens()` in `src/black/lines.py`. This function decides whether optional parentheses around the RHS can be safely removed. It returned `False` for simple expressions with operators (like `10 - 5`) because they don't start or end with brackets — falling through to the default `return False`.

The fix adds a targeted check: when the RHS head ends with `=` and the LHS contains brackets (indicating an indexed/subscript assignment), and the body expression is short enough, allow omitting the optional parens. The downstream `_prefer_split_rhs_oop_over_rhs()` still makes the final decision on whether the resulting layout is better.

The `line_length // 2` threshold ensures we only omit parens for genuinely short RHS expressions — long expressions that need their own line splitting still get wrapped as before.

## Test plan

- Added test cases in `tests/data/cases/prefer_rhs_split.py` (idempotent formatting)
- Added input→output test case in `tests/data/cases/prefer_rhs_split_reformatted.py`
- Verified fix against the exact reproducer from #4349


src/black/lines.py
```diff
@@ -1350,6 +1350,7 @@ def can_be_split(line: Line) -> bool:
 def can_omit_invisible_parens(
     rhs: RHSResult,
     line_length: int,
+    mode: Mode = Mode(),
```
cobaltt7: I'd prefer this parameter to be required so it's not mistakenly forgotten anywhere `can_omit_invisible_parens` is called

src/black/lines.py
```diff
@@ -1479,6 +1480,23 @@ def can_omit_invisible_parens(
...
+    # whether it actually produces better output.
+    if (
+        Preview.fix_unnecessary_parens_in_indexed_assignment in mode
+        and len(rhs.head.leaves) >= 2
+        and rhs.head.leaves[-2].type == token.EQUAL
+        and any(leaf.type in BRACKETS for leaf in rhs.head.leaves[:-2])
+    ):
+        # Only when the body is short enough to fit on the tail line
+        # after the LHS bracket split (e.g. `] = 10 - 5`).
+        body_length = str_width(str(line).strip("\n"))
```
cobaltt7: `str` and `str_width` are rather performance-heavy, and it's a somewhat hacky way to do this here. Is there an AST-only way to do this? (Perhaps you add optional params and update the logic that decides which ones are visible or not?)

src/black/lines.py
ParamChordiya: Done — made `mode` a required parameter (removed the `= Mode()` default). It's only called in one place (`linegen.py:1069`) which already passes it explicitly, so no other call sites needed updating. This also fixes the flake8 B008 lint failure.

src/black/lines.py
ParamChordiya: Good call — replaced `str_width(str(line).strip("\n"))` with an AST-based approach using `line.enumerate_with_length()`, which sums leaf lengths directly. This follows the same pattern used by `_can_omit_closing_paren` just above.

docs/the_black_code_style/future_style.md
```diff
@@ -189,6 +192,33 @@ def foo(x: str) -> str: ...
...
 ```
 
+(labels/fix-unnecessary-parens-indexed-assignment)=
+
+### Unnecessary parentheses in indexed assignments
+
+When an assignment target contains brackets (e.g. indexed access like `x[key] = expr`),
+Black would previously wrap the right-hand side expression in unnecessary parentheses
+when the line was too long. With this feature enabled, Black removes the unnecessary
+parentheses when the RHS expression is short enough.
```
cobaltt7: Could we make it clearer that this was a bug? Maybe "Previously, Black would incorrectly wrap the right-hand side..."

src/black/lines.py
```diff
@@ -1479,6 +1480,25 @@ def can_omit_invisible_parens(
...
+        and rhs.head.leaves[-2].type == token.EQUAL
+        and any(leaf.type in BRACKETS for leaf in rhs.head.leaves[:-2])
+    ):
+        # Only when the body is short enough to fit on the tail line
+        # after the LHS bracket split (e.g. `] = 10 - 5`).
+        body_length = 4 * line.depth
+        for _index, _leaf, leaf_length in line.enumerate_with_length():
+            body_length += leaf_length
+        if body_length <= line_length // 2:
+            return True
```
cobaltt7: I think there's a better heuristic here, probably by leveraging the optional parentheses logic.
For example, this is still be parenthesized -
```py
dictionary_of_arrays["long_key_name_for_the_example"][
    very_long_index_name, index_zero
] = (1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1)
```

docs/the_black_code_style/future_style.md
ParamChordiya: updated the wording to say - incorrectly wrap

src/black/lines.py
ParamChordiya: moved the check before the delimiter_count > 1 early return (which was blocking multi-operator expressions), replaced line_length // 2 with line_length, and added a has_brackets check. matching the approach of _can_omit_opening_paren/_can_omit_closing_paren

cobaltt7 (dismissed):

cobaltt7:
Almost missed this:
There's a few side-effects here that don't look like an improvement to me
image
image
image


Black would previously prefer to wrap long lines in parentheses to get it under the line length:
```py
# -l 80
middle.children[1].prefix = (
    middle.children[0].prefix + middle.children[1].prefix
)
```
With this change, Black always prefers to split lines instead of wrap them, which isn't an improvement
```py
# -l 80
middle.children[1].prefix = middle.children[0].prefix + middle.children[
    1
].prefix
```

ParamChordiya:
Hey @cobaltt7 , thanks for catching that, really appreciate the thorough review. You were right, and it took a fair bit of digging to get to the bottom of it.                                                                                              
                                                                                                                                                                                                                                                              
What was happening                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                              
The regression you spotted (middle.children[1].prefix = ... splitting mid-chain at the subscript instead of paren-wrapping the whole RHS) was happening because _prefer_split_rhs_oop_over_rhs in linegen.py was choosing the OOP (omit-optional-parens) split path even when the result was ugly. The OOP split would break at middle.children[ and put the index 1 on its own line, which is clearly wrong for a chained attribute access. 
To fix it, I added a guard in _prefer_split_rhs_oop_over_rhs that says: if the OOP tail starts with ] followed by a . (meaning we'd be splitting inside a subscript access chain like obj[idx].attr), and the = is in the head (so the split is on the RHS, not the LHS), prefer paren-wrapping instead. 

The first round of issues 
After gating this guard behind Preview.fix_unnecessary_parens_in_indexed_assignment (to avoid affecting stable mode), the diff-shades corpus check was clean, but "Format ourselves" started failing on lint and the uvloop platforms. Turned out the guard was slightly too broad the condition len(rhs_oop.tail.leaves) > 1 was firing for list literal assignments like expected: list[Path] = [items] because the tail had ] plus a NEWLINE token (count=2, not just ] alone). That caused those list assignments to get incorrectly paren-wrapped. Fixed by tightening to rhs_oop.tail.leaves[1].type == token.DOT only suppress OOP when the character immediately after ] is actually a dot (attribute access). 

The second bug, the trickier one
Even after that, "Format ourselves" was still failing because __init__.py kept getting reformatted. Traced it down to a separate bug in can_omit_invisible_parens in lines.py. The indexed-assignment early-return block checks whether there are brackets in the LHS:   any(leaf.type in BRACKETS for leaf in rhs.head.leaves[:-2])                                                                                                                                                                                                 
The problem: Python's AST wraps multi-target assignments like root, method = expr with invisible parentheses — these show up as empty-value LPAR/RPAR tokens in the leaf list. Those invisible tokens were matching the bracket condition and causing the indexed-assignment OOP path to fire for completely unrelated tuple assignments. That led to the outer parens being removed from the ternary in __init__.py, which is stable-idempotent but not unstable-idempotent, creating a formatting loop.   

The fix was a one-character addition: and leaf.value, only count a bracket if it has an actual visible value, which correctly excludes the implicit/invisible tokens that the parser inserts.                                                              
                                                                                                                                                                                                                                                              
  How we verified                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                              
  - All 211 format tests pass, including the prefer_rhs_split regression cases added in the original PR                                                                                                                                                       
  - test_piping passes (reads __init__.py in stable mode and expects idempotency)                                                                                                                                                                             
  - python -m black --check src/ tests/ with the project's own unstable = true config reports 57 files unchanged                                                                                                                                              
  - __init__.py is idempotent under both stable and unstable mode now
  
The net change across both files is two lines, but it took some real archaeology to get there, the invisible-bracket issue in particular was subtle since nothing in the usual formatting output hinted that those tokens even existed.

cobaltt7:
Please check the shades changes - it's still splitting within function calls.

cobaltt7:
Hey @ParamChordiya, are you still updating this PR?

ParamChordiya:
> Hey @ParamChordiya, are you still updating this PR?

Hey @cobaltt7  sorry i got caught up with something else and lost track of this PR. let me circle back soon.

ParamChordiya:
Hey @cobaltt7, back on this — thanks for your patience. You were right that the earlier approach was still too aggressive, so I reworked the heuristic instead of patching around it.

**What was causing the bad shades changes**

The check in `can_omit_invisible_parens` had an escape hatch that allowed omitting the parens whenever the RHS contained *any* bracket, even when the RHS didn't fit on one line. The splitter would then break at the last bracket inside the expression — which is exactly the "splitting within function calls" you flagged (ternaries like `kwargs["empty_label"] = kwargs.get(...) if ... else None`, and method chains like `_get_totp(...).generate(...).decode()`). It was also matching things that aren't indexed assignments at all: parenthesized tuple targets (`(self.next,) = ...`) and annotated assignments (`self.args: list[str] = ...`), because it only looked for "any bracket before the `=`".

**The new approach**

The paren-omission path now only fires when all of these hold:

- the target genuinely ends in a subscript — a visible `]` immediately before the `=`
- it isn't an annotated assignment (so the `]` of `list[str]` doesn't count)
- the LHS is too long to fit on one line, so Black is forced to split at the subscript brackets regardless
- the RHS fits on the resulting `] = expr` tail line

In other words, it only removes the parens in the exact #4349 situation: the line has to split at the LHS anyway, and the parens on the tail are pure noise. When the LHS fits, paren-wrapping the RHS stays the preferred style, untouched. With the bad candidates eliminated at the source, the subscript-chain guard I'd added to `_prefer_split_rhs_oop_over_rhs` became dead code, so it's removed — the diff is smaller than before.

**How I verified**

- I reconstructed every hunk from the last diff-shades report (all 88 hunks / 758 changes across the 9 projects) and ran the new build over them: zero changes remain, so the only diff-shades delta left should be the intended #4349 shape.
- All format tests pass, and I added regression cases for each failure mode you surfaced (ternary RHS, method chains, tuple targets, annotated assignments, and a too-long RHS that should still wrap).
- Black's own sources stay clean and idempotent under stable, preview, and unstable modes.
- A generated sweep of ~1000 assignment variants (slices, augmented/chained/annotated targets, walrus, lambdas, comments, wide chars) shows no stable-vs-preview differences outside the intended improvement, and no idempotency issues.

Also merged `main` and resolved the conflicts, so the branch is mergeable again. Would appreciate another look when you get a chance!

cobaltt7:
Thanks!

PR #5096, opened 2026-04-12
Prevent string merger from creating unsplittable long lines

### LINKED ISSUE #4510: string_processing: Multi-line strings with type ignore pragma get merged into a single line
**Describe the bug**

There is a clash between the process of reflowing multi-line strings and the rule that prevents formatting lines with `# type: ignore` pragma at the end. Here is what happens:

1. A multi-line string gets first merged into a single-line string which can be arbitrarily long.
2. Because of the pragma comment at the end of the line, the line is considered unsafe to split. Hence the final result is a single huge line.

**To Reproduce**

Here is a simple example:

```python
(
    "A very very very very very very very very very very very very long string "
    "annotated with a type ignore pragma gets merged into a single very long line "
    "which is against the line length rule."
)  # type: ignore
```

Running `black --unstable example.py` produces:

```python
("A very very very very very very very very very very very very long string annotated with a type ignore pragma gets merged into a single very long line which is against the line length rule.")  # type: ignore
```

**Expected behavior**

Probably the best option would be that the multi-line string wouldn't be reflown at all in cases there is a type ignore comment at the end. But the formatting definitely shouldn't produce code that breaks the line length rule.

**Environment**

- Black's version: `24.10.0`
- OS and Python version: Linux, Python 3.11.9

**Additional context**

In practice this issue can appear if users do some f-string operations inside a multi-line string and want to suppress type checking. E.g. the following line also gets incorrectly reflown into a single huge line:

```python
message = (
    "This is an error message that also does an f-string operation for which "
    f"type checks might have to be ignored: {value1 + value2} ")  # type: ignore
```

ParamChordiya:
## Summary

Fixes #4510.

When multi-line strings are followed by a pragma comment (like `# type: ignore`) on a surrounding bracket, the `StringMerger` would merge them into a single long line. Because pragma comments prevent line splitting, this produced lines that violate the line length limit with no way to fix them.

**Before** (`black --unstable`):
```python
# Input:
(
    "A very very very very very very very very very very long string "
    "continued on the next line."
)  # type: ignore

# Output (one huge unsplittable line):
("A very very very very very very very very very very long string continued on the next line.")  # type: ignore
```

**After:** The multi-line string is left as-is, preserving the line length constraint.

## Root Cause

`_validate_msg()` in `StringMerger` only checked for pragma comments (`# type: ignore`, `# noqa`, `# pylint:`) on the string leaves themselves. It did not check surrounding tokens like closing brackets `)`, `]`, `}`. The pragma comment on `)` was invisible to the validation.

## Fix

Extended `_validate_msg()` to also scan tokens following the string group (walking through closing brackets) for pragma comments. If any are found, the merge is refused.

## Test plan

- Added regression test in `tests/data/cases/preview_long_strings__regression.py`
- Verified both examples from #4510
- Verified normal string merging (without pragma comments) still works

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,8 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Prevent string merger from creating unsplittable long lines when a pragma comment
+  (e.g. `# type: ignore`) follows the closing bracket (#5096)
 - Improve heuristics around whether blank lines should appear before, within and after
   groups of same-name decorated functions (such as `@overload` groups) in `.pyi` stub
   files (#5021)
--- a/src/black/trans.py
+++ b/src/black/trans.py
@@ -845,6 +845,24 @@ class StringMerger(StringTransformer, CustomSplitMapMixin):
                 f"Not enough strings to merge (num_of_strings={num_of_strings})."
             )
 
+        # Also check for pragma comments on tokens that follow the string
+        # group (e.g. a closing bracket).  Merging strings when a pragma
+        # comment like `# type: ignore` follows would produce an unsplittable
+        # long line.
+        is_valid_index = is_valid_index_factory(line.leaves)
+        next_idx = string_idx + num_of_strings
+        while is_valid_index(next_idx):
+            next_leaf = line.leaves[next_idx]
+            if id(next_leaf) in line.comments:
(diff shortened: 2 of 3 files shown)


cobaltt7:
Thanks!

PR #5097, opened 2026-04-12
Fix crash when f-string follows # fmt: off inside brackets

### LINKED ISSUE #4511: Error formatting f-string with `# fmt: off` inside a pair of brackets
**Describe the bug**

If there is a f-string (single-line or multi-line) following a `# fmt: off` (with or without a pairing `# fmt: on`) and both the f-string and `# fmt: off` are inside a pair of brackets (`()`, `[]`, `{}`), the code cannot be formatted.

The error reported is something like:

```plain
error: cannot format ???.py: {' ', 'r', 'f', 'o', ':', 'm', 't', '#', '\n'} is NOT a subset of {'U', 'r', 'f', 'u', 'F', 'R', 'b', 'B'}.
```

**To Reproduce**

Sample:

```python
(
# fmt: off
f"""
"""
# fmt: on
)
```

**Expected behavior**

No error is reported and the code is formatted as if the f-string is a normal string.

**Environment**

- Black's version: 24.10.0
- OS and Python version: Windows and Linux / Python 3.13.0

**Additional context**

None

ParamChordiya:
## Summary

Fixes #4511.

Black crashes with an `AssertionError` when formatting code that contains an f-string after a `# fmt: off` comment inside brackets:

```python
(
# fmt: off
f"""
"""
# fmt: on
)
```

Error:
```
error: cannot format: {' ', 'r', 'f', 'o', ':', 'm', 't', '#', '\n'} is NOT a subset of {'U', 'r', 'f', 'u', 'F', 'R', 'b', 'B'}.
```

## Root Cause

`normalize_fmt_off` runs before line generation and converts content between `# fmt: off`/`# fmt: on` into `STANDALONE_COMMENT` nodes. When this content is inside an f-string node, the f-string's children (FSTRING_START, FSTRING_MIDDLE, FSTRING_END) get replaced with a single STANDALONE_COMMENT leaf.

When `visit_fstring` is later called for this modified node, it calls `fstring_tstring_to_string()` which does `str(node)[len(node.prefix):]`. This produces a value like `# fmt: off\nf"""\n"""` — the comment text becomes part of the "string value", which fails the string prefix assertion.

## Fix

Added a guard in `visit_fstring` (and `visit_tstring`) to detect when the node contains a STANDALONE_COMMENT child (indicating it was modified by `normalize_fmt_off`). In that case, fall back to `visit_default` instead of attempting the f-string-to-string conversion.

## Test plan

- Added regression tests in `tests/data/cases/fmtonoff6.py` for both triple-quoted and single-line f-strings with `# fmt: off` inside brackets
- Verified the fix produces idempotent output

### DIFF
--- a/tests/data/cases/fmtonoff6.py
+++ b/tests/data/cases/fmtonoff6.py
@@ -27,3 +27,63 @@ dependencies = {
     f"{x}"
     # fmt: on
 ]
+
+# Same as above but unindented inside brackets.
+(
+# fmt: off
+f"""
+"""
+# fmt: on
+)
+
+[
+# fmt: off
+f"{x}"
+# fmt: on
+]
+
+# output
+
+# Regression test for https://github.com/psf/black/issues/2478.
+def foo():
+    arr = (
+        (3833567325051000, 5, 1, 2, 4229.25, 6, 0),
+        # fmt: off
+    )
+
+
+# Regression test for https://github.com/psf/black/issues/3458.
+dependencies = {
+    a: b,
+    # fmt: off
+}
+
+
+# Regression test for https://github.com/psf/black/issues/4511.
+# fmt: off inside brackets with an f-string should not crash.
+(
+    # fmt: off
+    f"""
+"""
+    # fmt: on
+)
+
+[
+    # fmt: off
+    f"{x}"
+    # fmt: on
+]
+
+# Same as above but unindented inside brackets.
+(
+    # fmt: off
+f"""
+"""
+    # fmt: on
+)
+
+[
+    # fmt: off
+f"{x}"
+    # fmt: on
+]
--- a/tests/data/cases/fmtonoff6.py
+++ b/tests/data/cases/fmtonoff6.py
@@ -11,3 +11,19 @@ dependencies = {
     a: b,
     # fmt: off
 }
+
+
+# Regression test for https://github.com/psf/black/issues/4511.
+# fmt: off inside brackets with an f-string should not crash.
+(
+    # fmt: off
+    f"""
+"""
+    # fmt: on
+)
+
+[
+    # fmt: off
+    f"{x}"
+    # fmt: on
+]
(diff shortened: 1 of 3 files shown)


tests/data/cases/fmtonoff6.py
```diff
@@ -11,3 +11,19 @@ def foo():
     a: b,
     # fmt: off
 }
+
+
+# Regression test for https://github.com/psf/black/issues/4511.
+# fmt: off inside brackets with an f-string should not crash.
```
cobaltt7, line 17: Please also add unintended tests, ie
```py
(
# fmt: off
f"""
"""
# fmt: on
)
```

ParamChordiya: Added unindented test cases in c86c2f2. Thanks for the suggestion!

tests/data/cases/fmtonoff6.py
```diff
@@ -11,3 +11,79 @@ def foo():
...
+
+[
+    # fmt: off
+    f"{x}"
+    # fmt: on
+]
+
+# Same as above but unindented inside brackets.
+(
+    # fmt: off
```
cobaltt7, line 79: Ideally the comments wouldn't move. Would that be a lot of extra work? If so it's fine like this.

Please also add a changelog entry

ParamChordiya: The comment movement (re-indenting `# fmt: off`/`# fmt: on` to match bracket depth) is existing Black behavior for any comment inside brackets — it's not introduced by this PR. Changing that would require modifying the general comment normalization/line generation logic, which feels out of scope for this crash fix. Happy to look into it as a follow-up if you think it's worth pursuing.

Added a changelog entry under **Stable style** in 67adc97.

JelleZijlstra (approved):

tests/data/cases/fmtonoff6.py
cobaltt7: Sounds good. Thank you!

PR #5114, opened 2026-04-22
fix(fmt: skip): respect skip directive on multi-line compound statements

### LINKED ISSUE #5112: `# fmt: skip` regression
**Describe the bug**

`# fmt: skip` is not handled properly in all cases with black 26.1 and newer.

**To Reproduce**

Our project has this code:

```python
class Body(model.BaseBody[
    "Keyword", "For", "While", "Group", "If", "Try", "Var", "Return", "Continue",
    "Break", "model.Message", "Error"
]):  # fmt: skip
    __slots__ = ()
```

Older black versions have left that alone as expected, but starting from black 26.1 the result is this:

```python
class Body(model.BaseBody["Keyword", "For", "While", "Group", "If", "Try", "Var", "Return", "Continue", "Break", "model.Message", "Error"]):  # fmt: skip
    __slots__ = ()
```

**Expected behavior**

I'd expect the code not to be touched as earlier.

**Environment**

- Black's version: 26.1 - 26.3.2.dev31+g8dcc4865c
- OS and Python version: Linux, Python 3.13

**Additional context**

Other `# fmt: skip` usages seem to work fine also with new black versions. For example, we have the following code right above the earlier example and it is not modified:

```python
BodyItemParent = Union[
    "TestSuite", "TestCase", "UserKeyword", "For", "While", "If", "IfBranch",
    "Try", "TryBranch", "Group", None
]  # fmt: skip
```

### LINKED ISSUE #5122: Fail to parse multiline `case` statement with `# fmt: skip`
**Describe the bug**

Parsing a multiline `case` statement with `# fmt: skip` fails with black >= 26.

**To Reproduce**

For example, take this code:

```python
# file.py
match (method, *path.split("/")):

    case ("GET", "parent", _, "resource", resource_id) \
            | ("GET", "resource", resource_id):  # fmt: skip
        pass

    case _:
        pass
```

```sh
$ black file.py
```

The resulting error is:

> error: cannot format file.py: Cannot parse: 6:42:         | ("GET", "resource", resource_id):  # fmt: skip

**Expected behavior**

The multi-line `case` statement remains unchanged.

**Environment**

- Black's version: 26.1.0 and up (no issue with 25.12.0).
- OS and Python version: [Windows/Python 3.13.2] -->

armorbreak001:
## Summary

Fixes #5112, closes #5117
fixes #5122, closes #5123

When `# fmt: skip` is placed on the colon line of a multi-line class (or def/if) statement, Black still reformatted it — collapsing multi-line type parameters onto a single line.

**Root cause:** In `_generate_ignored_nodes_from_fmt_skip()`, line 647 unconditionally sets `prev_sibling = parent.prev_sibling` when the leaf (NEWLINE) has no previous sibling. This causes the function to take the **per-sibling traversal path** (which stops at any leaf containing a newline in its prefix) instead of the **suite-sibling collection path** (the `elif` branch at line 738, which correctly collects ALL previous siblings of the suite node).

For a multi-line class like:
```python
class Foo(Base[\
    "A", "B"\
]):  # fmt: skip
```
The traversal starts at `:` and collects `]`, `)`, `:` — then hits `]` whose prefix contains `\n` (end of line), stopping early. The entire class header (`class Foo(Base[`, type params) is never marked as ignored, so Black reformats it.

**Fix:** Add `parent.type != syms.suite` condition on line 646 so that suite NEWLINEs fall through to the `elif` branch that properly collects all statement header nodes.

## Test plan

- [x] Reproduced issue #5112 with the exact example from the bug report — now left unchanged
- [x] Verified existing `# fmt: skip` cases still work (assignments, one-line compounds, etc.)
- [x] All 210 format tests pass
- [x] All 19 fmtskip format tests pass
- [x] All 3 skip-related unit tests pass

### DIFF
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -643,7 +643,7 @@ def _generate_ignored_nodes_from_fmt_skip(
     if not comments or comment.value != comments[0].value:
         return
 
-    if not prev_sibling and parent:
+    if not prev_sibling and parent and parent.type != syms.suite:
         prev_sibling = parent.prev_sibling
 
     if prev_sibling is not None:


pekkaklarck:
Thanks for taking a look at this and for the quick fix @armorbreak001! I don't know the rules of this project, but in general I believe it would be a good idea to add a new test that reproduces the bug and is fixed by this change.

cobaltt7:
Thanks @armorbreak001! Yes, please do add a new test case, as well as a changelog entry.

PR #5127, opened 2026-05-11
Add CI coverage for Python 3.15

JelleZijlstra:


hauntsaninja (approved):
Looks good, although I would not in the changelog that we don't ship accelerated wheels yet

PR #5135, opened 2026-05-18
Don't break before a comparator when the RHS bracket explodes

### LINKED ISSUE #4514: Line break added in list comprehension separates variable from its expression
**Black https://github.com/psf/black/commit/c47255**

[Playground link](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AEBAIJdAD2IimZxl1N_WmyLkeRBtS4hSHlEaQHnSb0yBGOHa3zzPX8cU0zwEgYPJ1lSxf7QQpvBFhKkPTxtpRBOxtEjPDiquhG9el77O5Vr4d8SFrDsvdsfYzWlILS5RYdQIBC-VUbe3-6PQCK6sTR_3xM8uQdomhdi-6Fk9h-v5UWQ0DswdwAAAACkO3dOoKwJeAABngGCAgAAgCCNnbHEZ_sCAAAAAARZWg==)

## Options

`--line-length=88`
`--safe`

## Input

```python
x = [
    t
    for t in y
    if t not in {
        LongNameOne,
        LongNameTwo,
        LongNameThree,
    }
]

```

## Output

```python
x = [
    t
    for t in y
    if t
    not in {
        LongNameOne,
        LongNameTwo,
        LongNameThree,
    }
]

```

## Expected

```python
x = [
    t
    for t in y
    if t not in {
        LongNameOne,
        LongNameTwo,
        LongNameThree,
    }
]

```

By moving the `not in` onto its own line it separates the variable from the expression it's a part of, breaking visual parsing.

Pedro-Muller29:
### Description

Fixes #4514.

```python
# black --line-length=88
x = [
    t
    for t in y
    if t not in {
        LongNameOne,
        LongNameTwo,
        LongNameThree,
    }
]
```

Black rewrites this to:

```python
x = [
    t
    for t in y
    if t
    not in {
        LongNameOne,
        LongNameTwo,
        LongNameThree,
    }
]
```

`if t` lands on its own line, visually disconnected from the `not in` and the set it belongs to. The same shape shows up with `==`, `<`, `in`, `is`, and well beyond comprehensions: inside `if`/`elif` conditions, `assert` statements, and any parenthesized expression where a comparator sits in front of a bracket that has to break.

## Root cause

After the surrounding construct splits (a comprehension on `for`/`if`, an `if` on `and`/`or`, an `assert` opening its outer parens), the sub-line is something like `if t not in {Long1, Long2, Long3,}`. It's `inside_brackets=True`, so `transform_line` runs `delimiter_split` first. The bracket tracker records exactly one delimiter at `COMPARATOR_PRIORITY = 10` (the `not`), so `delimiter_split` happily breaks before it. `right_hand_split`, which would have exploded the `{...,}` instead and kept `if t not in {` together, never gets a turn.

## Fix

In `delimiter_split`, bail out when the only delimiter is at `COMPARATOR_PRIORITY` and `right_hand_split` would produce a head that fits in the line length. `right_hand_split` then handles the bracket the way it would have anyway. Covers both the magic-trailing-comma case from the issue and the case where the RHS is long enough that Black has to break it anyway.

Gated behind `Preview.hug_comparator` since this changes already-formatted code.

Flagging that the check is narrow on purpose. `delimiter_count_with_priority == 1` keeps it to the "lone comparator with a bracketed RHS" case; chained comparisons (`a < b < c`) still split exactly as before. The fit check on `right_hand_split` also bails when exploding the bracket would just push the comparator back onto an overflowing line (e.g. an empty `decode()` paren).

### Checklist - did you ...


cobaltt7:
> One related shape stays broken on purpose for now: a bracket on the RHS _without_ a magic trailing comma, where the line is long enough that Black has to break it anyway. For example: ...

I think we should fix this here too. Otherwise, this is only a half-fix, and we're bound to get the same reports again.

Pedro-Muller29:
> > One related shape stays broken on purpose for now: a bracket on the RHS _without_ a magic trailing comma, where the line is long enough that Black has to break it anyway. For example: ...
> 
> I think we should fix this here too. Otherwise, this is only a half-fix, and we're bound to get the same reports again.

Fixed. Added a lookahead so we only defer to `right_hand_split` when its head fits, otherwise empty-bracket cases like `"..." in foo.decode()` would trade a stranded `if t` for an overflowing line.

docs/the_black_code_style/future_style.md
```diff
@@ -31,6 +31,10 @@ Currently, the following features are included in the preview style:
 - `pyi_blank_line_before_decorated_class`: In `.pyi` stub files, enforce a blank line
   before a decorated class definition when it follows a function definition.
   ([see below](labels/pyi-blank-line-before-decorated-class))
+- `avoid_splitting_comparator_with_bracketed_rhs`: Don't break a comparator expression
```
cobaltt7: (Nit) could the style name be shortened a bit?

cobaltt7 (commented):
Thanks a lot for this! The code looks good, and all the shades changes look excellent!
One last small note: looking at the shades changes, it affects a lot more than just comprehensions (in a good way). Could the docs examples and the tests be updated to show some of those changes as well? (Also, a lot of the current test cases feel a bit redundant since they all share the same general structure with comprehensions, and a few of them could probably be removed)

Pedro-Muller29:
> Thanks a lot for this! The code looks good, and all the shades changes look excellent! One last small note: looking at the shades changes, it affects a lot more than just comprehensions (in a good way). Could the docs examples and the tests be updated to show some of those changes as well? (Also, a lot of the current test cases feel a bit redundant since they all share the same general structure with comprehensions, and a few of them could probably be removed)

Thanks. All three:

- Flag renamed to `hug_comparator`.
- Docs and fixture now show non-comprehension shapes too: `if` in an `and` chain with `in`, plus `assert` with `is` against a call-chain RHS. Both lifted from the diff-shades comparison so they reflect real-world hits.
- Comprehension cases cut from 8 down to 4: one canonical `not in {,}` from the issue, one with `<` + call, the `or`-chained one, and one length-forced (`no_magic`).

Issue #4723, opened 2025-08-01
strange wrapping with long PEP 695 type and long signature


asottile-sentry:
**Describe the bug**

I expect the wrapping to look nicer

**To Reproduce**

For example, take this code:

```python
class Example:
    @overload
    @classmethod
    def create_from[TBool: (Literal[True], Literal[False])](
        cls: type[SearchConfig[Any]],
        search_config: SearchConfig[Any],
        *,
        allow_boolean: TBool,
        **overrides: Any,
    ) -> SearchConfig[TBool]: ...
```

And run it with these arguments:

```sh
$ black t.py --target-version py313 --line-length 100
```

and the result is:

```python
class Example:
    @overload
    @classmethod
    def create_from[TBool: (
        Literal[True],
        Literal[False],
    )](
        cls: type[SearchConfig[Any]],
        search_config: SearchConfig[Any],
        *,
        allow_boolean: TBool,
        **overrides: Any,
    ) -> SearchConfig[TBool]: ...
```

**Expected behavior**

(no change ideally?)

there's two weirdnesses
- the typevar was already pretty short so it should not have been wrapped
- the wrapping feels wrong -- it doesn't follow black's usual parameter indentation

**Environment**

- Black's version: 25.1.0
- OS and Python version: macos 3.13.1

**Additional context**

nope!

RizwaanShaik273:
Hi @asottile-sentry, thanks for reporting this.

I’ve taken an initial look, and this seems to be a formatting decision rather than a parsing issue. The wrapping appears to come from the line-splitting logic for bracketed expressions, where PEP 695 type parameter bounds are treated like generic containers and split even when they fit within the configured line-length.

I believe the fix will involve adjusting the splitting/indentation logic in src/black/linegen.py (and possibly brackets.py) so that short type parameter bounds either remain on one line or follow Black’s usual signature-style indentation. I can also add test coverage under tests/data/cases for this scenario.

If this sounds reasonable, I’d be happy to work on a fix.

Pedro-Muller29:
Already fixed on `main`. Bisected to 8203030 (PR #4777). Opening a PR with a regression test for this snippet.

PR #5143, opened 2026-05-19
Add regression test for #4723

Pedro-Muller29:
### Description

Regression test for #4723. Bisected the fix to 8203030 (PR #4777)

### Checklist - did you ...


cobaltt7:
I don't think a regression test for this exact situation is needed. #4777 already added tests for this generic shape. But thanks for the investigation!

Issue #4036, opened 2023-11-09
"Black produced code that is not equivalent to the source" error in preview mode on function call with two `type:ignore`


DetachHead:
```py
foo(  # type:ignore
    [  # type:ignore
        a
    ]
)
```
```
INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_qaa1dff6.log

--- src
+++ dst
@@ -29,8 +29,6 @@
         )  # /Call
     )  # /Expr
   type_ignores=
     TypeIgnore(
     )  # /TypeIgnore
-    TypeIgnore(
-    )  # /TypeIgnore
 )  # /Module
\ No newline at end of file
```
[playground](https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ACiAGJdAD2IimZxl1N_WlkOE2TjAlzwFHg1OzrKA2GPMsnbWWKAwnOpXnmNcPrWu0ypzX9sCUP2QqK_81z3KCstrgcBnvdB8Dyo8yqN4YwB6cegexiMRHo1hO6IuA3EkzU_E6TEMRIAAAAAoq3_awojetoAAX6jAQAAAE86Rn2xxGf7AgAAAAAEWVo=)

JelleZijlstra:
Thanks! cc @henriholopainen for the single-argument call hugging change.

bukzor:
The example above currently does not crash:

```console
$ black --version
black, 24.4.2 (compiled: yes)
Python (CPython) 3.12.3

$ cat example.py
foo(  # type:ignore
    [  # type:ignore
        a
    ]
)

$ black --enable-unstable-feature=hug_parens_with_braces_and_square_brackets  example.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

cobaltt7:
This example doesn't crash for me either, but here's one that does:
```py
items = [  # type: ignore
    (  # type: ignore
        {"key1": "val1", "key2": "val2", "key3": "val3"}
        if some_var == "longstring"
        else {"key": "val"}
    )
]
```

DetachHead:
i would recommend using ruff instead of black these days, which is able to format both examples without crashing

PR #5144, opened 2026-05-20
Fix crash on standalone comments in comprehensions and lambdas

### LINKED ISSUE #4296: Standalone comments still partially broken
**Describe the bug**

Specific placements of standalone comments, primarily in list comprehensions, will cause parsing errors.
These are actually formatting errors; black generates invalid code on the first run of `_format_str_once`, which then can't be parsed.

**To Reproduce**

I stumbled over two cases while working on #4026:

1. 
```py
[
    x for
    # comment
    x, y in ["AB",]
]
```
causes
> Cannot parse: 9:0: ]

This is because the code is first split at `# comment`, leaving 
```py
    x, y in ["AB",]
```
to [delimiter_split](https://github.com/psf/black/blob/836acad86371578527408a4c8f968cde1302e130/src/black/linegen.py#L1151), which doesn't know about the comprehension and adds a trailing comma after the `]`.

2.
```py
[
    [
        x for x
        # comment
        in [
            # comment
            "ABC"
        ]
    ]
]
```
This looks similar, but leads to a different error:
> Cannot parse: 7:0: EOF in multi-line statement

This is the code produced in the first formatting run:
```py
[
    [x for x
    # comment
    in [# comment"ABC"]]
]
```
Besides the broken comment, the style also looks pretty weird to me.
This is somehow related to the way [append_safe](https://github.com/psf/black/blame/836acad86371578527408a4c8f968cde1302e130/src/black/lines.py#L99-L117) doesn't allow comments if there is a comprehension or lambda in the line and can also be triggered with a lambda:
```py
[(lambda 
    # comment
    x: [
        # comment
        1
    ]
)]
```
There's other weirdness here. I haven't fully understood why this happens yet, but couldn't reduce the input further.

**Expected behavior**
The code should be formatted successfully and correctly.

**Suggestions**

I feel like there are probably more edge cases in which black expects a line to have a certain structure and crashes if standalone comments chop it up arbitrarily. It might be better to mostly ignore standalone comments and fix them as a last step.

Any other solution would have to give lines more context about their origin.
For example, the input 
```py
[
    x, y in ["AB",]
]
```
causes the exact same call to `delimiter_split` as the first example, but here the trailing comma is valid and correct.

**Environment**

```sh
$ black --version
black, 24.3.0 (compiled: yes)
Python (CPython) 3.12.1
```

**Additional context**

Pedro-Muller29:
### Description

Fixes #4296.

Three inputs from the issue crash on the first formatting pass:

```py
[
    x for
    # comment
    x, y in ["AB",]
]

[
    [
        x for x
        # comment
        in [
            # comment
            "ABC"
        ]
    ]
]

{(lambda
    # comment
    x: [
        # comment
    ]
)}
```

All of them blow up with `Cannot parse: ... bad input` because Black dumps a string where a standalone comment sits inline next to the following token, and Python's tokenizer swallows everything to the next newline.

## Root cause

Two distinct things.

For cases B and C, `transform_line` runs through its splitters and, when nothing succeeds, falls through to `yield line`. `Line.__str__` concatenates leaves without inserting newlines, so a `STANDALONE_COMMENT` mid-stream renders glued to whatever follows. `Line.append_safe` enforces "comment on its own line" only at bracket depth 0 or inside an open `for` / `lambda`. For comments living inside arbitrary nested brackets, there's no guard.

For case A, `standalone_comment_split` does pull the comment between `for` and the tuple target onto its own line, but the resulting sub-line `x, y in ["AB",]` no longer carries the `for` keyword. Its bracket tracker walks fresh, the comma between `x` and `y` lands at depth 0, gets `COMMA_PRIORITY`, and `delimiter_split` breaks the tuple apart. Black also tacks a stray trailing comma onto the iterable.

## Fix

- `linegen.py`: when no transformer succeeds and the line still carries standalone comments, fall back to `_force_standalone_comment_split`, which yields each comment on its own line. The alternative is invalid Python, so the unconditional split is safe.
- `brackets.py`: `is_split_after_delimiter` returns 0 for commas whose parent is `exprlist` and whose grandparent is `comp_for` / `old_comp_for` / `for_stmt`. Those commas are tuple-packing, not list delimiters.

The fixture covers the three issue cases plus async-for and a nested tuple target (`for (a, b), c in ...`) to lock both fixes in.

Flagging that case B comes out with loose indentation:

```py
[
    [x for x
    # comment
    in [
    # comment
    "ABC"]]
]
```

The inner `[...]` doesn't re-explode, so the comment and `"ABC"]]` sit at the wrong depth. Let me know if there's a target shape you'd want, or if this is fine for the crash fix.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,9 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Fix crash when a standalone comment sits between tokens of a comprehension or lambda
+  (#5144)
+
 ### Preview style
 
 <!-- Changes that affect Black's preview style -->
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,12 +16,12 @@
 <!-- Changes that affect Black's stable style -->
 
 <<<<<<< fix/4296-standalone-comments-comprehensions
+
 - Fix crash when a standalone comment sits between tokens of a comprehension or lambda
-  (#5144)
-=======
+  (#5144) =======
 - Fix a crash when splitting `case case if ...` match patterns at very small line
   lengths (#5147)
->>>>>>> main
+  > > > > > > > main
 
 ### Preview style
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,13 +15,10 @@
 
 <!-- Changes that affect Black's stable style -->
 
-<<<<<<< fix/4296-standalone-comments-comprehensions
-
 - Fix crash when a standalone comment sits between tokens of a comprehension or lambda
-  (#5144) =======
+  (#5144)
 - Fix a crash when splitting `case case if ...` match patterns at very small line
   lengths (#5147)
-  > > > > > > > main
 
 ### Preview style
 
--- a/src/black/brackets.py
+++ b/src/black/brackets.py
@@ -225,6 +225,17 @@ def is_split_after_delimiter(leaf: Leaf) -> Priority:
     Higher numbers are higher priority.
     """
     if leaf.type == token.COMMA:
+        # For-target commas are tuple-packing, not list delimiters; splitting
+        # on them produces invalid Python if `for` lands on another line (#4296).
+        parent = leaf.parent
+        if parent is not None and parent.type == syms.exprlist:
+            grandparent = parent.parent
+            if grandparent is not None and grandparent.type in {
+                syms.comp_for,
+                syms.old_comp_for,
+                syms.for_stmt,
+            }:
+                return 0
         return COMMA_PRIORITY
 
     return 0
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -833,7 +833,13 @@ def transform_line(
             break
 
     else:
-        yield line
+        # A leftover standalone comment would render inline and break the
+        # second pass with a parse error (#4296). Force a split so the
+        # output is at least valid Python.
+        if line.contains_standalone_comments():
+            yield from _force_standalone_comment_split(line)
+        else:
+            yield line
 
 
 def should_split_funcdef_with_rhs(line: Line, mode: Mode) -> bool:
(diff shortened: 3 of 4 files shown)


cobaltt7:
Thanks!

> Flagging that case B comes out with loose indentation: ...
> The inner `[...]` doesn't re-explode, so the comment and `"ABC"]]` sit at the wrong depth. Let me know if there's a target shape you'd want, or if this is fine for the crash fix.

Yeah, that's not ideal indentation. However, that's already handled with the `hug_parens_with_braces_and_square_brackets` style (even without this change, but the change is necessary for some of the other failures):
```py
[[
    x
    for x
    # comment
    in [
        # comment
        "ABC"
    ]
]]
```
On the other hand, that style is currently restricted to `--unstable` due to #4036 and some other issues.

If possible, it'd be nice to improve that here as well, but if it's a lot of extra code it's fine to merge as-is and depend on `hug_parens_with_braces_and_square_brackets`.

Pedro-Muller29:
Tried implementing it. The depth tracking itself is straightforward. Walk the leaves counting bracket nesting, emit each sub-line at `line.depth + nesting`. Applied to `standalone_comment_split`, case B comes out the way you described.

The problem is the closing brackets. To drop `]` back to the outer depth I had to force a new sub-line whenever a closer lowers nesting below the current sub-line's starting depth. That breaks `comments_in_lambda_default`: the trailing comma after `False` disappears, because `)` now lands on a fresh sub-line and the recursive `transform_line` pass no longer hits the path where `_ensure_trailing_comma` injects the comma for the outer `help(...)`.

Fixing that means threading magic trailing comma state into the new sub-line, which is reaching into things `hug_parens_with_braces_and_square_brackets` already handles.

cobaltt7:
Sounds good. Thank you for your investigation!

PR #5146, opened 2026-05-21
Do not use pip-install; add dependency cooldowns

JelleZijlstra:
Zizmor recommends against pip-install (https://docs.zizmor.sh/audits/#misfeature).

Also add dependency cooldowns to prevent supply-chain compromises.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,7 @@ updates:
     schedule:
       interval: daily
     cooldown:
-      default-days: 7
+      default-days: 2
 
   - package-ecosystem: "pip"
     directory: "/"
@@ -18,4 +18,4 @@ updates:
     schedule:
       interval: daily
     cooldown:
-      default-days: 7
+      default-days: 2
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     types: [opened, synchronize, labeled, unlabeled, reopened]
 
+env:
+  PIP_UPLOADED_PRIOR_TO: P2D
+
 permissions:
   contents: read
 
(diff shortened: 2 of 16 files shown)


cobaltt7:
Zizmor recommends against `pip-install` because it installs dependencies to the global scope and makes them vulnerable to dependency conflicts. This change does the same thing, just without being flagged by Zizmor. The correct fix to this would be to create a venv inside the workflow and install dependencies there. IMO that's overkill and unnecessary because there's only one `pip-install` per job and not much opportunity for conflicts

cobaltt7:
setup-python 7.0.0 has removed the input: https://github.com/actions/setup-python/pull/1336
Note again that this change is effectively the same as using pip-install, and installs into the runner-level environment. Again, I don't think this is an issue because we don't run multiple install commands or mix multiple tools. The proper fix to this would be to use a venv in each workflow, which to me feels like overkill, and IMO GHA should always have a quick-and-dirty way to install packages. It makes most sense to me to keep this until/unless it stops working


Would it be a bad idea to set `PIP_UPLOADED_PRIOR_TO=P2D` as a [repository env value](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#creating-configuration-variables-for-a-repository) instead? That avoids repetition in workflow files and the possibility of forgetting to include it.

JelleZijlstra:
A repository-level fix sounds good!

cobaltt7:
virtualenv 20.39.1 (via hatch) uses pip 26.0: https://github.com/pypa/virtualenv/blob/20.39.1/src/virtualenv/seed/wheels/embed/__init__.py#L39

virtualenv 21.3.0 bumps pip to 26.1: https://github.com/pypa/virtualenv/blob/21.3.0/src/virtualenv/seed/wheels/embed/__init__.py#L42

Working around this by manually calculating PIP_UPLOADED_PRIOR_TO until we can bump virtualenv #5023 (blocked by #5037, which is blocked by https://github.com/pypa/hatch/pull/2257)

cobaltt7:
It's better to set PIP_UPLOADED_PRIOR_TO inline after all, because otherwise it won't be included in workflows triggered by forks. I also added it to the Dockerfile.

PR #5149, opened 2026-05-23
Allow multiple cases per fixture file

### LINKED ISSUE #4506: Test organization improvements
Part of #2238
While working on my first PR to the repo, I've noticed that interacting with the test cases is not easy.
This is because of how much is in each file, with the longest being ~1200 lines long. Because it's a test that uses `# output`, the input and output for a specific case are ~600 lines apart. This is made worse by the fact that formatting changes line counts, so getting back to the source isn't as easy as dividing the line number by 2. Running the test case also doesn't help, as it only reports the end failure, and nothing about the original input.

With the current framework, the way I can see to solve this would be splitting cases into their own files, inside folders. For example, instead of one big `preview_long_strings__regression.py` file, there would be a `preview_long_strings__regression` folder with contents named `preview_long_strings__regression_1.py`, `preview_long_strings__regression_2.py`, etc.

This would be really nice for making all the tests more local, as if a set of tests fails, each one would have it's input directly next to it's output, and each test case would be individually re-runnable with pytest.

I'm not sure how this would affect the performance, since I don't know the cost of running a bunch of tiny files vs one large file.

I'm not sure if the current test runner supports running cases inside folders, but given the current structure I assume so.

If for whatever reason a test ever needed to be removed, this would make it a different type of difficult, as instead of having to find the input in a long file, now all test files after the one being removed would have to be renamed.

If this is a thing that is wanted, I'd be happy to start work on the PR.

Pedro-Muller29:
Fixes #4506.

### Description

In #4506, MeGaGiGaGon flagged that the longest fixture files in `tests/data/cases/` are painful to work with. `preview_long_strings.py` and `preview_long_strings__regression.py` are both 1000+ lines: input and output for any given case sit hundreds of lines apart, and when a test fails the assertion points at the end-of-file mismatch with nothing tying it back to the original case.

MeGaGiGaGon's first proposal was to split big files into folders of numbered files. JelleZijlstra pushed back on numbering ("renumbering files would be needlessly annoying") and suggested an alternative shape: sections within larger files, like [mypy's test data format](https://github.com/python/mypy/blob/master/test-data/unit/check-abstract.test).

This PR takes the mypy-style path. Fixture files can group multiple cases under `[case <name>]` headers, and the two longest files are migrated as the first consumers.

```
# flags: --unstable

[case assign_x_1]
x = "..."
# output
x = (
    "..."
)

[case assign_y]
y = "Short string"
# output
y = "Short string"
```

The format is opt-in. Files without a `[case ]` header at the top run through the legacy single-case loader unchanged. Cell names appear in pytest IDs (`test_simple_format[preview_long_strings::assign_x_1]`), and on failure the assertion message attaches the file path, the cell header line, and the `# output` marker line, so the traceback points at the case under test rather than the file boundary.

A `# flags: ` line above the first `[case ]` applies to every cell, and cell flags take precedence on overlap. That makes preview-to-stable migrations more granular: cells can graduate one at a time, while neighbors in the same file keep their preview flags.

Documentation in `docs/contributing/the_basics.md` covers the format and the merge semantics.

I picked `preview_long_strings*` for the first migrations because they're the worst offenders on file length.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,7 +62,7 @@
   `[case <name>]` headers, inspired by mypy's test data format. Files without a header
   continue to work unchanged. Failing cells report their file path, cell header line,
   and `# output` marker line. Migrates `preview_long_strings.py` and
-  `preview_long_strings__regression.py` as the first consumers (#5147)
+  `preview_long_strings__regression.py` as the first consumers (#5149)
 
 ## Version 26.5.1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,12 @@
 <!-- Major changes to documentation and policies. Small docs changes
      don't need a changelog entry. -->
 
+- Allow grouping multiple cases in one fixture file under `tests/data/cases/` using
+  `[case <name>]` headers, inspired by mypy's test data format. Files without a header
+  continue to work unchanged. Failing cells report their file path, cell header line,
+  and `# output` marker line. Migrates `preview_long_strings.py` and
+  `preview_long_strings__regression.py` as the first consumers (#5147)
+
 ## Version 26.5.1
 
 ### Stable style
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,12 +58,6 @@
 <!-- Major changes to documentation and policies. Small docs changes
      don't need a changelog entry. -->
 
-- Allow grouping multiple cases in one fixture file under `tests/data/cases/` using
-  `[case <name>]` headers, inspired by mypy's test data format. Files without a header
-  continue to work unchanged. Failing cells report their file path, cell header line,
-  and `# output` marker line. Migrates `preview_long_strings.py` and
-  `preview_long_strings__regression.py` as the first consumers (#5149)
-
 ## Version 26.5.1
 
 ### Stable style
--- a/docs/contributing/the_basics.md
+++ b/docs/contributing/the_basics.md
@@ -92,12 +92,57 @@ foo(bar(x, y))
 
 Rules:
 
-- Anything before the first `[case ]` is file-level prose and is ignored by the loader.
+- Anything before the first `[case ]` is file-level prose and is ignored by the loader,
+  with one exception: a single `# flags: ` line in that region is taken as the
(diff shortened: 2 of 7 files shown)


CHANGES.md
```diff
@@ -63,6 +58,12 @@
 <!-- Major changes to documentation and policies. Small docs changes
      don't need a changelog entry. -->
 
+- Allow grouping multiple cases in one fixture file under `tests/data/cases/` using
+  `[case <name>]` headers, inspired by mypy's test data format. Files without a header
+  continue to work unchanged. Failing cells report their file path, cell header line,
+  and `# output` marker line. Migrates `preview_long_strings.py` and
+  `preview_long_strings__regression.py` as the first consumers (#5149)
+
```
MeGaGiGaGon:
```suggestion

```

MeGaGiGaGon:
I like this idea. One other big benefit it has over my original idea is that it keeps the blame history semi-legible - while the input part would get clobbered on any full test file re-writes, the output part stays unchanged, leaving part of it in tact, which is nice.

docs/contributing/the_basics.md
```diff
@@ -67,6 +67,88 @@ files in the `tests/data/cases` directory. These files consist of up to three pa
...
+
+When most cells in a file share the same flags, lifting them above the first `[case ]`
+keeps the cells terse:
+
+```
+# flags: --preview --line-length=88
+
+[case keep_short_strings]
+x = "abc"
+# output
```
JelleZijlstra, line 114: Should we change this to `[output]`? The magical comment is a bit confusing.

MeGaGiGaGon: I think no, with the idea being that any of these smaller case cells has the same format as a pre-change test file, so they can share the same parsing/mental machinery

Pedro-Muller29: mypy's format (which this borrows from) uses a bracket marker for this (`[out]`). Should I move the multi-cell format the same way (`[case ]` + `[output]`) and leave single-case files on `# output`?

PR #5158, opened 2026-05-30
Add blank lines after docstring-only functions in stub files outside overload groups

AlexWaygood:
### Description

In `.pyi` stub files, Black currently does not consistently separate a function or method with a docstring-only body from the statement that follows it. It already inserts a blank line when another function definition follows directly, but not when the next statement is a comment, conditional block, variable annotation, or other non-function statement:

```py
# Before

class Example:
    def foo(self) -> None:
        """Documentation."""

    def method(self) -> None:
        """Documentation."""
    # comment for the next member
    attr: int
    def method2(self) -> None:
        """Documentation."""
    if sys.version_info < (3, 12):
        def method3(self) -> None:
            """Documentation"""
    attr: int
```

This PR adds a new preview feature (pyi_blank_line_after_function_docstring) that enforces a blank line after a function or method body that consists only of a docstring in .pyi files, making Black more consistent about inserting blank lines after function docstrings:

```py
# After (with `--preview`)

class Example:
    def foo(self) -> None:
        """Documentation."""

    def method(self) -> None:
        """Documentation."""

    # comment for the next member
    attr: int
    def method2(self) -> None:
        """Documentation."""

    if sys.version_info < (3, 12):
        def method3(self) -> None:
            """Documentation"""

    attr: int
```

The implementation teaches `EmptyLineTracker` to detect this case when the previous statement was a definition whose body ended in a docstring. It preserves the existing special handling for same-name decorated functions, so `@overload groups` and `@property` setters are still kept together without extra blank lines between the related definitions.

### Motivation

Most libraries don't include docstrings in their stub files, but it's increasingly popular to monkeypatch them in as a build step using tools like docstring-adder and docify. This is what ty does for its vendored typeshed stubs, for example; this patch greatly improves readability for our vendored, docstringified stubs, in my opinion. An example branch with the formatting from this PR applied is here: https://github.com/astral-sh/ruff/compare/main...AlexWaygood:alex/new-stub-formatting?expand=1

### Checklist - did you ...

- [x] Implement any code style changes under the --preview style, following the stability policy?
- [x] Add an entry in CHANGES.md if necessary?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,8 @@
   instead. This fixes the awkward break that was showing up in comprehension `if`
   clauses (#4514) as well as the same shape inside `if`, `elif`, `assert`, and
   parenthesized expressions (#5135)
+- In `.pyi` stub files, enforce a blank line after a function or method that has a
+  docstring-only body when another comment or statement follows it.
 
 ### Configuration
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,7 +32,7 @@
   clauses (#4514) as well as the same shape inside `if`, `elif`, `assert`, and
   parenthesized expressions (#5135)
 - In `.pyi` stub files, enforce a blank line after a function or method that has a
-  docstring-only body when another comment or statement follows it.
+  docstring-only body when another comment or statement follows it (#5158)
 
 ### Configuration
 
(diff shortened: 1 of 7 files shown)


CHANGES.md
```diff
@@ -31,6 +31,8 @@
   instead. This fixes the awkward break that was showing up in comprehension `if`
   clauses (#4514) as well as the same shape inside `if`, `elif`, `assert`, and
   parenthesized expressions (#5135)
+- In `.pyi` stub files, enforce a blank line after a function or method that has a
+  docstring-only body when another comment or statement follows it.
```
AlexWaygood:
```suggestion
  docstring-only body when another comment or statement follows it (#5158)
```

AlexWaygood:
Looks like the PyPy/diff-shades errors have been occurring on other PRs too; I don't think they're related to this PR

cobaltt7 (approved):

AlexWaygood:
Thanks @cobaltt7! 😃

PR #5161, opened 2026-06-02
Fix crash on `# fmt: skip` after an opening bracket with a standalone comment

sarathfrancis90:
### Description

`black` crashes with `AttributeError: 'Leaf' object has no attribute 'bracket_depth'` on valid Python when a `# fmt: skip` line is placed directly after an opening bracket and a standalone comment also appears among the bracket's contents. For example, all of the following crash:

```python
from m import (
# fmt: skip
    # comment
    a
)

f(
# fmt: skip
    # comment
    a
)

x[
# fmt: skip
    # comment
    a
]

from m import (  # fmt: skip
    # comment
    a
)
```

```
error: cannot format ...: 'Leaf' object has no attribute 'bracket_depth'
```

**Root cause.** The crash happens in `is_line_short_enough` (`src/black/lines.py`). When the rendered line spans multiple physical lines, the function walks every leaf and reads `leaf.bracket_depth`. That attribute is set by `BracketTracker.mark`, but `mark` deliberately early-returns for a closing bracket that is *continued* from a previous line break (a leftover `)` / `]` / `}` left on its own line) — see `BracketTracker.mark` in `src/black/brackets.py`. Such a leaf therefore never receives a `bracket_depth`, and accessing it raises `AttributeError`.

The combination of `# fmt: skip` right after an opening bracket plus a standalone comment is exactly what produces a line containing such an unmarked, continued closing bracket, so `is_line_short_enough` blows up.

**Fix.** A leaf that the bracket tracker skipped is, by construction, not inside any bracket that opened on the current line, so its effective depth is `0`. Read the attribute defensively with that default (`getattr(leaf, "bracket_depth", 0)`). This is a no-op for every already-marked leaf (behaviour is unchanged) and removes the crash, producing valid, AST-equivalent and idempotent output. Since this only fixes a crash (it does not change formatting of any code that previously formatted successfully), it is a stable-style fix per the contributing docs.

**Testing.** Added `tests/data/cases/fmtskip_after_bracket_with_comment.py` covering the import, call, subscript and inline `# fmt: skip` variants. The test fails on `main` (crash inside `is_line_short_enough`) and passes with this change. The produced output is AST-preserving and idempotent (formatting it a second time is a no-op).

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? (N/A — this is a crash fix only, no formatting changes to previously-working code, so it belongs in stable style)
- [x] Add new / update outdated documentation? (N/A — no user-facing behaviour or documented option changes)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,9 @@
   lengths (#5147)
 - Fix multiline docstring indentation when leading tabs are used inside indented
   docstrings (#5148)
+- Respect `# fmt: skip` on a line that opens a bracket (e.g. `from x import (  # fmt: skip`)
+  when a standalone comment is among the bracket's contents: the whole statement is now
+  preserved instead of being reformatted (and previously crashing) (#5161)
 
 ### Preview style
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,9 +21,10 @@
   lengths (#5147)
 - Fix multiline docstring indentation when leading tabs are used inside indented
   docstrings (#5148)
-- Respect `# fmt: skip` on a line that opens a bracket (e.g. `from x import (  # fmt: skip`)
-  when a standalone comment is among the bracket's contents: the whole statement is now
-  preserved instead of being reformatted (and previously crashing) (#5161)
+- Respect `# fmt: skip` on a line that opens a bracket (e.g.
+  `from x import (  # fmt: skip`) when a standalone comment is among the bracket's
+  contents: the whole statement is now preserved instead of being reformatted (and
+  previously crashing) (#5161)
 
 ### Preview style
 
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -786,6 +786,27 @@ def _generate_ignored_nodes_from_fmt_skip(
                 if header_nodes:
                     ignored_nodes = header_nodes + ignored_nodes
 
+        # If the nodes captured for the comment's physical line leave a bracket
+        # open, the `# fmt: skip` sits inside a multi-line bracketed statement
+        # (e.g. on the `from x import (` line). Skipping only that line would
+        # reformat the rest of the statement, so preserve the whole enclosing
+        # statement instead.
+        bracket_depth = 0
+        for node in ignored_nodes:
+            for ignored_leaf in node.leaves():
+                if ignored_leaf.type in OPENING_BRACKETS:
+                    bracket_depth += 1
+                elif ignored_leaf.type in CLOSING_BRACKETS:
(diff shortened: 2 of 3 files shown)


cobaltt7:
Thank you! One note: when formatting something like your first example (with the multiline import and the comment on its one line), Black with this fix formats it as:
```py
from m import (# fmt: skip  # comment
 a
)
```
While this is an improvement from a crash, it's still not the expected behavior - the line should be skipped. In this case, all of your examples should probably remain unchanged. Is that something you're able to fix here? If it's a lot of extra work, that's also fine, again this is definitely an improvement.

sarathfrancis90:
Thanks for the review, and good catch — you're right that this stops at "doesn't crash" rather than restoring the skip.

The reason it doesn't skip: in all of these the `# fmt: skip` sits *after* the opening bracket, so it's mid-statement, whereas `# fmt: skip` is really meant to trail a complete one-line statement. The logic that decides which nodes a skip covers keys off the nodes on the comment's own line; once the comment is inside the brackets (with a standalone comment between the bracket and the contents) it doesn't resolve to the enclosing statement, so the body gets formatted normally. Making the examples round-trip unchanged means teaching that traversal to recognize this case and ignore the whole enclosing statement — a behavior change rather than just hardening the crash, and it needs a decision on what a standalone `# fmt: skip` after `(` should even mean.

My suggestion would be to keep this PR scoped to the crash fix — it now produces valid, idempotent output instead of erroring — and handle the "actually skip it" behavior as a follow-up with its own tests. I'm happy to take that on, or to fold it in here if you'd rather have it all in one PR. Whichever you prefer.

cobaltt7:
As I said, please try to fix the issue here unless it'd be a significant amount of work that'd be better to put off and take on again later. I would expect all the test cases you provided to stay unchanged for the most part, especially the line with `# fmt: skip` on it.

sarathfrancis90:
Done — reworked it to actually respect the skip, so all the examples now stay unchanged.

The skip was only capturing the leaves on the comment's physical line (e.g. `from m import (`), which leaves the bracket open, so the rest of the statement got reformatted — and that's also what tripped the crash. When the captured line ends inside an open bracket, the `# fmt: skip` is really sitting inside a multi-line statement, so I now capture the whole enclosing statement and preserve it verbatim. That fixes the crash as a side effect too, so I dropped the earlier `is_line_short_enough` guard in favour of this root-cause fix. Updated the test case to expect the input unchanged; the full suite is green.

cobaltt7:
Thanks!

PR #5163, opened 2026-06-02
fix quadratic backtracking in unicode escape normalization

sahvx655-wq:
normalize_unicode_escape_sequences runs UNICODE_ESCAPE_RE over every string leaf, and the pattern matched a backslash run with (\\+) immediately followed by a required escape body. A long run of backslashes with no trailing u/U/x/N escape makes the engine retry the run from each restart position, so formatting a string of backslashes is quadratic: a 64KB literal already takes minutes, and blackd accepts far larger bodies over the wire. Making the escape body optional lets the whole run be consumed in a single match, which restores linear time and leaves the output identical.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -53,6 +53,8 @@
 
 <!-- Changes that improve Black's performance. -->
 
+- Improve performance on strings containing many consecutive backslashes (#5163)
+
 ### Output
 
 <!-- Changes to Black's terminal output and error messages -->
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Thanks, could you add a changelog entry? Something like "Improve performance on strings containing many consecutive backslashes"

sahvx655-wq:
Added it under Performance, thanks.

JelleZijlstra (approved):

PR #5167, opened 2026-06-03
fix: no spurious target-version warning when runtime version is included

### LINKED ISSUE #5164: Unexpected version warning
I'm getting 
```
Warning: Python 3.14 cannot parse code formatted for Python 3.15. To fix this: run Black with Python 3.15, set --target-version to py314, or use --fast to skip the safety check. Black's safety check verifies equivalence by parsing the AST, which fails when the running Python is older than the target version.
```
when running on 3.14, with `-t py310 -t py311 -t py312 -t py313 -t py314 -t py315`.

(the warning was introduced in #4983)

I would expect that warning if I only had `-t py315`, but why does it appear when I've asked it to make code formatted for 3.10->3.15 (which includes 3.14)? Should black always be run on the highest version it is formatting for?

BTW, there are no changes in output with and without `-t py315` for my use case.

wahajahmed010:
### Description

The `_target_versions_exceed_runtime` function was using `max()` to compare
target versions against the runtime version. This caused a spurious warning
when users specified the runtime version alongside higher target versions
(e.g., `--target-version py314 --target-version py315` on Python 3.14).

The fix uses `min()` instead: the warning now only fires when **all** target
versions exceed the runtime. If any target version is at or below the runtime,
the AST safety check can succeed for that version, making the warning
unnecessary.

Closes #5164.
Closes #5168.

### Approach

- Changed `max()` to `min()` in `_target_versions_exceed_runtime`
- Added regression test that verifies no warning when runtime version is
  included in the target set
- Updated the docstring to clarify the semantics

### AI assistance disclosure

This contribution was developed with the assistance of an AI coding agent
(Buck/OpenClaw). All code was reviewed and verified by a human contributor
before submission.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
      Please include the PR number in the changelog entry, not the issue number -->
 
 - Add support for NO_COLOR environment variable to disable ANSI output (#5129)
+- No spurious target-version warning when runtime version is included in
+  multiple --target-version flags (#5164)
 
 ### Highlights
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,8 +6,8 @@
      Please include the PR number in the changelog entry, not the issue number -->
 
 - Add support for NO_COLOR environment variable to disable ANSI output (#5129)
-- No spurious target-version warning when runtime version is included in multiple
-  --target-version flags (#5164)
+- No spurious target version warning when runtime version is included in a
+  --target-version flag (#5167)
 
 ### Highlights
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,8 +6,8 @@
      Please include the PR number in the changelog entry, not the issue number -->
 
 - Add support for NO_COLOR environment variable to disable ANSI output (#5129)
-- No spurious target-version warning when runtime version is included in
-  multiple --target-version flags (#5164)
+- No spurious target-version warning when runtime version is included in multiple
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -6,6 +6,8 @@
      Please include the PR number in the changelog entry, not the issue number -->
 
 - Add support for NO_COLOR environment variable to disable ANSI output (#5129)
+- No spurious target-version warning when runtime version is included in multiple
+  --target-version flags (#5164)
```
```suggestion
- No spurious target version warning when runtime version is included in a
  --target-version flag (#5167)
```
Grammar + corrected number

tests/test_black.py
```diff
@@ -3283,6 +3283,22 @@ def test_target_version_at_runtime_no_warning(self) -> None:
...
+        """Regression test for #5164: no spurious warning when target includes
+        the runtime version alongside higher versions."""
+        current_minor = sys.version_info[1]
+        higher_target = f"py3{current_minor + 1}"
+        runtime_target = f"py3{current_minor}"
+        code = "x = 1\n"
+        args = [
+            "--target-version", runtime_target,
+            "--target-version", higher_target,
+            "--code", code,
```
cobaltt7:
```suggestion
            "--target-version",
            runtime_target,
            "--target-version",
            higher_target,
            "--code",
            code,
```
format

PR #5170, opened 2026-06-06
Parenthesize singleton tuples in yield expressions

### LINKED ISSUE #3851: single element tuples in yield expressions are not parenthesized

wayyoungboy:
Fixes #3851.
## Summary
- normalize yield expressions so bare singleton tuple yields like `yield x,` become `yield (x,)`
- preserve required parentheses around yield walrus expressions
- add a formatting fixture for singleton tuple yield expressions

## Tests
- `tests/test_format.py::test_simple_format[yield_singleton_tuple]` (RED before fix)
- `tests/test_format.py`
- `tests/test_black.py::BlackTestCase::test_get_features_used`
- `git diff --cached --check`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,8 @@
 
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
   `x[key] = expr` (#5095)
+- Parenthesize tuple expressions in `yield` statements for consistency with function
+  calls and returns (#5170)
 - Stop splitting between a variable and its comparator (`not in`, `==`, `is`, ...) when
   the right-hand side is a bracketed expression. Black now lets the bracket explode
   instead. This fixes the awkward break that was showing up in comprehension `if`
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -41,6 +41,8 @@ Currently, the following features are included in the preview style:
 - `hug_comparator`: Don't break a comparator (`not in`, `==`, `is`, ...) away from its
   left operand when the right operand is a bracketed expression that has to break
   anyway; let the bracket explode instead. ([see below](labels/hug-comparator))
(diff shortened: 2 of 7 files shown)


cobaltt7:
Thanks! This change affects more than just singleton `yield` tuples, it makes `yield` statements with any tuple behave more like functions in general, but I think this is a good change overall for consistency.

This changes Black's formatting, so it needs to be gated to the Preview style. See https://github.com/psf/black/commit/01c29bd5fded1d6f17ebd4e6f2b7e3999c153bf8 for a good recent example of how to do this. Please also add a changelog entry.

wayyoungboy:
> 谢谢！这个变化不仅影响单元组，还让任何元组的语句更像函数，但我认为总体来说这是个好变化，有助于一致性。`yield``yield`
> 
> 这改变了Black的格式，因此需要将它限制为预览样式。请参见[01c29bd](https://github.com/psf/black/commit/01c29bd5fded1d6f17ebd4e6f2b7e3999c153bf8)，那里有一个很好的近期示例。请同时添加更新日志条目。

Makes sense. I’ll move this behind preview style and add the changelog entry.

wayyoungboy:
Updated.

cobaltt7:
Thanks!

PR #5171, opened 2026-06-08
avoid re-walking nested brackets in max_delimiter_priority_in_atom

sahvx655-wq:
**Quadratic paren normalisation on deeply nested atoms**

Profiling the formatting of an over-parenthesised assignment showed `maybe_make_parens_invisible_in_atom` recursing once per paren layer, and at every layer it calls `max_delimiter_priority_in_atom`, which walks every descendant leaf of the atom through the recursive `Node.leaves()` generator. For input shaped like `x = (((...)))` that means a full re-scan at each depth, so the work grows cubically once the per-leaf generator delegation is counted in. A 1.6 kB file with 800 nested parens took about 7s here, and the profile was almost entirely `Node.leaves()` and `BracketTracker.mark`.

Only depth-0 leaves can be a top-level delimiter, yet the old scan descended into every nested bracket pair whose contents sit at depth >= 1 and can never contribute. The fix keeps just the brackets of a nested pair to balance the depth tracker and skips the interior, so each call is proportional to the atom's own top level rather than its whole subtree; that same file now formats in about 170ms. Output is unchanged across the test suite, a multi-file corpus and fuzzed bracket-heavy inputs. Worth fixing because the scan runs during ordinary formatting and is reachable pre-auth through blackd, where a tiny request body can pin a worker for seconds.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -64,8 +64,8 @@
 - Improve performance on files with many `# fmt: skip`/`# fmt: off` comments by no
   longer re-walking the whole tree from the root for every directive (#5169)
 - Improve performance on deeply nested parenthesised expressions by no longer
-  re-scanning the whole atom for every nesting level in
-  `max_delimiter_priority_in_atom` (#5171)
+  re-scanning the whole atom for every nesting level in `max_delimiter_priority_in_atom`
+  (#5171)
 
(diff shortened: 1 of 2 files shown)


cobaltt7:
Thanks!

PR #5175, opened 2026-06-12
Preserve prefixes outside --line-ranges

### LINKED ISSUE #4430: `--line-ranges` formats lines outside of range
**Describe the bug**

Black's `--line-ranges` option does not completely prevent reformatting of code outside of the specified line ranges.

**To Reproduce**

For example, take this code:

```python
print("format me"  )
 # format whitespace
  # format whitespace
   # format whitespace
print( "format me" )
   # format whitespace
  # format whitespace
 # format whitespace
print(  "format me")



print("don't format me"  )
 # don't format whitespace
  # don't format whitespace
   # don't format whitespace
print( "don't format me" )
   # don't format whitespace
  # don't format whitespace
 # don't format whitespace
print(  "don't format me")
```

```sh
$ black --line-ranges 1-9 file.py
```

The result is:

```python
print("format me")
# whitespace
# whitespace
# whitespace
print("format me")
# whitespace
# whitespace
# whitespace
print("format me")


print("don't format me"  )
# don't format whitespace
# don't format whitespace
# don't format whitespace
print( "don't format me" )
# don't format whitespace
# don't format whitespace
# don't format whitespace
print(  "don't format me")
```

**Expected behavior**

The expected result would be the formatting constrains itself to lines 1 through 9, and does not modify whitespace outside of this line range:

```python
print("format me")
# format whitespace
# format whitespace
# format whitespace
print("format me")
# format whitespace
# format whitespace
# format whitespace
print("format me")



print("don't format me"  )
 # don't format whitespace
  # don't format whitespace
   # don't format whitespace
print( "don't format me" )
   # don't format whitespace
  # don't format whitespace
 # don't format whitespace
print(  "don't format me")
```

Note both the three-line separation as well as the indentation would be preserved as they are outside of the line range 1-9.

**Environment**

- Black's version: 
```
black, 24.8.0 (compiled: yes)
Python (CPython) 3.11.9
```
- OS and Python version:  OS is macOS 14.5 (23F79).  Python is 3.11.9

**Additional context**

It would be useful for `--line-ranges` to only format what is contained within the specified line ranges, as much as possible, to allow for incremental adoption of formatting in larger code bases.

cyphercodes:
### Description

Preserves comments and blank lines that are outside the requested `--line-ranges` instead of normalizing them while unchanged code is converted to standalone-comment leaves for range formatting.

This keeps the issue repro constrained to the requested lines: comments and separation before unchanged statements are preserved when the selected range ends immediately before them.

Fixes #4430.

### Checklist - did you ...

### Verification

- `python -m black --check src/black/comments.py src/black/linegen.py src/black/lines.py src/black/ranges.py tests/test_ranges.py`
- `python -m pytest tests/test_ranges.py tests/test_format.py -q`
- `git diff --check`


cobaltt7:
Thanks - this fixes issues where whitespace at the beginning of lines before comments is stripped when using `--line-ranges`. However, this doesn't address trailing whitespace on prefix lines, which I think was the main issue #4430 was implying.

As an example:
```py
   #  format whitespace   
print( "format me" )   
      

   #  don't format whitespace   
print("don't format me"  )     
      
```
Output with `--line-ranges 1-3`:
```py
#  format whitespace
print("format me")


   #  don't format whitespace
print("don't format me"  )     
[empty line at EOF]
```

Expected output:
```py
#  format whitespace
print( "format me" )


   #  don't format whitespace   
print("don't format me"  )     
      
```

cobaltt7:
I also found an edge case where this fix doesn't seem to be applied. Example with `--line-ranges 1-2`:
```py
print( "format me" )
   #  format whitespace

print("don't format me"  )
```
The spaces before the comment are retained, but they should be removed here. They are removed when running Black on the whole file.

cobaltt7:
@cyphercodes are you still updating this?

cyphercodes:
Updated — merged current `main` and resolved the `src/black/ranges.py` conflict by preserving the upstream batched replacement logic plus the PR's line-range conversion marker.

Verification:
- `python -m pytest tests/test_black.py -k 'line_ranges or preserve_outside_prefix' -q`
- `git diff --check`

cobaltt7:
@cyphercodes Please resolve my two review comments above.

cyphercodes:
Updated — addressed both line-ranges prefix cases from the review:

- unselected prefix comment lines now keep their original leading/trailing whitespace
- prefix comments that fall inside the selected `--line-ranges` are still normalized
- trailing whitespace-only blank lines outside the selected ranges are preserved

Verification:
- `.venv/bin/python -m black src/black/__init__.py src/black/comments.py src/black/linegen.py src/black/ranges.py src/blib2to3/pytree.py tests/test_black.py`
- `.venv/bin/python -m pytest tests/test_black.py -k 'line_ranges or preserve_outside_prefix' -q`
- `.venv/bin/python -m pytest tests/test_format.py -k line_ranges -q`
- `git diff --check`

tests/test_black.py
```diff
@@ -2126,6 +2126,47 @@ def test_line_ranges_with_source(self) -> None:
             """)
             assert expected == formatted
 
+    def test_line_ranges_preserves_unselected_prefix_trailing_whitespace(self) -> None:
```
cobaltt7, line 2172: Is there a reason these aren't their own test case files? They should be if possible.

src/blib2to3/pytree.py
```diff
@@ -484,6 +484,10 @@ class Leaf(Base):
     # code, and `fmt_pass_converted_first_leaf` points to the first Leaf in the
     # converted code.
     fmt_pass_converted_first_leaf: Optional["Leaf"] = None
+    # True when this Leaf is converted from unchanged code for --line-ranges.
+    line_ranges_converted: bool = False
+    line_ranges_first_lineno: int = 0
+    line_ranges_selected: Optional[set[int]] = None
```
cobaltt7: Wouldn't `line_ranges_first_lineno != 0` or `line_ranges_selected is not None` always be equal to `line_ranges_converted`? Could we replace `line_ranges_converted` with one of those?

src/black/__init__.py
```diff
@@ -1267,7 +1267,36 @@ def f(
...
+) -> str:
+    """Restore trailing whitespace-only lines outside requested line ranges."""
+    src_lines = src_contents.splitlines(keepends=True)
+    dst_lines = dst_contents.splitlines(keepends=True)
+    if len(dst_lines) >= len(src_lines):
+        return dst_contents
+
+    selected: set[int] = set()
+    for start, end in lines:
+        selected.update(range(start, end + 1))
```
cobaltt7: Could we use comparison operators (`<=`/`>=`) instead of creating a set and checking inclusion?

src/blib2to3/pytree.py
cobaltt7: Also, when I first read through, I assumed that `min(line_ranges_selected)` would be the same as `line_ranges_first_lineno`, but I don't think that's always the case after reading through the code. It may be worth adding comments describing the use of both and the differences.

cobaltt7:
@cyphercodes are you still updating this?

cyphercodes:
Updated — merged current `main`, resolved the merge conflicts, and addressed the review feedback:

- moved the selected-prefix comment regression into a data case
- removed `line_ranges_converted` in favor of existing line-range metadata
- documented why `line_ranges_first_lineno` can differ from the selected line range
- replaced the temporary selected-line `set` with direct range comparisons

The trailing-whitespace preservation regression stays inline because it requires literal trailing spaces, which would fail `git diff --check` in a data case file.

Verification:
- `.venv/bin/python -m black src/black/__init__.py src/black/comments.py src/black/linegen.py src/black/lines.py src/black/ranges.py src/blib2to3/pytree.py tests/test_black.py`
- `.venv/bin/python -m pytest tests/test_black.py -k 'line_ranges or preserve_outside_prefix' -q`
- `.venv/bin/python -m pytest tests/test_format.py -k line_ranges -q`
- `git diff --check`

cobaltt7:
Thanks!

PR #5185, opened 2026-06-17
Fix normalize_string_prefix not lowercasing uppercase T t-string prefix

JSap0914:
### Description

`normalize_string_prefix` is documented to "make all string prefixes lowercase", and it already normalises `F`→`f` and `B`→`b`. However, when t-string support was added in #4805 the `T`→`t` normalisation was accidentally omitted.

**Bug:** Black leaves `T"..."` unchanged instead of normalising it to `t"..."`:

```python
# Before (Black ≥ 26.5 with --target-version py314)
T"hello"   →   T"hello"   # wrong — should be t"hello"
RT"hello"  →   RT"hello"  # wrong — should be Rt"hello"
rT"hello"  →   rT"hello"  # wrong — should be rt"hello"
```

**Fix:** Add a single `.replace("T", "t")` step to the existing chain in `normalize_string_prefix`.

**Verification:**

```
$ pytest tests/test_format.py -k pep_750 -v
PASSED  tests/test_format.py::test_simple_format[pep_750]
```

> AI-assisted: this PR was prepared with the help of an AI coding tool.

### Checklist - did you ...

### DIFF
--- a/src/black/strings.py
+++ b/src/black/strings.py
@@ -6,10 +6,15 @@ import re
 import sys
 from functools import lru_cache
 from re import Match, Pattern
-from typing import Final
+from typing import TYPE_CHECKING, Final
+
 
 from black._width_table import WIDTH_TABLE
 from blib2to3.pytree import Leaf
+
+if TYPE_CHECKING:
+    from black.mode import Mode
+
 
 STRING_PREFIX_CHARS: Final = "fturbFTURB"  # All possible string prefix characters.
 STRING_PREFIX_RE: Final = re.compile(
@@ -142,7 +147,7 @@ def assert_is_leaf_string(string: str) -> None:
     ), f"{set(string[:quote_idx])} is NOT a subset of {set(STRING_PREFIX_CHARS)}."
 
 
-def normalize_string_prefix(s: str) -> str:
+def normalize_string_prefix(s: str, mode: "Mode | None" = None) -> str:
     """Make all string prefixes lowercase."""
     match = STRING_PREFIX_RE.match(s)
     assert match is not None, f"failed to match string {s!r}"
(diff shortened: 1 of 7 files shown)


src/black/strings.py
```diff
@@ -150,6 +150,7 @@ def normalize_string_prefix(s: str) -> str:
     new_prefix = (
         orig_prefix.replace("F", "f")
         .replace("B", "b")
+        .replace("T", "t")
```
JelleZijlstra: This needs to go into the preview style.

JSap0914:
T→t prefix normalization is now gated behind `Preview.normalize_tstring_prefix`, so it only applies in `--preview` mode as requested.

src/black/strings.py
```diff
@@ -153,13 +157,21 @@ def normalize_string_prefix(s: str) -> str:
         .replace("U", "")
         .replace("u", "")
     )
+    if mode is None or _is_preview_tstring_normalization(mode):
```
cobaltt7, line 160: Can you inline the `Preview.normalize_tstring_prefix in mode` call instead of making a utility function?

src/black/strings.py
```diff
@@ -142,7 +146,7 @@ def assert_is_leaf_string(string: str) -> None:
     ), f"{set(string[:quote_idx])} is NOT a subset of {set(STRING_PREFIX_CHARS)}."
 
 
-def normalize_string_prefix(s: str) -> str:
+def normalize_string_prefix(s: str, mode: "Mode | None" = None) -> str:
```
cobaltt7, line 149: I'd prefer for mode to be required, just so we don't accidentally forget it anywhere. Black's public API isn't considered stable, so it's not a breaking change.

src/black/strings.py
```diff
@@ -153,13 +157,21 @@ def normalize_string_prefix(s: str) -> str:
...
+        new_prefix = new_prefix.replace("T", "t")
 
     # Python syntax guarantees max 2 prefixes and that one of them is "r"
     if len(new_prefix) == 2 and new_prefix[0].lower() != "r":
         new_prefix = new_prefix[::-1]
     return f"{new_prefix}{match.group(2)}"
 
 
+def _is_preview_tstring_normalization(mode: "Mode") -> bool:
+    from black.mode import Preview
```
JelleZijlstra, line 170: Do we need this inline import? Why can't we just import `black.mode` normally?

cobaltt7:
@JSap0914 are you still updating this?

cobaltt7:
Closing as stale. Feel free to open a new PR following the guidelines I set above. I'll also create a new issue for this so it may be picked up by someone else instead.

PR #5189, opened 2026-06-20
Fix fmt off block before with statement

### LINKED ISSUE #5187: INTERNAL ERROR crash; syntax error in log
**Describe the bug**

Given particular inputs, Black crashes with an INTERNAL ERROR that I have not made sense of.

**To Reproduce**

Put this code in `bug.py`:
```python
with open("bug.txt", "r", encoding="utf-8") as fh1:
    pass
# fmt: off
# fmt: on
# xxx
# fmt: off
# fmt: on
with open("bug.txt", "r", encoding="utf-8") as fh2:
    pass
def fn(str: str) -> float:
    match str:
        case _:
            pass
```
And run `black bug.py`.

The resulting error is:

```
error: cannot format bug.py: INTERNAL ERROR: Black 26.5.2.dev29+gb74b23013 on Python (CPython) 3.14.5 produced invalid code: unindent does not match any outer indentation level (<unknown>, line 10). Please report a bug on https://github.com/psf/black/issues.  This invalid output might be helpful: /run/user/1000/tmpdir/blk_tnip9iro.log

Oh no! 💥 💔 💥
1 file failed to reformat.
```
The `.log` file reads
```
  File "/home/lukeshu/Projects/sbc-harness/sbc-harness-hardware/build-aux/venv/lib/python3.14/site-packages/black/__init__.py", line 1652, in assert_equivalent
    dst_ast = parse_ast(dst)
  File "/home/lukeshu/Projects/sbc-harness/sbc-harness-hardware/build-aux/venv/lib/python3.14/site-packages/black/parsing.py", line 162, in parse_ast
    raise SyntaxError(first_error)
with open("bug.txt", "r", encoding="utf-8") as fh1:
    pass
# fmt: off
# fmt: on


# xxx
# fmt: off
# fmt: on
 with open("bug.txt", "r", encoding="utf-8") as fh2:
    pass


def fn(str: str) -> float:
    match str:
        case _:
            pass
```

**Expected behavior**

I expected Black to not crash.

**Environment**

Observed with both
```
$ black --version
black, 26.5.1 (compiled: no)
Python (CPython) 3.14.5
```
and
```
(venv) $ black --version
black, 26.5.2.dev29+gb74b23013 (compiled: no)
Python (CPython) 3.14.5
```
On Parabola GNU/Linux-libre (like Arch Linux).

**Additional context**

Running `python bug.py` returns without error---CPython does not encounter a syntax error.  While minimizing the reproducer, I often got a case instead of an INTERNAL ERROR, Black told me I had invalid input with a syntax error, but I believed the input to be valid syntax.

sjh9714:
### Description

Fixes #5187.

Comment-only `# fmt: off` / `# fmt: on` blocks before a `with` statement could be emitted on the same logical line as the following `with` statement after a preceding standalone comment. When the line was split, the separator before `with` became a line prefix, so Black produced invalid code with a leading space at top level.

This normalizes the prefix copied onto comment-only fmt blocks and flushes those blocks when they follow Black's invisible context-manager parenthesis, while leaving fmt blocks inside real bracketed expressions on the existing path.

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? (N/A: this fixes invalid output, not a style change.)
- [x] Add new / update outdated documentation? (N/A.)

### Testing

- `/tmp/black-pr-5187-venv/bin/python -m pytest tests/test_format.py -k 'fmtonoff_comment_only_with or fmtonoff6 or fmtonoff3 or fmtonoff2 or fmtskip11'`
- `/tmp/black-pr-5187-venv/bin/python -m pytest tests/test_format.py`
- `/tmp/black-pr-5187-venv/bin/python -m black --check src/black/linegen.py src/black/comments.py`
- `env -u NO_COLOR TERM=xterm-256color /tmp/black-pr-5187-venv/bin/python -m pytest tests --run-optional no_jupyter`

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@
 
 - Fix crash when a standalone comment sits between tokens of a comprehension or lambda
   (#5144)
+- Fix crash when a comment-only `# fmt: off`/`# fmt: on` block is followed by a
+  `with` statement after another standalone comment (#5189)
 - Fix a crash when splitting `case case if ...` match patterns at very small line
   lengths (#5147)
 - Fix multiline docstring indentation when leading tabs are used inside indented
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,8 +19,8 @@
 
 - Fix crash when a standalone comment sits between tokens of a comprehension or lambda
   (#5144)
-- Fix crash when a comment-only `# fmt: off`/`# fmt: on` block is followed by a
-  `with` statement after another standalone comment (#5189)
+- Fix crash when a comment-only `# fmt: off`/`# fmt: on` block is followed by a `with`
+  statement after another standalone comment (#5189)
 - Fix a crash when splitting `case case if ...` match patterns at very small line
   lengths (#5147)
 - Fix multiline docstring indentation when leading tabs are used inside indented
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -300,9 +300,11 @@ def _handle_comment_only_fmt_block(
         # This preserves all comments and content before the fmt:off directive
         pre_fmt_off_consumed = all_comments[fmt_off_idx - 1].consumed
 
-    standalone_comment_prefix = (
-        original_prefix[:pre_fmt_off_consumed] + "\n" * comment.newlines
-    )
+    preceding_prefix = original_prefix[:pre_fmt_off_consumed]
+    if fmt_off_idx > 0:
+        preceding_prefix = preceding_prefix.lstrip("\r\n")
+
+    standalone_comment_prefix = preceding_prefix + "\n" * comment.newlines
 
     fmt_off_prefix = original_prefix.split(comment.value)[0]
     if "\n" in fmt_off_prefix:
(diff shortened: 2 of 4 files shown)


maflAT:
This issue can be reduced to a much simpler case.

```python
if False is True or False:  #fmt: skip
    pass#this comment should begin with a space
``` 

> 2026-06-22 08:30:07.037 [info] error: cannot format c:\_WIP_\black.py: Cannot parse: 2:26:     False is True or False:  #fmt: skip
> 
> Oh no! 💥 💔 💥
> 1 file failed to reformat.

Black seems to crash on any slightly complex condition with a `fmt: skip` comment.
The following 2 cases work perfectly fine:
```python
if False is True :  #fmt: skip
    pass  # this comment should begin with a space

if False is True or False:
    pass  # this comment should begin with a space
```

cobaltt7:
@maflAT I could not reproduce a crash with either of those snippets on main nor this PR. Please open a new issue with a reproduction if you're experiencing further issues.

@sjh9714 Thanks for the PR!

PR #5192, opened 2026-06-20
Handle empty cache files

CodingFeng101:
### Summary

Treat empty or truncated cache files the same way as other malformed cache files.

### Details

`Cache.read()` already ignores malformed pickle data so a later cache write can repair the file, but an empty cache file raises `EOFError`, which was not caught. This can happen if cache creation is interrupted or an external process truncates the file.

This change catches `EOFError` while reading the cache and adds a regression test for an empty cache file.

### Test

- `python -m pytest tests/test_black.py::TestCaching::test_cache_empty_file tests/test_black.py::TestCaching::test_cache_broken_file -q`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,9 +56,6 @@
   `srcs` are given) is now resolved before the `lru_cache` key is computed, so each
   directory gets the correct `pyproject.toml` (#5152)
 - Add validation for --line-ranges values (#5107)
-
-### Caching
-
 - Ignore empty cache files like other malformed cache files instead of raising an
   `EOFError` (#5192)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,11 @@
   directory gets the correct `pyproject.toml` (#5152)
 - Add validation for --line-ranges values (#5107)
 
+### Caching
+
+- Ignore empty cache files like other malformed cache files instead of raising an
+  `EOFError` (#5192)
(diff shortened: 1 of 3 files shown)


cobaltt7:

CodingFeng101:
Added the requested changelog entry in fcf00f6.

cobaltt7:
Could you use one of the pre-existing categories instead of adding a new one? (Caching changes have gone in Configuration in the past)

CodingFeng101:
Moved the entry under the existing Configuration category in 66e62ac.

DTiming24 (commented):
Catching EOFError seems like the right symmetry with malformed pickle handling. One subtle edge case: if an external process truncates the file while Black is running, silently treating it as a cache miss is probably preferable to surfacing cache corruption to the user. The regression test covers the repair path well.

DTiming24:
Nice fix. Since empty cache files now fall back the same way as other malformed cache data, a regression test for a zero-byte file created by an interrupted write would help lock in the new behavior.

CodingFeng101:
Thanks for taking a look. The current PR already has a zero-byte cache regression test: `test_cache_empty_file` creates the cache path with `cache_file.touch()` and asserts `Cache.read(mode).file_data == {}`. That is the interrupted/truncated-write case this change is meant to keep on the same recovery path as malformed pickle data.

PR #5193, opened 2026-06-20
Validate include and force-exclude config types

CodingFeng101:
### Summary

Reject non-string `include` and `force-exclude` values in `pyproject.toml`.

### Details

Black already validates that `exclude` and `extend-exclude` config values are strings before passing them through regex validation. The same regex-backed options `include` and `force-exclude` did not have that type check, so list values could be silently stringified and accepted with surprising regex semantics.

This adds matching `BadOptionUsage` checks for `include` and `force_exclude`, plus a regression test for both config keys.

### Test

- `python -m pytest tests/test_black.py::BlackTestCase::test_read_pyproject_toml_rejects_non_string_regex_configs tests/test_black.py::BlackTestCase::test_read_pyproject_toml -q`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,6 +56,8 @@
   `srcs` are given) is now resolved before the `lru_cache` key is computed, so each
   directory gets the correct `pyproject.toml` (#5152)
 - Add validation for --line-ranges values (#5107)
+- Reject non-string `include` and `force-exclude` values in `pyproject.toml`
+  (#5193)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,8 +56,7 @@
   `srcs` are given) is now resolved before the `lru_cache` key is computed, so each
   directory gets the correct `pyproject.toml` (#5152)
 - Add validation for --line-ranges values (#5107)
-- Reject non-string `include` and `force-exclude` values in `pyproject.toml`
-  (#5193)
+- Reject non-string `include` and `force-exclude` values in `pyproject.toml` (#5193)
 
 ### Packaging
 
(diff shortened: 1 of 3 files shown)


cobaltt7:

CodingFeng101:
Added the requested changelog entry in 4baa9c8.

DTiming24 (commented):
Nice catch on list values being stringified. I like that this matches the existing exclude/extend-exclude validation. It may be worth making the error text explicitly mention include and force-exclude so users don't assume regex lists are accepted here.

DTiming24:
Good validation. Matching `exclude`'s failure mode makes the config errors much clearer; it may be worth checking whether the same message path is used when config is merged from CLI + pyproject, so users see the error at the source of the bad value.

CodingFeng101:
Thanks for the suggestion. I added a CLI-level regression test in `2db81d2` that writes a `pyproject.toml` with non-string `include` and `force-exclude` values, invokes Black through `--config`, and asserts the CLI exits with code 2 while surfacing the corresponding `Config key ... must be a string` message.

Validation run locally:

```text
python -m pytest tests/test_black.py::BlackTestCase::test_read_pyproject_toml_rejects_non_string_regex_configs tests/test_black.py::BlackTestCase::test_cli_rejects_non_string_pyproject_regex_configs tests/test_black.py::BlackTestCase::test_read_pyproject_toml -q
# 3 passed, 4 subtests passed

python -m black --check tests/test_black.py
python -m isort --check-only tests/test_black.py
python -m flake8 tests/test_black.py
git diff --check
```

I also tried `python -m pre_commit run --files tests/test_black.py`, but on this Windows environment it timed out while installing hook environments, so I ran the relevant checks directly.

cobaltt7:
Thanks!

PR #5200, opened 2026-06-26
Keep nested parens around lambda/ternary comprehension iterables (preview)

sarathfrancis90:
While poking at `wrap_comprehension_in` after #5176, I hit a case it doesn't cover. #5176 keeps the parens around a lambda or ternary used as a comprehension's iterable so the trailing `for`/`if` clauses aren't absorbed into it. But when those parens are redundantly nested, e.g.

```python
[x for x in ((lambda: 0))]
```

`maybe_make_parens_invisible_in_atom` recurses and strips *both* pairs, leaving `[x for x in lambda: 0]` — invalid code, which crashes with an `ASTSafetyError` at the default line length. Same for a nested ternary and for generator/dict/set comprehensions, and at deeper nesting.

The guard only inspected the atom's direct child, so a paren-wrapped atom slipped past it. I made it look through the redundant nested parens before deciding whether the iterable is a lambda/ternary, so at least one pair is kept (matching what the stable style does with such input).

Added the nested lambda/ternary cases to the existing `preview_wrap_comprehension_in` fixture — they crashed before and pass now. The full test suite, mypy, flake8 and the self-format check are green.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,9 +48,10 @@
 - Keep the parentheses around a lambda used as the iterable of a comprehension (e.g.
   `[x for x in (lambda: 0) if x]`). They were previously stripped by
   `wrap_comprehension_in`, which produced invalid code and crashed Black (#5176)
-- Keep redundant nested parentheses around a lambda or conditional expression used as a
-  comprehension's iterable (e.g. `[x for x in ((lambda: 0))]`). The inner pair was still
-  being stripped, leaving the bare expression and crashing Black (#5200)
+- Collapse redundant nested parentheses around a lambda or conditional expression used
+  as a comprehension's iterable down to a single pair (e.g. `[x for x in ((lambda: 0))]`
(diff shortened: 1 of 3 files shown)


cobaltt7:
Thanks for this fix! Would it be non-trivial to continue stripping the outer parentheses, only leaving one pair always? If that'd be a lot of extra work, I'd be fine deferring that and merging the PR as-is, but I'd appreciate that change if possible.

sarathfrancis90:
Done! Extended it to collapse all the way down to a single pair, so `[i for i in ((lambda: 0))]` now formats to `[i for i in (lambda: 0)]` (and deeper nesting like `(((lambda: 0)))` reduces the same way). It ended up being small: I let the usual paren-stripping run and then just restore one visible pair, so it reuses the existing comment/prefix handling rather than special-casing anything. Added the collapsed outputs plus a triple-nested case to the fixture; full suite, self-format and the idempotency check are green.

cobaltt7:
Thanks!

PR #5201, opened 2026-06-27
feat: add --cache-dir command-line option

### LINKED ISSUE #5174: Allow cache dir configuration via command-line argument

I work with a monorepo, and all my Python-related tools have their cache directories under `python/out/{tool_name}/cache`. I've configured my usages of black with the `BLACK_CACHE_DIR` environment variable appropriately (IDE run configurations, for instance).

Recently, I've added black as a pre-commit hook. Unfortunately, this has made it start writing its cache directory on the project root. I've browsed the issues of pre-commit, and the author seems _pretty adamant_ about **not** introducing an `env` configuration option at the hook level (for some valid reasons, might I add).

I would like Black to accept the path it should use for its cache directory as a command-line argument. It should probably take precedence over the environment variable, in accordance with how the `BLACK_NUM_WORKERS/--workers` pair currently operates. Additionally, this might make the `--no-cache-dir` option “redundant”, if the new option is configured to interpret an empty path string as “do not attempt to read/write to the cache”.

I've conceded to using `--no-cache-dir` as an extra argument when running black as a pre-commit hook. Alternative solutions, such as using a custom binary or maybe altering the git hook that pre-commit runs with, seem either too error-prone or difficult to propagate throughout a team strictly via VCS (but I'm open to suggestions!).

valter-silva-au:
### Description

Adds a `--cache-dir` command-line option that sets the directory where _Black_ reads
and writes its cache. It takes **precedence over** the `BLACK_CACHE_DIR` environment
variable, mirroring how `--workers` relates to `BLACK_NUM_WORKERS`.

Resolution precedence (highest first): `--cache-dir` → `BLACK_CACHE_DIR` →
`XDG_CACHE_HOME` → platform-specific user cache directory.

This is implemented by reassigning the module-level `black.cache.CACHE_DIR` constant
once in `main()` before files are dispatched — the same global the cache read/write
helpers (and the existing test fixture) already consult at runtime. The single- and
multi-file paths both read/write the cache in the parent process, so one override
covers both.

### Motivation

Closes #5174. The reporter runs _Black_ as a pre-commit hook in a monorepo. Pre-commit
deliberately won't expose an `env` knob at the hook level, so `BLACK_CACHE_DIR` can't be
set for that invocation and the cache lands in the project root. A CLI option lets the
cache location be configured directly and committed to VCS via the hook args.

Scope is intentionally narrow: this only **relocates** the formatting cache. It does not
overload an empty path as a cache-disable switch — the existing `--no-cache` flag
already serves that purpose.
Closes #5268

### Checklist - did you ...

### DIFF
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -26,8 +26,9 @@ from mypy_extensions import mypyc_attr
 from pathspec import GitIgnoreSpec
 from pathspec.patterns.gitignore import GitIgnorePatternError
 
+import black.cache
 from _black_version import version as __version__
-from black.cache import Cache
+from black.cache import Cache, get_cache_dir
 from black.comments import normalize_fmt_off
 from black.const import (
     DEFAULT_EXCLUDES,
@@ -540,6 +541,21 @@ def validate_regex(
         " included files."
     ),
 )
+@click.option(
+    "--cache-dir",
+    type=click.Path(
+        file_okay=False,
+        dir_okay=True,
+        writable=True,
+        path_type=str,
+    ),
+    default=None,
+    help=(
+        "The directory where Black should read and write its cache. This takes"
+        " precedence over the BLACK_CACHE_DIR environment variable. Defaults to the"
+        " platform-specific user cache directory."
+    ),
+)
 @click.pass_context
 def main(
     ctx: click.Context,
@@ -572,6 +588,7 @@ def main(
     src: tuple[str, ...],
     config: str | None,
     no_cache: bool,
+    cache_dir: str | None,
 ) -> None:
     """The uncompromising code formatter."""
     ctx.ensure_object(dict)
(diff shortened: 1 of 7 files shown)


src/black/__init__.py
```diff
@@ -697,6 +714,12 @@ def main(
 
     report = Report(check=check, diff=diff, quiet=quiet, verbose=verbose)
 
+    if cache_dir is not None:
+        # Redirect the on-disk cache for this run. This takes precedence over the
+        # BLACK_CACHE_DIR environment variable. Reassigning the module-level constant
+        # is what the cache read/write helpers consult at runtime.
+        black.cache.CACHE_DIR = get_cache_dir(override=Path(cache_dir))
```
cobaltt7, line 721: (Nit) If this is the only place where it's called with the override, why can't you run `Path(cache_dir) / __version__` inline?

JelleZijlstra:
> This is implemented by reassigning the module-level black.cache.CACHE_DIR constant
once in main()

That's the wrong approach. We should instead pass the cache dir around as an option. We should not reassign constants.

cobaltt7:
@valter-silva-au Are you still updating this?

cobaltt7:
Closing as stale. Please feel free to open a new one if you can follow the guidelines we provided above.

PR #5207, opened 2026-07-01
Write formatted files atomically to avoid corruption on write failure

### LINKED ISSUE #2479: black in-place reformat wipes or corrupts target when disk is full
**Describe the bug**

When performing an in-place reformat, if the disk is full and black cannot write to disk, instead of leaving the file untouched it wipes the entire file, or corrupts it.

**To Reproduce**

1. Fill up the disk
2. Attempt to reformat a target python file
3. Observe:
```
python3 -m black target.py
error: cannot format target.py: [Errno 28] No space left on device
Oh no! \U0001f4a5 \U0001f494 \U0001f4a5
1 file failed to reformat.
```
4. The file is now corrupt or in some cases completely wiped.

Example:
```
qlyoung@host ~> md5sum target.py
9640bcda965d5b519ae09ac7d3e87a91  target.py

qlyoung@host ~> df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda         20G   19G   78M 100% /

qlyoung@host ~> cat /dev/zero > bingus
cat: write error: No space left on device
qlyoung@host ~> df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda         20G   19G     0 100% /

qlyoung@host ~> python3 -m black target.py
error: cannot format target.py: [Errno 28] No space left on device
Oh no! \U0001f4a5 \U0001f494 \U0001f4a5
1 file failed to reformat.

qlyoung@host ~> md5sum target.py
9640bcda965d5b519ae09ac7d3e87a91  target.py
```

In this case the end of the file was truncated. In the case that caused me to notice this bug, the entire file was wiped.

**Expected behavior**

The file should be left with its original contents

```
~> lsb_release -a
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.3 LTS
Release:        20.04
Codename:       focal

~> python --version
Python 3.8.11

~> python -m black --version
__main__.py, version 20.8b1
```

I'm too sad to find out

**Additional context**

:cry:

ParamChordiya:
### Description

Fixes #2479.

Right now `format_file_in_place` opens the target file directly with `open(src, "w")` before writing the reformatted contents. That call truncates the file immediately, so if the write fails partway through for any reason (disk fills up, permissions get revoked mid-run, process gets killed, etc.) the original file is left wiped or half-written with no way to get it back. Since most people run Black on files that aren't always safely committed yet, that's a pretty rough way to lose code.

This was actually discussed in the issue back in 2021 - the suggestion from qlyoung (with JelleZijlstra agreeing) was to write to a temp file and swap it in, so that's what this PR does:

- Write the reformatted contents to a temp file created in the same directory as the target (so it's guaranteed to be on the same filesystem, which is what makes the swap atomic).
- `flush()` + `fsync()` the temp file before doing anything else, so we know the data actually made it to disk.
- Copy the original file's permission bits onto the temp file, then `os.replace()` it over the original. `os.replace` is atomic on both POSIX and Windows, so there's never a window where the file is half-written.
- If anything raises along the way, the temp file gets cleaned up and the original file is never touched.

One thing that took a bit of extra care: if the path being formatted is a symlink, `os.replace` operates on the symlink itself rather than what it points to, which would silently turn the symlink into a regular file. So there's a check for that - if `src` is a symlink, we resolve it first and write through to the real target, leaving the symlink itself alone.

Worth calling out (as was already flagged in the issue thread): this does mean Black now needs write permission on the containing directory, not just the file, since it has to create the temp file there. That seems like a reasonable trade-off for not losing people's work, but flagging it in case it's a concern for some setups.

Since JelleZijlstra asked in the thread whether other formatters handle this the same way, I went and checked a few:

- clang-format actually does exactly this - it writes to a temp file through LLVM's `TempFile` utility and renames it over the target (`llvm/lib/Support/Path.cpp`).
- gofmt doesn't rename a temp file over the target, but it does something in the same spirit: it copies the original file to a backup path before writing, and restores from that backup if the real write fails.
- rustfmt, Prettier, yapf, and autopep8 all just open the target file and write over it directly, as far as I could tell - I didn't find a fallback for this failure mode in any of them.

So it's a mixed bag rather than an industry standard - but there's real precedent for it in at least one widely-used formatter, and gofmt clearly cares about the same failure mode even though it solves it differently. Given Black is often run against files that aren't committed yet, I think it's worth fixing here regardless of what the others do.

On cooperlees' question about IO cost on large repos: I wasn't able to put together a reliable large-scale benchmark on my end, so take this as reasoning rather than a measured number. The extra work per file is one additional file creation, one `fsync`, and one `rename`, all constant-time metadata operations against the same directory regardless of the file's size. Parsing and re-rendering the file - which already happens today - should dominate the per-file cost by a good margin for anything but a near-empty file. Happy to help put together a proper before/after benchmark against a large real-world codebase if that would help move this along, or run whatever perf harness the project already uses if there is one.

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? (N/A - this doesn't change any formatting output, just how the result gets written to disk)
- [x] Add new / update outdated documentation? (N/A - no user-facing behavior or docs to update)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,8 +10,8 @@
   --target-version flag (#5167)
 - Fix in-place formatting truncating or corrupting the source file if writing the
   reformatted contents fails partway through, e.g. because the disk is full. Black now
-  writes to a temporary file in the same directory and atomically replaces the
-  original once the write has fully succeeded (#5207)
+  writes to a temporary file in the same directory and atomically replaces the original
+  once the write has fully succeeded (#5207)
 
 ### Highlights
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
 - Add support for NO_COLOR environment variable to disable ANSI output (#5129)
 - No spurious target version warning when runtime version is included in a
   --target-version flag (#5167)
+- Fix in-place formatting truncating or corrupting the source file if writing the
+  reformatted contents fails partway through, e.g. because the disk is full. Black now
+  writes to a temporary file in the same directory and atomically replaces the
+  original once the write has fully succeeded (#5207)
 
 ### Highlights
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -1,8 +1,11 @@
 import io
 import json
+import os
 import platform
 import re
+import shutil
 import sys
+import tempfile
 import tokenize
 import traceback
 from collections.abc import (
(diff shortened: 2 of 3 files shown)


JelleZijlstra:
Given that most other leading formatters don't do this, I'm not enthusiastic about the extra complexity here, and I suspect there's some risk of losing metadata associated with the original file.

PR #5211, opened 2026-07-01
Validate BLACK_NUM_WORKERS values

### LINKED ISSUE #5210: Invalid BLACK_NUM_WORKERS value can produce a raw ValueError traceback
### Summary

`BLACK_NUM_WORKERS` is documented in the `--workers` help as an alternate way to configure the same worker-count setting, but invalid environment variable values bypass the Click validation used by `--workers` and can produce a raw traceback.

### Reproduction

```bash
mkdir -p /tmp/black-workers-repro
printf 'x=1\n' >/tmp/black-workers-repro/a.py
printf 'y=2\n' >/tmp/black-workers-repro/b.py

BLACK_NUM_WORKERS=abc python -m black --check /tmp/black-workers-repro
```

### Actual behavior

The command fails with a Python traceback ending in:

```text
ValueError: invalid literal for int() with base 10: 'abc'
```

### Expected behavior

Since this is the environment-variable form of the same `workers` setting, invalid values should be reported as a user-facing configuration/parameter error rather than an internal traceback. For example, it could say that `BLACK_NUM_WORKERS` must be an integer greater than or equal to 1.

If the intended product semantics are that `BLACK_NUM_WORKERS` is only a best-effort low-level override and not equivalent to `--workers`, it may be worth documenting that explicitly; otherwise the help text currently suggests they share the same contract.

### Source of the mismatch

The CLI option is validated by Click as an integer range:

```python
@click.option(
    "-W",
    "--workers",
    type=click.IntRange(min=1),
    default=None,
    help=(
        "When Black formats multiple files, it may use a process pool to speed up"
        " formatting. This option controls the number of parallel workers. This can"
        " also be specified via the BLACK_NUM_WORKERS environment variable. Defaults"
        " to the number of CPUs in the system."
    ),
)
```

But the environment variable path parses the value directly:

```python
if workers is None:
    workers = int(os.environ.get("BLACK_NUM_WORKERS", 0))
    workers = workers or os.cpu_count() or 1
```

As a result, `--workers abc` is handled by Click's validation layer, while `BLACK_NUM_WORKERS=abc` raises `ValueError` directly.

### Possible fix

Parse `BLACK_NUM_WORKERS` through equivalent validation semantics, or catch `ValueError` / range errors and raise a Black/Click-style error that names the environment variable.

Regression tests could cover:

- `BLACK_NUM_WORKERS=abc`
- `BLACK_NUM_WORKERS=0`
- `BLACK_NUM_WORKERS=-1`

---
Submitted with PilotDeck ([OpenBMB/PilotDeck](https://github.com/OpenBMB/PilotDeck)).

MicroMilo:
## Summary

- validate `BLACK_NUM_WORKERS` before using it to size the worker pool
- report malformed, zero, or negative values as Click parameter errors that name the environment variable
- add regression coverage for invalid `BLACK_NUM_WORKERS` values

## Test Plan

- `python -m pytest tests/test_black.py::BlackTestCase::test_invalid_black_num_workers tests/test_black.py::BlackTestCase::test_works_in_mono_process_only_environment -q`
- `python -m black src/black/concurrency.py tests/test_black.py --check`

Fixes #5210

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -61,6 +61,8 @@
 - Ignore empty cache files like other malformed cache files instead of raising an
   `EOFError` (#5192)
 - Reject non-string `include` and `force-exclude` values in `pyproject.toml` (#5193)
+- Validate `BLACK_NUM_WORKERS` values and report invalid values as usage errors
+  instead of crashing (#5211)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -61,8 +61,8 @@
 - Ignore empty cache files like other malformed cache files instead of raising an
   `EOFError` (#5192)
 - Reject non-string `include` and `force-exclude` values in `pyproject.toml` (#5193)
-- Validate `BLACK_NUM_WORKERS` values and report invalid values as usage errors
-  instead of crashing (#5211)
+- Validate `BLACK_NUM_WORKERS` values and report invalid values as usage errors instead
+  of crashing (#5211)
 
 ### Packaging
 
--- a/src/black/concurrency.py
+++ b/src/black/concurrency.py
@@ -18,6 +18,7 @@ from multiprocessing import Manager
 from pathlib import Path
 from typing import Any
 
+import click
 from mypy_extensions import mypyc_attr
 
 from black import WriteBack, format_file_in_place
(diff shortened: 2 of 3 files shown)


cobaltt7:

PR #5213, opened 2026-07-01
batch --line-ranges node conversion to avoid quadratic removal

sahvx655-wq:
### Description

**Quadratic node removal in the `--line-ranges` conversion**

Profiling `black --line-ranges` over a synthetic module with a few thousand `elif` branches, `convert_unchanged_lines` dominated the run and `blib2to3`'s `Base.remove` alone accounted for roughly half of it. When a range leaves most of a file unchanged, every unchanged block is turned into a single `STANDALONE_COMMENT` by `_convert_node_to_standalone_comment` / `_convert_nodes_to_standalone_comment`, and those call `remove` and `insert_child` on the block's parent. Each call walks and shifts the parent's whole children list, so a parent that owns many sibling blocks (a long `if`/`elif` chain, a `match` with many cases, or a module full of top-level definitions) is O(n^2). A 4000-branch chain sat at about 3.6s and grew quadratically; left unfixed it is a cheap way to make `--line-ranges` on a large file crawl.

All the blocks under a parent are known up front, so there is no need to mutate the tree one block at a time. I collect the conversions in a small `_NodeReplacements` helper and rewrite each affected parent's children list in a single pass, splicing the comment leaves in place of the removed runs; nodes already scheduled for conversion are skipped so the deferred pass keeps the old "first conversion wins" behaviour (an `async` block keeps its `ASYNC` leaf on a different parent, and stacked decorators collapse to one node, both of which the single-pass rewrite handles). Output is byte-identical: I diffed the formatter over the `tests/data/cases` corpus and a batch of generated `if`/`elif`, `match`, decorator, `try`, `with` and mixed modules against many random line ranges, and the existing `test_ranges` suite still passes. The same 4000-branch chain now formats in about 0.6s and scales linearly.

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? (no style change; output is unchanged)
- [ ] Add / update tests if necessary? (behaviour is unchanged, so this is covered by the existing `test_ranges` suite)
- [x] Add new / update outdated documentation? (not applicable)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -110,6 +110,11 @@
   `--preview` string processing of `"%s ..." % (a, b, c, ...)` or a string with a
   backslash continuation) by resuming the child lookup in `append_leaves` instead of
   rescanning each leaf's parent from the start (#5199)
+- Improve performance of `--line-ranges` on files with many sibling blocks (a long
+  `if`/`elif` chain, a `match` with many cases, or many top-level definitions) by
+  splicing the unchanged blocks into each parent's child list in a single pass rather
+  than removing and re-inserting each one, which rescanned and shifted the whole child
+  list on every conversion (#5213)
 
 ### Output
 
--- a/src/black/ranges.py
+++ b/src/black/ranges.py
@@ -190,9 +190,78 @@ def convert_unchanged_lines(src_node: Node, lines: Collection[tuple[int, int]]) 
     lines_set: set[int] = set()
     for start, end in lines:
         lines_set.update(range(start, end + 1))
-    visitor = _TopLevelStatementsVisitor(lines_set)
+    replacements = _NodeReplacements()
+    visitor = _TopLevelStatementsVisitor(lines_set, replacements)
     _ = list(visitor.visit(src_node))  # Consume all results.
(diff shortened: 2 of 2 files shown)


cobaltt7:
Tests are failing, please take a look

sahvx655-wq:
Sorted. The failure was `line_ranges_decorator_edge_case`: the batched conversion mangled single-line decorated blocks, emitting `@overload\n\n\ndef foo(): ...` instead of keeping them together.

Root cause was an ordering the old per-node path quietly relied on. It converted the decorator in place first, so by the time the enclosing `decorated` node was converted the decorator was already a single leaf with its trailing newline stripped and its prefix migrated onto the standalone. Batching the conversions dropped that in-place step, so re-inserting a newline doubled the blank line and the outer conversion then read an empty prefix off the leaf the decorator run had already cleared. I removed the now-redundant newline insert, kept each first leaf's taken prefix so the inner and outer conversions reuse it, and zeroed that leaf's lineno to mirror the line-0 standalone the in-place path used to leave behind. `test_ranges` and `test_format` both pass locally now.

PR #5219, opened 2026-07-03
fix: rename unused loop variable to _i

decembercomposer697-hue:
B007: loop control variable `i` in conv.py was assigned but never used.
✅ 226 tests pass.


cobaltt7:
Unnecessary change.

PR #5220, opened 2026-07-04
avoid quadratic per-leaf copy in StringMerger.do_transform

sahvx655-wq:
### Description

Profiling `--preview` string merging on `"%s ..." % (a, b, c, ...)` still showed a quadratic after #5199 linearised `append_leaves`: the run sat around 4s for 8000 operands with `len` called about 32 million times. `StringMerger.do_transform` rebuilds the merged line by walking every leaf of the old line and copying each leaf that is not part of a merged string with its own `append_leaves(new_line, line, [leaf])` call. The resume hint added in #5199 only persists inside a single `append_leaves` call, so a one-leaf call always restarts the search for the leaf's position from the front. A `%` expression's operand tuple is a single parent holding every operand, so copying its n leaves one at a time rescans that child list from index 0 each time and is O(n²). It is reachable pre-auth from `blackd` with `X-Preview` and `X-Enable-Unstable-Feature: string_processing`.

I gather the consecutive leaves that fall outside any merged string group and hand each contiguous run to `append_leaves` in one call, which is the shape the resume hint was built for, so the copy stays linear. The per-leaf calls were doing nothing the batched call does not, comments included, so the new line is built identically. Output is byte-identical across the test suite and the black source tree in stable, preview and unstable modes plus crafted `%`-format and implicit-concat inputs; the 8000-operand case drops from roughly 4.1s to 0.8s and the curve flattens from quadratic to linear.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -120,6 +120,10 @@
   splicing the unchanged blocks into each parent's child list in a single pass rather
   than removing and re-inserting each one, which rescanned and shifted the whole child
   list on every conversion (#5213)
+- Improve performance of `--preview` string merging on lines such as
+  `"%s ..." % (a, b, c, ...)` by copying the leaves that surround the merged string in
+  one `append_leaves` call instead of one call per leaf, which rescanned the shared
+  parent's child list from the start every time (#5220)
 
 ### Output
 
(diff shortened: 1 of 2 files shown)


sahvx655-wq:
Merged main in to clear the CHANGES.md conflict, no code changes; the n=4000 timing still holds locally.

cobaltt7:
Thanks! Sorry this one got lost

PR #5222, opened 2026-07-04
Memory Optimizations

### LINKED ISSUE #5216: Potential memory optimizations
### POC
I have added `__slots__` to `pytree.Leaf` and `pytree.Node`.

These changes reduced memory usage to one-third, i.e. 3x improvement, also gave additional 10% speed up.

```
class Leaf(Base):
    __slots__ = ('value', 'fixers_applied', 'bracket_depth', 'used_names')

class Node(Base):
    __slots__ = ('fixers_applied', 'used_names', 'prev_sibling_map', 'next_sibling_map')
```

### Expectation
Optimizing the tool for memory and henceforth improving the performance.
1. Add slots and frozen to frequently used classes or make them `dataclasses`. (This seems to be more impactful here)
2. Using iterators/ comprehensions.

rsb-23:
### Description

**Reference** : issue #5216
1. Reduces per-instance memory of Node and Leaf (the two most frequently instantiated objects;
 every token and AST node in a formatted file is one of these) 
and cuts allocation overhead across the hot formatting path.
2. Removed dead fields (`used_names` and `fixers_applied`), never read anywhere.
3. Simplified unnecessary string and other object creations to reduce memory allocations.
4. Removed avoidable intermediate allocations in hot paths like `r = r0 | r1`

> Measured on a synthetic benchmark: peak memory 75.325 MB → 20.107 MB after these changes.

### Checklist - did you ...

### DIFF
--- a/src/black/brackets.py
+++ b/src/black/brackets.py
@@ -56,7 +56,7 @@ class BracketMatchError(Exception):
     """Raised when an opening bracket is unable to be matched to a closing bracket."""
 
 
-@dataclass
+@dataclass(slots=True, eq=False)
 class BracketTracker:
     """Keeps track of brackets on a line."""
 
@@ -156,7 +156,7 @@ class BracketTracker:
             return 0
 
         priority = priority or self.max_delimiter_priority()
-        return sum(1 for p in self.delimiters.values() if p == priority)
+        return sum(p == priority for p in self.delimiters.values())
 
     def maybe_increment_for_loop_variable(self, leaf: Leaf) -> bool:
         """In a for loop, or comprehension, the variables are often unpacks.
--- a/src/black/brackets.py
+++ b/src/black/brackets.py
@@ -156,7 +156,7 @@ class BracketTracker:
             return 0
 
         priority = priority or self.max_delimiter_priority()
-        return sum(p == priority for p in self.delimiters.values())
+        return sum(1 for p in self.delimiters.values() if p == priority)
 
     def maybe_increment_for_loop_variable(self, leaf: Leaf) -> bool:
         """In a for loop, or comprehension, the variables are often unpacks.
--- a/src/black/handle_ipynb_magics.py
+++ b/src/black/handle_ipynb_magics.py
@@ -460,7 +460,9 @@ class MagicFinder(ast.NodeVisitor):
             if node.value.func.attr == "getoutput":
                 src = f"!{args[0]}"
             elif node.value.func.attr == "run_line_magic":
-                src = "%" + " ".join(filter(None, args))
+                src = f"%{args[0]}"
+                if args[1]:
+                    src += f" {args[1]}"
             else:
                 raise AssertionError(
                     f"Unexpected IPython magic {node.value.func.attr!r} found. "
@@ -498,7 +500,9 @@ class MagicFinder(ast.NodeVisitor):
                 elif args[0] == "pinfo2":
                     src = f"??{args[1]}"
                 else:
-                    src = "%" + " ".join(filter(None, args))
+                    src = f"%{args[0]}"
+                    if args[1]:
+                        src += f" {args[1]}"
             elif node.value.func.attr == "system":
                 src = f"!{args[0]}"
             elif node.value.func.attr == "getoutput":
--- a/src/black/handle_ipynb_magics.py
+++ b/src/black/handle_ipynb_magics.py
@@ -460,9 +460,7 @@ class MagicFinder(ast.NodeVisitor):
             if node.value.func.attr == "getoutput":
                 src = f"!{args[0]}"
             elif node.value.func.attr == "run_line_magic":
-                src = f"%{args[0]}"
-                if args[1]:
-                    src += f" {args[1]}"
+                src = "%" + " ".join(filter(None, args))
             else:
                 raise AssertionError(
                     f"Unexpected IPython magic {node.value.func.attr!r} found. "
@@ -500,9 +498,7 @@ class MagicFinder(ast.NodeVisitor):
                 elif args[0] == "pinfo2":
                     src = f"??{args[1]}"
                 else:
-                    src = f"%{args[0]}"
-                    if args[1]:
-                        src += f" {args[1]}"
+                    src = "%" + " ".join(filter(None, args))
             elif node.value.func.attr == "system":
                 src = f"!{args[0]}"
             elif node.value.func.attr == "getoutput":
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -197,9 +197,7 @@ class Line:
         if not self or self.leaves[0].type != token.STRING:
             return False
         value = self.leaves[0].value
-        if value.startswith(('"""', "'''")):
-            return True
-        return value.startswith(("r'''", 'r"""', "R'''", 'R"""'))
+        return value.startswith(('"""', "'''", "r'''", 'r"""', "R'''", 'R"""'))
 
     @property
     def is_docstring(self) -> bool:
@@ -481,9 +479,9 @@ class Line:
             return "\n"
 
         indent = "    " * self.depth
-        first, *rest = self.leaves
+        first = self.leaves[0]
 
-        rest_str = "".join(map(str, rest))
+        rest_str = "".join(map(str, self.leaves[1:]))
         comments_str = "".join(
             str(comment) for group in self.comments.values() for comment in group
         )
@@ -1607,6 +1605,7 @@ def can_omit_invisible_parens(
         # a leading opening bracket and a trailing closing bracket.  If the
         # opening bracket doesn't match our rule, maybe the closing will.
 
+    penultimate = line.leaves[-2]
     last = line.leaves[-1]
 
     if (
@@ -1620,7 +1619,6 @@ def can_omit_invisible_parens(
             and last.parent.type != syms.trailer
         )
     ):
-        penultimate = line.leaves[-2]
         if penultimate.type in OPENING_BRACKETS:
             # Empty brackets don't help.
             return False
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -38,7 +38,7 @@ LeafID = int
 LN = Union[Leaf, Node]
 
 
-@dataclass
+@dataclass(slots=True, eq=False)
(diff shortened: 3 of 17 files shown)


src/blib2to3/pytree.py
```diff
@@ -61,16 +62,13 @@ class Base:
     A node may be a subnode of at most one parent.
     """
 
-    # Default values for instance variables
-    type: int  # int: token number (< 256) or symbol number (>= 256)
-    parent: Optional["Node"] = None  # Parent node pointer, or None
-    children: list[NL]  # List of subnodes
-    was_changed: bool = False
```
cobaltt7, line 68: Could you migrate these comments into `__init__`?

src/blib2to3/pytree.py
cobaltt7: Is this argument now unused?

cobaltt7 (commented):
Thanks! Could you add a high-level changelog entry for one or two of the biggest changes?

tusharsadhwani:
Did you compare this under the mypyc build? slots shouldn't make that big of a difference after being compiled

src/blib2to3/pytree.py
rsb-23: Yes, `fixer_applied` is now unused in both `Node` and `Leaf`. But since Node had unused 'context' param, so I kept this param as is.

src/black/brackets.py
```diff
@@ -156,7 +156,7 @@ def delimiter_count_with_priority(self, priority: Priority = 0) -> int:
             return 0
 
         priority = priority or self.max_delimiter_priority()
-        return sum(1 for p in self.delimiters.values() if p == priority)
+        return sum(p == priority for p in self.delimiters.values())
```
JelleZijlstra: Why would this reduce memory usage? Python won't allocate a new `1` object.

src/black/handle_ipynb_magics.py
```diff
@@ -460,9 +460,7 @@ def visit_Assign(self, node: ast.Assign) -> None:
             if node.value.func.attr == "getoutput":
                 src = f"!{args[0]}"
             elif node.value.func.attr == "run_line_magic":
-                src = f"%{args[0]}"
-                if args[1]:
-                    src += f" {args[1]}"
+                src = "%" + " ".join(filter(None, args))
```
JelleZijlstra: My intuition is that this is likely slower and uses more memory.

src/black/handle_ipynb_magics.py
```diff
@@ -500,9 +498,7 @@ def visit_Expr(self, node: ast.Expr) -> None:
                 elif args[0] == "pinfo2":
                     src = f"??{args[1]}"
                 else:
-                    src = f"%{args[0]}"
-                    if args[1]:
-                        src += f" {args[1]}"
+                    src = "%" + " ".join(filter(None, args))
```
JelleZijlstra: Same

src/black/lines.py
```diff
@@ -199,9 +199,7 @@ def _is_triple_quoted_string(self) -> bool:
         value = self.leaves[0].value
         if value.startswith(('"""', "'''")):
             return True
-        if value.startswith(("r'''", 'r"""', "R'''", 'R"""')):
-            return True
-        return False
+        return value.startswith(("r'''", 'r"""', "R'''", 'R"""'))
```
JelleZijlstra: Can also be combined with the previous startswith call on lines 200-201

src/black/lines.py
```diff
@@ -483,22 +481,21 @@ def __str__(self) -> str:
...
 
         indent = "    " * self.depth
-        leaves = iter(self.leaves)
-        first = next(leaves)
-        res = f"{first.prefix}{indent}{first.value}"
-        res += "".join(str(leaf) for leaf in leaves)
-        comments_iter = itertools.chain.from_iterable(self.comments.values())
-        comments = [str(comment) for comment in comments_iter]
-        res += "".join(comments)
+        first, *rest = self.leaves
```
JelleZijlstra: The new code overall seems nicer here but note this materializes `rest` into a new list and therefore might use more memory.

src/black/output.py
```diff
@@ -89,8 +89,7 @@ def diff(a: str, b: str, a_name: str, b_name: str) -> str:
         if line[-1] == "\n":
             diff_lines.append(line)
         else:
-            diff_lines.append(line + "\n")
-            diff_lines.append("\\ No newline at end of file\n")
+            diff_lines.extend((line + "\n", "\\ No newline at end of file\n"))
```
JelleZijlstra, line 92: Making an intermediate tuple likely means more memory usage.

src/black/output.py
```diff
@@ -113,9 +112,7 @@ def color_diff(contents: str) -> str:
 
 
 def style_output(message: str, **styles: Any) -> str:
-    if not _color_enabled():
```
JelleZijlstra, line 116: This shouldn't affect memory usage and the old code is clearer.

src/black/trans.py
```diff
@@ -360,7 +360,7 @@ def _get_key(string: str) -> CustomSplitMapKey:
             A unique identifier that is used internally to map @string to a
             group of custom splits.
         """
-        return (id(string), string)
+        return id(string), string
```
JelleZijlstra: This definitely doesn't affect memory usage.

src/black/trans.py
```diff
@@ -1397,11 +1397,13 @@ def _toggle_fexpr_quotes(fstring: str, old_quote: str) -> str:
     escaping, once Black figures out how to parse the new grammar.
     """
     new_quote = "'" if old_quote == '"' else '"'
-    parts = []
+    parts: list[str] = []
     previous_index = 0
     for start, end in iter_fexpr_spans(fstring):
-        parts.append(fstring[previous_index:start])
-        parts.append(fstring[start:end].replace(old_quote, new_quote))
+        parts.extend((
```
JelleZijlstra, line 1405: old code is better

src/blib2to3/pgen2/conv.py
```diff
@@ -168,7 +168,7 @@ def parse_graminit_c(self, filename):
         mo = re.match(r"static dfa dfas\[(\d+)\] = {$", line)
         assert mo, (lineno, line)
         ndfas = int(mo.group(1))
-        for i in range(ndfas):
+        for _ in range(ndfas):
```
JelleZijlstra: doesn't affect memory

src/blib2to3/pytree.py
```diff
@@ -51,7 +52,7 @@ def type_repr(type_num: int) -> str | int:
 RawNode = tuple[int, Optional[str], Optional[Context], Optional[list[NL]]]
 
 
-class Base:
+class Base(ABC):
```
JelleZijlstra, line 55: Making this an ABC may well have negative performance effects.

JelleZijlstra (changes requested):
This makes a lot of stylistic changes, many of which are unlikely to affect memory usage.

You claim this reduced memory usage significantly. How did you measure this? Did you compile Black with mypyc? And which specific changes can be proven to help?

src/blib2to3/pytree.py
rsb-23: Added locally, will push soon

rsb-23:
@tusharsadhwani @JelleZijlstra
I have used the default build, I will confirm for mypyc. 

I took few of the slow testcases to profile for memory and check hot classes.
Here is 1 of the profiler scripts [memory_profier.py](https://github.com/rsb-23/black/blob/optimize-memory/memory_profile.py)

src/blib2to3/pgen2/conv.py
rsb-23: true, just that it isn't used in the loop.

src/black/output.py
rsb-23: Will revert

src/black/output.py
rsb-23: Both append and extend makes a method lookup, so 1 is generally better. 

Also, it won't add to memory usage as Cython uses cache memory for tuple even if there's no tuple.
Rest, is handled by gc. Henceforth the change.

src/black/trans.py
JelleZijlstra: We're unlikely to accept a big change that's just your personal style preferences.

src/black/lines.py
rsb-23: made the change

src/black/lines.py
rsb-23: Thanks to point it out.
Since self.leaves is already a list, hence now using it directly instead of creating iter overhead.

src/black/handle_ipynb_magics.py
rsb-23: reverted

src/black/handle_ipynb_magics.py
rsb-23: reverted

rsb-23:
These results are based on the profiler script [memory_profier.py](https://github.com/rsb-23/black/blob/optimize-memory/memory_profile.py)
OS: Linux
Python: 3.12

### Performance (Negligible)
| Build | Pure Python | mypyc|
|---|---|---|
|Upstream|21.6s|14.4s|
|This PR|21.0s|13.6s|

### Memory 
Slight improvement in mypyc build, while Pure python is much better
> Current \ Peak allocations (in MB)

| Build | Pure Python | mypyc|
|---|---|---|
|Upstream|12.174  \ 41.601 |9.538  \ 25.711|
|This PR|1.481 \ 20.075| 7.157 \ 24.568 |

src/blib2to3/pytree.py
rsb-23: There is non-negative effect indicated in stats below.
[mypyc ABC support](https://mypyc.readthedocs.io/en/stable/native_classes.html#metaclasses)

src/black/trans.py
rsb-23: Not a style preference, but a micro-optimization.
Reason is as for 'output.py' changes.

src/blib2to3/pytree.py
cobaltt7: Could you remove both while we're here? I imagine unused variables would also contribute to high memory usage?

PR #5223, opened 2026-07-05
fix: add conditional stripping for Linux executables

### LINKED ISSUE #2291: Debug Linux self contained binary size + reduce if possible
For some reason, the Linux self contained black binary is double the size of MacOS + Windows. This seems weird.

1) Lets try workout why / root cause
2) If possible, try clean it up / reduce the size

Screen Shot 2021-05-31 at 7 37 41 AM

vaibhavxlr:
### Description

This essentially fixes https://github.com/psf/black/issues/2291. 

I tried building the Linux executable on my machine ( Ubuntu 24.04 x86_64 desktop running Python 3.14 ) and found that executable generated was around ~12M, roughly the same size of PE and Mach-O generated by the Github Action workflow. 

On analysing the executable a bit, I found the main difference was the size of Python interpreter being bundled by PyInstaller. Following is the size comparison-

On my local-
```text
-rw-r--r-- 1 root root 8.7M Mar 24 00:34 /usr/lib/x86_64-linux-gnu/libpython3.14.so.1.0
```

On Github Action runner-
```text
-rwxrwxrwx 1 runner packer 31M Jun 28 21:54 /opt/hostedtoolcache/Python/3.14.6/x64/lib/libpython3.14.so.1.0
```

And this is what file command returned-

On my local-
```text
/usr/lib/x86_64-linux-gnu/libpython3.14.so.1.0: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=0d148f3b8bdf451ed0e7e416957307e9b691032b, stripped
```

On Github Action runner-
```text
/opt/hostedtoolcache/Python/3.14.6/x64/lib/libpython3.14.so.1.0: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=e6fdb4bf77689f4fbc54f28b6562a4c5121963df, with debug_info, not stripped
```

As we can see Github Action libpython is not stripped and carries DWARF data, as compared to what gets shipped on Ubuntu systems by default. It didn't make sense to me that release builds carried debug information anyway. I looked up a bit and found that PyInstaller offered [--strip](https://pyinstaller.org/en/stable/usage.html#cmdoption-s), which made sense over manually stripping the libpython provided by the runner's environment.

As it is not recommended to be used on Windows and the Mac executable seemed fine, I have added the appropriate matrix based check. 

I ran the workflow on my fork and these are the results-
image

I also did some sanity tests locally by trying to format some projects using the above x86_64 build just to be on safer side.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,7 +69,7 @@
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
 - Reduce the size of Linux standalone binaries by stripping debug symbols during the
-  PyInstaller release build (#5233)
+  PyInstaller release build (#5223)
 
 ### Parser
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -68,6 +68,8 @@
 
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
+- Reduce the size of Linux standalone binaries by stripping debug symbols during the PyInstaller release build (#5233)
+
 ### Parser
 
 <!-- Changes to the parser or to version autodetection -->
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -68,7 +68,8 @@
 
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
-- Reduce the size of Linux standalone binaries by stripping debug symbols during the PyInstaller release build (#5233)
+- Reduce the size of Linux standalone binaries by stripping debug symbols during the
+  PyInstaller release build (#5233)
 
 ### Parser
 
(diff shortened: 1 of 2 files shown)


cobaltt7:
Thanks! Can you add a changelog entry?

vaibhavxlr:
> Thanks! Can you add a changelog entry?

Hey, it's done.

CHANGES.md
```diff
@@ -68,6 +68,9 @@
 
 <!-- Changes to how Black is packaged, such as dependency requirements -->
 
+- Reduce the size of Linux standalone binaries by stripping debug symbols during the
+  PyInstaller release build (#5233)
```
```suggestion
  PyInstaller release build (#5223)
```
corrected PR number

PR #5238, opened 2026-07-15
Preserve blank lines before fmt-off classes

### LINKED ISSUE #5225: Inconsistent number of lines after import
**Describe the bug**

If the first thing after the import is a class, optionally with an arbitrary comment - black makes the separation as 2 lines - in line with PEP-8.
Unless the comment is `# fmt: off` - then black ignores that there's a class, and makes it one empty line, conflicting with various other hooks in our pre-commit.

**To Reproduce**

```python
import sys


# fmt: off
class C:
    def foo(self):
        print(sys.platform)
```

**Expected behavior**

Always 2 lines in this case. 

**Environment**

- Black's version: 26.5.1
- OS and Python version: Windows 11/Python 3.12.10

**Additional context**

* Both flake8 and isort insist on 2 lines here.
* While flake8 can be pacified, isort can't be and cannot reproduce this behavior with any settings.
* PEP-8 says 2 lines.
* We usually toggle black off for the test classes to make them readable.

sjh9714:
### Description

Fixes #5225.

When a top-level class starts inside a `# fmt: off` block after an import, the
converted block looks like a standalone comment to the empty-line tracker. This caused
Black to keep only one blank line instead of the two required before a top-level class.

Preserve the converted block's original `class` token when deciding import separation.
The change is limited to a new preview feature, so stable formatting remains unchanged,
and a focused fixture covers the reported input.

### Testing

- `.venv/bin/pytest tests/test_format.py -k preview_fmt_off_class_blank_lines -q`
- `.venv/bin/pytest tests/test_format.py -q`
- `env -u NO_COLOR .venv/bin/pytest -q`
- `env -u NO_COLOR .venv/bin/pre-commit run mypy --files src/black/lines.py src/black/mode.py`
- `env -u NO_COLOR .venv/bin/tox -e run_self`
- `.venv/bin/tox -e generate_schema`

`env -u NO_COLOR .venv/bin/pre-commit run --all-files` passed every hook except
repository-wide mypy, which reports an unchanged `action/main.py` `tomllib`
ignore/stub mismatch under Python 3.14. The targeted mypy hook above passes for both
changed Python modules.

Note: I used Codex while preparing this change, reviewed the final diff, and ran the
checks listed above locally.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,8 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
+  after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
   `x[key] = expr` (#5095)
 - Parenthesize tuple expressions in `yield` statements for consistency with function
--- a/docs/the_black_code_style/future_style.md
+++ b/docs/the_black_code_style/future_style.md
@@ -43,6 +43,8 @@ Currently, the following features are included in the preview style:
   anyway; let the bracket explode instead. ([see below](labels/hug-comparator))
 - `parenthesize_tuple_in_yield`: Add parentheses around tuple expressions in `yield`
   statements.
+- `fmt_off_class_blank_lines`: Preserve two blank lines before a top-level class whose
+  definition starts inside a `# fmt: off` block after an import.
 
 (labels/wrap-comprehension-in)=
 
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -1114,6 +1114,17 @@ class EmptyLineTracker:
             )
 
         if (
+            Preview.fmt_off_class_blank_lines in self.mode
+            and self.previous_line.is_import
+            and self.previous_line.depth == 0
+            and current_line.depth == 0
+            and current_line.is_fmt_pass_converted(
+                first_leaf_matches=lambda leaf: leaf.value == "class"
+            )
+        ):
+            return 2, 0
+
+        if (
             self.previous_line.is_import
             and self.previous_line.depth == 0
             and current_line.depth == 0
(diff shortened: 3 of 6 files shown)


cobaltt7:
What logic normally handles newlines before classes? Why isn't it triggering here? Would it make morse sense to update it?

sjh9714:
Good question. I dug into where this comes from.

The logic that normally handles blank lines before classes is `EmptyLineTracker._maybe_empty_lines`, which routes any line whose `is_decorator`/`is_def`/`is_class` is true to `_maybe_empty_lines_for_class_or_def` (src/black/lines.py:1111).

It doesn't trigger here because `normalize_fmt_off` collapses the whole `# fmt: off` region into a single `STANDALONE_COMMENT` leaf (`convert_one_fmt_off_pair` in src/black/comments.py), keeping the original first token only in `fmt_pass_converted_first_leaf`. So the line's first leaf is a `STANDALONE_COMMENT`, not a `NAME` `class` — `is_class` is `False`, the class/def dispatch is skipped, and the line falls through to the top-level import rule just below (src/black/lines.py:1128), which returns 1 blank line. That's exactly the reported behavior.

On updating the normal logic instead: I considered teaching `is_class` (or the dispatch) to see through fmt-pass-converted lines, but `is_class`/`is_def` are consulted in many other places (stub-class handling, `previous_defs` tracking, docstring logic), and routing an opaque fmt:off block into `_maybe_empty_lines_for_class_or_def` would let other rules act on content that `# fmt: off` promises to leave alone. The two import rules directly below already special-case `is_fmt_pass_converted(first_leaf_matches=is_import)`, so I kept the fix in the same shape and place: a narrow guard for the import → fmt-converted-class boundary.

That said, if you'd rather see this handled inside `_maybe_empty_lines_for_class_or_def` (or generalized beyond the after-imports case), I'm happy to rework it that way.

cobaltt7:
Sounds good. Thank you for investigating & fixing this!

cobaltt7:
Whoops, wrong button

PR #5245, opened 2026-07-18
Add test coverage for Python 3.12+ type params and PEP 701 f-strings

Emilianocarfu88:
### Description

Adds test coverage for Python 3.12+ syntax that was thin in the suite, focused on how
**PEP 695 type parameters** (incl. PEP 696 defaults, `*TypeVarTuple`/`**ParamSpec`) and
**PEP 701 f-strings** interact with Black's line splitting and comment handling.

New data-driven cases (auto-discovered by `test_simple_format`, so each also runs Black's
built-in AST-equivalence, idempotency/stability, preview, and line-length-1 checks):

- `tests/data/cases/type_params_comments.py` — comments inside type-param brackets
  (inline, leading/standalone, on bounds and constraint tuples).
- `tests/data/cases/type_param_defaults_comments.py` — PEP 696 defaults with comments and
  wrapping.
- `tests/data/cases/fstring_pep701_comments.py` — comments in multiline interpolations,
  nested/quote-reuse f-strings, format specs, `=` debug, call-argument splitting.
- `tests/data/cases/type_params_line_split.py` — overflowing type-param lists,
  bounds/constraints, combined type-param + argument + return-type splitting.

Property-based tests:

- `tests/test_pep695_701_properties.py` — targeted Hypothesis strategies for type-param
  lists and f-strings asserting **idempotency** (`assert_stable`) and **AST safety**
  (`assert_equivalent`) across varied modes (line lengths incl. 1, preview, magic-comma,
  string normalization). Every generated example is `compile()`-checked first so a
  strategy bug can't masquerade as a Black bug. Mirrors the deterministic settings of
  `scripts/fuzz.py`.

Regression test:

- `tests/test_regression_type_params.py` — a strict-`xfail` pinning #5243 (a magic trailing
  comma fails to explode a type-param list once it contains `*TypeVarTuple`/`**ParamSpec`).
  It asserts the desired exploded output, so when the bug is fixed the test XPASSes and the
  strict marker turns that into a failure, prompting promotion to a normal data case.

All expected `# output` blocks were generated by running Black itself, never hand-written.

### Checklist

- [x] Added tests
- [ ] No user-facing changes, so no `CHANGES.md` entry (happy to add one if preferred)

Relates to #5243.

### DIFF
--- a/tests/data/cases/fstring_pep701_comments.py
+++ b/tests/data/cases/fstring_pep701_comments.py
@@ -1,0 +1,80 @@
+# flags: --minimum-version=3.12
+x = f"result {
+    value  # inline comment in interpolation
+} done"
+
+
+y = f"nested {f"inner {a}"} outer"
+
+
+z = f"reuse {"double quotes inside"} works now"
+
+
+spec = f"padded {value:{width}d} end"
+
+
+conv = f"repr {obj!r} str {obj!s} ascii {obj!a}"
+
+
+debug = f"debug {2 + 2 = } and {name = }"
+
+
+multi = f"line one {
+    first_value  # comment one
+} middle {
+    second_value  # comment two
+} last"
+
+
+call = function(f"prefix {some.method(argument_one, argument_two, argument_three)} suffix", other)
+
+
+nested_spec = f"a{2 + 2:=^{foo(x + y**2):another spec}}b"
+
+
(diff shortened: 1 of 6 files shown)


cobaltt7:
This does not fix anything. These constructs are already tested.

PR #5246, opened 2026-07-21
Preserve single-element tuple comma under --skip-magic-trailing-comma

sarathfrancis90:
### Description

Running with `--skip-magic-trailing-comma`, Black could delete the required trailing comma from a single-element tuple, turning `(x,)` into `(x)` and changing the code's meaning. It shows up when the tuple is a lambda parameter default and a standalone comment forces it across multiple lines, e.g.:

```python
foo(lambda x=(
    # comment
    1,
): x)
```

Black rewrote the `1,` to `1`, which made the formatted output non-equivalent to the source and tripped the AST safety check (`INTERNAL ERROR ... produced code that is not equivalent`).

The trailing comma of a one-element tuple is syntactically required, so it is never a magic trailing comma. `Line.has_magic_trailing_comma` normally recognizes this, but its one-tuple check (`is_one_sequence_between`) looks for the opening bracket in the current line's leaves. When a standalone comment inside the tuple splits the opening `(` onto an earlier line, it is no longer among those leaves, so the tuple is misread as an ordinary collection and the comma gets dropped during line construction.

I guarded the comma removal in `Line.append` with a tree-based `is_one_tuple` check, which does not depend on where the brackets landed. The magic-trailing-comma (default) code path is untouched, so stable formatting elsewhere is unchanged; genuine magic commas (single-argument calls, lists, sets, multi-element tuples) are still removed as before.

### How I tested

- Added input/output cases to `tests/data/cases/skip_magic_trailing_comma.py`; confirmed they fail before the change and pass after.
- Ran the full test suite (`pytest tests/`) — green.
- Fuzzed the fix by running every `tests/data/cases` input through many line lengths and modes (with and without `--skip-magic-trailing-comma`, stable and preview) and checking AST equivalence and stability — no divergences remain.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,10 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Fix dropping the required trailing comma from a single-element tuple used as a lambda
+  parameter default under `--skip-magic-trailing-comma` when a standalone comment forces
+  the tuple across multiple lines; removing the comma turned the tuple into a bare
+  expression and failed Black's AST safety check (#5246)
 - Fix unstable formatting when an inline comment sits on optional parentheses (for
   example a parenthesized assert message or assignment RHS) (#5241)
 - Fix crash when a standalone comment sits between tokens of a comprehension or lambda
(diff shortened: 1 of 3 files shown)


cobaltt7:
Thank you!

PR #5252, opened 2026-07-22
Prefer splitting whole boolean conditions

### LINKED ISSUE #3629: Parenthesize whole conditional expression with function calls rather than wrap last function call
**Describe the style change**

```python
from typing import Sequence


very_very_very_very_very_very_very_very_long_expression = 'foo'
if isinstance(very_very_very_very_very_very_very_very_long_expression, Sequence) and not isinstance(
    very_very_very_very_very_very_very_very_long_expression, str
):
    print('foo')


another_very_very_very_very_very_very_very_very_long_expression = False
if (
    very_very_very_very_very_very_very_very_long_expression
    and another_very_very_very_very_very_very_very_very_long_expression
):
    print('foo')

```

**Desired style**

```python
from typing import Sequence


very_very_very_very_very_very_very_very_long_expression = 'foo'
if (
    isinstance(very_very_very_very_very_very_very_very_long_expression, Sequence)
    and not isinstance(very_very_very_very_very_very_very_very_long_expression, str)
):
    print('foo')


another_very_very_very_very_very_very_very_very_long_expression = False
if (
    very_very_very_very_very_very_very_very_long_expression
    and another_very_very_very_very_very_very_very_very_long_expression
):
    print('foo')
```

**Additional context**

`black` can automatically add parenthesis around long conditions and wrap at `and` and `or`.

```diff
 very_very_very_very_very_very_very_very_long_expression = 'foo'
 another_very_very_very_very_very_very_very_very_long_expression = False
-if very_very_very_very_very_very_very_very_long_expression and another_very_very_very_very_very_very_very_very_long_expression:
+if (
+    very_very_very_very_very_very_very_very_long_expression
+    and another_very_very_very_very_very_very_very_very_long_expression
+):
     print('foo')
```

But `black` wraps the last clause rather than the whole condition expression if the last clause is a function call.

zhangli091011:
### Description

When an `if`, `elif`, or `while` condition contains top-level boolean operators and function calls, prefer splitting the condition's optional outer parentheses before splitting an inner call. This adds a narrowly scoped preview-style RHS candidate and falls back to the existing split order if that candidate does not fit.

The stable style is unchanged.

Fixes #3629.

### Tests

- `python -m pytest tests/test_format.py -q` (229 passed)
- `python -m tox -e py` (475 passed, 1 skipped; 73 Jupyter tests passed; 94% coverage)
- pre-commit hooks: isort, flake8, mypy, repository checks, end-of-file-fixer, trailing-whitespace
- `npx prettier@3.9.6 --check CHANGES.md docs/the_black_code_style/future_style.md`
- explicit stable/preview output-isolation check

### AI disclosure

This contribution was developed with assistance from OpenAI GPT-5.6-sol. I reviewed and tested all changes.

### DIFF
--- a/tests/data/cases/preview_parenthesize_whole_conditional_expression.py
+++ b/tests/data/cases/preview_parenthesize_whole_conditional_expression.py
@@ -1,0 +1,94 @@
+# flags: --preview
+
+if isinstance(very_very_very_very_very_very_long_expression, Sequence) and not isinstance(very_very_very_very_very_very_long_expression, str):
+    pass
+
+if first_condition:
+    pass
+elif matches_very_very_very_very_very_very_long_expression(first_value) or matches_very_very_very_very_very_very_long_expression(second_value):
+    pass
+
+while is_ready_very_very_very_very_very_very_long_expression(first_value) and not is_finished_very_very_very_very_very_long_expression(second_value):
+    pass
+
+# Short conditions do not need parentheses.
+if check(first) and check(second):
+    pass
+
+# A condition containing only one call still splits the call.
+if check_very_very_very_very_very_very_very_very_very_very_long_expression(first, second):
+    pass
+
+# Boolean operators nested in a call do not make the condition a boolean expression.
+if check_very_very_very_very_very_very_very_very_very_very_long_expression(first and second):
+    pass
+
+# Boolean operators nested in explicit parentheses do not affect the outer condition.
+if check_very_very_very_very_very_very_very_very_very_long_expression((first and second)):
+    pass
+
+# A trailing comma still explodes the call normally.
+if enabled and check_very_very_very_very_very_very_very_very_long_expression(first, second,):
+    pass
+
+# The split preference is limited to conditional statements.
+result = check_very_very_very_very_very_long_expression(first) and check_very_very_very_very_very_long_expression(second)
+
+# output
+
+if (
+    isinstance(very_very_very_very_very_very_long_expression, Sequence)
+    and not isinstance(very_very_very_very_very_very_long_expression, str)
+):
+    pass
+
+if first_condition:
+    pass
+elif (
+    matches_very_very_very_very_very_very_long_expression(first_value)
+    or matches_very_very_very_very_very_very_long_expression(second_value)
+):
+    pass
+
+while (
+    is_ready_very_very_very_very_very_very_long_expression(first_value)
+    and not is_finished_very_very_very_very_very_long_expression(second_value)
+):
+    pass
+
+# Short conditions do not need parentheses.
+if check(first) and check(second):
+    pass
+
+# A condition containing only one call still splits the call.
+if check_very_very_very_very_very_very_very_very_very_very_long_expression(
+    first, second
+):
+    pass
+
+# Boolean operators nested in a call do not make the condition a boolean expression.
+if check_very_very_very_very_very_very_very_very_very_very_long_expression(
+    first and second
+):
+    pass
+
+# Boolean operators nested in explicit parentheses do not affect the outer condition.
+if check_very_very_very_very_very_very_very_very_very_long_expression(
+    (first and second)
+):
+    pass
+
+# A trailing comma still explodes the call normally.
+if (
+    enabled
+    and check_very_very_very_very_very_very_very_very_long_expression(
+        first,
+        second,
+    )
+):
+    pass
+
+# The split preference is limited to conditional statements.
+result = check_very_very_very_very_very_long_expression(
+    first
+) and check_very_very_very_very_very_long_expression(second)
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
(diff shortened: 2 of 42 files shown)


zhangli091011:
The branch is now updated to the latest `main` (`311f9b399a`). The new fork workflow runs are awaiting maintainer approval, so the previous self-format failures cannot yet be re-evaluated against the updated baseline. Could a maintainer please approve the latest Actions runs?

tests/data/cases/preview_parenthesize_whole_conditional_expression.py
```diff
@@ -0,0 +1,94 @@
...
+# Boolean operators nested in explicit parentheses do not affect the outer condition.
+if check_very_very_very_very_very_very_very_very_very_long_expression((first and second)):
+    pass
+
+# A trailing comma still explodes the call normally.
+if enabled and check_very_very_very_very_very_very_very_very_long_expression(first, second,):
+    pass
+
+# The split preference is limited to conditional statements.
+result = check_very_very_very_very_very_long_expression(first) and check_very_very_very_very_very_long_expression(second)
```
cobaltt7, line 35: IMO it should apply everywhere. Part of the point of a code formatter is consistency. Was there specific reasoning behind this?
Similarly, does this only affect Boolean operators, or does it affect math operators too? I could see that one going either way (don't make this change without discussion and a consensus! 😅)

On the other hand, this does make it even more intrusive than it already is...

cobaltt7 (commented):
Thanks! I personally really like this change, and the diff-shades changes don't seem to include anything bad. However, this is a pretty intrusive change in the sense that it affects a lot of code that would need reformatting, and it's always a net increase in LOC.

As such, Black needs to be run on this repository to fix failing tests!

Would appreciate input from other maintainers!

tests/data/cases/preview_parenthesize_whole_conditional_expression.py
zhangli091011: The current scope is intentional. Issue #3629 is accepted specifically for parenthesizing whole conditional expressions, and its accepted examples are conditional headers. The implementation therefore adds a candidate only for if/elif/while statement conditions, and only when the optional outer parentheses contain top-level and/or delimiters. That gives the requested whole-condition split before the existing RHS call split, while retaining the existing fallback when it does not fit. Applying this everywhere would require a broader policy for expression contexts such as assignments and returns, and extending it to arithmetic would also change operator-precedence and style behavior well beyond the accepted issue. I have not made either semantic expansion without maintainer consensus. If the intended decision is to broaden the accepted scope, I am happy to follow that direction in a subsequent revision, but I think that decision should be explicit first.

zhangli091011:
Updated to current main and ran this PR's Black over the repository as requested. The resulting 15-file self-format change is committed separately, and the piping test now explicitly uses preview to match the repository style. Local self-format, lint, full format fixtures, no-Jupyter, and Jupyter suites pass. I also replied to the open scope thread without expanding semantics pending maintainer consensus. Ready for re-review.

JelleZijlstra:
I like this in many cases but it's a very disruptive change; it would probably require a ton of reformatting for users of Black. I'm also not sure I like it in all cases; sometimes the extra lines make the code harder to see at once.

tests/data/cases/preview_parenthesize_whole_conditional_expression.py
cobaltt7: I agree math operators doesn't make as much sense. But for boolean operators I think it should apply everywhere, if we are to implement this.
Right now it sounds like the main question is whether or not this style is a good change in general, so you can hold off on making that change until we decide to move forward with this PR, and until other maintainers have a chance to discuss whether it should apply outside if/elif/while conditions.

cobaltt7:
I personally find that the cases where it can decrease visibility are mostly changes such as this:
image
(here the trailing comma can be manually removed to condense the second term into one line, which IMO looks best overall)
```py
            if (
                cache is not None
                and write_back not in (WriteBack.DIFF, WriteBack.COLOR_DIFF)
            ):
```

Or cases where one of the terms is just too long and has to stay split anyway:
image

One possible way to handle that is to *not* have Black split before the boolean and retain previous behavior if one of the terms is still split across multiple lines with this style. That could also help slightly decrease the scope.

See also the similar #5259, which also made a similar change to exclude arrays that split across multiple lines, and it significantly decreased scope while significantly increasing readability.

cobaltt7:
@zhangli091011, could you try implementing the change I suggested above?
> One possible way to handle that is to not have Black split before the boolean and retain previous behavior if one of the terms is still split across multiple lines with this style. That could also help slightly decrease the scope.

cobaltt7:
@zhangli091011 Are you still updating this?

zhangli091011:
> 你还在更新这个吗？

I apologize, I am a student from China, and I have been too busy with my studies to update the code recently. I will try to update the code using this method as soon as possible.

zhangli091011:
Thanks for the nudge. I've updated the branch to the latest main and implemented the suggested behavior:

- Black now prefers the whole-condition split only when every top-level boolean operand fits on one line.
- If an operand would split internally, including due to an active magic trailing comma, Black retains the previous split behavior.
- Added regression cases for both an over-long operand and a magic trailing comma.

Local checks: focused fixture passes, the full suite reports 555 passed / 2 skipped, and \lack --check .\ passes. The only local failures are two Windows-specific ANSI color assertions. Pushed as 8116a73 for CI.

PR #5258, opened 2026-07-24
Ignore permission errors when reading cache

cobaltt7:
### Description

Looking in to flakiness on tests, this might be a cause. Specifically with grammar files, since those are read/written from multiple processes, but I included it on other cache files too.

### Checklist - did you ...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -181,7 +181,7 @@ skip = [
 
 test-groups = ["tests"]
 test-command = 'pytest {project} -k "not incompatible_with_mypyc" -n auto'
-test-extras = ["d", " jupyter"]
+test-extras = ["d", "jupyter"]
 # Skip trying to test arm64 builds on Intel Macs. (so cross-compilation doesn't
 # straight up crash)
 test-skip = ["*-macosx_arm64", "*-macosx_universal2:arm64"]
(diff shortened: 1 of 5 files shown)


src/blib2to3/pgen2/driver.py
```diff
@@ -282,27 +294,14 @@ def _newer(a: str, b: str) -> bool:
...
     """
-    if os.path.isfile(grammar_source):
-        gp = _generate_pickle_name(grammar_source, cache_dir) if cache_dir else None
-        return load_grammar(grammar_source, gp=gp)
-    pickled_name = _generate_pickle_name(os.path.basename(grammar_source), cache_dir)
-    data = pkgutil.get_data(package, pickled_name)
-    assert data is not None
-    g = grammar.Grammar()
-    g.loads(data)
-    return g
```
cobaltt7, line 305: This is dead code, since `load_packaged_grammar()` is only ever called with full file paths that are guaranteed to exist (all calls are in pygram.py below). I removed it because there wasn't an obvious way to handle errors in `g.loads()`. I can leave it as-is if desired

pyproject.toml
```diff
@@ -181,7 +181,7 @@ skip = [
 
 test-groups = ["tests"]
 test-command = 'pytest {project} -k "not incompatible_with_mypyc" -n auto'
-test-extras = ["d", " jupyter"]
```
cooperlees, line 184: 😳

cooperlees (approved):
Sounds a sane thing to do to me.

PR #5259, opened 2026-07-25
Format symmetric list concatenations in preview

### LINKED ISSUE #260: List concatenation, multiple lines
Probably not a bug, but I was wondering if the code below shouldn't be formatted more like call chains are.
I did not expect black to change anything in this example

```
-    search_fields = (
-        ['file__%s' % field for field in FileAdmin.search_fields] +
-        ['resource__%s' % field for field in ResourceAdmin.search_fields]
-    )
+    search_fields = ["file__%s" % field for field in FileAdmin.search_fields] + [
+        "resource__%s" % field for field in ResourceAdmin.search_fields
+    ]
```

Operating system: OSX 10.11.6
Python version: 3.6.5
Black version: 18.5b0

ColumbusLabs:
### Description

Closes #260.

When a long expression is exactly two list displays or comprehensions joined by one top-level `+`, preserve its optional parentheses under `--preview` only when each operand fits on its own delimiter-split line. This lets the existing delimiter splitter place each operand on its own line instead of exploding only the right-hand list. Long operands, active magic-trailing-comma layouts, stable style, short expressions, chained concatenations, and mixed operands keep their current behavior.

The patch adds the preview feature flag, native format fixtures, schema entry, future-style documentation, changelog entry, and the required self-format dogfood change.

Validation:
- `env -u NO_COLOR tox -e py` (480 passed, 1 skipped; Jupyter 73 passed)
- focused symmetric-list fixtures (4 passed)
- `tox -e run_self,generate_schema`
- strict Sphinx build with `-W`
- full pre-commit suite
- 16,000-case list/line-length/magic-comma idempotence review matrix
- `git diff --check`
- independent review completed with no remaining findings

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,6 +69,7 @@
 <!-- Changes that affect Black's preview style -->
 
 - Remove redundant parentheses around generator expressions (#5304)
+- Format long concatenations of two list displays symmetrically (#5259)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,7 +69,9 @@
 <!-- Changes that affect Black's preview style -->
 
 - Remove redundant parentheses around generator expressions (#5304)
-- Format long concatenations of two list displays symmetrically (#5259)
+- Format long concatenations of two list displays symmetrically when both operands fit
+  on their own delimiter-split line; retain the existing layout when either operand
+  needs an internal split, including for an active magic trailing comma (#5259)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,9 +69,9 @@
 <!-- Changes that affect Black's preview style -->
 
 - Remove redundant parentheses around generator expressions (#5304)
-- Format long concatenations of two list displays symmetrically when both operands fit
-  on their own delimiter-split line; retain the existing layout when either operand
-  needs an internal split, including for an active magic trailing comma (#5259)
+- Format long binary operations between matching list, tuple, dictionary, or set
+  displays symmetrically when both operands fit on their own delimiter-split line
+  (#5259)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,7 +58,9 @@
 
 <!-- Changes that affect Black's preview style -->
 
-- Format long concatenations of two list displays symmetrically (#5259)
+- Format long concatenations of two list displays symmetrically when both operands fit
+  on their own delimiter-split line; retain the existing layout when either operand
+  needs an internal split, including for an active magic trailing comma (#5259)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,7 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Format long concatenations of two list displays symmetrically (#5259)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,9 +69,8 @@
 <!-- Changes that affect Black's preview style -->
 
 - Remove redundant parentheses around generator expressions (#5304)
-- Format long binary operations between matching list, tuple, dictionary, or set
-  displays symmetrically when both operands fit on their own delimiter-split line
-  (#5259)
+- Format long binary operations between collection displays symmetrically when both
+  operands fit on their own delimiter-split line (#5259)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
--- a/src/black/lines.py
+++ b/src/black/lines.py
@@ -10,16 +10,19 @@ from black.mode import Mode, Preview
 from black.nodes import (
     BRACKETS,
     CLOSING_BRACKETS,
+    MATH_OPERATORS,
     OPENING_BRACKETS,
     STANDALONE_COMMENT,
     TEST_DESCENDANTS,
     child_towards,
     first_leaf,
     is_docstring,
+    is_generator,
     is_import,
     is_multiline_string,
     is_one_sequence_between,
     is_one_tuple,
+    is_tuple,
     is_type_comment,
     is_type_ignore_comment,
     is_with_or_async_with_stmt,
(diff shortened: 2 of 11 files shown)


src/blib2to3/pgen2/tokenize.py
```diff
@@ -62,11 +62,14 @@
...
 from . import token as _token
 
-__all__ = [x for x in dir(_token) if x[0] != "_"] + [
-    "tokenize",
-    "generate_tokens",
-    "untokenize",
-]
+__all__ = (
+    [x for x in dir(_token) if x[0] != "_"]
+    + [
```
cobaltt7: If one of the arrays is split across multiple lines (either because of line length or force-split with magic trailing comma), I think it should be formatted as it was before. This example adds an extra level of nesting and extra lines for minimal symmetry.

Edit: Actually, I'm noticing you already have a test for this, "A body that fits inside optional parentheses keeps the existing bracket split." Somehow this isn't following that

src/black/lines.py
```diff
@@ -1509,6 +1509,42 @@ def _is_annotated_assignment(head: Line) -> bool:
     return False
 
 
+def _is_symmetric_list_concatenation(line: Line) -> bool:
+    """Is `line` exactly two list displays joined by a top-level `+`?"""
```
cobaltt7: 3 lists are already formatted symmetrically, correct? Is it then necessary to search for specifically 2, rather than just 2+? Would changing this to 2+ simplify the code?

src/black/lines.py
```diff
@@ -1509,6 +1509,42 @@ def _is_annotated_assignment(head: Line) -> bool:
...
+    if len(line.bracket_tracker.delimiters) != 1:
+        return False
+
+    delimiter_id = next(iter(line.bracket_tracker.delimiters))
```
cobaltt7, line 1533: Is there a functional or performant difference versus this?
```suggestion
    delimiter_id = line.bracket_tracker.delimiters[0]
```

src/blib2to3/pgen2/tokenize.py
cobaltt7: Another example from shades, where the first array is the long one
image

cobaltt7: An even clear downgrade where both arrays are too long
image

cobaltt7 (commented):
Similar to #5252, I really like this change and I think it looks cleaner, but it's a net LOC increase. Would appreciate input from other maintainers!

On the other hand, it looks like once the multiple lines issue is fixed, this won't end up affecting much already written code, because the second array that's currently broken across multiple lines has a trailing comma and won't be condensed. Hence in practice this only affects people running with `-C`, list comprehensions, and unformatted code. With the decreased scope it makes it easier to merge quickly

src/blib2to3/pgen2/tokenize.py
ColumbusLabs: Good catch. I narrowed the feature so it only retains the outer parentheses when both operands fit on their delimiter-split lines. Long operands and lists forced open by active magic trailing commas now keep the previous asymmetric formatting. I added fixtures for long left/right/both operands and both trailing-comma orientations.

src/black/lines.py
ColumbusLabs: Three-or-more-list chains already receive the normal symmetric delimiter split. This helper intentionally handles exactly one top-level delimiter, so broadening it to 2+ would add scope rather than simplify the implementation.

src/black/lines.py
ColumbusLabs: `BracketTracker.delimiters` is keyed by leaf ID rather than position, so `[0]` would normally raise `KeyError`. I added a comment clarifying why `next(iter(...))` is used.

src/black/lines.py
cobaltt7: If 3+ list chains already follow this style, why would broadening the function add scope? I don't imagine it would format them differently from how it has been?

ColumbusLabs: You are right: chained list concatenations already keep the optional parentheses via the normal delimiter splitter, and the regression fixture covers that unchanged formatting. The special case is only needed for exactly two operands, where omitting those parentheses would otherwise expand one list. Making this predicate 2+ would not improve the chained output, but would broaden a guard that does not need to participate there. I rebased the branch onto current main and kept the narrowly scoped two-operand condition.

tests/data/cases/preview_symmetric_list_concatenation.py
```diff
@@ -0,0 +1,121 @@
...
+    ["first_long_value", "second_long_value", "third_long_value"]  # first list
+    + ["fourth_long_value", "fifth_long_value", "sixth_long_value"]
+)
+
+# A body that fits inside optional parentheses keeps the existing bracket split.
+values = [first_value, second_value, third_value] + [
+    fourth_value,
+    fifth_value,
+    sixth_value,
+]
```
cobaltt7: Why is this example not formatted like this?
```suggestion
values = (
    [first_value, second_value, third_value]
    + [fourth_value, fifth_value, sixth_value]
)
```

CHANGES.md
```diff
@@ -58,6 +58,9 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Format long concatenations of two list displays symmetrically when both operands fit
+  on their own delimiter-split line; retain the existing layout when either operand
+  needs an internal split, including for an active magic trailing comma (#5259)
```
```suggestion
- Format long concatenations of two lists symmetrically when both operands fit on their
  own delimiter-split line (#5259)
```
Shorten it a bit

cobaltt7:
```py
values = (
    ["first_value", "second_value", "third_value"]  # abc
    + ["fourth_value", "fifth_value", "sixth_value"]
)

values = (
    ["first_value", "second_value", "third_value"]
    + ["fourth_value", "fifth_value", "sixth_value"]  # abc
)
```
Comments at the end of the line should not restrict splitting

tests/data/cases/preview_symmetric_list_concatenation.py
ColumbusLabs: Updated in 12b570b: the fixture now uses the suggested symmetric layout. I also added left and right end-of-line comment cases; the focused formatter suite passes 235 tests.

CHANGES.md
ColumbusLabs: Updated in 12b570b with the suggested concise changelog wording.

cobaltt7 (approved):

JelleZijlstra (commented):
There are closely analogous code patterns with tuples (`+`), dicts (`|`), and sets (`|` or `&`). Are those formatted consistently with this change?

cobaltt7 (approved):

src/black/lines.py
```diff
@@ -1509,6 +1509,75 @@ def _is_annotated_assignment(head: Line) -> bool:
...
+        return False
+
+    first = line.leaves[0]
+    left_closing = line.leaves[left_closing_index]
+    delimiter = line.leaves[delimiter_index]
+    right_opening = line.leaves[delimiter_index + 1]
+    last = line.leaves[-1]
+
+    display_operation = (
+        (token.LSQB, token.RSQB, token.PLUS),
```
JelleZijlstra: I think we don't need to really check the exact binary operation between them, we should just do this for every binop with two collections operands.

ColumbusLabs: Addressed in d65ce98: this now applies to every arithmetic/bitwise Python BinOp between collection displays, including mixed collection kinds. The regression fixture covers every supported operator plus scalar and generator exclusions; `pytest tests/test_format.py` passed (236 tests) locally. Fresh GitHub CI is pending.

tests/data/cases/preview_symmetric_list_concatenation.py
```diff
@@ -12,6 +12,24 @@
 set_union = {"first_long_value", "second_long_value"} | {"third_long_value", "fourth_long_value"}
 set_intersection = {"first_long_value", "second_long_value"} & {"third_long_value", "fourth_long_value"}
 
+# Every arithmetic and bitwise binary operator gets the same treatment when both operands
```
JelleZijlstra, line 15: Can you check comparison ops too? This is not theoretical, people can write test cases like `assert [a, b] == [c, d]` or `assert [f(x) for x in xs] == [1, 2, 3]`

PR #5261, opened 2026-07-26
Treat Ruff ignore comments as pragmas

### LINKED ISSUE #5260: `ruff:ignore` should be processed as a pragma in `contains_pragma_comment`
**Describe the bug**

**To Reproduce**

For example, take this code:

```python
```

```sh
$ black x.py
```

The resulting output is:

```diff
--- x.py        2026-07-26 00:36:44.849442+00:00
+++ x.py        2026-07-26 00:36:45.724865+00:00
@@ -1 +1,3 @@
-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx = 5  # ruff: ignore[bweh]
+xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx = (
+    5  # ruff: ignore[bweh]
+)
```

**Expected behavior**

**Environment**

- Black's version:  26.5.1

**Additional context**

Baltsat:
## Summary

- recognize canonical `# ruff:ignore[...]` comments as Black pragmas
- preserve original physical lines for Ruff ignores, matching `type: ignore`
- add direct and end-to-end regressions for #5260

Fixes #5260, closes #5266

## Testing

- `pytest tests/test_trans.py tests/test_format.py -k ruff`
- `pre-commit run -a`
- `env -u NO_COLOR tox -e py` — 476 passed, 1 skipped; 73 Jupyter tests passed

## AI-assisted contribution

AI assistance was used for investigation, implementation, testing, and review. I reviewed the final diff and verified all reported commands locally.

## Checklist

- [x] The change includes regression coverage.
- [x] Add the PR-numbered entry to `CHANGES.md`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Preserve physical lines ending in canonical Ruff ignore comments such as
+  `# ruff:ignore[...]`, matching existing pragma handling (#5261)
 - Fix dropping the required trailing comma from a single-element tuple used as a lambda
   parameter default under `--skip-magic-trailing-comma` when a standalone comment forces
   the tuple across multiple lines; removing the comma turned the tuple into a bare
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -13,6 +13,7 @@ from black.nodes import (
     WHITESPACE,
     container_of,
     first_leaf_of,
+    is_ruff_ignore_comment_string,
     is_type_comment_string,
     make_simple_prefix,
     preceding_leaf,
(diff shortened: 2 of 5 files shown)


Baltsat:
The current corpus run shows that the unconditional implementation changes stable output in 134 files. The issue spans two formatter paths: line-preservation via `is_type_ignore_comment` and string-transform protection via `contains_pragma_comment`. I can rework both behind a single `--preview` feature, requiring zero stable diff-shades while retaining focused stable/preview regressions. Is that the preferred scope? I have marked the PR draft while this is clarified.

cobaltt7:

PR #5262, opened 2026-07-26
don't re-escape an already escaped quote in normalize_string_quotes

sahvx655-wq:
### Description

`normalize_string_quotes` escapes a trailing `"` in the body so that it is not glued to
the closing `"""`, but it never checks whether that quote is already escaped. For
`x = '''\'''\"'''` (valid Python, value `'''"`) the handler drops the quote and appends
`\"`, so the backslash that was escaping the quote ends up escaping a backslash and the
quote is left bare; the literal it emits, `"""'''\\""""`, terminates early and strands a
`"`. Black's forced second pass then re-parses its own output and fails with
`InvalidInput: Cannot parse: 1:4 ... TokenError: Failed to parse: UnterminatedString`. I
hit this while fuzzing single string tokens through the normaliser, and enumerating
short triple-quoted bodies turns up 42 distinct shapes that reach it; it reproduces on
26.1.0. Left alone, one such literal makes a whole file unformattable, `blackd` answers
400 for valid input, and the parse error printed points at a line nobody wrote.

The check belongs in the escaping helper rather than at the call sites, since every
caller hands it a raw literal and has no way to know whether the final quote already
carries a backslash. `_ends_with_unescaped_quote` counts the backslash run in front of
the quote and only escapes on an even count, which also leaves `'''\\"'''` alone: there
the backslashes escape each other, the quote really is bare, and escaping it would add
an escape. `normalize_fstring_quotes` repeats the same edge case so it shares the helper
now; that path is unreachable today because its call site is commented out, but the
defect is identical and leaving half of it fixed seemed worse. On behaviour: comparing
the old and the new function over 2.89M string literals from `tests/data`, Black's own
sources and the CPython stdlib gives byte-identical output, so nothing that formats
today moves. The single delta is that a body ending in an already escaped quote now
converts to `"""` instead of being corrupted or declined, because the spurious backslash
used to inflate `new_escape_count`. I read that as part of the same defect rather than a
style change, but say the word if you would rather see it preview-gated.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,10 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Fix unparseable output for a triple-quoted string whose body ends in an already
+  escaped double quote (for example `'''\'''\"'''`). Switching to `"""` escaped the
+  backslash instead of the quote, leaving the closing quotes bare, so Black failed on
+  its own output (#5262)
 - Fix dropping the required trailing comma from a single-element tuple used as a lambda
   parameter default under `--skip-magic-trailing-comma` when a standalone comment forces
   the tuple across multiple lines; removing the comma turned the tuple into a bare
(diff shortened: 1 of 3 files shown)


cobaltt7:
Thanks!

PR #5272, opened 2026-07-29
Keep the required trailing comma of a one-element subscript under --skip-magic-trailing-comma

LuShadowX:
### Description

`--skip-magic-trailing-comma` drops the trailing comma from a one-element subscript when the line is long enough to be split and contains a power operator, which turns `a[x,]` (indexing with a one-tuple) into `a[x]` and trips Black's own AST safety check:

```py
value = alpha[beta,]() ** gamma * delta << epsilon | zeta < eta ^ theta * iota + kappa + mu
```

```console
$ black --skip-magic-trailing-comma file.py
error: cannot format file.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.
```

The comma survives on a short line — `tests/data/cases/skip_magic_trailing_comma.py` opens by asserting exactly that — so this is the same rule failing on a longer one. Under `--fast` there is no safety net and the changed code is written out.

`hug_power_op` is the path. It rebuilds the line with `leaf.clone()` for every leaf, and a clone has no parent, so both tree-based guards that protect a required comma stop working: `has_magic_trailing_comma`'s single-element-subscript branch needs `closing.parent.type == syms.trailer`, and the one-tuple guard in `Line.append` needs `leaf.parent`. Neither can fire, the comma looks magic, and `remove_trailing_comma()` takes it.

Only the operands either side of a hugged `**` actually need their prefix rewritten, so this stops cloning the rest and reuses the leaf, keeping it attached to the tree. Reusing leaves across lines is what `bracket_split_build_line` already does, and with `preformatted=True` `Line.append` does not touch the prefix — it only re-runs `BracketTracker.mark`, which recomputes the same metadata for an identical leaf sequence.

The one-tuple guard the same clone breaks (`foo(lambda x=(1,): x)`, already in that case file) is fixed by this too; I could not build a failing example for it because the lambda cases are short enough not to reach `hug_power_op`.

Found with a randomised run of `scripts/fuzz.py`'s property — the committed harness pins `derandomize=True, max_examples=1000`, so it re-explores the same thousand inputs on every CI run and never reaches this one.

### Verification

- Case added to `tests/data/cases/skip_magic_trailing_comma.py`. It fails on `main` (`alpha[beta]()` in the output) and passes with the change.
- `pytest`: 474 passed, 3 skipped, unchanged from `main`. `black --check .`, `flake8`, `isort`, `mypy` clean.
- Formatted 716 files from scrapy, pydantic, Pillow and fontTools with `--skip-magic-trailing-comma` before and after, comparing output hashes: **zero differences**, no errors either way. Power-operator hugging is unaffected (`x = a ** b` still becomes `x = a**b`).

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? — no style change; this restores the documented behaviour for a case that was crashing
- [ ] Add new / update outdated documentation? — no documented behaviour changes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,11 @@
   expression and failed Black's AST safety check (#5246)
 - Fix unstable formatting when an inline comment sits on optional parentheses (for
   example a parenthesized assert message or assignment RHS) (#5241)
+- Fix `--skip-magic-trailing-comma` dropping the syntactically required trailing comma
+  of a one-element subscript (`a[x,]`) when the line is long enough to be split and
+  contains a power operator. `hug_power_op` rebuilt the line from cloned leaves, whose
+  missing parent kept the guards for required commas from firing, so the index stopped
+  being a one-tuple and Black failed its own AST safety check (#5272)
 - Fix crash when a standalone comment sits between tokens of a comprehension or lambda
   (#5144)
 - Respect the magic trailing comma in a PEP 695 type parameter list containing a
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,11 +31,9 @@
   expression and failed Black's AST safety check (#5246)
 - Fix unstable formatting when an inline comment sits on optional parentheses (for
   example a parenthesized assert message or assignment RHS) (#5241)
-- Fix `--skip-magic-trailing-comma` dropping the syntactically required trailing comma
-  of a one-element subscript (`a[x,]`) when the line is long enough to be split and
-  contains a power operator. `hug_power_op` rebuilt the line from cloned leaves, whose
-  missing parent kept the guards for required commas from firing, so the index stopped
-  being a one-tuple and Black failed its own AST safety check (#5272)
+- Fix `--skip-magic-trailing-comma` dropping the trailing comma of a one-element
+  subscript (`a[x,]`) when the line is long enough to be split and contains a power
+  operator (#5272)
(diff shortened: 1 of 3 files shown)


LuShadowX:
One more data point: 6000 randomised examples through `scripts/fuzz.py`'s property on top of this branch — the same run that produced the failing input on `main` — come back clean.

CHANGES.md
```diff
@@ -31,6 +31,11 @@
   expression and failed Black's AST safety check (#5246)
 - Fix unstable formatting when an inline comment sits on optional parentheses (for
   example a parenthesized assert message or assignment RHS) (#5241)
+- Fix `--skip-magic-trailing-comma` dropping the syntactically required trailing comma
+  of a one-element subscript (`a[x,]`) when the line is long enough to be split and
+  contains a power operator. `hug_power_op` rebuilt the line from cloned leaves, whose
+  missing parent kept the guards for required commas from firing, so the index stopped
+  being a one-tuple and Black failed its own AST safety check (#5272)
```
```suggestion
- Fix `--skip-magic-trailing-comma` dropping the trailing comma of a one-element
  subscript (`a[x,]`) when the line is long enough to be split and contains a power
  operator (#5272)
```

LuShadowX: Thanks, applied — you had already committed it, so nothing further from me here.

LuShadowX:
Thanks for the review and merge, @cobaltt7.

PR #5276, opened 2026-07-31
Preserve ruff ignore pragmas when splitting lines

### LINKED ISSUE #5260: `ruff:ignore` should be processed as a pragma in `contains_pragma_comment`

happykawayigt:
### Description

Fixes #5260.

Ruff's line-scoped `ruff:ignore` pragmas could be moved onto a nested expression when Black split an overlong line, changing which diagnostic the pragma suppresses. This change recognizes Ruff directives as pragmas and treats `ruff:ignore` comments like `type: ignore` comments when deciding whether a physical line can be split.

Both `# ruff:ignore[...]` and `# ruff: ignore[...]` forms are covered by regression tests.

Validation:

- `pytest tests/test_format.py -q` (232 passed)
- pre-commit on changed files: isort, flake8, mypy, EOF, trailing whitespace passed
- `git diff --check`

AI assistance was used during implementation. I independently reproduced the formatting change, reviewed the final diff, and ran the validation above.

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? (No style change.)
- [x] Add new / update outdated documentation? (No documentation change is needed.)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Preserve line-scoped `ruff:ignore` pragmas instead of moving them into nested
+  expressions when splitting long lines (#5276)
 - Fix unparseable output for a t-string whose replacement field contains a quote (for
   example `t'\'{a["b"]}\''`). The guards that keep quote normalisation away from the
   inside of an f-string replacement field were never reached for t-strings, so the
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -907,7 +907,7 @@ def contains_pragma_comment(comment_list: list[Leaf]) -> bool:
         pylint).
     """
     for comment in comment_list:
-        if comment.value.startswith(("# type:", "# noqa", "# pylint:")):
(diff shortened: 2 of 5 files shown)


cobaltt7:

PR #5288, opened 2026-08-02
docs: mention Vim PEP 8 indentation plugin

### LINKED ISSUE #3151: vim plugin: provide black-compatible `indentexpr`

tomatotomata:
## Summary

- document `vim-python-pep8-indent` as an option for Black-style insert-mode indentation
- clarify that the Vim plugin provides `indentexpr` while Black remains the formatter
- include a Vundle installation example

## Verification

- `git diff --check`
- documentation-only change; no runtime code paths changed

Closes #3151

### DIFF
--- a/docs/integrations/editors.md
+++ b/docs/integrations/editors.md
@@ -353,11 +353,11 @@ nnoremap <F9> :Black<CR>
 
 ### Vim indentation plugin
 
-For insert-mode indentation that follows the same hanging-bracket style as _Black_,
-you can also use the
-[`vim-python-pep8-indent`](https://github.com/Vimjas/vim-python-pep8-indent) plugin.
-It provides Vim's `indentexpr`; _Black_ remains responsible for formatting the file.
-For example, with [vim-plug](https://junegunn.github.io/vim-plug/):
+For insert-mode indentation that follows the same hanging-bracket style as _Black_, you
+can also use the
+[`vim-python-pep8-indent`](https://github.com/Vimjas/vim-python-pep8-indent) plugin. It
+provides Vim's `indentexpr`; _Black_ remains responsible for formatting the file. For
+example, with [vim-plug](https://junegunn.github.io/vim-plug/):
 
 ```vim
 Plug 'Vimjas/vim-python-pep8-indent'
--- a/docs/integrations/editors.md
+++ b/docs/integrations/editors.md
@@ -210,6 +210,16 @@ global setting, follow these steps:
 
 ### Official plugin
 
+For insert-mode indentation that follows the same hanging-bracket style as _Black_,
+you can also use the
+[`vim-python-pep8-indent`](https://github.com/Vimjas/vim-python-pep8-indent) plugin.
+It provides Vim's `indentexpr`; _Black_ remains responsible for formatting the file.
+For example, with Vundle:
+
+```vim
+Plugin 'Vimjas/vim-python-pep8-indent'
+```
+
 Commands and shortcuts:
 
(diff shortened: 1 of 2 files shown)


cobaltt7:
Thanks! I don't think this is an appropriate location in the docs for this, it's right in the middle of the docs for the official Black plugin. Could you move it down towards the bottom of the file and make it its own section?

tomatotomata:
I moved the Vim indentation plugin into its own section at the bottom of the editor guide, so the official Black plugin instructions stay together. The branch is updated at `18b5e13`; `git diff --check` is clean.

docs/integrations/editors.md
```diff
@@ -210,8 +210,6 @@ global setting, follow these steps:
 
 ### Official plugin
 
-Commands and shortcuts:
```
cobaltt7, line 213: Shouldn't have been removed

cobaltt7:
Sorry, I meant a new subsection within the Vim section. Could you also add an example for vim-plug? Thanks!

tomatotomata:
I moved the indentation plugin into its own subsection inside the Vim section and added both vim-plug and Vundle examples. The change is in `f2bdb08`, and the diff is whitespace-clean.

Let me know what you think.

tomatotomata:
I restored the “Commands and shortcuts” lead-in under the official Vim plugin section in `83da43a`. The indentation plugin stays in its own subsection, with both vim-plug and Vundle examples. `git diff --check` is clean. Let me know what you think.

docs/integrations/editors.md
```diff
@@ -353,6 +353,24 @@ nnoremap <F9> :Black<CR>
    let g:ale_fixers.python = ['black']
    ```
 
+### Vim indentation plugin
+
+For insert-mode indentation that follows the same hanging-bracket style as _Black_, you
+can also use the
+[`vim-python-pep8-indent`](https://github.com/Vimjas/vim-python-pep8-indent) plugin. It
+provides Vim's `indentexpr`; _Black_ remains responsible for formatting the file. For
+example, with [vim-plug](https://junegunn.github.io/vim-plug/):
```
cobaltt7:
```suggestion
example, with vim-plug:
```
The link is unnecessary since we already link in the vim-plug section above

PR #5296, opened 2026-08-06
Refactor release process to be more automated and use immutable releases

### LINKED ISSUE #5120: Please consider making releases immutable
Particularly because this repository includes a github action. This will improve security of the ecosystem.

Docs: https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases

Thank you!

cobaltt7:
### Description

Resolves #5120

Enabling immutable releases adds a few more steps to the release process. Instead of making it more complicated for releasers, I took the opportunity to automate most of the steps within GitHub Actions, hopefully simplifying the process.

From the new docs:
> The 10,000-foot view of the release process is that you trigger a workflow to create a PR automating the release chores. Then, merge it, triggering release automation that builds all release artifacts and publishes them to the various platforms we publish to.

The hope is that for most non-major releases there's nothing the releaser has to do other than trigger the workflow, merge the PR, and make sure everything's green.
For major releases and others that require some cleanups, other changes, or prereleases cut before merge, the releaser is able to check out the autocreated "prepare release" branch and do all necessary work directly on the branch.

I also included some other changes that aren't directly related, but help streamline the release process.

I don't know that it's still neccessary to include the `check-pre-commit-rev-in-example` pre-commit hook, since it's done automatically and should be impossible to skip with the new release process. But it's also not super hard to maintain and doesn't hurt too much to keep it.
Similarly, I don't know that it's necessary to have `release_tests.yml`, it feels a bit odd to me to have tests for only one of the scripts we have & use regularly, but it similarly doesn't have much harm to keep beyond maintainance.

Another maintainer will have to enable immutable releases within GitHub settings after this is merged and before the next release.

### Checklist - did you ...


.github/workflows/test.yml
```diff
@@ -29,13 +30,6 @@ concurrency:
 
 jobs:
   test:
-    # We want to run on external PRs, but not on our own internal PRs as they'll be run
-    # by the push to the branch. Without this if check, checks are duplicated since
-    # internal PRs match both the push and pull_request events.
-    if:
-      github.event_name == 'push' || github.event.pull_request.head.repo.full_name !=
-      github.repository
-
```
cobaltt7, line 38: This comment is correct in theory. However, there's some times when it's not always triggered on internal PRs.

If a commit to an internal PR modifies source files, CI will run on that commit (and not on the PR due to this check). If another commit is added that only modifies CHANGES.md, for example, this action doesn't run on the commit nor on the PR, because CHANGES.md is not captured by the `paths:` filter. This requires the reviewer to check the CI status on all the commits, and not just the most recent. Removing the check ensures that all the desired checks run on the PR. This does unfortunately mean that checks can be duplicated on internal PRs, but prevents checks being missed.

An alternative fix would be removing the `push.paths` filter (the `pull_requests.paths` filter can stay), but that means unnecessary checks run on some pushes to main, for instance tests would be run on docs commits. IMO I think there's less internal PRs, so the solution I implemented would affect less.

This change is relevant to this PR because it's important to make sure tests (and other CI) passes on the release PR. I added a `pull_request.paths` entry to make sure it is, but with the check it wouldn't run. Even if I also added it to `push.paths`, it wouldn't be rerun if someone committed to CHANGES.md.

I kept this check on actions without `push.paths`.

.github/workflows/pypi_upload.yml
```diff
@@ -39,14 +39,14 @@ jobs:
...
-            | pyp 'json.dumps({"only": x, "os": "windows-11-arm"})'
-          } | pyp 'json.dumps(list(map(json.loads, lines)))' > /tmp/matrix
+              | jq -Rn '[{"only": inputs, "os": "ubuntu-latest"}]'
+            cibuildwheel --print-build-identifiers --platform macos \
+              | jq -Rn '[{"only": inputs, "os": "macos-latest"}]'
+            cibuildwheel --print-build-identifiers --platform windows \
+              | jq -Rn '[{"only": inputs, "os": "windows-latest"}]'
+            cibuildwheel --print-build-identifiers --platform windows --archs ARM64 \
+              | jq -Rn '[{"only": inputs, "os": "windows-11-arm"}]'
+          } | jq -cs 'add' > /tmp/matrix
```
cobaltt7, line 49: This refactor removes the dependency on pypyp

docs/contributing/release_process.md
```diff
@@ -1,111 +1,190 @@
...
+of the initial chores before releasing.
+
+#### create
+
+This job runs `scripts/release.py` to clean up CHANGES.md and bump most references to
+old _Black_ versions, then creates a PR with the changes for your review. The PR
+creation is authenticated in order to trigger test CI.
+
+```{note}
+_Currently this workflow uses a GitHub API Token associated with @TODO's account._
```
cobaltt7, line 132: This API token is used to create the "prepare release" and "update changelog" PRs, and to change the GitHub Release's status from "draft" to "published". Since it's not used to create the release and only to publish it, the release shows up as "by github-actions", not by this user. The only place where this user will show up is as the author of the release PRs and associated commits.

Like it says, it's TBD whose token this will be.

.github/workflows/pypi_upload.yml
```diff
@@ -57,9 +57,11 @@ jobs:
...
         run: |
           {
-            CIBW_BUILD="cp310-*" cibuildwheel --print-build-identifiers --platform linux | pyp 'json.dumps({"only": x, "os": "ubuntu-latest"})'
-            CIBW_BUILD="cp314-*" cibuildwheel --print-build-identifiers --platform windows | pyp 'json.dumps({"only": x, "os": "windows-latest"})'
-          } | pyp 'json.dumps(list(map(json.loads, lines)))' > /tmp/matrix
+            CIBW_BUILD="cp310-*" cibuildwheel --print-build-identifiers --platform linux \
+              | jq -Rn '[{"only": inputs, "os": "ubuntu-latest"}]'
+            CIBW_BUILD="cp314-*" cibuildwheel --print-build-identifiers --platform windows \
+              | jq -Rn '[{"only": inputs, "os": "windows-latest"}]'
+          } | jq -cs 'add' > /tmp/matrix
```
cobaltt7, line 64: These version/OS combos run on every PR seem somewhat arbitrary, but I'm not sure if there's a better selection to use that stays small but covers a wide range. Maybe one more for MacOS?

scripts/release.py
```diff
@@ -14,6 +14,8 @@
 from pathlib import Path
 from subprocess import run
 
+from packaging.version import InvalidVersion, Version
```
cobaltt7, line 17: I tried to improve this without vendoring `packaging`, but I decided it'd be easier to vendor it just to cover all the edge cases.

.github/workflows/cut_release.yml
```diff
@@ -0,0 +1,223 @@
...
+          if [ "$REF" == "refs/heads/main" ]; then
+            echo "version=$version" >> "$GITHUB_OUTPUT"
+          else
+            echo "version=$(python scripts/release.py prerelease $version)" >> "$GITHUB_OUTPUT"
+          fi
+        env:
+          HEAD: ${{ github.head_ref || github.ref_name }}
+          REF: ${{ github.ref }}
+
+      - name: Delete previous release if exists
```
JelleZijlstra, line 55: Should we do this? It goes against making releases immutable. I'd just increment the version if a release ends up incomplete.

.github/workflows/pypi_upload.yml
JelleZijlstra: I think this is fine. We should have some cross-version/OS coverage, but Black doesn't have a ton of OS-dependent code so running the full matrix is wasteful.

docs/contributing/release_process.md
```diff
@@ -1,111 +1,190 @@
...
+     in the box provided (not the version of the prerelease). The prerelease version
+     will be determined automatically, following these rules:
+     - Prereleases use the `YY.MMaN` format.
+     - The first alpha should start `N` at `1` (not `0`).
+     - Example: the first alpha released in December 2025 and to be stabilized in
+       January 2026 is `26.1a1`, but `Version_Override` should be set to `26.1.0`.
+
+1. You will be auto-assigned to the created PR. **Check the changelog diff in the
+   description and copyedit `CHANGES.md`.** The CI will update it after each push, so
+   double-check it again before you merge. Things to check include:
```
JelleZijlstra, line 60: A repeated issue has been that some changelog entries end up in the wrong month (because on rebase they are targeted wrongly). I've dealt with this in the past by running `git diff <last tag>` or similar. It occurs to me this could also be done automatically by the release script.

scripts/release.py
JelleZijlstra: Is it vendored? I don't see it in the PR.

Maybe this should be a PEP 723 script ("scroll") and we run it with `uv run`.

.github/workflows/cut_release.yml
cobaltt7: At this point the release would still be an unpublished draft release, only visible to maintainers. GitHub doesn't let you delete published immutable releases.

docs/contributing/release_process.md
cobaltt7: Forgot to mention that but yes, `git diff stable` is run by the script. The output is included in the "prepare release" PR description and updated on each commit.

scripts/release.py
cobaltt7: Sorry, vendored might've been the wrong word. I added it as a new dependency group:
https://github.com/cobaltt7/black/blob/9a508f1ff88108501fa7793c7db4bd0d697fda03/pyproject.toml#L106
https://github.com/cobaltt7/black/blob/9a508f1ff88108501fa7793c7db4bd0d697fda03/.github/workflows/prepare_release.yml#L40

I'll look in to PEP 723 scripts.

.github/workflows/pypi_upload.yml
cobaltt7: For sure. Do you think adding a MacOS build would be beneficial, or no?

scripts/release.py
JelleZijlstra: Ah if it's in a dependency group that's fine, no need to bother with PEP 723 if what you have now works already.

.github/workflows/pypi_upload.yml
JelleZijlstra: Probably yes

Issue #2877, opened 2022-02-10
Black removes empty lines in fmt:off sections


amyreese:
**Describe the bug**

For files with a `#fmt:off` blocks, black will still remove empty newlines in these otherwise-unformatted sections. This is kind of pedantic, but it defies expectations that black would change anything at all here.

**To Reproduce**

Given the source file (note especially the trailing newlines at the end):

```python
something = [1,2,3]


# fmt:off
other = [1,2,3
        ]



# fmt:on
another = [1,2,3]



# fmt:off
last = [1,2,3
        ]


```

And run it with these arguments:

```sh
$ black --diff foo.py
--- foo.py      2022-02-10 21:51:33.647821 +0000
+++ foo.py      2022-02-10 21:54:13.748438 +0000
@@ -1,19 +1,15 @@
-something = [1,2,3]
+something = [1, 2, 3]


 # fmt:off
 other = [1,2,3
         ]


-
 # fmt:on
-another = [1,2,3]
-
+another = [1, 2, 3]


 # fmt:off
 last = [1,2,3
         ]
-
-
would reformat foo.py
All done! ✨ 🍰 ✨
1 file would be reformatted.
```

**Expected behavior**

Black should leave everything as-is inside `#fmt:off` blocks, including whitespace and newlines.

**Environment**

- Black's version:
```
(venv39) jreese@butterfree ~/scratch » black --version
black.macos, version 21.4b2
```
- OS and Python version: macOS 12.1, Python 3.9.3

cooperlees:
This is a bug and we should fix, but, just out of curiosity, what do the new lines at the end help with here? 

I don't think it would happen if they were not at the end of the file.

jreese:
We noticed this when generated files with a `#fmt:off` comment were committed and triggered daily formatting changes. The diff above shows that it also happens in sections that aren't at the end.

alliasgher:
Opened #5300 for the mid-file half of this. The end-of-file half is a different mechanism and is already tracked in #496, so I left it out rather than bundle two fixes.

cobaltt7:
Fixed in #5300, see #496 for EOF newlines

Issue #496, opened 2018-09-03
empty lines at end of file trimmed even with fmt: off


matejcik:
Operating system: Linux Mint
Python version: 3.6.6
*Black* version: latest
Does also happen on master: yes

The following file:
```
# fmt: off

print("hello")

```
gets trimmed to:
```
# fmt: off

print("hello")
```

IOW, if the file ends with empty lines, they are removed.

zsol:
Sounds like a bug. Want to take a look at fixing it?

scolby33:
The problem here appears to happen in `convert_one_fmt_off_pair`, where the node enters as `# fmt: off\n\nprint("hello")\n\n` and exits as `# fmt: off\n\nprint("hello")\n`. I am continuing to investigate.

scolby33:
The function `generate_ignored_nodes` creates a list of nodes that is just `# fmt: off\n\nprint("hello")\n`. This is where the final newline is lost.

Unfortunately, I've reached the  end of my experience/time to look at this today. I hope this has given someone more versed in the project a better starting place for fixing this bug.

felix-hilden:
At least our documentation states that `fmt: off` should be followed by a `fmt: on`, but supporting a single off switch is a reasonable request.

tagadearpit:
I reproduced this on the current development branch (26.5.2.dev98+g8947c48ef, CPython 3.12.3).

Minimal input:

```python
# fmt: off

print("hello")

```

Running black file.py removes the final blank line and produces:

```python
# fmt: off

print("hello")
```

The issue is that the parser stores trailing blank lines after the final statement in the END marker prefix, while fmt-off conversion currently discards that prefix. I have a focused fix that preserves the END marker prefix for an unclosed # fmt: off block at EOF, together with a regression test through Black’s file-formatting path.

PR #5300, opened 2026-08-08
Preserve blank lines before a closing fmt: on comment

alliasgher:
### Description

Blank lines immediately before a closing `# fmt: on` are collapsed to the usual maximum, even though they are inside the opted-out region:

```python
x = 1
# fmt: off
a = 1
                 # <- three blank lines
# fmt: on
y = 2
```

comes back with two. Reproduced on current `main` (74371e2).

`convert_one_fmt_off_pair` → `_handle_regular_fmt_block` collapses the region into a single `STANDALONE_COMMENT` whose `hidden_value` is verbatim, which is why blanks *between* statements survive. But the blank run that *terminates* the region is not part of that value — it lives in the prefix of the following `# fmt: on` comment, which is owned by the normal blank-line machinery. So `_maybe_empty_lines` applies `before = min(before, max_allowed)` to it like any ordinary blank run.

I confirmed the mechanism by instrumenting `EmptyLineTracker._maybe_empty_lines` rather than reading it off: for the input above the `# fmt: on` line logs `raw_nl=4 -> before=2`.

The loss is one-shot — the output is stable on a second pass — which matches the report of a fresh diff every time the file is regenerated.

### Scope

The issue bundles two defects and this takes only the first. The second symptom, blank lines lost at end of file, is a different mechanism (end-of-file newline normalisation, in `generate_ignored_nodes`) and is separately tracked as still-open #496. `tests/util.read_data` also strips the final newline, so it cannot be covered by a `tests/data/cases` fixture as-is. I did not want to bury an unrelated fix in this PR.

Fixes the mid-file half of #2877.

### Checklist

- [x] Added a test case (`tests/data/cases/fmtonoff9.py`, no `# output` section, so it asserts the input is left untouched)
- [x] Added entry to `CHANGES.md`

### Testing

I checked the fixture is not vacuous: with `src/black/lines.py` reverted, `test_simple_format[fmtonoff9]` fails; with it applied, it passes.

Covered in the fixture: three and four blank lines, nested inside a `def`, `# yapf: disable`/`# yapf: enable`, and the no-space `# fmt:off`/`# fmt:on` spelling. I also checked separately that a bare `# fmt: on` with no preceding `# fmt: off` is still capped normally — those blanks are not in an opted-out region, so they should be.

Full suite: 478 passed, 3 skipped. No existing fixture depended on the old behaviour (7 `fmtonoff*`, 3 `line_ranges_fmt_off*`, 20+ `fmtskip*` all unchanged). `black --check` on the two changed files is clean.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,10 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Preserve blank lines that come immediately before a `# fmt: on` comment. They are
+  inside the opted-out region, but they live in the prefix of the `# fmt: on` comment
+  rather than in the verbatim block, so they were being capped like any ordinary blank
+  run (#5300)
 - Stop treating a t-string in docstring position as a docstring (for example
   `t"  spam  "` as the first statement of a module, class or function). t-strings
   evaluate to `Template`, never `str`, so stripping and reindenting one changed the
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,10 +17,7 @@
 
 <!-- Changes that affect Black's stable style -->
 
-- Preserve blank lines that come immediately before a `# fmt: on` comment. They are
-  inside the opted-out region, but they live in the prefix of the `# fmt: on` comment
-  rather than in the verbatim block, so they were being capped like any ordinary blank
-  run (#5300)
+- Preserve blank lines that come immediately before a `# fmt: on` comment (#5300)
 - Stop treating a t-string in docstring position as a docstring (for example
   `t"  spam  "` as the first statement of a module, class or function). t-strings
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -17,6 +17,10 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Preserve blank lines that come immediately before a `# fmt: on` comment. They are
+  inside the opted-out region, but they live in the prefix of the `# fmt: on` comment
+  rather than in the verbatim block, so they were being capped like any ordinary blank
+  run (#5300)
```
```suggestion
- Preserve blank lines that come immediately before a `# fmt: on` comment (#5300)
```

cobaltt7:
Thanks!

PR #5304, opened 2026-08-12
Remove redundant generator parentheses

### LINKED ISSUE #2943: Remove redundant parentheses from generator expression inside function call
**Describe the style change**

I think that redundant parentheses should be removed from generators or at least be allowed to stick to next/previous parenthesis. Consider the following source example:

```python
any((
    this_very_long_method_name(obj) or another_long_method_name(obj)
    for obj in this_iterable
))
```

Now black (22.1.0) formats it to

```python
any(
    (
        this_very_long_method_name(obj) or another_long_method_name(obj)
        for obj in this_iterable
    )
)
```
I think that added extra indentation level here makes source look worse.

**Desired style**

```python
any(
    this_very_long_method_name(obj) or another_long_method_name(obj)
    for obj in this_iterable
) 
```

Or at least

```python
any((
    this_very_long_method_name(obj) or another_long_method_name(obj)
    for obj in this_iterable
))
```

I would prefer the first approach (since AST of both results is the same), because this parenthesis are redundant and can be safely removed.

Related issue: #2522

Boulea7:
### Description

Closes #2943.

Under `--preview`, this removes redundant extra parentheses around generator expressions, including standalone expressions, tuple/list/dict values, nested calls, and call arguments where Python still requires one visible pair.

### Checklist - did you ...

### Tests

- `.venv/bin/python -m pytest -n 4 -q` (551 passed, 2 skipped, 8 subtests passed)
- `.venv/bin/python -m pytest tests/test_format.py -q` (233 passed)
- `.venv/bin/pre-commit run --all-files`
- `.venv/bin/sphinx-build -a -b html -W docs/ docs/_build/`
- Direct and extra-parenthesized generator cases before attribute, subscript, and call trailers

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,8 +58,7 @@
 
 <!-- Changes that affect Black's preview style -->
 
-- Remove redundant parentheses around a generator expression passed as the sole argument
-  to a call (#5304)
+- Remove redundant parentheses around generator expressions (#5304)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,8 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Remove redundant parentheses around a generator expression passed as the sole argument
+  to a call (#5304)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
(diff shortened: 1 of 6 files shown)


cobaltt7:
Thanks!

Would it be simple to extend this to all structures? Here's some examples I found:
```py
((item.is_valid() for item in items))  # alone at top level
(((item.is_valid() for item in items)),)  # tuple
[((item.is_valid() for item in items)), ""]  # array (it's already handled when it's the only element)
{"foo": ((item.is_valid() for item in items))}  # dictionary value
```

Boulea7:
Yes, that fit cleanly. I expanded this to the structures you listed and added regression cases for top-level expressions, tuple/list/dict values, nested calls, keyword/starred call args, and multi-arg calls where one pair is still required.

Latest checks:
- `.venv/bin/python -m pytest tests/test_format.py -q`
- in-memory AST/idempotence checks for nested, walrus, async, attribute/subscript, starred, keyword, and multi-arg cases

cobaltt7 (approved):

PR #5286, opened 2026-07-31
Preserve function docstring spacing with line ranges

happykawayigt:
### Description

Fixes #5285.

When line-range formatting replaces out-of-range code with `fmt: pass` placeholders, those placeholders are represented as comment lines. The blank-line tracker could therefore scan past the enclosing function boundary and misclassify its docstring as a module docstring, adding a blank line that was outside the requested formatting range.

Treating a converted `fmt: pass` line as a scope boundary preserves the existing spacing. The regression fixture exercises every single-line range across a formatted function whose docstring has no following blank line.

Validation:
- `pytest tests/test_format.py -q` (230 passed)
- pre-commit on changed files: isort, flake8, mypy, prettier, EOF, trailing whitespace passed
- `git diff --check`

AI assistance was used during implementation. I independently reviewed the line-range conversion path and final diff, reproduced the regression through the line-by-line fixture, and ran the validation above.

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? (No style change.)
- [x] Add an entry in `CHANGES.md` if necessary?
- [x] Add / update tests if necessary?
- [x] Add new / update outdated documentation? (No documentation change is needed.)

Issue #2370, opened 2021-07-12
Blank line after docstring... or not.


PabloLec:
**Describe the style change**

Currently, black does not check blank lines between functions docstring and body. 
That means both these options are valid:
``` python
def foo():
    """docstring"""
    
   return 0
```

``` python
def foo():
    """docstring"""
   return 0
```

As black is `uncompromising` and `opinionated`, I guess this should be taken into account.

Personally I tend to put a blank line but looking at the only potential reference [PEP 257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings), there is no blank lines in given examples.

Whether we enforce the blank line or not, black should take care of this as this results in easy style inconsistency.

I'd say we should follow PEP 257 no blank line example.
**Comment your opinion so that this issue can be refered to in a PR fixing this.**

If there is no consensus, we could organize a jousting match to settle this. :goberserk:

hukkin:
Isn't this a case of (referring the docs):
> Black will allow single empty lines inside functions, and single and double empty lines on module level left by the original editors, except when they’re within parenthesized expressions.

So working as intended and documented.

> we could organize a jousting match to settle this

Reasonable proposal, I like this!

PabloLec:
Well, indeed the doc describes this behavior, but I think this should be treated separately.
As on module level, black does not allow beginning blank lines and enforce a final blank line, this same logic should apply at function level.

e.g. this code looks inconsistent and should be reformatted:
``` python
def foo():
    """docstring"""
   return 0

def bar():
    """docstring"""

   return 0

def baz():
    """docstring"""
   return 0
```

tolomea:
I agree and would also like to see Black have an opinion in the case where there is no docstring, my personal preference is no line between function def and first line of the body, but having it take a position one way or the other matters more to me than which position.

echoing @PabloLec this code looks inconsistent and should be reformatted.

```python
def foo():
   return 0

def bar():

   return 0

def baz():
   return 0
```

stinos:
Also: https://github.com/psf/black/issues/1872

felix-hilden:
@tolomea Your suggestion (without the docstrings) is being tracked in #902.

PabloLec:
Just a quick bump, I tried reaching out  and offer help on main communication means a few times within the past months but no response so, yeah.
I guess there are more important matters and blocking issues to consider first but this one seems close to black main purpose to me.
Again, whatever the formatting will be, there need to be some.

felix-hilden:
[...]

PR #5312, opened 2026-08-16
Fix --line-ranges inserting a blank line after a docstring

### LINKED ISSUE #5285: Empty new line added after docstring when using `--line-ranges`
**Describe the bug**

Based on the discussion in [this issue](https://github.com/psf/black/issues/2370), Black has no opinion on whether a blank line should or should not exist after a docstring. Both are permitted. However, Black will add an empty new line after a docstring when using `--line-ranges` when the lines are within the lines containing the docstring.

**To Reproduce**

For example, take this code in "example.py":

```python
def function():
    """This is a docstring."""
    return "Hi There!"
```

```sh
$ black --line-ranges 2-2 example.py
```

"example.py" is changed to:

```python
def function():
    """This is a docstring."""

    return "Hi There!"
```

**Expected behavior**

I expect no change to the original file.

**Environment**

- Black's version: black, 26.5.2.dev73+gc7801d9e2
- OS and Python version: MacOS/Python 3.14.6

**Additional context**

This unexpected behavior was introduced in version 26.1.0 (and earlier in preview mode, but I didn't check when that occured).

I notice this error because the [VS Code Black Formatter Extension](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter) often used the `--line-ranges` option. I notice it a lot when pasting text into a docstring because I have configured VS Code to format-on-paste.

aryansk:
Fixes #5285. closes #5286

## What

`--line-ranges` inserts an empty line after a docstring when the range covers only the docstring:

```python
def function():
    """This is a docstring."""
    return "Hi There!"
```

`black --line-ranges 2-2` changes it to:

```python
def function():
    """This is a docstring."""

    return "Hi There!"
```

Black deliberately has no opinion on blank lines after (function) docstrings (see #2370), so the file should be left unchanged.

## Root cause

With `--line-ranges`, unchanged statements are converted to `STANDALONE_COMMENT` leaves. When only the docstring is in range, the `def` header (and everything else outside the range) becomes a comment, so `_line_is_module_docstring` sees a docstring whose only preceding "statements" are comments and concludes it is a *module* docstring — which Black always separates from the next statement with one empty line.

A module docstring is by definition the first statement of a module, i.e. at depth 0. A function/class/method docstring is at depth ≥ 1. Requiring depth 0 fixes the misclassification.

## Test

Added `tests/data/line_ranges_formatted/function_docstring.py`; the existing line-by-line harness formats every single line of the file and asserts no change. It fails on main and passes with this change. All of `test_black.py` (166 tests + 8 subtests), `test_format.py` and `test_ranges.py` pass, and the file is self-formatted.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 - Add support for NO_COLOR environment variable to disable ANSI output (#5129)
 - No spurious target version warning when runtime version is included in a
   --target-version flag (#5167)
+- `--line-ranges` no longer inserts an empty line after a docstring when the
+  range covers only the docstring itself (#5312)
 
 ### Highlights
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,8 @@
 - Add support for NO_COLOR environment variable to disable ANSI output (#5129)
 - No spurious target version warning when runtime version is included in a
   --target-version flag (#5167)
-- `--line-ranges` no longer inserts an empty line after a docstring when the
-  range covers only the docstring itself (#5312)
+- `--line-ranges` no longer inserts an empty line after a docstring when the range
+  covers only the docstring itself (#5312)
 
 ### Highlights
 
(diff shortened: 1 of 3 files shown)


cobaltt7:
Thanks!

aryansk:
Thank you for the review and merge @cobaltt7!

Issue #5307, opened 2026-08-12
Crash on `async def`/`with` and a semicolon on the same line


cobaltt7:
**Describe the bug**

Black crashes on code such as:
```py
async with give_me_async_context(): print("a");  # fmt: skip
async def give_me_async_context(): print("a");  # fmt: skip
```

Found via #5303, which was closed for not clearly solving a valid issue.

**Environment**

- Black's version: main
- OS and Python version: Windows 11, Python 3.12.4

**Additional context**

@Dharshika-11 feel free to request #5303 to be reopened if you're going to clearly target this issue. The issue you were previously targeting wasn't valid, leading to the PR having an unclear scope.

Dharshika-11:
I'd like to work on this issue. I'll investigate the crashes caused by **async def`/** async with` one-liners with a semicolon and **# fmt: skip**, and add regression tests along with the fix.

PR #5314, opened 2026-08-17
fix: handle async with/def/for + semicolon + `# fmt: skip` crash (issue #5307)

HosnainRafi:
## Summary

Black crashes with a `ParseError` when formatting a one-line `async with` or `async def` statement that has a semicolon-separated body and a `# fmt: skip` comment:

```python
async with give_me_async_context(): print("a");  # fmt: skip
async def give_me_async_context(): print("a");  # fmt: skip
```

## Root Cause

The `# fmt: skip` handling in `comments.py` preserves the entire compound statement header inline when the body contains semicolons. However, for `async` statements, the `ASYNC` token is a sibling of the compound statement inside the `async_stmt` wrapper node — it is not a child of the `with_stmt`/`funcdef`. This means the header collection logic in `_get_compound_statement_header` was missing the `async` keyword, causing Black to split it onto its own line during the first pass. On the second pass, `async` alone on a line is not valid Python, resulting in a parse error.

## Fix

1. In `_get_compound_statement_header`: when the compound statement's parent is an `async_stmt`, also include the `ASYNC` token in the header leaves.

2. In `_find_compound_statement_context`: added a new Case 3 to handle the `async_stmt` wrapper structure (`async_stmt -> compound_stmt -> simple_stmt`).

## Tests

- Added regression test cases to `tests/data/cases/fmtskip10.py` for `async def` and `async with` with semicolon bodies and `# fmt: skip`.
- All 398 existing tests pass.

### DIFF
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -634,6 +634,18 @@ def _find_compound_statement_context(parent: Node) -> Node | None:
     if parent.parent.type in _COMPOUND_STATEMENTS:
         return parent
 
+    # Case 3: Async one-line form.
+    # The async keyword wraps the compound statement:
+    #     async def f(): print("a"); print("b")  # fmt: skip
+    #     async with ctx(): print("a"); print("b")  # fmt: skip
+    # Structure: async_stmt -> compound_stmt -> simple_stmt
+    if (
+        isinstance(parent.parent, Node)
+        and parent.parent.parent is not None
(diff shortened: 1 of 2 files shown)


cobaltt7:
Thanks! Sorry, but I ended up going with #5311 due to the more concise and targeted fix.

PR #5322, opened 2026-08-19
avoid quadratic child rescan in normalize_invisible_parens

sahvx655-wq:
Formatting a long `if`/`elif` chain scales quadratically. Profiling a chain of a few thousand `elif` clauses put nearly all the time in `blib2to3`'s `Base.remove`, reached from `normalize_invisible_parens`: after each `if`/`elif`/`while`/`for`/`return` keyword it wraps the following test expression in invisible parentheses via `wrap_in_parentheses(node, child)`, which locates the child with `child.remove()`. The whole chain is a single `if_stmt` node, so its child list grows with the number of clauses and every `remove()` rescans it from the start, so wrapping n clauses is O(n^2). This runs in the default stable style and before any line splitting, so it is reachable pre-auth through `blackd`.

The enclosing `enumerate` loop already holds each child's position, so I pass it as the `index` that `wrap_in_parentheses` gained in #5184 and let it swap the atom in place with `set_child` instead of remove plus re-insert. Keeping the change in the wrapper call means every clause type flowing through this loop benefits without its own bookkeeping. Output is byte-identical across the test suite and the black plus stdlib sources in stable, preview and skip-string-normalisation modes; a 4000-clause chain drops from ~3.1s to ~1.0s and the curve flattens to linear.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -179,11 +179,8 @@
   `"%s ..." % (a, b, c, ...)` by copying the leaves that surround the merged string in
   one `append_leaves` call instead of one call per leaf, which rescanned the shared
   parent's child list from the start every time (#5220)
-- Improve performance on long `if`/`elif` chains (and other compound statements with
-  many clauses) by passing the known child index when wrapping each clause's test in
-  invisible parentheses in `normalize_invisible_parens`, so the child is swapped in
-  place instead of being located with a full rescan of the statement's child list
-  (#5322)
+- Improve performance on long `if`/`elif` chains and other compound statements with
+  many clauses (#5322)
 
 ### Output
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -179,6 +179,11 @@
   `"%s ..." % (a, b, c, ...)` by copying the leaves that surround the merged string in
   one `append_leaves` call instead of one call per leaf, which rescanned the shared
   parent's child list from the start every time (#5220)
+- Improve performance on long `if`/`elif` chains (and other compound statements with
+  many clauses) by passing the known child index when wrapping each clause's test in
+  invisible parentheses in `normalize_invisible_parens`, so the child is swapped in
+  place instead of being located with a full rescan of the statement's child list
+  (#5322)
```
```suggestion
- Improve performance on long `if`/`elif` chains and other compound statements with
  many clauses (#5322)
```

PR #5329, opened 2026-08-20
Fix a crash when `# type: ignore` is lost during formatting

### LINKED ISSUE #5328: Formatting a dictionary with a long string and type ignore fails on --unstable
**Describe the bug**

Black crashes with INTERNAL ERROR when trying to format a combination of a dictionary, long string, and a type ignore comment.

**To Reproduce**

Take this code:

```python
my_dict = {
    "key": (
        "A very very very very very very very very very very very very very very very very long string literal"
    )  # type: ignore
}
```

```sh
$ black example.py --unstable
```

The resulting error is:

```
error: cannot format example.py: INTERNAL ERROR: Black 0.1.dev2313+gc835e9db6 on Python (CPython) 3.14.3 produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_vleux34f.log

Oh no! 💥 💔 💥
```

**Expected behavior**

Code should get formatted to something like:

```python
my_dict = {
    "key": (
        "A very very very very very very very very very very very very very very very"
        " very long string literal"
    )  # type: ignore
}
```

**Environment**

- Reproduced both with Black `26.5.1` and the latest code on the `main` branch

**Additional context**

Here's the content of the `/tmp/blk_vleux34f.log` log file:
```
--- src
+++ dst
@@ -28,8 +28,6 @@
                 'A very very very very very very very very very very very very very very very very long string literal',  # str
             )  # /Constant
         )  # /Dict
     )  # /Assign
     type_ignores=
-    TypeIgnore(
-    )  # /TypeIgnore
 )  # /Module
\ No newline at end of file
```

The crash happens because there is a bug that makes the `# type: ignore` comment just disappear.

AleksMat:
### Description

In the transformation process of formatting a complicated expression with a `# type: ignore` comment at the end, the comment was getting lost. This was in turn crashing Black on the `--unstable` style.

Fix the issue by moving comments from `old_rpar_leaf` to the new last line of the transformed code after a transformation.  It seems this was missed in the initial implementation.

This PR is a direct fix for #5328.

### Checklist - did you ...

- [ ] ~Add new / update outdated documentation?~

FIXES: #5328

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,6 +63,7 @@
   previously crashing) (#5161)
 - Fix crash when `# fmt: skip` is used on one-line `async def`, `async with`, and
   `async for` statements containing a semicolon (#5311)
+- Fix crash when `# type: ignore` is lost during formatting (#5328)
 
 ### Preview style
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,7 +63,7 @@
   previously crashing) (#5161)
 - Fix crash when `# fmt: skip` is used on one-line `async def`, `async with`, and
   `async for` statements containing a semicolon (#5311)
-- Fix a crash when `# type: ignore` is lost during formatting (#5328)
+- Fix a crash when `# type: ignore` is lost during formatting (#5329)
 
 ### Preview style
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,7 +63,6 @@
   previously crashing) (#5161)
 - Fix crash when `# fmt: skip` is used on one-line `async def`, `async with`, and
   `async for` statements containing a semicolon (#5311)
-- Fix a crash when `# type: ignore` is lost during formatting (#5329)
 
 ### Preview style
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -63,6 +63,7 @@
   previously crashing) (#5161)
 - Fix crash when `# fmt: skip` is used on one-line `async def`, `async with`, and
   `async for` statements containing a semicolon (#5311)
+- Fix a crash when `# type: ignore` is lost during formatting (#5329)
```
cobaltt7: Please slightly elaborate on this - something like "...during formatting of a long parenthesized string". Since this bug is exclusive to the preview style, please also move it to that section.

Otherwise LGTM thanks!

AleksMat: Done - moved to preview and expanded the description as suggested.

cobaltt7:
Sorry, one more quick note - since this is only reproducible under `--unstable`, please make sure the test is in a file that enables `# flags: --unstable` on the first line.

cobaltt7:
@AleksMat are you still updating this?

AleksMat:
> @AleksMat are you still updating this?

Sorry, I missed your previous message. Now I moved the test case to `preview_comments7.py` which has the `--unstable` flag.

cobaltt7:
Thanks!

PR #5335, opened 2026-08-21
Fix long keyword argument values exceeding line length

### LINKED ISSUE #4040: Black fails to keep within line length limit when using named parameters
**Describe the bug**

If a long named parameter expression is passed to a function call, black exceeds the line length limit. Adding parentheses fixes the issue, but it feels like Black should do that automatically.

**To Reproduce**

```python
foo(
    this_is_a_very_looooooooooooooong_parameter_name=some_very_loooooooooooooooooong_variable_name
)
```

**Expected output**

```python
foo(
    this_is_a_very_looooooooooooooong_parameter_name=(
        some_very_loooooooooooooooooong_variable_name
    )
)
```

Which stays within the 88-character line length limit.

**Actual output**

```python
foo(
    this_is_a_very_looooooooooooooong_parameter_name=some_very_loooooooooooooooooong_variable_name
)
```

Which exceeds the 88-character line length limit.

**Workaround**

The issue can be worked around by adding the parentheses manually and then running black.

**Additional information**

Worth noting this appears to be specific to named parameters, and does not affect, say variable assignments. Indeed, the following:

```python
this_is_a_very_looooooooooooooong_parameter_name=some_very_loooooooooooooooooong_variable_name
```

Is correctly formatted to:

```python
this_is_a_very_looooooooooooooong_parameter_name = (
    some_very_loooooooooooooooooong_variable_name
)
```

Black's behaviour appears to be inconsistent between variable assignments and named parameters in this regard.

dajinzhu:
## Summary

Fix formatting of long keyword argument values so the value can be parenthesized and split when the complete call would otherwise exceed the configured line length.

## Changes

- Add invisible parentheses around eligible long keyword argument values.
- Add a regression case covering a long keyword argument value.
- Keep strings and values that already contain opening brackets on the existing path.

## Testing

- `pytest tests/test_format.py` — 237 passed
- `pytest tests/test_black.py` — 166 passed, 8 subtests passed
- Focused regression case — passed
- Source self-formatting check — passed
- `git diff --check` — passed

Resolves #4040.

### DIFF
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -243,7 +243,9 @@ class LineGenerator(Visitor[Line]):
                 leaf.prefix + leaf.value for leaf in value_leaves
             ).strip()
             continuation_indent = 4 * (self.current_line.depth + 1)
-            key_width = str_width(node.children[0].value) + 1
+            key = node.children[0]
+            assert isinstance(key, Leaf)
+            key_width = str_width(key.value) + 1
             value_width = str_width(value_text)
             if (
                 value_text
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -234,6 +234,34 @@ class LineGenerator(Visitor[Line]):
 
             yield from self.visit(child)
 
+    def visit_argument(self, node: Node) -> Iterator[Line]:
+        """Add potential parentheses around long keyword argument values."""
+        if len(node.children) == 3 and node.children[1].type == token.EQUAL:
+            value = node.children[2]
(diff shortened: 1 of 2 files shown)


cobaltt7:
Looking at the diff-shades changes, this fix parenthesizes values even if they're already broken within. Parenthesizing values should only be the last-try solution to reduce line length if there's no pre-existing split that fits.

Also, this needs to be gated to the preview style (see example bd622eda1aab7c8c6604282e2fa78381ee6370f3) and a changelog entry should be added.

PR #5336, opened 2026-08-21
Preserve trailing blank lines in unclosed # fmt: off blocks

### LINKED ISSUE #496: empty lines at end of file trimmed even with fmt: off

tagadearpit:
Closes #496

### Summary

Black currently drops trailing blank lines when a file ends inside an unclosed `# fmt: off` region. The parser keeps those newlines in the `ENDMARKER` prefix, but the fmt-off conversion discarded them. This change preserves whitespace-only `ENDMARKER` prefixes for an unclosed fmt-off block at EOF while avoiding comment text that may also be attached to the end marker.

### Testing

- Added `test_fmt_off_preserves_trailing_blank_lines` to exercise the real file-formatting path.
- Full test suite: `482 passed, 3 skipped, 8 subtests passed`.
- `black --check .` passed.
- `git diff --check` passed.

This is a focused bugfix with no user-facing style-policy change, so no documentation or changelog update is needed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 <!-- PR authors:
      Please include the PR number in the changelog entry, not the issue number -->
 
+- Preserve trailing blank lines at EOF in unclosed `# fmt: off` blocks (#5336)
 - Add support for NO_COLOR environment variable to disable ANSI output (#5129)
 - No spurious target version warning when runtime version is included in a
   --target-version flag (#5167)
--- a/src/black/comments.py
+++ b/src/black/comments.py
@@ -409,6 +409,7 @@ def convert_one_fmt_off_pair(
                 is_fmt_skip,
                 lines,
                 leaf,
+                mode,
                 search_hints,
             )
             return idx
@@ -426,6 +427,25 @@ def _is_attached(leaf: Leaf, root: Node) -> bool:
     return False
 
 
+def _get_endmarker_and_dedents(node: LN) -> tuple[LN | None, list[LN]]:
+    """Return the ENDMARKER and terminal DEDENTs following an unclosed fmt block."""
+    dedents: list[LN] = []
+    current: LN | None = node
+    while current is not None:
+        sibling = current.next_sibling
+        if sibling is None:
+            current = current.parent
+            continue
+        if sibling.type == token.DEDENT:
+            dedents.append(sibling)
+            current = sibling
+            continue
+        if sibling.type == token.ENDMARKER:
+            return sibling, dedents
+        return None, []
+    return None, []
+
+
 def _handle_regular_fmt_block(
     ignored_nodes: list[LN],
     comment: ProtoComment,
@@ -433,6 +453,7 @@ def _handle_regular_fmt_block(
     is_fmt_skip: bool,
     lines: Collection[tuple[int, int]],
     leaf: Leaf,
+    mode: Mode,
     search_hints: dict[int, int] | None = None,
 ) -> None:
     """Handle fmt blocks with actual AST nodes."""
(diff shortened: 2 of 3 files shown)


src/black/comments.py
```diff
@@ -485,6 +485,19 @@ def stringify_node(n: LN) -> str:
             parts.append(str(node))
 
     hidden_value = "".join(parts)
+
+    # The parser stores blank lines after the final statement in the END marker
+    # prefix rather than in the statement node. Preserve those newlines for an
+    # unclosed fmt: off block at EOF; otherwise they are silently discarded.
+    if not is_fmt_skip:
+        endmarker = ignored_nodes[-1].next_sibling
```
yangfan-yf-yf: `next_sibling` is not the `ENDMARKER` when the unclosed block is inside an indented suite; it is a `DEDENT`. For `def f():\n    # fmt: off\n    x  =  1\n\n`, be2acd5 removes the final blank line. The same happens with deeper nesting and multiple trailing blank lines. Please walk past the relevant `DEDENT` nodes without crossing the ignored region incorrectly, and add an indented regression.

src/black/comments.py
```diff
@@ -485,6 +485,19 @@ def stringify_node(n: LN) -> str:
...
+        if (
+            endmarker is not None
+            and endmarker.type == token.ENDMARKER
+            and not endmarker.prefix.strip()
```
yangfan-yf-yf: This condition treats any non-whitespace EOF prefix as all-or-nothing, but a prefix can contain both an ordinary comment and trailing whitespace. On be2acd5, `# fmt: off\nx=  1\n# tail\n\n` loses the final blank line; CRLF behaves the same. Please preserve the trailing whitespace suffix without duplicating the comment text, and add LF/CRLF coverage for this EOF shape.

tests/test_black.py
```diff
@@ -205,6 +205,16 @@ def test_one_empty_line_ff(self) -> None:
                 os.unlink(tmp_file)
             self.assertFormatEqual(expected, actual)
 
+    def test_fmt_off_preserves_trailing_blank_lines(self) -> None:
+        source = expected = '# fmt: off\n\nprint("hello")\n\n'
+        tmp_file = Path(black.dump_to_file(source, ensure_final_newline=False))
```
yangfan-yf-yf: The regression fails on Windows before validating the formatter behavior. `dump_to_file()` writes the string in text mode, so LF becomes CRLF, while `expected` remains LF; the subsequent byte read therefore differs on be2acd5. Please write explicit bytes, as `test_one_empty_line_ff` does above, or parameterize LF/CRLF with matching expected bytes.

yangfan-yf-yf (changes requested):
I found three blockers on be2acd5: indented unclosed regions and EOF comment prefixes still lose trailing blank lines, and the added regression fails on Windows before it checks the formatter behavior. Please also add the user-facing bugfix entry required by the contributing guide, or have a maintainer apply `ci: skip news` if they consider it unnecessary.

tagadearpit:
The changes had been successfully done

tests/test_black.py
```diff
@@ -205,6 +205,66 @@ def test_one_empty_line_ff(self) -> None:
                 os.unlink(tmp_file)
             self.assertFormatEqual(expected, actual)
 
+    def test_fmt_off_preserves_trailing_blank_lines(self) -> None:
```
cobaltt7, line 208: Is there a reason these are new methods instead of new test case files?

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 <!-- PR authors:
      Please include the PR number in the changelog entry, not the issue number -->
 
+- Preserve trailing blank lines at EOF in unclosed `# fmt: off` blocks (#5336)
```
cobaltt7, line 8: Please put this in the "Stable Style" section

cobaltt7:
@tagadearpit are you still updating this? Please check out my review comments above

PR #5339, opened 2026-08-24
Clarify project-root exclusion patterns

### LINKED ISSUE #2750: Documentation example `--extend-exclude ^/foo.py` does not exclude root file `foo.py`
**Describe the bug**
Black does not exclude the file`foo.py` root when asked to exclude it with `^/foo.py`. This particular example was used in the [documentation](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html?highlight=.venv#configuration-format):

> ```
> extend-exclude = '''
> # A regex preceded with ^/ will apply only to files and directories
> # in the root of the project.
> ^/foo.py  # exclude a file named foo.py in the root of the project (in addition to the defaults)
> '''
> ```

**To Reproduce**
Start with an empty directory

```sh
$ echo "s='test'" > foo.py

$ black --extend-exclude "^/foo.py" --check .
would reformat foo.py
Oh no! 💥 💔 💥
1 file would be reformatted.
```

Same behavior using `--exclude` flag

**Environment**

- Black's version: 21.12b0
- OS and Python version: OSX 11.6, Python 3.9.6

nightcityblade:
### Description

Fixes #2750.

Clarify that exclusion options use regular expressions against paths relative to Black's discovered project root. Update the root-file example to escape the dot and anchor the filename, and note how to inspect root discovery with `--verbose`.

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? (N/A, documentation only.)
- [x] Add an entry in `CHANGES.md` if necessary? (Not necessary for this documentation clarification.)
- [x] Add / update tests if necessary? (`pre-commit` and the warning-as-error Sphinx build pass.)


cobaltt7:
Thanks!

Issue #3569, opened 2023-02-14
Consistent handling of too long lines due to directives (no special treatment of `# type: ignore`)


ruestefa:
_(I initially wrote the issue as a bug report, hence the issue template, but it's probably closer to a feature request.)_

**Describe the ~bug~**

Lines that are too long because of trailing comments are handled inconsistently depending on whether they start with a mypy `type: ignore` directive.

Specifically, lines where the comment begins with `# type: ignore` are left as they are, whereas other such lines are broken up over multiple lines, which includes lines that contain `# type: ignore` but not in the beginning, and others that only contain directives from other tools.

Input:

```python
from my_module import some_function  # type: ignore [import]  # pylint: ignore=no-name-in-module
from my_module import some_function  # pylint: ignore=no-name-in-module  # type: ignore [import]
from my_module import some_function  # type: ignore [import]  # pylint: ignore=no-name-in-module  # pylint: ignore=a_rule
from my_module import some_function  # pylint: ignore=no-name-in-module  # pylint: ignore=a_rule  # type: ignore [import]
from my_module import some_function  # pylint: ignore=no-name-in-module  # pylint: ignore=a_rule
```

Output:

```python
from my_module import some_function  # type: ignore [import]  # pylint: ignore=no-name-in-module
from my_module import (
    some_function,
)  # pylint: ignore=no-name-in-module  # type: ignore [import]
from my_module import some_function  # type: ignore [import]  # pylint: ignore=no-name-in-module  # pylint: ignore=a_rule
from my_module import (
    some_function,
)  # pylint: ignore=no-name-in-module  # pylint: ignore=a_rule  # type: ignore [import]
from my_module import (
    some_function,
)  # pylint: ignore=no-name-in-module  # pylint: ignore=a_rule
```

_I'm using import statements in the example because that's how I became aware of this behavior (and because using `# pylint: disable-next:...` instead is complicated due to how `isort` handles comments, not that that's black's problem), but I assume it is universal._

**Expected behavior**

Ideally, I'd like the `# pylint: disable` directives to be treated like the `# type: ignore` directives, such that all lines remain unchanged. Also, I'd like the `# type: ignore` directive also to be detected at the end of the line. However, I'm aware that this would require parsing the comments for directives beyond a simple `.startswith("# type: ignore")`, which you might not want to get into.

[...]

PR #5352, opened 2026-08-29
Keep inline comments with the string when wrapping it in parentheses (#3802)

### LINKED ISSUE #3802: `--preview` makes it impossible to use a comment to ignore a static type violation
**Describe the bug**

**To Reproduce**

For example, take this code:

```python
    def __str__(self):
        return f"{self.get_name_display()}" # pyright: ignore[reportGeneralTypeIssues] -- This is some special Django magic that we can't easily tell Pyright about
```

```sh
$ black --preview
```

The resulting error is:

> Unnecessary "# pyright: ignore" rule: "reportGeneralTypeIssues" (Pylance)
```python
    def __str__(self):
        return (  # pyright: ignore[reportGeneralTypeIssues] -- This is some special Django magic that we can't easily tell Pyright about
            f"{self.get_name_display()}"
        )
```
If I move the comment to the middle line, `black` moves it back up to the top.

**Expected behavior**

Even if `black` breaks my comment in some cases, there should be at least one possible place for me to leave the comment that will suppress my type error without `black` moving it.

**Environment**

- Black's version: `23.7.0`
- OS and Python version: MacOS, 3.11

**Additional context**

This also occurs if the comment starts on the line with the violation

Mtok2002:
Closes #3802

When `StringParenWrapper` wraps a return/assign/assert value in parentheses so
the string can go on its own line, it moved any inline comment that followed the
string up to the opening-paren line. That silently changed which expression a
suppression comment applied to, so `# pyright: ignore[...]` (and similar) stopped
working — and there was no place the comment could be written that survived
formatting.

The comment now stays on the same line as the string it was written after.
Comments after a trailing comma or a pre-existing `(` are still moved up, as
before.

This changes the expected output of three existing unstable-style cases, updated
here: `preview_comments7`, `preview_long_strings` and
`preview_long_strings__regression`. A regression case for this issue is added as
`preview_comment_after_wrapped_string`.

Verified on this base: `tests/test_format.py` + `tests/test_black.py` go from 3
failures to 403 passed, with no newly failing test; re-running Black on the
formatted output is idempotent.

This patch was produced with the help of an autonomous coding system I build and
supervise. I review and test every change before submitting and follow up on
review feedback personally. If automated contributions are unwelcome in this
project, please say so and I will stop submitting them.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -68,11 +68,10 @@
 
 <!-- Changes that affect Black's preview style -->
 
-- Keep an inline comment that follows a string on the same line as that string
-  when the string is wrapped in parentheses, instead of moving the comment up to
-  the opening parenthesis. This lets suppression comments such as
-  `# pyright: ignore[...]` keep applying to the expression they were written for
-  (#3802)
+- Keep an inline comment that follows a string on the same line as that string when the
+  string is wrapped in parentheses, instead of moving the comment up to the opening
+  parenthesis. This lets suppression comments such as `# pyright: ignore[...]` keep
+  applying to the expression they were written for (#3802)
 - Remove redundant parentheses around generator expressions (#5304)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -68,6 +68,11 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Keep an inline comment that follows a string on the same line as that string
+  when the string is wrapped in parentheses, instead of moving the comment up to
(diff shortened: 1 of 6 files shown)


cobaltt7:
There's a bigger discussion here about tooling comments in general. Many tooling comments we explicitly special-case to avoid moving them, but we want to find a better heuristic for either identifying tooling comments, or for handling comments in general, so we can stop playing a cat-and-mouse game of special-casing more and more tools as they come up.

Related issues: #3569, #3946, #5260

cobaltt7:
Please discuss on the related issues and come to a consensus with maintainers and the community before opening a PR.

PR #5357, opened 2026-08-29
Long string in return annotation does not get split if there's parameters (#3926)

### LINKED ISSUE #3926: Long string in return annotation does not get split if there's parameters
**Describe the bug**
This was first reported in #2699, and thought resolved in #2990, except that only solved the specific case of the function having no parameters. 

**To Reproduce**
Run the test `return_annotation_brackets_string`:

https://github.com/psf/black/blob/3a2d76c7bcf39e852f3b379b76537d7847ed4225/tests/data/preview/return_annotation_brackets_string.py#L1-L23

The second function should be formatted exactly the same as the first one.

Mtok2002:
Closes #3926

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -68,6 +68,7 @@
 
 <!-- Changes that affect Black's preview style -->
 
+- Long string in return annotation does not get split if there's parameters (#3926)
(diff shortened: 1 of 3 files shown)


JelleZijlstra:

PR #5361, opened 2026-08-29
gitignore patterns with trailing slashes are incorrectly handled (#3694)

Mtok2002:
Closes #3694

This patch was produced with the help of an autonomous coding system I build and supervise. I review and test every change before submitting and follow up on review feedback personally. If automated contributions are unwelcome in this project, please say so and I will stop submitting them.

PR #5363, opened 2026-08-29
Improve `adjusted_lines`'s diff algorithm used in `--line-ranges`. (#4052)

### LINKED ISSUE #4052: Improve `adjusted_lines`'s diff algorithm used in `--line-ranges`.
Currently, _Black_ performs two formatting passes, and also performs an extra formatting pass for stability check in `--safe` mode.

When performing a second formatting pass with `--line-ranges`, we can't simply use the original lines specified by the user. For example, unformatted:

```python
def restrict_to_this_line(arg1,
  arg2,
  arg3):
    print  ( "This should not be formatted." )
    print  ( "This should not be formatted." )
```

If we let it format lines `1-3`, after the first pass it becomes:

```python
def restrict_to_this_line(arg1, arg2, arg3):
    print  ( "This should not be formatted." )
    print  ( "This should not be formatted." )
```

If we use the original `1-3` lines in a second pass, it would format all the lines now.

To solve this, we have an [`adjusted_lines`](https://github.com/psf/black/blob/58f31a70efe6509ce8213afac998bc5d5bb7e34d/src/black/ranges.py#L48) function to calculate the new line ranges for the second pass. It uses the diffing algorithm from `difflib.SequenceMatcher`. Unfortunately it could produce undesired results for certain edge cases like a list of unformatted lines with the same content.

For example:

```python
print ( "format me" )
print ( "format me" )
print ( "format me" )
print ( "format me" )
print ( "format me" )
```

Using `--line-ranges=2-3`, the result of the first pass is

```python
print ( "format me" )
print("format me")
print("format me")
print ( "format me" )
print ( "format me" )
```

`adjusted_lines` will return `5-5` with the input of `2-3` and the original & formatted code. After the second pass the code becomes:

```python
print ( "format me" )
print("format me")
print("format me")
print ( "format me" )
print("format me")
```

The end result is

1) `--line-range` might format extra lines when there are unformatted lines.
2) The stability check fails in `--safe` mode.

In #4034, we are disabling the stability check for now. This issue tracks the improvement of `adjusted_lines` so that we can eventually fix the underlying issue and re-enable the stability check.

Mtok2002:
Closes #4052

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Improve `adjusted_lines`'s diff algorithm used in `--line-ranges`. (#4052)
 - Preserve blank lines that come immediately before a `# fmt: on` comment (#5300)
 - Keep the parentheses around the target of an annotated assignment (for example
   `(x): int = 5`). They make the target non-simple, so CPython leaves the name out of
--- a/src/black/ranges.py
+++ b/src/black/ranges.py
@@ -3,6 +3,7 @@
 import difflib
 from collections.abc import Collection, Iterator, Sequence
 from dataclasses import dataclass
+from typing import Any
 
 from black.nodes import (
     LN,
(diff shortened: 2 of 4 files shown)


JelleZijlstra:

PR #5365, opened 2026-08-29
global config not possible if project contains a `pyproject.toml` (#4411)

### LINKED ISSUE #4411: global config not possible if project contains a `pyproject.toml`
**Describe the bug**

We have multiple Python Git projects, which each have their own `pyproject.toml`, but we don't want to have to copy and paste the same `[tool.black]` section in each of them and would prefer to be able to have it configured in our user's `~/.black` global config file. Unfortunately, [black.files.find_pyproject_toml()](https://github.com/psf/black/blob/8827accf560c7be418000962565f1fd6fa7b01cb/src/black/files.py#L110C1-L115C62) doesn't check if the project's `pyproject.toml` file has a `[tool.black]` section before returning it, so that causes it to never check to see if there's a `~/.black` global config file.

**To Reproduce**

Have a Python Git project with a `pyproject.toml` file that does not have a `[tool.black]` section, and have a `~/.black` global config file in your user's directory. The file in your user's directory will be ignored.

**Expected behavior**

If a project's `pyproject.toml` file does not have a `[tool.black]` section, then `black` should look for a user-level `~/.black` global config file.

**Additional context**

NOTE: this is similar to Issue #2863, but this asks for the `[tool.black]` section-searching to continue and not stop at the project level.

Mtok2002:
Closes #4411

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- global config not possible if project contains a `pyproject.toml` (#4411)
 - Preserve blank lines that come immediately before a `# fmt: on` comment (#5300)
 - Keep the parentheses around the target of an annotated assignment (for example
   `(x): int = 5`). They make the target non-simple, so CPython leaves the name out of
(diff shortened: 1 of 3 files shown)


JelleZijlstra:

PR #5369, opened 2026-08-30
Format a call whose sole argument is a parenthesized generator under --preview

Kayvan-Zahiri:
### Description

`black --preview` refuses to format `outer(inner(((x for x in y))))` when the generator is long enough to split:

```
error: cannot format g.py: INTERNAL ERROR: Black produced different code on the second pass of the formatter.
```

`visit_power` only spots a sole-argument generator when the call's argument is the generator atom itself, so with a second redundant pair it does nothing and `visit_atom` removes just the outer pair. The pair left behind keeps the generator inside a bracket, so the `for` never registers as a delimiter and the line breaks inside the element expression instead. Re-parsing that output drops the surviving pair and splits at the `for`, hence the two passes disagree.

`visit_power` now also accepts a nested redundant-paren atom, and `maybe_make_parens_invisible_in_atom` forwards `remove_generator_parens` into its recursion so the inner pair goes too. The result matches what the one-pair spelling already produced.

Preview only. Formatting 2589 files across three modes gives byte-identical output before and after, and the stable style is untouched.

### Checklist - did you ...

- [x] Add new / update outdated documentation? (`future_style.md` already describes the feature; no behavior to add there)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,6 +69,11 @@
 <!-- Changes that affect Black's preview style -->
 
 - Remove redundant parentheses around generator expressions (#5304)
+- Remove every layer of redundant parentheses around a generator passed as the sole
+  argument of a call (e.g. `outer(inner(((x for x in y))))`). Only the outermost pair
+  was removed, and the pair left behind hid the `for` from the line splitter, so a
+  generator long enough to split came out differently on the second pass and Black
+  reported unstable formatting (#5329)
 - Preserve two blank lines before a top-level class starting inside a `# fmt: off` block
   after an import (#5238)
 - Fix unnecessary parentheses around short RHS expressions in indexed assignments like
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -69,6 +69,11 @@
 <!-- Changes that affect Black's preview style -->
 
 - Remove redundant parentheses around generator expressions (#5304)
+- Remove every layer of redundant parentheses around a generator passed as the sole
+  argument of a call (e.g. `outer(inner(((x for x in y))))`). Only the outermost pair
+  was removed, and the pair left behind hid the `for` from the line splitter, so a
+  generator long enough to split came out differently on the second pass and Black
+  reported unstable formatting (#5329)
```
```suggestion
- Remove redundant parentheses around generator expressions (#5304, #5329)
```

This isn't a regression on stable, so I don't think it's necessary to include the full details.

CHANGES.md
```diff
@@ -68,7 +68,7 @@
 
 <!-- Changes that affect Black's preview style -->
 
-- Remove redundant parentheses around generator expressions (#5304)
+- Remove redundant parentheses around generator expressions (#5304, #5329)
```
```suggestion
- Remove redundant parentheses around generator expressions (#5304, #5369)
```

cobaltt7:
Thanks!

PR #5375, opened 2026-08-31
Fix delimiter splitting after standalone comments

### LINKED ISSUE #4026: Single line with division operator in commented list causes additional line breaks
**Describe the bug**

Inside a list, a comment followed by a single line with division operator causes additional lines breaks before each division operator.

**To Reproduce**

For example, take this code:

```python
import pathlib

foobar = [
    # comment
    pathlib.Path("foo") / "bar" / "baz",
]

fizzbuzz = [
    # comment
    23 / 7
]
```

And run black on it with the default config.

The resulting formatted code is:

```python
import pathlib

foobarbaz = [
    # comment
    pathlib.Path("foo")
    / "bar"
    / "baz",
]

fizzbuzz = [
    # comment
    23
    / 7
]
```

**Expected behavior**

The formatting doesn't change (apart from maybe the trailing comma).

**Environment**

- Black's version: 23.10.1 (compiled: yes)
- OS and Python version: Fedora Linux 38 / Python 3.11.6

**Additional context**

Formatting does not occur if the expression is enclosed in parentheses, e.g.

```python
(pathlib.Path("foo") / "bar" / "baz")
```

rootsec1:
Fixes #4026.

A standalone comment can force a bracketed expression through the splitting pipeline even when the expression itself fits on one line. In a one-element collection, `/` then becomes the highest-priority delimiter, so Black expands short paths and fractions unnecessarily.

This changes `delimiter_split()` to check the result of `standalone_comment_split()` before splitting on the operator. It defers the operator split only when the remaining non-comment lines fit, and keeps the existing behavior for comma-separated elements and expressions that genuinely need wrapping. The no-trailing-comma case is limited to multiplicative precedence so unrelated boolean expressions, unions, and comprehensions are unaffected.

The regression cases cover the original path and fraction examples with and without trailing commas, `--preview`, `--skip-magic-trailing-comma`, custom line lengths, and a long-expression control that must still wrap.

Tests: `483 passed, 4 skipped, 8 subtests passed`, plus the repository's isort, flake8, mypy, Prettier, and hygiene hooks.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Keep short binary expressions on one line when preceded by a standalone comment in a
+  bracketed expression (#5375)
 - Preserve blank lines that come immediately before a `# fmt: on` comment (#5300)
 - Keep the parentheses around the target of an annotated assignment (for example
   `(x): int = 5`). They make the target non-simple, so CPython leaves the name out of
--- a/src/black/linegen.py
+++ b/src/black/linegen.py
@@ -14,6 +14,7 @@ from black.brackets import (
     COMMA_PRIORITY,
     COMPARATOR_PRIORITY,
     DOT_PRIORITY,
+    MATH_PRIORITIES,
     STRING_PRIORITY,
     get_leaves_inside_matching_brackets,
     max_delimiter_priority_in_atom,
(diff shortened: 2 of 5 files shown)


cobaltt7:
What's the root cause of the issue? It might make more sense to change whatever condition is causing the split instead of adding guards to undo it later on.

cobaltt7:
@rootsec1 are you updating this? Also, this needs to be gated to the preview style.

PR #5381, opened 2026-09-01
Handle UTF-8 offsets when masking notebook magics

### LINKED ISSUE #5378: Notebook assignment magics fail to parse when preceded by non-ASCII characters
**Describe the bug**

Black fails to format notebook cells where an assignment magic is preceded by non-ASCII characters, for example `变量 = %env FOO`.

While masking the transformed IPython magic, Black slices the line using `ast.col_offset`. CPython reports this offset in UTF-8 bytes, but slicing a Python string uses character indexes. As a result, the slice ends in the middle of `get_ipython(...)`, producing invalid code such as `变量 = get_b"…"` and eventually raising a `ParseError`.

The equivalent ASCII-only assignment (`x = %env FOO`) works, as do magics at the beginning of a line such as `%matplotlib inline`.

**To Reproduce**

Install the Jupyter extra with `pip install "black[jupyter]"`, then create this notebook:

```python
# write nb.ipynb
import json
from pathlib import Path

nb = {
    "nbformat": 4,
    "nbformat_minor": 5,
    "metadata": {"language_info": {"name": "python"}},
    "cells": [
        {
            "cell_type": "code",
            "execution_count": None,
            "id": "a",
            "metadata": {},
            "outputs": [],
            "source": ["变量 = %env FOO\n"],
        }
    ],
}
Path("nb.ipynb").write_text(
    json.dumps(nb, indent=1, ensure_ascii=False) + "\n", encoding="utf-8"
)
```

```sh
$ black --check nb.ipynb
```

Observed (CPython 3.11.2, Black `8947c48`, `black[jupyter]`):

```
error: cannot parse: nb.ipynb:1:10
    变量 = get_b"4ed19"
             ^
ParseError: bad input

Oh no! 💥 💔 💥
```
The hexadecimal mask changes on each run, but the malformed `get_b"` prefix is consistent.

Replacing the source line with the ASCII-only equivalent, `x = %env FOO\n`, succeeds:

```
would reformat nb.ipynb
```

The difference between byte and character offsets can also be demonstrated without IPython:

```python
import ast

s = '变量 = get_ipython().run_line_magic("env", "FOO")'
col = ast.parse(s).body[0].value.col_offset
print(col, s.index("get_ipython"), repr(s[:col]))
```

```
9 5 '变量 = get_'
```

`变量` is two characters but six bytes in UTF-8. Here, `line[:col_offset]` retains `get_`, so the mask replaces only the remainder of `get_ipython(...)`.

**Expected behavior**

Black should format the cell successfully and preserve the magic as `变量 = %env FOO`.

**Environment**

- Black's version: `8947c48ef2077c3a301b03c1e814dc2e3f78436e` (main)
- OS and Python version: Linux / CPython 3.11.2
- Extra: `black[jupyter]` (IPython)

**Additional context**

`replace_magics` in `src/black/handle_ipynb_magics.py`:

```python
col_offset, magic = (
    offsets_and_magics[0].col_offset,
    offsets_and_magics[0].magic,
)
mask = get_token(src, magic, existing_tokens)
line = line[:col_offset] + mask
```

`col_offset` comes from `ast.Call` after IPython's `TransformerManager.transform_cell`. CPython documents `col_offset` as a UTF-8 byte offset.

The relevant path is `format_cell` -> `mask_cell` -> `replace_magics` -> `format_str`; the final step attempts to parse the malformed line.

Converting the byte offset to a character index before slicing should avoid the corruption. The issue is not limited to CJK identifiers; accented identifiers such as `café = %env FOO` are affected as well.

HyunsikParker:
### Description

Fixes #5378.

`ast.col_offset` counts UTF-8 bytes, but `replace_magics()` used the value as a Python string index. When non-ASCII characters appeared before an assignment magic, Black kept part of the transformed `get_ipython()` call and produced source that could not be parsed.

Build the preserved prefix from the UTF-8 bytes before replacing the magic with its mask. The regression test covers a CJK identifier in both the default and custom cell-magic modes.

### Validation

- `tox -e py`
- `pre-commit run -a`
- Black self-format check

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@
 
 <!-- Changes that affect Black's stable style -->
 
+- Fix parsing Jupyter notebook assignment magics when non-ASCII characters appear
+  earlier on the line (#5381)
 - Preserve blank lines that come immediately before a `# fmt: on` comment (#5300)
 - Keep the parentheses around the target of an annotated assignment (for example
   `(x): int = 5`). They make the target non-simple, so CPython leaves the name out of
--- a/src/black/handle_ipynb_magics.py
+++ b/src/black/handle_ipynb_magics.py
@@ -299,7 +299,9 @@ def replace_magics(src: str) -> tuple[str, list[Replacement]]:
             mask = get_token(src, magic, existing_tokens)
             replacements.append(Replacement(mask=mask, src=magic))
             existing_tokens.add(mask)
-            line = line[:col_offset] + mask
+            # AST column offsets are UTF-8 byte offsets, not character indices.
+            prefix = line.encode("utf-8")[:col_offset].decode("utf-8")
+            line = prefix + mask
         new_srcs.append(line)
     return "\n".join(new_srcs), replacements
 
--- a/tests/test_ipynb.py
+++ b/tests/test_ipynb.py
@@ -138,6 +138,11 @@ def test_cell_magic_noop() -> None:
         pytest.param(
             "env =  %env var", "env = %env var", id="Assignment to environment variable"
         ),
+        pytest.param(
+            "变量 =  %env var",
+            "变量 = %env var",
+            id="Assignment after non-ASCII characters",
+        ),
         pytest.param("env =  %env", "env = %env", id="Assignment to magic"),
     ),
 )


cobaltt7:
Thanks!

Issue #134, opened 2018-04-16
Gradual adoption: option to only format PEP8 errors


maxim-lian:
If I were starting a project from scratch, I'd start with enforcing `black`. But most code is written on existing projects, and converting everything to `black` requires a high degree of confidence and share of agreement within contributors.

What do you think about enabling an option in `black` to reformat only:
- Code that violates PEP8 / some rules
- Code that has been changed

One of the nice attributes of `autopep8` is that it only changes code where there's a problem. This has two big advantages to adoption:
- The impact is limited to strict, unambiguous improvements
- Projects don't require everyone to agree that `autopep8` is the correct tool - an individual / subset can start using it for their contributions 

Thanks again for the project

ambv:
Black by design doesn't enable reformatting parts of the file. This functionality is against PEP 8 which states that internal consistency within the file is more important than any particular style. If you used Black for this purpose and somebody else kept using autopep8 or YAPF, your styles would fight with each other.

Black is both a tool and a code style. You are adopting the style and the tool is merely an automated way of enforcing it. Yes, this requires agreement within your project that this is the style your team wants to enforce. If you decide that you don't like the particular style and you'd rather keep formatting by hand instead, that's fine by me. If you decide that the style is fine but the *tool* is not stable enough yet, hopefully this will be gradually resolved by increased adoption of the tool and squashing bugs one by one.

As far as responding to particular pycodestyle violations and only formatting that code, I fully agree with YAPF's take on it which is that this is a suboptimal strategy. It also famously leads to situations where one fix may introduce other pycodestyle violations.

Summing up, this is deliberately not in scope for this project.

maxim-lian:
OK, thanks for the full response!

wbolster:
fyi, i wrote https://github.com/wbolster/black-macchiato

DylanYoung:
I don't think that's how time works.... It's continuous ;)

akaihola:
The [darker](https://github.com/akaihola/darker) tool applies Black formatting only to lines which have changed in the Git working tree since the last commit.

pauloneves:
Sure the formatting consistency in the file is important and black shouldn't fight with other formatters. My use case is different. My coworkers didn't use any formatter. I persuaded the team to use black as standard formatter, but we have a lot of already written code. 

[...]

PR #5382, opened 2026-09-01
Honor negations in nested .gitignore files

### LINKED ISSUE #5376: Nested .gitignore negations can never re-include what an ancestor excluded
**Describe the bug**

A negation in a nested `.gitignore` can never re-include something an ancestor `.gitignore`
excluded. Git decides the other way round: the *deepest* gitignore file that has a matching
pattern wins, and only inside that file does last-match-win. Black ORs across gitignore files, so
the deepest one — the only one that can say `!` — is often never consulted.

Effect: `black .` silently skips files that git tracks, and `black --check` passes in CI without a
word about it. rc=0, no warning, no diff.

**To Reproduce**

```
.gitignore                                  ->  .vscode
packages/.gitignore                         ->  !playground/.vscode
packages/playground/.vscode/settings.py
packages/playground/keep.py                 (control)
```

```console
$ git --version
git version 2.55.0
$ git init -q && git add . --dry-run
add '.gitignore'
add 'packages/.gitignore'
add 'packages/playground/.vscode/settings.py'
add 'packages/playground/keep.py'

$ black --check -v .
Identified `/tmp/x` as project root containing a .git directory.
Found input source directory: "/tmp/x"
/tmp/x/.git ignored: matches the --exclude regular expression
/tmp/x/packages/playground/.vscode ignored: matches a .gitignore file content
/tmp/x/packages/playground/keep.py already well formatted, good job.

All done! ✨ 🍰 ✨
```

git tracks `packages/playground/.vscode/settings.py`; Black prunes the directory and never sees
it. `keep.py` is the control — neither tool ignores it, so the fixture isn't just ignoring
everything.

**Cause**

`src/black/files.py`, `_path_is_ignored`:

```python
for gitignore_path, pattern in gitignore_dict.items():
    relative_path = ...
    if pattern.match_file(relative_path):
        return True
return False
```

The dict is ordered least-specific → most-specific (the comment above it says so), and the loop
returns `True` on the *first* match. So an exclusion at the root short-circuits before the nested
file gets a vote. By construction no nested negation can ever undo an ancestor's exclusion.

This is not `pathspec`'s doing. I checked: fed the rules of the file that git says decides, plus
the path relative to *that* directory, `pathspec` answers exactly like git on all of these cases,
in both directions. The divergence lives in the combination layer here.

That matters because #5306 mentions this same divergence in passing and guesses it "likely belongs
in pathspec/traversal upstream". As far as I can measure, it doesn't — it's these lines.

**How often it bites**

I ran Black 26.5.1 against `git check-ignore` over 4,373 queries built from the real `.gitignore`
trees of 43 popular repos. Nine divergences. Eight of them are this bug, in five repos —
`facebook/react` (`packages/.gitignore` with `!playground/.vscode`), `python/cpython`
(`Platforms/Android/testbed/.gitignore` with `!.idea/`), `supabase/supabase`
(`docker/.gitignore` with `!volumes/functions/deno.json*`), and two more. The ninth is a
`pathspec` bug I reported separately (cpburnz/python-pathspec#134).

So it's rare per-query but concentrated: the projects that use nested `.gitignore` negations hit
it every time, and they never find out, because the failure mode is silence.

**A note on the fix**

It is *not* "reverse the iteration order" — that would break the ordinary case where the root
ignores and the nested file says nothing. The rule git implements is: for a given path, find the
deepest gitignore file with any pattern matching it, and let that file's last matching pattern
decide; files closer to the root only get asked if no deeper one matched. Directory pruning during
the walk has to follow the same rule, or the directory disappears before the deeper file is read.
That's more than a four-line change, which is probably why #3694 has been open a while.

Happy to send a PR with the fixture above as a regression test if you'd like it — say the word and
I'll do it against `main`. Not opening one unprompted; I saw the note on #5361.

**Environment**

- Black's version: 26.5.1
- OS and Python version: Linux / Python 3.14.7
- git 2.55.0

RRXXZZYY:
### Description

Fixes #5376.

Black stopped checking `.gitignore` files as soon as an ancestor rule matched. That meant a more specific negation in a nested `.gitignore` could never re-include the path. This keeps the last explicit match while walking from the repository root toward the file, so a deeper rule can override an earlier one and a non-match leaves the current state alone.

The regression test uses a `generated` directory rather than the issue's `.vscode` example because `.vscode` is also covered by Black's default exclusions.

### Validation

- The regression test fails on the previous implementation and passes with this change.
- All 24 applicable file-collection tests pass. Three other cases require Windows symlink privileges that are unavailable in this environment.
- The remaining test suite passes when five environment-specific Windows cases (three symlink privilege checks and two ANSI color expectations) are excluded: 476 regular tests and 73 Jupyter tests, with 93% coverage.
- Full pre-commit suite passes.
- The self-format check leaves all 68 files unchanged. The `tox -e run_self` wrapper itself splits a Windows workspace path containing spaces, so I ran its installed Black command with `.` instead.
- `git check-ignore` and `git add --dry-run` both confirm the nested negation re-includes the fixture path.

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? (No formatting style change.)
- [x] Add new / update outdated documentation? (No documentation change needed.)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -97,6 +97,8 @@
 
 <!-- Changes to how Black can be configured -->
 
+- Honor negations in nested `.gitignore` files when an ancestor rule also matches
+  (#5382)
 - Fix `find_project_root` returning a stale cached result when `--code` is used from
   different working directories in the same process. The CWD fallback (used when no
   `srcs` are given) is now resolved before the `lru_cache` key is computed, so each
--- a/src/black/files.py
+++ b/src/black/files.py
@@ -304,6 +304,7 @@ def _path_is_ignored(
     gitignore_dict: dict[Path, GitIgnoreSpec],
 ) -> bool:
     path = root / root_relative_path
+    ignored = False
     # Note that this logic is sensitive to the ordering of gitignore_dict. Callers must
     # ensure that gitignore_dict is ordered from least specific to most specific.
     for gitignore_path, pattern in gitignore_dict.items():
@@ -313,9 +314,12 @@ def _path_is_ignored(
                 relative_path = relative_path + "/"
         except ValueError:
             break
-        if pattern.match_file(relative_path):
-            return True
-    return False
+        # A matching rule in a deeper .gitignore overrides an earlier match. A
+        # None result means this file had no applicable rule and leaves the state alone.
+        match = pattern.check_file(relative_path).include
+        if match is not None:
+            ignored = match
+    return ignored
 
 
 def path_is_excluded(
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -2844,6 +2844,18 @@ class TestFileCollection:
         )
         assert sorted(expected) == sorted(sources)
 
+    def test_nested_gitignore_reincludes_parent_match(self, tmp_path: Path) -> None:
+        packages = tmp_path / "packages"
+        source = packages / "playground" / "generated" / "settings.py"
+        source.parent.mkdir(parents=True)
+        source.write_text("x=1\n", encoding="utf-8")
+        (tmp_path / ".gitignore").write_text("generated\n", encoding="utf-8")
+        (packages / ".gitignore").write_text(
+            "!playground/generated\n", encoding="utf-8"
+        )
+
+        assert_collected_sources([tmp_path], [source], root=tmp_path)
+
     def test_nested_gitignore_directly_in_source_directory(self) -> None:
         # https://github.com/psf/black/issues/2598
         path = Path(DATA_DIR / "nested_gitignore_tests")


KaizenShogun:
I opened #5376, so take this as an interested party's number rather than a review: I measured
this PR against an independent bench before saying anything about it, because a change to how
`.gitignore` files are resolved deserves a number that doesn't come from the test suite shipping
in the same commit.

**Setup.** 500 randomly generated nested-`.gitignore` trees (five directories, a 14-pattern
vocabulary, ~35% of the patterns negated), 4,953 (tree, path) queries, `git check-ignore
--no-index` as the oracle throughout. Each candidate walks the tree the way Black does —
accumulating specs on the way down and *pruning* ignored directories — so directory decisions are
scored too, not just the files that survive the prune. I copied your `_path_is_ignored` verbatim
into the harness instead of paraphrasing it.

| | agrees with git |
|---|---|
| `main` today | 4,885 / 4,953 = **98.63 %** |
| this PR (`c750aa7`) | 4,952 / 4,953 = **99.98 %** |
| union the pattern files with `chain()` (the alternative floated on #5376) | 3,205 / 4,953 = **64.71 %** |

Compared as sets rather than counts, this PR is **67 fixed and 0 regressions**. The third row is
there because that alternative reads well and measures badly: `PathSpec.from_lines` knows nothing
about which file a pattern came from, so every nested pattern leaks sideways as if it lived at the
root. (It's the same harness as the numbers in my #5376 comment; this run is a different size, so
the counts differ slightly from the ones there.)

**On the two implementations.** I proposed walking the dict in reverse and returning at the first
`.gitignore` that has anything to say; you accumulate forward and keep the last one. They agree on
all 4,953 queries here, and yours doesn't materialise a list per path, so mine is the one that
should be dropped. `check_file(...).include is not None` is also the right predicate — plain
`match_file` collapses "no rule applies" and "a negation applies" into the same `False`, which is
exactly the distinction this fix needs.

**The one remaining disagreement is not yours.** Root `.gitignore` with `*.txt`, `a/.gitignore`
with `!b/`, path `a/b/note.txt`: git ignores it, this PR doesn't. A trailing-slash pattern matches
a *directory*, so `!b/` re-includes the directory `b` and leaves `a/b/note.txt` to be judged on its
own, where the root's `*.txt` still catches it. `pathspec` instead lets `b/` match everything
underneath — correct for exclusion, wrong for the negated form — so the re-inclusion leaks onto the
file. That's a pathspec-level asymmetry (same neighbourhood as cpburnz/python-pathspec#129 and
\#134), it predates this PR, and no arrangement of the loop in `files.py` fixes it. It shouldn't
hold this up.

One thing I have **not** done: run Black's own test suite. Black isn't installed on the machine I
measure on, and I'd rather say that than imply coverage I don't have. Everything above is
`files.py`'s resolution logic exercised directly.

— Midas

PR #5385, opened 2026-09-03
Raise UsageError when sources share no common project root

### LINKED ISSUE #5379: `find_project_root` raises `ValueError` for sources on different Windows drives
**Describe the bug**

**To Reproduce**

```sh
pip install black
```

Observed:

```
 ...
    value = find_pyproject_toml(
    )
    common_base = max(
        key=lambda path: path.parts,
    )
```

```
intersection     -> empty
```

**Expected behavior**

**Environment**

**Additional context**

```python
```

Dmao233:
### Description

Fixes #5379.

`find_project_root` passed an empty parent-path intersection to `max()`, which raises `ValueError` when sources live on different Windows drives. It now raises `click.UsageError` instead. Same-drive and POSIX paths that share a parent are unchanged.

### Checklist - did you ...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -108,6 +108,8 @@
 - Validate `BLACK_NUM_WORKERS` values and report invalid values as usage errors instead
   of crashing (#5211)
 - Ignore permission errors when reading cache (#5258)
+- Raise a usage error instead of crashing when sources on different drives share
+  no common project root (#PR)
 
 ### Packaging
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -109,7 +109,7 @@
   of crashing (#5211)
 - Ignore permission errors when reading cache (#5258)
 - Raise a usage error instead of crashing when sources on different drives share
-  no common project root (#PR)
+  no common project root (#5385)
 
 ### Packaging
 
--- a/src/black/files.py
+++ b/src/black/files.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from re import Pattern
 from typing import TYPE_CHECKING, Any, Union
 
+import click
 from mypy_extensions import mypyc_attr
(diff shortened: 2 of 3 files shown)


shashb27:
Verified this on Windows, where the crash lives — thanks for picking it up. The crash-to-usage-error conversion works:

```pycon
>>> from black.files import find_project_root
>>> find_project_root(("C:/work/app/a.py", "D:/tmp/b.py"), None)
UsageError: Sources on different drives share no common project root.
```

One gap against the requirement @cobaltt7 stated for this route ("passing `--config` should still allow formatting to continue"): `main()` calls `find_project_root(src, stdin_filename)` unconditionally whenever `code is None` (`src/black/__init__.py:616-618`), before and independent of the `--config` parameter, so `black C:\...\a.py D:\...\b.py --config my.toml` would still raise. (I couldn't produce a literal two-drive CLI run — only one drive with media on this machine — but the call flow is unconditional; the `--stdin-filename` path doesn't hit this since `-` is substituted with the single filename, so that intersection is never empty.)

There's also a second consumer of the root that either route needs to deal with for formatting to continue: `get_sources`/`gen_python_files` use it for gitignore handling, exclude matching, and the symlink-outside-root guard, and `child.relative_to(root)` can't work across drives with a single root.

I've opened #5386 implementing the other route from cobaltt7's comment (treat it as no config found, warn, use defaults), with a per-source-root fallback that covers both points — feel free to use whatever parts of it are useful, or close it if the error route is preferred and this PR gets amended instead. Happy to test any revision on Windows.

Dmao233:
@shashb27 thanks for the Windows check. You're right that `--config` still hits `find_project_root` unconditionally, so this UsageError route doesn't meet cobaltt7's requirement.

Closing in favor of #5386 (no-root → defaults + warn), which matches the preferred option.

Dmao233:
Superseded by #5386.

PR #5386, opened 2026-09-03
Fall back to the default configuration when sources share no project root

### LINKED ISSUE #5379: `find_project_root` raises `ValueError` for sources on different Windows drives

shashb27:
## Description

Closes #5379
Closes #5422

Fixes the crash in #5379: when the given sources share no common parent directory — for example, they live on different drives on Windows — `find_project_root` raised `ValueError: max() iterable argument is empty` during CLI startup, before any file was formatted.

This implements the option @cobaltt7 preferred in the issue discussion: treat the situation as *no config found* rather than an error.

- `find_project_root` returns `(None, None)` when the sources' parent sets do not intersect, instead of crashing.
- No project-level `pyproject.toml` is discovered in that case; the user-level config applies as usual, otherwise Black's defaults are used.
- A warning is printed (`No project root could be identified: ... Pass --config ... or --quiet ...`), silenced by passing `--config` explicitly or by `--quiet`. (The issue comment mentioned a `--silent` flag; Black's closest equivalent is `--quiet`.)
- Formatting continues.

One structural point that the "no config found" framing alone doesn't cover: the project root is not used only for config discovery — it also anchors gitignore handling, `--exclude`/`--force-exclude` matching, and the symlink-outside-root guard in `get_sources`/`gen_python_files`, and `child.relative_to(root)` can never work across drives with a single root. So when no common root exists, this PR falls back to a **per-source root**: each top-level source (its directory, for files) acts as the root for its own discovery/exclusion handling. Same-drive invocations are completely unchanged.

Note #5385 implements the alternative route from the same comment (a usage error). Whichever route the maintainers prefer is fine by me — the per-source-root fallback here is also what lets formatting continue when `--config` is passed explicitly, which that comment asked for from the error route as well. Happy to rebase, adjust, or reshape this PR either way.

## Checklist

- [x] The change is adequately tested (new tests cover the `(None, None)` return, config discovery skipping the project root, the warning and its `--config`/`--quiet` gating, and end-to-end formatting with per-source roots; verified on Windows against real cross-drive paths, where the original crash reproduces on `main`).

Refs #5379

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,9 +10,9 @@
   --target-version flag (#5167)
 - `--line-ranges` no longer inserts an empty line after a docstring when the range
   covers only the docstring itself (#5312)
-- Fall back to the default configuration, with a warning, when the given sources
-  share no common project root (for example, they are on different drives on
-  Windows) instead of crashing (#5386)
+- Fall back to the default configuration, with a warning, when the given sources share
+  no common project root (for example, they are on different drives on Windows) instead
+  of crashing (#5386)
 
 ### Highlights
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@
   --target-version flag (#5167)
 - `--line-ranges` no longer inserts an empty line after a docstring when the range
   covers only the docstring itself (#5312)
+- Fall back to the default configuration, with a warning, when the given sources
+  share no common project root (for example, they are on different drives on
+  Windows) instead of crashing (#5386)
 
 ### Highlights
 
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -618,6 +618,20 @@ def main(
     )
     ctx.obj["root"] = root
 
+    if (
+        code is None
+        and root is None
+        and not quiet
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -10,6 +10,9 @@
   --target-version flag (#5167)
 - `--line-ranges` no longer inserts an empty line after a docstring when the range
   covers only the docstring itself (#5312)
+- Fall back to the default configuration, with a warning, when the given sources share
+  no common project root (for example, they are on different drives on Windows) instead
+  of crashing (#5386)
```
cobaltt7: Can you move this to the Configuration section please?

shashb27: Done — moved it to the Configuration section (cb57b84). Thanks!

cobaltt7:
Thank you!

PR #5390, opened 2026-09-06
Skip symlink tests when the platform forbids creating one

### LINKED ISSUE #5389: Three symlink tests fail on Windows without symlink privilege (regression of #287)
Describe the bug

Three tests in TestFileCollection create symlinks with no guard, so they fail on Windows for any user who is not elevated and does not have Developer Mode enabled:

FAILED tests/test_black.py::TestFileCollection::test_get_sources_symlink_and_force_exclude
FAILED tests/test_black.py::TestFileCollection::test_get_sources_with_stdin_symlink_outside_root
FAILED tests/test_black.py::TestFileCollection::test_get_sources_with_stdin_filename_and_force_exclude_and_symlink
E   OSError: [WinError 1314] A required privilege is not held by the client
Full suite on Windows (excluding test_blackd.py, which needs the d extra): 3 failed, 442 passed, 4 skipped. The three failures are the only ones.

This is the same bug as [#287](https://github.com/psf/black/issues/287)

[#287](https://github.com/psf/black/issues/287) ("Skip symlink test if can't create one", 2018) reported exactly this for test_broken_symlink, and it was fixed by guarding the call:

try:
    symlink.symlink_to("nonexistent.py")
except (OSError, NotImplementedError) as e:
    self.skipTest(f"Can't create symlinks: {e}")
That guard is still there and still works. The three tests above were added later without it, so the same failure came back for the same reason.

Why CI doesn't catch it

The Windows job's runner can create symlinks, so these pass in CI. The failure only appears on an ordinary Windows contributor's machine — green in CI, red locally, permanently. That asymmetry is why a bug fixed in 2018 could silently return.

To Reproduce

On Windows, as a normal user with Developer Mode off:

git clone https://github.com/psf/black && cd black
pip install -e . pytest
pytest tests/test_black.py -k symlink
Expected behavior

The tests skip with a clear reason, as test_broken_symlink already does.

Environment

Windows-11-10.0.26200-SP0
Python 3.13.15
black at HEAD
Suggested fix (implemented and verified)
Add a small module-level helper beside the other test utilities and use it at all five symlink_to call sites, matching the existing skip behaviour:

def symlink_or_skip(link: Path, target: Path | str) -> None:
    """Create a symlink, or skip the test where the platform forbids one.

    Windows refuses symlink creation unless the process is elevated or Developer
    Mode is enabled, so these tests cannot run for an ordinary Windows
    contributor. Same treatment as test_broken_symlink, which has guarded this
    since #287.
    """
    try:
        link.symlink_to(target)
    except (OSError, NotImplementedError) as e:
        pytest.skip(f"Can't create symlinks: {e}")
pytest.skip rather than self.skipTest because TestFileCollection is a plain pytest class, not a unittest.TestCase. Path | str rather than Union[...] because the file does not import Union and requires-python is >=3.10.

Result on Windows: 3 failed / 442 passed → 0 failed / 442 passed, with the three tests reporting a clear skip reason instead of an obscure OSError.

Patch: patches/black/0001-tests-skip-symlink-tests-when-the-platform-forbids-s.patch

Centralising the helper also means the next symlink test added picks up the guard by default, which is what would have prevented this recurrence.

istoolsfox:
Fixes #5389.

## What

Three tests in `TestFileCollection` create symlinks without any guard, so the test suite fails with `OSError: [WinError 1314]` for any Windows contributor who is neither elevated nor running with Developer Mode enabled:

- `test_get_sources_symlink_and_force_exclude`
- `test_get_sources_with_stdin_symlink_outside_root`
- `test_get_sources_with_stdin_filename_and_force_exclude_and_symlink`

This is the same problem #287 fixed in 2018 for `test_broken_symlink` — that guard is still in place and working, but the three tests above were added later without it. CI cannot catch the regression because the Windows runner can create symlinks; only an ordinary local Windows checkout fails.

## Change

Add a module-level `symlink_or_skip(link, target)` helper next to the other test utilities and use it at all five `symlink_to` call sites in `TestFileCollection`:

```python
def symlink_or_skip(link: Path, target: Path | str) -> None:
    try:
        link.symlink_to(target)
    except (OSError, NotImplementedError) as e:
        pytest.skip(f"Can't create symlinks: {e}")
```

Notes:

- `pytest.skip` instead of `self.skipTest`, since `TestFileCollection` is a plain pytest class, not a `unittest.TestCase` (`test_broken_symlink` keeps its existing `skipTest` guard).
- Centralising the helper means the next symlink test picks up the guard by default, which is what would have prevented this recurrence.
- The fix follows the approach suggested in the issue (thanks @iam-kira for the detailed report and verification).

## Verification

- Before: `3 failed, 442 passed` on Windows (WinError 1314) — matches the issue report.
- After: `0 failed / 442 passed`, with the three tests reporting `SKIPPED (Can't create symlinks: ...)`.
- Full suite: `442 passed, 7 skipped, 8 subtests passed` (`pytest --ignore=tests/test_blackd.py`).
- `black --check tests/test_black.py`, `flake8`, `isort --check-only`, and `mypy` all pass.

## Changelog

Test-only change with no user-facing behaviour impact, so I believe no `CHANGES.md` entry is needed — happy to add one under the appropriate heading if you prefer.

### DIFF
--- a/tests/test_black.py
+++ b/tests/test_black.py
@@ -2644,6 +2644,20 @@ class TestCaching:
             assert len(set(keys)) == len(modes)
 
 
+def symlink_or_skip(link: Path, target: Path | str) -> None:
+    """Create a symlink, or skip the test where the platform forbids one.
+
+    Windows refuses symlink creation unless the process is elevated or
+    Developer Mode is enabled, so these tests cannot run for an ordinary
+    Windows contributor. Same treatment as test_broken_symlink, which has
+    guarded this since GH #287.
+    """
+    try:
+        link.symlink_to(target)
+    except (OSError, NotImplementedError) as e:
+        pytest.skip(f"Can't create symlinks: {e}")
+
+
 def assert_collected_sources(
     src: Sequence[str | Path],
     expected: Sequence[str | Path],
@@ -3039,7 +3053,7 @@ class TestFileCollection:
             actual = tmp / "actual"
             actual.mkdir()
             symlink = tmp / "symlink"
-            symlink.symlink_to(actual)
+            symlink_or_skip(symlink, actual)
 
             actual_proj = actual / "project"
             actual_proj.mkdir()
@@ -3063,7 +3077,7 @@ class TestFileCollection:
 
                 # a few tricky tests for force_exclude
                 flat_symlink = symlink_proj / "symlink_module.py"
-                flat_symlink.symlink_to(actual_proj / "module.py")
+                symlink_or_skip(flat_symlink, actual_proj / "module.py")
                 assert_collected_sources(
                     src=[flat_symlink],
                     root=symlink_proj.resolve(),
@@ -3074,7 +3088,7 @@ class TestFileCollection:
                 target = actual_proj / "target"
                 target.mkdir()
                 (target / "another.py").write_text("print('hello')", encoding="utf-8")
-                (symlink_proj / "nested").symlink_to(target)
+                symlink_or_skip(symlink_proj / "nested", target)
 
                 assert_collected_sources(
                     src=[symlink_proj / "nested" / "another.py"],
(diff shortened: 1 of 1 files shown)


istoolsfox:
Test-only change — no user-facing behavior is touched, so there's nothing to report in CHANGES.md. Could a maintainer add the `ci: skip news` label? Thanks!

cobaltt7:
Thanks!

PR #5393, opened 2026-09-09
Treat 'ruff: ignore' comments as unsplittable pragmas

### LINKED ISSUE #5260: `ruff:ignore` should be processed as a pragma in `contains_pragma_comment`

jabrailkhalil:
## Summary

Black was reformatting lines that end with a `# ruff: ignore[code]` comment, even though those lines are exempt from ruff linting. Splitting such a line moves the pragma onto a different subexpression, silently breaking the suppression — the same reason Black never splits a line ending in `type: ignore`.

Changes:

- `src/black/nodes.py`: added `is_ruff_ignore_comment_string`, mirroring `is_type_ignore_comment_string`. It accepts both `ruff: ignore` and the spacing-free `ruff:ignore` form, with or without a bracketed rule list.
- `src/black/lines.py`: `Line.contains_unsplittable_type_ignore` now also treats a trailing `ruff: ignore` comment as unsplittable, so the RHS-splitting and other line-splitting paths leave these lines intact.
- `src/black/comments.py`: `contains_pragma_comment` recognizes `# ruff:` comments (for example `# ruff: noqa`) as pragmas, so string merging/splitting also leaves them alone.

## Tests

Added `tests/data/cases/ruff_ignore_pragma.py` covering both the spaced and spacing-free `ruff: ignore` forms on a line that previously got parenthesized.

## Verification

```bash
pytest tests/test_format.py tests/test_black.py -q   # 403 passed, 8 subtests passed
python -m black --check src/black/lines.py src/black/nodes.py src/black/comments.py   # clean
python -m mypy src/black/lines.py src/black/nodes.py src/black/comments.py            # clean
```

Fixes #5260.


jabrailkhalil:
The CI workflows for the current head `9fe2423086` are waiting for contributor workflow approval (`action_required`): [diff-shades](https://github.com/psf/black/actions/runs/34509019524), [test](https://github.com/psf/black/actions/runs/34509019560), [docs](https://github.com/psf/black/actions/runs/34509019547), [lint and format](https://github.com/psf/black/actions/runs/34509019604), [changelog](https://github.com/psf/black/actions/runs/34509019640). Could a maintainer approve the pending runs when convenient? These runs have not executed their checks yet; I will address any failures once they run.

cobaltt7:

PR #5395, opened 2026-09-11
Keep a trailing comment on optional parentheses that contain own-line comments

### LINKED ISSUE #3741: Black moves same-block trailing comment inside expression wrapped with parentheses
**Describe the bug**

`black` formats the following code improperly handling `# Trailing same-block`.
```py
x = 1
# Leading comment
del (
    # First inner own-line
    x  # Trailing comment same-line
    # Trailing comment own-line last inner
)  # Trailing same-block
# Trailing own-line
```

**Result**

```diff
@@ -4,5 +4,6 @@ del (
     # First inner own-line
     x  # Trailing comment same-line
     # Trailing comment own-line last inner
-)  # Trailing same-block
-# Trailing own-line
\ No newline at end of file
+    # Trailing same-block
+)
+# Trailing own-line
```

**Environment**

```
black, 23.3.0 (compiled: yes)
Python (CPython) 3.11.3
```
MacOS 13.3.1

ekanshul:
### Description

Fixes #3741.

When optional parentheses wrap an expression that contains own-line comments, an inline comment after the closing bracket was moved inside the brackets:

```python
del (
    x
    # last inner comment
)  # trailing same-block
```

became

```python
del (
    x
    # last inner comment
    # trailing same-block
)
```

and with the own-line comment at the start (`del (\n    # first\n    x\n)  # trailing`) the trailing comment ended up on `x` instead. The same happens for assignment right-hand sides, `return`, `assert` and `for ... in`, since it's the optional-parentheses machinery rather than `del` specifically.

`Line.append_comment()` pins a comment that follows an invisible closing paren to the wrapped leaf (so that `x = (\n    1\n)  # c` still collapses to `x = 1  # c` without the comment migrating between passes). That pin assumes the leaf before the paren is the wrapped expression, but when the parentheses contain standalone comments the previous leaf may be one of those comments. Those parentheses are going to stay visible anyway, so the comment now stays on the closing paren in that case. The tail line of the resulting split consists of just the invisible paren, which the old code turned into a standalone comment; that line is only ever emitted after `ensure_visible()`, so the comment is kept on the paren there too and comes out as `)  # trailing`.

No change for parentheses without own-line comments (they still collapse and the comment moves with the expression), and the output of the new test case is stable across passes.

### Checklist - did you ...

      [stability policy](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy)?
      (N/A: this fixes a comment being moved to a different place, not a deliberate style change; listed under "Stable style" in `CHANGES.md` like the other recent comment-placement fixes. Happy to move it under preview if you'd rather.)
      (N/A)


cobaltt7:
Thanks!

PR #5396, opened 2026-09-12
Avoid looking at __class__.__name__ in run_transformer (#2432)

### LINKED ISSUE #2432: Avoid looking at __name__ attribute in `black.linegen.run_transformer`
Right the transformer code looks at the transformer's `__name__` attribute when deciding to return early: https://github.com/psf/black/blob/ef7c45f28132a7704a4549072ce5b272034fa196/src/black/linegen.py#L964-L967

This is kinda hacky and becomes quite a problem when integrating mypyc since nested functions don't keep that attribute due a bug (https://github.com/mypyc/mypyc/issues/884). The current workaround is to dynamtically create a wrapping class at runtime (via type() mind you :upside_down_face:) which do keep their `__name__` attribute. Let's avoid this entire situation if feasible ^^

**Additional context:**

> I guess it would be a bigger refactor but I'd like it better if we just stopped using `__name__`.

by @JelleZijlstra in https://github.com/psf/black/pull/2431#discussion_r691367047

> That's a bit hacky; maybe we should turn the Transformer type into a pair instead to store this extra information.

by @JelleZijlstra in https://github.com/psf/black/pull/1621#discussion_r476033948

Prateek99-s:
## Summary of changes

    - Promoted `_rhs` from an inner nested function inside `transform_line` to a top-level module function
  `right_hand_split_with_omits`.
    - Removed the `type("rhs", ...)` dynamic class hack and the dummy `self` parameter.
    - Replaced `transform.__class__.__name__ != "rhs"` with `transform is not right_hand_split_with_omits` in `run_transformer`.
    - Closes #2432.

    ## Test Plan

    - Ran pytest on formatting and transformation test suite: `pytest tests/test_format.py tests/test_trans.py` (241 passed)
    - Verified static typing: `mypy src/black/linegen.py` (no issues found)

     Helpful links:

     - PSF COC: https://www.python.org/psf/conduct/
     - Contributing docs: https://black.readthedocs.io/en/latest/contributing/index.html
     - Chat on Python Discord: https://discord.gg/RtVdv86PrH
     - Stability policy: https://black.readthedocs.io/en/stable/the_black_code_style/index.html -->


src/black/linegen.py
```diff
@@ -800,39 +800,7 @@ def transform_line(
...
-            # trailing commas (those have to be exploded).
-            yield from right_hand_split(line, mode, features=features)
-
-        # HACK: nested functions (like _rhs) compiled by mypyc don't retain their
-        # __name__ attribute which is needed in `run_transformer` further down.
-        # Unfortunately a nested class breaks mypyc too. So a class must be created
-        # via type ... https://github.com/mypyc/mypyc/issues/884
-        rhs = type("rhs", (), {"__call__": _rhs})()
-
+        rhs = right_hand_split_with_omits
```
cobaltt7: Could we just use `right_hand_split_with_omits` without renaming it?

CHANGES.md
```diff
@@ -10,6 +10,8 @@
   --target-version flag (#5167)
 - `--line-ranges` no longer inserts an empty line after a docstring when the range
   covers only the docstring itself (#5312)
+- Avoid dynamic class creation and class name inspection in line transformer dispatch
+  (#5396)
```
cobaltt7: This doesn't need a changelog entry

Prateek99-s: Oh ok, removed the changelog entry. Thanks!

src/black/linegen.py
Prateek99-s: Yeah, that makes sense. Updated to use right_hand_split_with_omits directly in the latest commit.

cobaltt7:
Thank you!

PR #5399, opened 2026-09-13
Support version specifiers in `tool.black.required-version` for GitHub Action (fixes #4406)

### LINKED ISSUE #4406: Allow valid version specifier when specifying Black version in `pyproject.toml`

This could be a user error, and if so, please let me know how to achieve the desired result :)

The [documentation](https://black.readthedocs.io/en/stable/integrations/github_actions.html#usage) states that if we use `version` in the GitHub (GH) action workflow, valid version specifier per Python packaging standards can be used. This means we can do:

```
- uses: psf/black@stable
  with:
    options: "--check --verbose"
    src: "./src"
    version: ">= 23.4"
```

However, if we have a `pyproject.toml` file for other configurations in addition to version and wish GH action to use the same configuration for checks, we have to use `tool.black.required-version` field because as I discovered through trial and error, `version` in GH version workflow and `use_pyproject` are _mutually exclusive_.

```
> Run psf/black@stable
> Run # Even when black fails, do not close the shell
Error: 'with.version' and 'with.use_pyproject' inputs are mutually exclusive.

Error: Process completed with exit code 1.
```

And when I tried to do `required-version: >=24.3` in `pyproject.toml`, I encountered the following issue on GH actions:

```
> Run psf/black@stable
> Run # Even when black fails, do not close the shell
Error: Failed to install Black.
Installing black[colorama]==>=24.3...
ERROR: Could not find a version that satisfies the requirement black[colorama]==>=24.3 (from versions: 18.3a0, 18.3a1, 18.3a2, 18.3a3, 18.3a4, 18.4a0, 18.4a1, 18.4a2, 18.4a3, 18.4a4, 18.5b0, 18.5b1, 18.6b0, 18.6b1, 18.6b2, 18.6b3, 18.6b4, 18.9b0, 19.3b0, 19.10b0, 20.8b0, 20.8b1, 21.4b0, 21.4b1, 21.4b2, 21.5b0, 21.5b1, 21.5b2, 21.6b0, 21.7b0, 21.8b0, 21.9b0, 21.10b0, 21.11b0, 21.11b1, 21.12b0, 22.1.0, 22.3.0, 22.6.0, 22.8.0, 22.10.0, 22.12.0, 23.1a1, 23.1.0, 23.3.0, 23.7.0, 23.9.0, 23.9.1, 23.10.0, 23.10.1, 23.11.0, 23.12.0, 23.12.1, 24.1a1, 24.1.0, 24.1.1, 24.2.0, 24.3.0, 24.4.0, 24.4.1, 24.4.2)
ERROR: No matching distribution found for black[colorama]==>=24.3

Notice:  A new release of pip available: 22.3 -> 24.1.2
Notice:  To update, run: python -m pip install --upgrade pip
Error: Process completed with exit code 1.
```

The use-case here is that I would like to have a way to define a more relaxed version requirement for black in `pyproject.toml`, so we can have the same functionality and options as the `version` option in the GH actions workflow.

Allow valid version specifier when specifying Black version in `pyproject.toml`, so the syntax will be equivalent to the `version` in GH actions workflow.

I will admit that I have not looked into the codebase to determine how difficult the enhancement is. If this suggestion is accepted I would be happy to look into it.

We resolved the issue by downgrading the required version in `pyproject.toml` to 23.4.0 for now.

**Additional context**

How we discovered this was that the readme for the `black` repo states that the latest is 24.4.2. We set up our GH actions workflow to use 24.4.2 but only recently discovered that the release version of the VS code black formatted extension still uses 23.4.0.

Screenshot 2024-07-19 at 2 13 00 PM

This causes issues locally as the stated required version in `pyproject.toml` file is not the same as the VS code extension.

The pre-release version, as of July 19th, 2024, uses 24.4.2.

Screenshot 2024-07-19 at 2 31 11 PM

I understand that as the new release comes out, this should not be an issue as the extension will support _slightly_ older versions. However, this feels like a bit of inconsistency and would be very nice to have equivalent notations between the 2 fields.

Prateek99-s:
#### Description:

    ## Summary of changes

    When using `use_pyproject: true` in the GitHub Action, `read_version_specifier_from_pyproject()` previously assumed `required-
  version` was always a bare version number (e.g. `24.4.2`), prefixing `==` if a dot was present. This caused version specifiers
  with operators (e.g. `>=24.3` or `~=24.0`) to become invalid strings like `==>=24.3` and fail during `pip install`.

    This PR updates `read_version_specifier_from_pyproject()` in `action/main.py` to check if `version` starts with a digit before
  prefixing `==` or `~=`, matching the behavior of `determine_version_specifier()` and preserving full PEP 440 version specifiers
  as documented.

    Closes #4406.

    ## Test Plan

    - Tested version parsing across standard bare versions (`24.4.2`, `24`), exact versions (`==24.4.2`), and version constraints
  (`>=24.3`, `~=24.3.0`, `!=24.1.0,>=23.12`).
    - Ran `mypy action/main.py` (0 errors).
    - Ran pytest suite: 238 passed.


CHANGES.md
```diff
@@ -10,6 +10,8 @@
   --target-version flag (#5167)
 - `--line-ranges` no longer inserts an empty line after a docstring when the range
   covers only the docstring itself (#5312)
+- Support PEP 440 version specifiers in `tool.black.required-version` for the GitHub
+  Action (#5399)
```
cobaltt7: Can you move this to the Integrations section please?

cobaltt7:
Thanks!

PR #5401, opened 2026-09-13
Keep paired fmt: skip comments on compound headers

### LINKED ISSUE #5397: Paired `fmt: skip` directives make a valid parenthesized if header fail to format
**Describe the bug**

Safe formatting raises `InvalidInput` for valid Python with `# fmt: skip` on both ends of a parenthesized `if` header.

**To Reproduce**

Run this Python script with Black 26.5.1:

```python
import black

src = """if (  # fmt: skip
    condition
):  # fmt: skip
    pass
"""
compile(src, "<input>", "exec")  # succeeds
black.format_file_contents(src, fast=False, mode=black.Mode())
```

The formatting call raises `InvalidInput`:

```text
Cannot parse: 3:1
    )
    ^
ParseError: bad input
```

**Expected behavior**

Format successfully while keeping the skipped header valid.

**Environment**

- Black: 26.5.1 and main at `20622e1259c29bda81831962ace1348ba1921c84`, run from source.
- Ubuntu, CPython 3.12.3.
- Default `Mode()`, `fast=False`.

**Additional context**

Repeated four times on each revision. Trailing-only skip and no-directive controls succeed. The first internal pass separates `)` from `:`, and the next pass rejects it; the demonstrated outcome is refusal to format.

Related: #3608, #4731 and #5161. This example still fails on the tested main revision.

GPT-6 Astra assisted; I manually reviewed this report.

GruffElixir:
### Description

Keep paired `# fmt: skip` comments on a parenthesized compound header adjacent so the structural colon remains parseable. This fixes valid input being rejected when skipped nodes are rendered.

Fixes #5397

### Checklist

- [x] This is an internal bug fix, not a code style change.
- [x] Tests added for the regression.
- [x] Documentation is not needed for this internal fix.

### Tests

- `python -m pytest tests/test_format.py -q`


torjan0:
Tested this locally: it fixes the crash and the full suite passes (482 passed, 3 skipped).

One case is left over. The same newline logic exists in the nested `stringify_node` branch below, which this patch doesn't touch:

```python
if (
    a in (  # fmt: skip
        1,
    )  # fmt: skip
):  # fmt: skip
    pass
```

With this PR applied, the middle directive is moved onto its own line at two-space indent:

```python
if (
    a in (  # fmt: skip
        1,
    )
  # fmt: skip
):  # fmt: skip
    pass
```

No crash, but skipped content gets reformatted. Applying the same condition at both sites keeps it byte-identical.

Claude Opus 5 assisted with the reproduction.

GruffElixir:
Nice catch. I applied the same sibling-aware handling in the nested stringify_node path and added the exact paired-header case to the fixture. 	ests/test_format.py is green (237 passed).

GruffElixir:
Added the required Unreleased changelog entry for this fix in c418ebc. The formatter regression suite is green locally: tests/test_format.py, 237 passed. Could you take another look?

src/black/comments.py
```diff
@@ -466,17 +477,45 @@ def _handle_regular_fmt_block(
...
+                def stringify_node(n: LN, next_node: LN | None = None) -> str:
                     if isinstance(n, Leaf):
                         if n.type == STANDALONE_COMMENT:
                             result = n.prefix + n.value
+                            if (
+                                isinstance(next_node, Leaf)
+                                and next_node.type in CLOSING_BRACKETS
+                                and "\n" in n.value
+                            ):
+                                comment_start = result.rfind("# fmt")
```
cobaltt7: Wouldn't just `result.find("#")` be more correct here? `#  fmt` with 2+ spaces is also valid, as well as `# foobar # fmt`. Since this is a `STANDALONE_COMMENT` node, I don't think we need to worry about `#` in strings, etc. Please let me know if my assumptions are incorrect! (and we may want to add a comment to clarify that then)

src/black/comments.py
```diff
@@ -466,17 +477,45 @@ def _handle_regular_fmt_block(
...
+                        return "".join(
+                            stringify_node(
+                                child,
+                                (
+                                    n.children[index + 1]
+                                    if index + 1 < len(n.children)
+                                    else None
+                                ),
+                            )
+                            for index, child in enumerate(n.children)
```
cobaltt7, line 520: [nit] This is a lot of indentation, could some of these variables be moved out?

GruffElixir: Done. I moved the newline handling into a helper to keep the recursive serializer simpler.

src/black/comments.py
GruffElixir: Yep, I now search the last serialized line for the first `#`, so earlier header comments do not interfere. The regression fixture covers both spaced and compound directives.

src/black/comments.py
cobaltt7: I was specifically talking about the `stringify_node()` call

GruffElixir: I moved the sibling pairing out of the recursive call.

src/black/comments.py
```diff
@@ -426,6 +426,21 @@ def _is_attached(leaf: Leaf, root: Node) -> bool:
...
+    comment_line_start = result.rfind("\n") + 1
+    comment_start = result.find("#", comment_line_start)
+    if comment_start < 0:
+        return result
+
+    newline_before_comment = result.rfind("\n", 0, comment_start)
+    if (
+        newline_before_comment >= 0
+        and result[newline_before_comment + 1 :].lstrip().startswith("#")
+    ):
```
cobaltt7:
```suggestion
    if newline_before_comment >= 0 and result[
        newline_before_comment + 1 :
    ].lstrip().startswith("#"):
```

cobaltt7:
Thanks!

GruffElixir:
Thanks @cobaltt7

PR #5405, opened 2026-09-14
Fix unparseable output for `# fmt: skip` on compound header bracket (#5402)

### LINKED ISSUE #5402: `# fmt: skip` on an opening bracket alone makes `if`, `while`, `for` and `with` headers fail to parse
**Describe the bug**

On `main`, a `# fmt: skip` on the opening line alone of a parenthesized `if`, `while`, `for` or `with` header raises `InvalidInput`. Released 26.5.1 formats the same input fine, so this is an unreleased regression.

**To Reproduce**

```python
import black

src = "if (  # fmt: skip\n    cond\n):\n    pass\n"
compile(src, "<input>", "exec")  # succeeds
black.format_file_contents(src, fast=False, mode=black.Mode())
```

```text
InvalidInput: cannot parse: 3:1
    )
    ^
ParseError: bad input
```

`while (`, `for i in (` and `with (` behave the same. `def`, `class` and a plain call are unaffected on `main`.

**Expected behavior**

Format successfully, as 26.5.1 does.

**Environment**

Black `main` at `20622e12`; Ubuntu, CPython 3.12.3; default `Mode()`, `fast=False`.

**Additional context**

`git bisect` across the 105 commits in `26.5.1..20622e12` points at 7e8ef6bc (#5144). Its parent `650983f7` formats the input; 7e8ef6bc raises. `_force_standalone_comment_split` splits at the standalone-comment boundary, separating `)` from the `:` that has to stay on that line.

The behaviour is inverted across the release boundary: on 26.5.1 these four succeed while `def`, `class` and call forms raise `ASTSafetyError`. #5161 fixed those, and this appeared.

#5397 and #5401 are the paired-directive form of the same split; #5401 does not address this one.

Claude Opus 5 assisted with the bisect and reproduction.

katariyaVivek:
## Problem

When `# fmt: skip` is placed on the opening bracket of a compound statement header (such as `if`, `elif`, `while`, `for`, `with`, `try/except`, `async for`, `async with`, `match/case`), Black crashes with `InvalidInput: cannot parse ... ParseError: bad input`.

For example:
```python
if (  # fmt: skip
    cond
):
    pass
```
fails with:
```
black.parsing.InvalidInput: cannot parse: 9:1
    )
    ^
ParseError: bad input
```

## Root Cause

In `src/black/linegen.py`, `_force_standalone_comment_split()` splits lines at standalone comment boundaries. When an opening bracket carries `# fmt: skip`, the comment block ends at the closing bracket `)`. The split logic treated the boundary after the comment as an unconditional split point, placing the subsequent `:` on its own line:
```python
if (  # fmt: skip
    cond
)
:
    pass
```
In Python grammar, a colon cannot begin a statement line on its own, producing unparseable code that trips Black's AST safety checks.

## Solution

In `_force_standalone_comment_split()`, ensure that when `current_line.is_comment` is true, we do not split before a trailing `token.COLON`. This keeps the header colon attached to the closing bracket of the compound statement header, preserving valid syntax while maintaining the `# fmt: skip` suppression.

## Testing

- Added `tests/data/cases/fmtskip_header_bracket_5402.py` covering `if`, `elif`, `while`, `for`, `with`, `try/except`, `async for`, `async with`, and `match/case` compound headers.
- Verified pre-fix failure: raises `InvalidInput: cannot parse ... ParseError: bad input` on unpatched code.
- Verified post-fix pass: formats idempotently with 0 AST differences and 0 formatting regressions.
- Ran full test suites:
  - `tests/test_format.py`: 237 passed
  - `tests/test_ranges.py`: 24 passed
  - `tests/test_tokenize.py`: 16 passed

## Verification

- **Targeted Tests**: PASS (`pytest tests/test_format.py -k fmtskip_header_bracket_5402`)
- **Module Tests**: PASS (`pytest tests/test_format.py`, 237 passed)
- **Regression Tests**: PASS (Fails pre-fix with `ParseError: bad input`, passes post-fix)
- **Formatting**: PASS (`black --check src/black/linegen.py`)
- **Linting**: PASS (`flake8 --config=.flake8 src/black/linegen.py`)
- **Type Checking**: PASS (`mypy src/black/linegen.py`)
- **Adversarial Testing**: 8 cases covering nested blocks, slices with colons, multi-colon headers, and normal standalone comments.

## Impact

Fixes an unreleased regression where `# fmt: skip` on opening brackets of compound statement headers produced invalid Python syntax.

## Compatibility

Safe for stable style. Fixes crashes on valid Python syntax without modifying formatting of standard (non-skipped) code.

Fixes #5402
Closes #5403
Closes #5406


cobaltt7:
Thanks!

PR #5407, opened 2026-09-15
Avoid overlong commented RHS splits

### LINKED ISSUE #3925: Black 23.9.1 reformat already clean code expression and creates a line too long
**Describe the bug**

Black 23.9.1 reformat already clean but complex code expression and creates a line too long.
The original expression does not have lines longer than 79 characters.
But black reformat it and creates a line of 94 characters.

**To Reproduce**

Create a file named bug.py with the following code :
```python
if True:
    if True:
        if True:
            abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqr = (
                abcdefghijklmnopqrstuvwxyzabcd
                | (
                    # Blablablablablablablablablablablablablablablablablablabl
                    # blablablablablablablablablablablablablabla - (
                    abcdefghijk - (
                        abcdefghijklmnopqrstuvwxyzabcd
                        | abcdefghijklmnopqrstuvwxyza
                    )
                )
            )


```

And run black with these arguments:

```sh
$ black --check --diff bug.py
```

The resulting display is:
```text
--- bug.py	2023-10-06 10:06:30.801786+00:00
+++ bug.py	2023-10-06 10:12:04.476521+00:00
@@ -1,15 +1,9 @@
 if True:
     if True:
         if True:
-            abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqr = (
-                abcdefghijklmnopqrstuvwxyzabcd
-                | (
-                    # Blablablablablablablablablablablablablablablablablablabl
-                    # blablablablablablablablablablablablablabla - (
-                    abcdefghijk - (
-                        abcdefghijklmnopqrstuvwxyzabcd
-                        | abcdefghijklmnopqrstuvwxyza
-                    )
-                )
+            abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqr = abcdefghijklmnopqrstuvwxyzabcd | (
+                # Blablablablablablablablablablablablablablablablablablabl
+                # blablablablablablablablablablablablablabla - (
+                abcdefghijk
+                - (abcdefghijklmnopqrstuvwxyzabcd | abcdefghijklmnopqrstuvwxyza)
             )
-
would reformat bug.py

Oh no! 💥 💔 💥
1 file would be reformatted.
```

**Expected behavior**

Black should keep it as is or at least propose something that does not go beyond the default line length.

**Environment**

- Black's version: black, 23.9.1 (compiled: yes)
- OS and Python version: Ubuntu 23.04 Python (CPython) 3.11.4

**Additional context**

GruffElixir:
## Summary

- Prevent a commented, parenthesized RHS from being joined to an overlong assignment prefix.
- Add a regression case and a changelog entry.

Fixes #3925

## Validation

- python -m pytest tests/test_format.py -q
- mypy src/black/linegen.py
- python -m pytest tests/test_black.py -q (5 existing Windows environment failures: 2 color-output checks and 3 symlink checks)


cobaltt7:
This needs to be gated to the preview style. Additionally, the diff-shades changes do not look like improvements.

GruffElixir:
Yep, gated the issue-3925 guard behind mode.preview and moved the changelog entry to Preview style. The regression fixture now opts into preview too; 	ests/test_format.py is green (237 passed).

GruffElixir:
Rebased onto current `main` in `f4a6aef` and resolved the changelog conflict. The preview-only guard and Preview style changelog entry are preserved. `tests/test_format.py` passes locally with 244 tests, and `git diff --check` is clean. Could you take another look?

cobaltt7:
This is not how you correctly gate a feature to preview. See 64631f6623868d9cecdb6c98586522316ccb8dc3 for a correct example.

Additionally, the style change is still a regression. See the diff-shades report for examples of the nonimproved formatting.

GruffElixir:
Good catch. I moved the guard to a dedicated `Preview` feature and restricted it to parenthesized assignment RHSs, so unrelated commented expressions keep their existing formatting. All 244 formatter tests and the preview-doc check pass locally. Pushed as `5a61672`.

GruffElixir:
Rebased onto current `main` to resolve the conflict and preserved the newer changelog entry. The pre-commit hooks pass and `tests/test_format.py` passes (245 tests).

docs/the_black_code_style/future_style.md
```diff
@@ -13,6 +13,8 @@ experimental, feedback and issue reports are highly encouraged!
 
 Currently, the following features are included in the preview style:
 
+- `wrap_long_rhs_with_standalone_comments`: Avoid joining an overlong assignment prefix
+  to a parenthesized, commented right-hand side.
```
cobaltt7: This description doesn't make much sense without a knowledge of Black internals. Could it be changed? The changelog entry is a good description.

The feature name is also a bit long if it could be shortened to 3-4 words

GruffElixir:
The latest diff-shades run is clean for both preview and stable.

src/black/linegen.py
```diff
@@ -1219,6 +1222,28 @@ def _prefer_split_rhs_oop_over_rhs(
     Returns whether we should prefer the result from a split omitting optional parens
     (rhs_oop) over the original (rhs).
     """
+    # Do not join an overlong assignment prefix to a binary expression that starts a
+    # parenthesized, commented RHS. This can happen when standalone comments force a
+    # nested split to be considered after the first RHS expression (#3925).
+    if Preview.wrap_long_rhs_with_standalone_comments in mode and (
+        # Keep this limited to a parenthesized assignment RHS. Applying it to other
+        # commented expressions changes established preview formatting.
```
cobaltt7: This check seems overly-specific and it's only seems to target this specific shape. Usually bugs are caused by faulty or missing heuristics, and fixing them improve more formatting than just the one specific reported issue. Could you give some examples of what "established preview formatting" this changes?

Also, we should change the order of these checks - for instance unsplittable type ignores should always be the first check.

cobaltt7: Oh, I guess the "established preview formatting" would be the changes diff-shades flagged that I previously commented on. If this check is really the best way to fix the issue, it's fine, I'm just wary of uber-specific checks because most issues usually hint at bigger problems. Reducing the reproduction to a MRE may help figure out if there is one

GruffElixir:
Rebased onto current `main` and resolved the conflict while keeping both preview flags. `tests/test_format.py` passes (246 tests), and pre-commit passes.

src/black/linegen.py
GruffElixir: Moved the unsplittable type-ignore check ahead of this heuristic and reduced the fixture to a single assignment at line length 24. The 246 formatter tests and pre-commit pass.

docs/the_black_code_style/future_style.md
GruffElixir: Renamed the flag to `wrap_commented_rhs` and rewrote the description around the user-visible line-length effect.

src/black/linegen.py
cobaltt7: So is this the best way to fix the issue? Have you looked into a root cause that causes the issue, or a more generalize heuristic that could be changed or added? Are there other "parenthesized expressions with standalone comments" that this affects?

PR #5408, opened 2026-09-16
Add is_formatted output to GitHub Action (#3526)

### LINKED ISSUE #3526: Github action could return number of changed/formatted files

The official psf/black GitHub Action does not seem to return information on its run. Other related actions such as https://github.com/rickstaa/action-black returned an is_formatted flag which could be used by subsequent steps.

If you use an "open pull request" action after the formatting step, and if black did format any file, then you can automatically open a pull request and when approved the formatted code would be committed to the original branch.

Without such a parameter if you queue an "open a pull request" github action after a "format with black" action, you would ALWAYS have a pull request, even if no files had been changed

Have psf/black return a status or a number of files it changed.

none

**Additional context**

Sample workflow in github actions:

```
name: Formatting

on:
  push:
    branches: [ "main" ]
  pull_request:
    branches: [ "main" ]

jobs:
  black:
    runs-on: ubuntu-latest
    name: Format python
    
    steps:
    - name: Check out
      uses: actions/checkout@v3

    - name: Set up Python 3.10
      uses: actions/setup-python@v4
      with:
        python-version: "3.10"

    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        pip install black autopep8 isort
        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi

    - name: Format with black
      id: action_black
      uses: psf/black@stable
      with:
        options: "--verbose"
        src: "./src"

    - name: Create Pull Request
      uses: peter-evans/create-pull-request@v4
      with:
        title: "Format Python code with psf/black push"
        commit-message: "Format Python code with psf/black"
        body: "Format Python code with psf/black"
        base: main
        branch: black-format-github-action
        delete-branch: true

```

Prateek99-s:
#### Description:

    ## Summary of changes

    - Added `outputs.is_formatted` to `action.yml` and attached `id: black` to the composite execution step.
    - Updated `action/main.py` to write `is_formatted=true` / `is_formatted=false` to `$GITHUB_OUTPUT` based on Black's exit code
  and report output.
    - Documented the new output in `docs/integrations/github_actions.md`.
    - Closes #3526.

    ## Test Plan

    - Ran `mypy action/main.py` (clean, 0 errors).
    - Ran `pytest tests/test_docs.py` (passed).


sylvesterkaczmarek (commented):
`--quiet` seems to make this output wrong in normal write mode. Black can reformat files, return 0, and suppress the `reformatted` summary, which leaves `is_formatted=false`. Since the action lets users pass `--quiet`, could we derive this without parsing human stdout, or at least cover quiet mode in a regression test?

Prateek99-s:
Good catch! When --quiet was passed, Black's stdout summary was completely silenced in write mode, causing is_formatted to default to false.
  
  To resolve this reliably without needing external filesystem diffing:
   1. We strip -q/--quiet from the arguments passed to the internal Black subprocess so Black always produces its standard return code and summary internally.
   2. is_formatted is derived accurately using returncode == 1 or checking for reformatted file counts via regex.
   3. If --quiet was requested by the user, we filter the output prior to printing/writing to OUTPUT_FILE (suppressing summary banners while preserving diff output and error diagnostics).
  4. Added a regression test suite in tests/test_action.py covering write, check, diff, quiet, and error scenarios.

cobaltt7:
See #3531 - the preferred implementation would not be reading stdout/stderr which is meant for humans, not machines. See POC commit https://github.com/rasjani/black/commit/367beb839046471fd2a5c803ad916ede008e6295

sylvesterkaczmarek:
Agreed. Using a machine-readable signal rather than parsing Black's human stdout/stderr is the cleaner direction and also removes the need to strip/replay --quiet just to infer the result. That addresses the concern I raised more robustly; I'd follow the #3531/POC approach from here.

Prateek99-s:
Thanks for pointing me to #3531! I've refactored this to follow the POC approach: Report.write_github_outputs() now write directly to $GITHUB_OUTPUT when present, based on Black's internal change_count. This completely removes the need for stdout parsing, regex matching, and flag stripping in action/main.py. I also added tests in tests/test_black.py covering normal, quiet, and check modes. Ready for review!

sylvesterkaczmarek (approved):
Rechecked current 92a7ffaf. The implementation now derives the GitHub Action output directly from Black's internal Report.change_count instead of parsing stdout/stderr, so --quiet no longer hides the signal. The CLI regressions cover normal, quiet and --check modes, and the current CI matrix is green. This resolves the issue I raised.

src/black/report.py
```diff
@@ -120,3 +120,19 @@ def __str__(self) -> str:
...
         return ", ".join(report) + "."
+
+    def write_github_outputs(self, output: Path) -> None:
+        is_formatted = "true" if self.change_count > 0 else "false"
+        _outputs = {
+            "is_formatted": is_formatted,
+            "is-formatted": is_formatted,
+            "change_count": str(self.change_count),
+            "same_count": str(self.same_count),
+            "failure_count": str(self.failure_count),
```
cobaltt7, line 130: Why have both `is_formatted` and `is-formatted`?

I like having the other outputs as well, but they need to be documented and added to action.yml.

sylvesterkaczmarek (approved):
Rechecked current 1c5ee61f. The action still derives its machine-readable result from Report state rather than human stdout/stderr, so the --quiet issue remains resolved. The current head also exposes change_count/same_count/failure_count through the composite step, documents them, and keeps regression coverage for normal, quiet and --check modes. The full current CI matrix is green. No blocker from me.

PR #5411, opened 2026-09-17
Don't require sys.stdout.buffer when writing formatted output or diffs

### LINKED ISSUE #2516: Running in Jupyter Notebook raises 'OutStream' object has no attribute 'buffer'
**Describe the bug**

When running the function `reformat_code` (and code provided as a string) inside a Jupyter Notebook (or `ipython`), it raises the following exception:

```python
  File "C:\Python37-32\lib\site-packages\black\__init__.py", line 585, in reformat_code
    content=content, fast=fast, write_back=write_back, mode=mode
  File "C:\Python37-32\lib\site-packages\black\__init__.py", line 843, in format_stdin_to_stdout
    sys.stdout.buffer, encoding=encoding, newline=newline, write_through=True
AttributeError: 'OutStream' object has no attribute 'buffer'
error: cannot format <string>: 'OutStream' object has no attribute 'buffer'
```

We can find many similar issue on Github, including [this one](https://github.com/Yelp/mrjob/issues/1441#issuecomment-249267658), which 
contains the following analysis:

> looks like stdout and stderr are ipykernel.iostream.OutStream objects, which don't have a buffer attribute, but which seem perfectly happy to accept bytes as well as unicode.

This function works perfectly fine on a regular Python file.

**To Reproduce**

You can check the following notebook:
https://colab.research.google.com/gist/jfthuong/f8732badb0e19b2109e6e05346ca3b1b/formatting-code-with-black.ipynb

Or create a new one, with:
1. `from black import WriteBack, Report, Mode, reformat_code, TargetVersion, reformat_one`
2. Create a string with some source code
3. Create `WriteBack`, `Report`, and `Mode` objects
4. Call the function `reformat_code`

```python
reformat_code(
    content=code, fast=True, write_back=write_back, mode=mode, report=report
)
```

**Expected behavior**

Reformatted code. But error is raised.

**Environment:**

- Version: black-21.9b0
- OS and Python version: Jupyter on Windows / Linux (Python 3.7)

Yes.

(installed with `!pip install -q git+https://github.com/psf/black/`).

**Additional context**

Inside Jupyter Notebook / ipython. OK from "regular" python `.py` file.

ethanstoner:
### Description

Fixes #2516

**Bug:** Calling `black.reformat_code()` / `black.format_stdin_to_stdout()` from a Jupyter notebook fails with `AttributeError: 'OutStream' object has no attribute 'buffer'`. The diff output of `format_file_in_place()` (`WriteBack.DIFF` / `COLOR_DIFF`) has the same problem.

**Root cause:** Both functions wrap `sys.stdout.buffer` in `io.TextIOWrapper` so they can control the encoding and newline translation. `TextIOBase.buffer` is [not part of the `TextIOBase` API](https://docs.python.org/3/library/io.html#io.TextIOBase.buffer) and ipykernel's `OutStream` doesn't have it. `format_stdin_to_stdout()` builds that wrapper in its `finally` block for every `write_back` mode, so even `--check`-style calls crash. The issue thread notes that just using `getattr(sys.stdout, "buffer", sys.stdout)` doesn't work either, because `OutStream.write()` only accepts `str`.

**Change:** Both code paths now go through a small `_write_to_stdout()` helper:

- If `sys.stdout` has a `buffer`, it does what the code did before: wraps it in `io.TextIOWrapper(..., write_through=True)`, wraps it for Windows colors where needed, writes, then detaches.
- If there is no `buffer`, it writes text straight to `sys.stdout`. It does the `\n` -> `newline` translation itself (this is what `TextIOWrapper` would do, and the content has already been normalized to `\n` at that point) and leaves encoding to the stream. The stream is never detached or closed.

`format_stdin_to_stdout()` now only touches stdout when `write_back` is `YES`/`DIFF`/`COLOR_DIFF`. `wrap_stream_for_windows()`'s annotation is widened from `io.TextIOWrapper` to `TextIO` so it accepts `sys.stdout`. At runtime it already accepted any stream.

**Tests:** I added three tests that patch `sys.stdout` with an `io.StringIO`, which has no `buffer` and only accepts `str` (like `OutStream`). They cover `format_stdin_to_stdout` with `YES`/`CHECK`/`NO`/`DIFF`/`COLOR_DIFF`, `reformat_code`, and the `format_file_in_place` diff path with both `\n` and `\r\n` files. All three failed with the `AttributeError` before the fix and pass after it.

Validation (macOS, Python 3.14):
- `pytest -n auto tests`: 484 passed, 3 skipped
- `pre-commit run -a`: everything passes except mypy, which reports an `action/main.py:60` `tomllib` error that also shows up on unmodified `main` in my environment. It has nothing to do with this change, and there are no mypy errors in the touched files.
- `black --check` on the changed files: clean

### Checklist - did you ...

- [x] Implement any code style changes under the `--preview` style, following the stability policy? (N/A, no style changes)
- [x] Add new / update outdated documentation? (N/A)

Generated with Claude Code on behalf of @ethanstoner.


cobaltt7:
Disregard my previous comment, the patch wasn't properly downloading. Confirmed this IS working. Thanks for the fix!

PR #5422, opened 2026-09-21
Fix ValueError in find_project_root for files on multiple Windows drives (#5379)

### LINKED ISSUE #5379: `find_project_root` raises `ValueError` for sources on different Windows drives

Prateek99-s:
Resolves #5379.

    **Description:**
    When passing sources located on completely different Windows drives (e.g., `black C:\app\a.py D:\tmp\b.py`), `find_project_root` computes the common parents of all input paths. Because paths on separate Windows drives have no common `Path` parent, the resulting set intersection is empty. Previously, passing this empty intersection to `max()` caused an unhandled `ValueError` crash.

    This PR fixes the crash by checking if the intersection is empty before calling `max()`. If there is no common base (as is the
  case with files on different drives), it gracefully falls back to using the resolved current working directory (`Path.cwd()`) as
  the base to search for configuration files like `pyproject.toml`.

    **Changes:**
    - Updated `src/black/files.py` to handle an empty parent intersection and fallback to `Path.cwd()`.
    - Added `test_find_project_root_multiple_drives` regression test in `tests/test_black.py` (skipped on non-Windows platforms).


cobaltt7:
> I believe the project root is only used for finding a config file. I think the best option is to treat it as no config found, use the default config file, and probably warn that no project root was found and the default config was used. Passing `--config` with a file path manually should silence the warning, as well as passing `--silent`.
> Alternatively, Black could just error with a clear message. If this route is chosen, passing `--config` should still allow formatting to continue. 

 _Originally posted by @cobaltt7 in [#5379](https://github.com/psf/black/issues/5379#issuecomment-5517373573)_

PR #5427, opened 2026-09-22
Fix splitting stringified return annotations with parameters

### LINKED ISSUE #3926: Long string in return annotation does not get split if there's parameters

Mohamedajab:
### Description

Fixes #3926.

Long stringified return annotations were split when a function had no parameters, but adding parameters caused only the parameter list to split, leaving the annotation over the configured line length. This routes direct stringified return annotations through the right-hand-side transformation and preserves the function colon when wrapping and splitting the string.

### Checklist - did you ...

- [x] Implement the style change under `--unstable`, following the stability policy?
- [x] Add / update tests?
- [x] Add new / update outdated documentation? No user documentation change is needed.

Validation:
- `pytest tests/test_format.py -q` (243 passed)
- `pytest tests/test_trans.py -q` (1 passed)
- repository pre-commit hooks (isort, flake8, mypy, whitespace checks)
- `git diff --check`


cobaltt7:
Some other things to test:
- pragmas/type:ignore/fmt:skip shouldn't split the string
- long strings without spaces can't be split

Also please run black on itself.
Thank you!

cobaltt7:
Thanks!

